PR #119, opened 2021-02-11
Build and manage extensions separately

ankitkumarr:
To test this, make sure you have the latest `azure-functions-core-tools` released yesterday (>= `3.0.3284`). 
- Run `tools\devpack.ps1` (you may need to change the feed path, or add that path to your nuget local feed)
- Go to `samples\FunctionApp`
- Run `func host start`
- Watch it work :)

Some pending items (may come separately) --
1. I will update the E2E test for Cosmos once Sdk is released. 
2. [Discuss] Names of new attributes -- I have placeholders as `HttpEventTrigger` (all `Triggers` are made to `EventTriggers` to avoid confusion).
3. Add to build and release pipelines

I have few more changes to follow-up, along with other extensions. But this PR is already too big and may not be best to combine more.

Good News: Size of the new package that's published is like 6 MB :)

### DIFF
--- a/extensions/Worker.Extensions.CosmosDB/CosmosDBOutputAttribute.cs
+++ b/extensions/Worker.Extensions.CosmosDB/CosmosDBOutputAttribute.cs
@@ -1,0 +1,103 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+﻿using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.CosmosDB
+{
+    public class CosmosDBOutputAttribute : OutputBindingAttribute
+    {
+        /// <summary>
+        /// Constructs a new instance.
+        /// </summary>
+        /// <param name="databaseName">The CosmosDB database name.</param>
+        /// <param name="collectionName">The CosmosDB collection name.</param>
+        public CosmosDBOutputAttribute(string name, string databaseName, string collectionName) : base(name)
+        {
+            DatabaseName = databaseName;
+            CollectionName = collectionName;
+        }
+
+        /// <summary>
+        /// The name of the database to which the parameter applies.
+        /// May include binding parameters.
+        /// </summary>
+        public string DatabaseName { get; private set; }
+
+        /// <summary>
+        /// The name of the collection to which the parameter applies.
+        /// May include binding parameters.
+        /// </summary>
+        public string CollectionName { get; private set; }
+
+        /// <summary>
+        /// Optional.
+        /// Only applies to output bindings.
+        /// If true, the database and collection will be automatically created if they do not exist.
+        /// </summary>
+        public bool CreateIfNotExists { get; set; }
+
+        /// <summary>
+        /// Optional. A string value indicating the app setting to use as the CosmosDB connection string, if different
+        /// than the one specified in the <see cref="CosmosDBOptions"/>.
+        /// </summary>
+        public string? ConnectionStringSetting { get; set; }
+
+        /// <summary>
+        /// Optional. The Id of the document to retrieve from the collection.
+        /// May include binding parameters.
+        /// </summary>
+        public string? Id { get; set; }
+
+        /// <summary>
+        /// Optional.
+        /// When specified on an output binding and <see cref="CreateIfNotExists"/> is true, defines the partition key 
+        /// path for the created collection.
+        /// When specified on an input binding, specifies the partition key value for the lookup.
+        /// May include binding parameters.
+        /// </summary>
+        public string? PartitionKey { get; set; }
+
+        /// <summary>
+        /// Optional.
+        /// When specified on an output binding and <see cref="CreateIfNotExists"/> is true, defines the throughput of the created
+        /// collection.
+        /// </summary>
+        public int CollectionThroughput { get; set; }
+
+        /// <summary>
+        /// Optional.
+        /// When specified on an input binding using an <see cref="System.Collections.Generic.IEnumerable{T}"/>, defines the query to run against the collection.
+        /// May include binding parameters.
+        /// </summary>
+        public string? SqlQuery { get; set; }
+
+        /// <summary>
+        /// Optional.
+        /// Enable to use with Multi Master accounts.
+        /// </summary>
+        public bool UseMultipleWriteLocations { get; set; }
+
+        /// <summary>
+        /// Optional.
+        /// Enables the use of JsonConvert.DefaultSettings in the monitored Azure Cosmos DB collection.
+        /// <remarks>
+        /// This setting only applies to the monitored collection and the consumer to setup the serialization used in the monitored collection.
+        /// The JsonConvert.DefaultSettings must be set during the initialization process.
+        /// This is achieved by deriving a class from <see cref="CosmosDBWebJobsStartup"/> and adding a <see cref="WebJobsStartupAttribute"/>
+        /// to the assembly that specifies the derived class
+        /// </remarks>
+        /// </summary>
+        public bool UseDefaultJsonSerialization { get; set; }
+
+        /// <summary>
+        /// Optional.
+        /// Defines preferred locations (regions) for geo-replicated database accounts in the Azure Cosmos DB service.
+        /// Values should be comma-separated.
+        /// </summary>
+        /// <example>
+        /// PreferredLocations = "East US,South Central US,North Europe"
+        /// </example>
+        public string? PreferredLocations { get; set; }
+    }
+}
--- a/extensions/Worker.Extensions.CosmosDB/Worker.Extensions.CosmosDB.csproj
+++ b/extensions/Worker.Extensions.CosmosDB/Worker.Extensions.CosmosDB.csproj
@@ -1,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\..\build\Common.props" />
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <LangVersion>preview</LangVersion>
+    <AssemblyName>Microsoft.Azure.Functions.Worker.Extensions.CosmosDB</AssemblyName>
+    <RootNamespace>Microsoft.Azure.Functions.Worker.Extensions.CosmosDB</RootNamespace>
+    <SignAssembly>true</SignAssembly>
+    <AssemblyOriginatorKeyFile>..\..\key.snk</AssemblyOriginatorKeyFile>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Worker.Extensions.Abstractions\Worker.Extensions.Abstractions.csproj" />
+  </ItemGroup>
+
+</Project>
--- a/extensions/Worker.Extensions.Storage/Blobs/BlobInputAttribute.cs
+++ b/extensions/Worker.Extensions.Storage/Blobs/BlobInputAttribute.cs
@@ -1,0 +1,33 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+﻿using System;
+using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Storage
+{
+    public sealed class BlobInputAttribute : InputBindingAttribute
+    {
+        private readonly string _blobPath;
+
+        /// <summary>Initializes a new instance of the <see cref="BlobInputAttribute"/> class.</summary>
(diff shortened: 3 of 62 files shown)


samples/FunctionApp/FunctionApp.csproj
```diff
@@ -13,15 +13,14 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.0.0-preview2" />
```
ankitkumarr, line 16: Will fix this before merge.

ankitkumarr:
Build is failing because `FunctionApp` Sample project is referencing the `Sdk` packages that's not released. I can release it in a private feed once changes are approved, and use that to merge this PR.

samples/FunctionApp/FunctionApp.csproj
```diff
@@ -13,15 +13,14 @@
...
-    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Http" Version="3.0.8" />
-    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Storage" Version="4.0.3" />
-    <PackageReference Include="Microsoft.Azure.WebJobs.Script.Abstractions" Version="1.0.0-preview" />
-    <PackageReference Include="Microsoft.Azure.WebJobs.Script.ExtensionsMetadataGenerator" Version="1.2.0" />
```
ankitkumarr, line 20: We don't need any packages from `Microsoft.Azure.WebJobs.*` now :)

mhoeger:
+ @anthonychu on naming discussion?

samples/FunctionApp/FunctionApp.csproj
fabiocav: That was the goal! Nice

sdk/Sdk/Tasks/EnrichExtensionsMetadata.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        // From ExtensionsMetadataGenerator
+        // https://github.com/Azure/azure-functions-host/blob/51e7a2ead2327889da27c28681f89e60bd75a46f/tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator.Console/ExtensionsMetadataGenerator.cs#L71
+        private static string GenerateExtensionsJson(IEnumerable<ExtensionReference> extensionReferences)
+        {
+            var referenceObjects = extensionReferences.Select(r => string.Format("{3}    {{ \"name\": \"{0}\", \"typeName\":\"{1}\", \"hintPath\":\"{2}\"}}", r.Name, r.TypeName, r.HintPath, Environment.NewLine));
+            string json = string.Format("{{{1}  \"extensions\":[{0}{1}  ]{1}}}", string.Join(",", referenceObjects), Environment.NewLine);
+            return json;
+        }
```
brettsam, line 42: Why'd you have to do your own string formatting? Could you let JSON Serialization handle all this for you?

brettsam (commented):
This is great!  a few comments but nothing major...

extensions/Worker.Extensions.CosmosDB/CosmosDBEventTriggerAttribute.cs
```diff
@@ -0,0 +1,156 @@
+﻿using System;
+using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.CosmosDB
+{
+    public sealed class CosmosDBEventTriggerAttribute : TriggerBindingAttribute
```
SeanFeldman, line 6: nit: should the name be `CosmosDbEventTriggerAttribute`? 
Similar to [`DbException`]( https://docs.microsoft.com/en-us/dotnet/api/system.data.common.dbexception.-ctor?view=net-5.0) or [`SqlDbType`](https://docs.microsoft.com/en-us/dotnet/api/system.data.sqldbtype?view=net-5.0).

extensions/Worker.Extensions.CosmosDB/CosmosDBOutputAttribute.cs
```diff
@@ -0,0 +1,100 @@
+﻿using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.CosmosDB
+{
+    public class CosmosDBOutputAttribute : OutputBindingAttribute
```
SeanFeldman, line 5: Similar question about naming, `CosmosDbOutputAttribute`.

extensions/Worker.Extensions.Storage/Blobs/BlobInputAttribute.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// <summary>
+        /// Gets the path of the blob to which to bind.
+        /// </summary>
+        public string BlobPath
+        {
+            get { return _blobPath; }
+        }
+
+        /// <summary>
+        /// Gets or sets the app setting name that contains the Azure Storage connection string.
```
SeanFeldman, line 26: Would it be beneficial to elaborate on why the property is nullable?

extensions/Worker.Extensions.Storage/Queues/QueueEventTriggerAttribute.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Storage
+{
+    public sealed class QueueEventTriggerAttribute : TriggerBindingAttribute
```
SeanFeldman, line 6: I never liked the name `QueueTrigger`. ServiceBus has queues. And those can act as triggers. Perhaps `StorageQueueEventTriggerAttribute`? That way, the code is more intuitive:

```c#
[FunctionName("MyFunc")]
public static void QueueTrigger([StorageQueueTrigger("myqueue")] string item, ILogger log)
{
  log.LogInformation($"C# function processed: {item}");
}
```

extensions/Worker.Extensions.Storage/Queues/QueueOutputAttribute.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Storage
+{
+    public sealed class QueueOutputAttribute : OutputBindingAttribute
```
SeanFeldman, line 6: Same naming comment as for trigger.

sdk/Sdk/Tasks/EnrichExtensionsMetadata.cs
ankitkumarr: I don't need to. But I just copied the logic for how we write in ExtensionsMetadataGenerator ([here](https://github.com/Azure/azure-functions-host/blob/51e7a2ead2327889da27c28681f89e60bd75a46f/tools/ExtensionsMetadataGenerator/src/ExtensionsMetadataGenerator.Console/ExtensionsMetadataGenerator.cs#L71)) to avoid disparity. Had added a comment [here](https://github.com/Azure/azure-functions-dotnet-worker/pull/119/files/a105f1ee09b2e091dc2fdf544e0251169af07d07#diff-2a1b9b73617945b8cb93b0719e4b1d78fe442fda27784c48e5e593bd86c2897eR36). But yeah, there shouldn't be any disparity anyway, it's just a JSON file.

I will update to use the JsonWriter.

extensions/Worker.Extensions.CosmosDB/CosmosDBEventTriggerAttribute.cs
ankitkumarr: That's a good point! Though, I've been trying to keep it consistent with the types we have in `WebJobs` that are used today by .NET Core 3 functions. Example - https://github.com/Azure/azure-webjobs-sdk-extensions/blob/dev/src/WebJobs.Extensions.CosmosDB/CosmosDBAttribute.cs

extensions/Worker.Extensions.Storage/Queues/QueueEventTriggerAttribute.cs
ankitkumarr: Azure Storage sdk calls it `Queue` for all their types ([Example](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/storage/Azure.Storage.Queues/src/QueueClient.cs)). I think there's value in keeping parity between the terminology used by the Storage sdk.

And these `Queue` types would be part of the Storage extension package which would be different from the service bus extension package. I think you can also make the same argument for another Storage service that may have queue trigger support for Azure Functions. :)

I've tried to keep these types same as our WebJobs types except some semantical difference to tell them apart easily (Example - Trigger vs EventTrigger). That way it's easy for us to generate the metadata and keep it same across languages. So, unless there's a compelling argument, I would vote to keep the "WebJobs" types and these types consistent (with added convention to tell them apart),

extensions/Worker.Extensions.CosmosDB/Worker.Extensions.CosmosDB.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\..\build\Common.props" />
```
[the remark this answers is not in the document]
ankitkumarr: I had it this way for now, as it was easy for testing when I was trying it through package-reference (thanks to the LocalBuild Id you added).

I think versioning up to us though. Do we want them to match WebJobs? Or we could start at `1.0.0`

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -98,199 +119,226 @@ internal IEnumerable<SdkFunctionMetadata> GenerateFunctionMetadata(TypeDefinitio
...
+
+            return binding;
+        }
+
+        private static string GetBindingType(CustomAttribute attribute)
+        {
+            var attributeType = attribute.AttributeType.Name;
+
+            // TODO: fix this if we continue to use "<>EventAttribute" (questionable)
+            // TODO: Should "webjob type" be a property of the "worker types" and come from there?
```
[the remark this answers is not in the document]
ankitkumarr: Yeah, I can open an issue and address that separately I think once we settle on the "convention" we want to follow for these attributes.

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -98,199 +119,226 @@ internal IEnumerable<SdkFunctionMetadata> GenerateFunctionMetadata(TypeDefinitio
...
+            return attributeType
+                    .Replace("EventTriggerAttribute", "Trigger")
+                    .Replace("InputAttribute", string.Empty)
+                    .Replace("OutputAttribute", string.Empty)
+                    .Replace("Attribute", string.Empty);
```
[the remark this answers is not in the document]
ankitkumarr: Probably nothing right now. I think it was a left-over that I forgot to remove. I will update.

brettsam (approved):

samples/FunctionApp/Function3/Function3.cs
```diff
@@ -1,24 +1,27 @@
...
+        [QueueOutput("name", "functionstesting2", Connection = "AzureWebJobsStorage")]
         public static HttpResponseData Run([HttpTrigger(AuthorizationLevel.Anonymous, "get", "post", Route = null)] HttpRequestData req,
-            [Queue("functionstesting2", Connection = "AzureWebJobsStorage")] OutputBinding<string> name)
+            FunctionExecutionContext context)
         {
             var response = new HttpResponseData(HttpStatusCode.OK);
             response.Body = "Success!!";
 
-            name.SetValue("some name");
+            context.OutputBindings["name"] = "some name";
```
mhoeger, line 24: This might just be my opinion, but I think I prefer the "OutputBinding" implementation. A few concerns about the "QueueOutput" attribute: 1. the lack of type enforcement on the "name" property (might introduce some bugs javascript developers know too well) and 2. the lack of intellisense around context.OutputBindings (so it's not so intuitive that there would be a "name" property). I know IAsyncCollector was a bit wonky, but I wonder if it's also because the name wasn't intuitive (in how it relates to an output binding) and because it was one of many ways to do output bindings.

ankitkumarr: As discussed offline, we are thinking of keeping this as is (without OutputBinding type). Once we write some samples using different extensions, we'll come up with scenarios where we think one approach may work better than the other, and decide on that. We will be sure to that before we release these set of changes.

extensions/Worker.Extensions.CosmosDB/Worker.Extensions.CosmosDB.csproj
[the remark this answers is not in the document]
ankitkumarr: I will take care of versioning fixes in a separate PR.

PR #120, opened 2021-02-12
Implementing HTTP request data representation

### LINKED ISSUE #110: HTTP Body as bytes + no duplication
Enable the following capabilities:
- [RawHttpBodyBytes](https://github.com/Azure/azure-functions-host/wiki/Out-of-Proc-Worker-Capabilities#rawhttpbodybytes)
- [RpcHttpBodyOnly](https://github.com/Azure/azure-functions-host/wiki/Out-of-Proc-Worker-Capabilities#rpchttpbodyonly)
- [RpcHttpTriggerMetadataRemoved](https://github.com/Azure/azure-functions-host/wiki/Out-of-Proc-Worker-Capabilities#rpchttptriggermetadataremoved)

### LINKED ISSUE #112: Avoid buffer overrun on empt-valued HTTP properties by enabling "UseNullableValueDictionaryForHttp" capability
Read more here: https://github.com/Azure/azure-functions-host/wiki/Out-of-Proc-Worker-Capabilities#usenullablevaluedictionaryforhttp

### LINKED ISSUE #121: Update gRPC types with more recent proto file
The gRPC types were created with an outdated proto file, resulting in missing fields.

This is a blocker for https://github.com/Azure/azure-functions-dotnet-worker/issues/112

fabiocav:
Initial implementation, some cleanup still in progress. 

Resolves #110 
Resolves #112
Resolves #121 

Related to #122

- [ ] - Update the sample app with the new APIs to demonstrate usage.
- [ ] - Add helper methods to more easily read the request body.
- [ ] - Tests

Updates to the response will come in a separate PR.


src/DotNetWorker/Grpc/GrpcHttpRequestData.cs
```diff
@@ -0,0 +1,67 @@
...
+                if (_httpData.Body.DataCase != TypedData.DataOneofCase.Bytes)
+                {
+                    throw new NotSupportedException($"{nameof(GrpcHttpRequestData)} expects binary data only. The provided data type was '{_httpData.Body.DataCase}'.");
+                }
+
+                return _httpData.Body.Bytes.Memory;
+            }
+        }
+
+        public override IImmutableDictionary<string, string> Headers => _headers ??= _httpData.NullableHeaders.ToImmutableDictionary(kvp => kvp.Key, kvp => kvp.Value.Value);
```
brettsam, line 39: Do we need to make this something like `IHeaderDictionary`? I think @mhoeger mentioned that multiple identical keys wasn't supported... but maybe it could future-proof it?

src/DotNetWorker/Grpc/GrpcHttpRequestData.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        public override IReadOnlyCollection<IHttpCookie> Cookies => _httpData.Cookies;
+
+        public override Uri Url => _url ??= new Uri(_httpData.Url);
+
+        public override IEnumerable<ClaimsIdentity>? Identities
```
brettsam, line 45: Should we make this not-nullable and always return an empty collection if there aren't any? It'd make consuming this API a little nicer (no null-checks).

src/DotNetWorker/Http/SameSite.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    public enum SameSite
```
brettsam, line 6: Do we need to replicate every type like this? For example, we use `System.Net.HttpStatusCode` in responses.

src/DotNetWorker/Http/HttpRequestData.cs
```diff
@@ -1,18 +1,33 @@
...
         public abstract IImmutableDictionary<string, string> Headers { get; }
 
-        public abstract string Body { get; }
+        // <inheritdoc/>
+        public abstract IReadOnlyCollection<IHttpCookie> Cookies { get; }
+
+        // <inheritdoc/>
+        public abstract Uri Url { get; }
 
-        public abstract IImmutableDictionary<string, string> Params { get; }
```
mhoeger, line 14: I think it makes sense to bring back "Params"? (And not expose trigger_metadata in the future)

src/DotNetWorker/Grpc/GrpcHttpRequestData.cs
fabiocav: I like the idea

src/DotNetWorker/Grpc/GrpcHttpRequestData.cs
fabiocav: Done.

src/DotNetWorker/Http/IHttpCookie.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        double? MaxAge { get; }
+
+        string Name { get; }
+
+        string Path { get; }
+
+        SameSite SameSite { get; }
+
+        NullableBool Secure { get; }
```
SeanFeldman, line 24: Any specific reason why this is not a `bool?`?

src/DotNetWorker/Http/HttpRequestData.cs
fabiocav: After the discussion we had, I'm inclined to delay this for now to make sure this is the right model. Thoughts?

src/DotNetWorker/Http/IHttpCookie.cs
fabiocav: No. It was supposed to be. Updating. We won't be exposing those types as part of the public surface. There's some refactoring coming that will prevent that from happening.

src/DotNetWorker/Grpc/GrpcHttpRequestData.cs
```diff
@@ -0,0 +1,68 @@
...
+        private HttpHeadersCollection _headers;
+
+        public GrpcHttpRequestData(RpcHttp httpData)
+        {
+            _httpData = httpData ?? throw new ArgumentNullException(nameof(httpData));
+        }
+
+        public override ReadOnlyMemory<byte>? Body
+        {
+            get
```
mhoeger, line 28: Similar to brett's comment - should this not be not-nullable?

src/DotNetWorker/Http/IHttpRequestData.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Collections.Immutable;
+using System.Security.Claims;
+using Microsoft.Azure.Functions.Worker.Http;
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    /// <summary>
+    /// A representation of the HTTP request sent by the host.
+    /// </summary>
+    public interface IHttpRequestData
```
brettsam, line 15: Do we need this?

fabiocav: No. As discussed, this was added to support the approach of augmenting the gRPC types with the implementation, but that wasn't feasible and the interface was left behind. Removed now.

brettsam (approved):

PR #123, opened 2021-02-12
logging changes

brettsam:
A couple of notes about the approach here. I think that we should follow a similar pattern to gRPC: https://docs.microsoft.com/en-us/aspnet/core/grpc/diagnostics?view=aspnetcore-5.0. There's a mix of logging, `DiagnosticSource`, and metrics where necessary. This only gets us running with the logging part and the others can follow up later.

Also see this great comment about logging abstractions in .NET.: https://github.com/aspnet/Logging/issues/332#issuecomment-187836070. You can check out the entire issue -- lots of good stuff there.

Notably:
1. Context no longer has a Logger parameter.
2. Logger can be accessed via DI. This removes the ILogger param that we've had for WebJobs and in-proc that has caused confusion. `ILogger` isn't something typically injected by DI (.NET only supports `ILogger<T>`. It meant that in WebJobs/Functions we had to come up with a logger category that we passed to you -- which led to a bunch of other confusing issues for customers. Now you can access a logger by:
    - Injecting `ILogger<T>` into constructor
    - Using `GetLogger` extension method on context, which pulls the logger from the `InstanceServices` for that execution.
3. Internally, use `AsyncLocal` to flag logs that are coming from us, so we know they are System logs. **Note. I don't currently like this. I've commented in the PR and I'm working on a change to make this a more complete solution.** *Update: I've now taken care of this with the new `WorkerMessage.Define`. See note below*
4. Adds a logging scope with currently two properties (that we'll document like gRPC does). These provide context for any logging inside of an execution.


src/DotNetWorker/Diagnostics/FunctionInvocationScope.cs
```diff
@@ -0,0 +1,61 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Microsoft.Azure.Functions.Worker.Diagnostics
+{
+    internal class FunctionInvocationScope : IReadOnlyList<KeyValuePair<string, object>>
```
brettsam, line 10: I took this design from here: https://github.com/grpc/grpc-dotnet/blob/ec6de5c047/src/Grpc.Net.Client/Internal/GrpcCallScope.cs

src/DotNetWorker/Diagnostics/WorkerLoggerT.cs
```diff
@@ -0,0 +1,36 @@
...
+namespace Microsoft.Azure.Functions.Worker.Diagnostics
+{
+    internal class WorkerLogger<T> : ILogger<T>
+    {
+        private readonly WorkerLogger _inner;
+        private static Assembly _thisAssembly = typeof(WorkerLogger<>).Assembly;
+
+        public WorkerLogger(ILoggerFactory loggerFactory)
+        {
+            bool isSystemAssembly = typeof(T).Assembly == _thisAssembly;
```
brettsam, line 17: I'm not really loving this. It means that:
1. If someone were to create a `ILogger<HttpRequestData>` (or any of our types) for some reason, it'd suddenly be treated as a system log. 
2. If someone were to overwrite our `ILogger<>` service registration, we'd stop having system logs. We want to get out of the way of this as much as possible as it's bitten us before.
 
I have a different plan for this that'll be a little overhead for us logging internally but would eliminate both of these issues. Will update this PR.

src/DotNetWorker/Diagnostics/WorkerMessage.cs
```diff
@@ -0,0 +1,70 @@
...
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    internal static class WorkerMessage
+    {
+        private static readonly AsyncLocal<bool> _isSystemLog = new AsyncLocal<bool>();
+
+        internal static bool IsSystemLog => _isSystemLog.Value;
+
+        public static Action<ILogger, Exception?> Define(LogLevel logLevel, EventId eventId, string formatString)
```
brettsam, line 16: The new approach to making sure that we are flagging worker (system) logs internally:
1. Do not use `logger.Log____`. Instead, always use `WorkerMessage.Define` for logs, which is a wrapper around `LoggerMessage.Define`.
2. I want to follow up with an analyzer to make sure we're doing this everywhere in the future (we currently have almost no system logging).
3. If this gets missed -- the log will still be written, but it won't be treated as a system log.
4. Yes, it's a tiny little bit extra work on our side -- but it takes care of the issues I had with the other approach.
5. If we need to expand the overloads for `WorkerMessage` with more parameters... we can just add them here as needed.

src/DotNetWorker/RpcExtensions.cs
```diff
@@ -187,9 +187,9 @@ internal static TypedData ToRpcLongArray(this long[] arrLong)
 
             return new RpcException
             {
-                Message = exception.Message,
-                Source = exception.Source,
-                StackTrace = exception.StackTrace
+                Message = $"{exception.GetType()}: {exception.Message}",
```
brettsam, line 190: There was no way to tell the type of the exception... so I'm formatting the message like this to include those details. Maybe I should just do `exception.ToString()` here?

src/DotNetWorker/Diagnostics/FunctionInvocationScope.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        internal const string FunctionInvocationIdKey = "AzureFunctionInvocationId";
+        internal const string FunctionNameKey = "AzureFunctionName";
```
mhoeger, line 13: AzureFunctions_InvocationId
AzureFunctions_FunctionName
?

samples/FunctionApp/Function4/Function4.cs
```diff
@@ -18,7 +18,8 @@ public static class Function4
         public static HttpResponseData Run([HttpTrigger(AuthorizationLevel.Anonymous, "get", "post", Route = null)] HttpRequestData req,
             FunctionExecutionContext executionContext)
         {
-            var logger = executionContext.Logger;
+            var logger = executionContext.GetLogger(("FunctionApp.Function4"));
```
fabiocav, line 21: nit: Add more parenthesis here :)

src/DotNetWorker/Broker/FunctionBroker.cs
```diff
@@ -47,54 +53,70 @@ public async Task<InvocationResponse> InvokeAsync(FunctionInvocation invocation)
...
+                    (executionContext as IDisposable)?.Dispose();
+                }
             }
 
             return response;
         }
-    }
 
+        private static class Log
+        {
```
fabiocav, line 111: Should we consider moving this to a partial with `<file>.Log.cs` like in the host?(https://github.com/Azure/azure-functions-host/blob/13a43aaa7d7f8441a51233083e692cc584f49143/src/WebJobs.Script.WebHost/Middleware/HostAvailabilityCheckMiddleware.Log.cs)

src/DotNetWorker/Converters/ExactMatchConverter.cs
```diff
@@ -13,6 +13,15 @@ public bool TryConvert(ConverterContext context, out object? target)
                 return true;
             }
 
+            context.Parameter.Type.IsAssignableFrom(context.Source?.GetType());
+
+            // Special handling for the context.
+            if (context.Parameter.Type == typeof(FunctionExecutionContext))
```
fabiocav, line 19: Break this out into a dedicated converter?

src/DotNetWorker/Diagnostics/GrpcFunctionsHostLogger.cs
```diff
@@ -2,47 +2,75 @@
...
         {
-            return EmptyDisposable.Instance;
+            // The built-in DI wire-up guarantees that scope provider will be set.
+            return _scopeProvider!.Push(state);
         }
 
         public bool IsEnabled(LogLevel logLevel)
         {
-            return true;
+            return logLevel != LogLevel.None;
```
fabiocav, line 38: We should probably have a todo here as we'll eventually receive the log level from the host.

src/DotNetWorker/Diagnostics/GrpcFunctionsHostLoggerProvider.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Threading.Channels;
```
fabiocav, line 4: Move gRPC types to a different folder?

src/DotNetWorker/RpcExtensions.cs
fabiocav: Yeah, I think that would be preferable (actually, just exception)

src/DotNetWorker/Diagnostics/GrpcFunctionsHostLogger.cs
brettsam: I don't think that will change anything here, though. My idea was to add a Filter that uses that (which happens outside of this call).

fabiocav (approved):

PR #135, opened 2021-02-17
E2E Http Tests

mhoeger:
Resolves https://github.com/Azure/azure-functions-dotnet-worker/issues/85

### DIFF
--- a/test/E2ETests/E2ETests/HttpEndToEndTests.cs
+++ b/test/E2ETests/E2ETests/HttpEndToEndTests.cs
@@ -1,0 +1,73 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Newtonsoft.Json.Linq;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Microsoft.Azure.Functions.Tests.E2ETests
+{
+    [Collection(Constants.FunctionAppCollectionName)]
+    public class HttpEndToEndTests
+    {
+        private readonly FunctionAppFixture _fixture;
+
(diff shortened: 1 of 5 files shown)


test/E2ETests/E2ETests/HttpEndToEndTests.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            HttpResponseMessage response = await HttpHelpers.InvokeHttpTrigger("HttpTriggerSetsCookie");
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            List<string> cookies = response.Headers.SingleOrDefault(header => header.Key == "Set-Cookie").Value.ToList();
+        }
+
+        [Fact (Skip = "TODO: https://github.com/Azure/azure-functions-dotnet-worker/issues/134")]
+        public async Task HttpTriggerBindingDataTests()
+        {
+
```
brettsam, line 70: Can you add the test here (even if it fails) so we can see what it's supposed to do?

test/E2ETests/E2ETests/HttpEndToEndTests.cs
```diff
@@ -0,0 +1,73 @@
...
+            string responseBody = await response.Content.ReadAsStringAsync();
+
+            Assert.Equal(expectedStatusCode, response.StatusCode);
+            Assert.Equal(expectedBody, responseBody);
+        }
+
+        [Fact(Skip = "TODO: https://github.com/Azure/azure-functions-dotnet-worker/issues/133")]
+        public async Task HttpTriggerWithCookieTests()
+        {
+            HttpResponseMessage response = await HttpHelpers.InvokeHttpTrigger("HttpTriggerSetsCookie");
```
brettsam, line 62: I don't see this function -- is that b/c the functionality to write it doesn't exist yet?

brettsam (approved):
A couple of questions on the skipped tests... but otherwise this is great. We should get it in and then fix those bugs!

PR #136, opened 2021-02-17
Explicit reference new package in devpack.ps1

ankitkumarr:


tools/devpack.ps1
```diff
@@ -1,10 +1,38 @@
...
+    Get-ChildItem "$packagePath" -Name |
+    Foreach-Object {
+        if ($_ -Match "$packageName")
+        {
+            # Converts the nupkg file name to a version
+            $currentVersion = $_ -replace "$packageName.", ""
+            $currentVersion = $currentVersion -replace ".nupkg", ""
+
+            # Gets the highest version
+            if ($currentVersion -gt $packageVersion)
```
brettsam, line 21: Is this doing string comparison? There's some weirdness with that where something like `1.0` will be sorted as higher than `10.0`. I believe this package gives you a way to sort with nuget's semantic versioning: https://www.nuget.org/packages/NuGet.Versioning.

It's probably not a big deal, but when it sorts incorrectly locally it'll be really confusing.

ankitkumarr:
Closing as @brettsam handled this scenario in his changes recently in a better and simpler way -- https://github.com/Azure/azure-functions-dotnet-worker/commit/ac6e7034c335226eb6a816d2c4fc9f29ac24e184

PR #137, opened 2021-02-17
Renaming FunctionExecutionContext

### LINKED ISSUE #107: Rename FunctionExecutionContext -> FunctionContext
This context is meant to be similar to `HttpContext`. And since `ExecutionContext` already has another meaning, the current name seems confusing. We think that `FunctionContext` makes the most sense.

fabiocav:
Resolves #107


src/DotNetWorker/Converters/Converter/DefaultConverterContext.cs
```diff
@@ -19,6 +19,6 @@ public DefaultConverterContext(FunctionParameter parameter, object? source, Func
 
         public override FunctionParameter Parameter { get; set; }
 
-        public override FunctionExecutionContext ExecutionContext { get; set; }
+        public override FunctionContext ExecutionContext { get; set; }
```
ankitkumarr, line 22: Should this property be called `Context` instead of `ExecutionContext`.

src/DotNetWorker/Context/FunctionContext.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Azure.Functions.Worker.Context;
+using Microsoft.Azure.Functions.Worker.Logging;
+
+namespace Microsoft.Azure.Functions.Worker.Pipeline
```
brettsam, line 9: Should we remove "Pipeline" from this namespace?

src/DotNetWorker/Converters/Converter/DefaultConverterContext.cs
fabiocav: Good catch! Will update.

src/DotNetWorker/Context/FunctionContext.cs
fabiocav: Will do.

ankitkumarr (approved):

PR #138, opened 2021-02-17
removing System.Text.Json dependency and instead using Azure.Core.Serialization.ObjectSerializer

### LINKED ISSUE #83: Change JSON serialization to use ObjectSerializer
See example here -- we should use `ObjectSerializer` from the Azure SDK to abstract away serialization technology. This likely should happen for GA because it could impact public API and slightly change behavior.

See here for an example: https://github.com/Azure/azure-sdk-for-net/blob/ef4430baed/sdk/core/Microsoft.Azure.Core.NewtonsoftJson/README.md. We'd likely add a `Serializer` property to our Options.

brettsam:
Fixes #83

### DIFF
--- a/src/DotNetWorker/Configuration/ServiceCollectionExtensions.cs
+++ b/src/DotNetWorker/Configuration/ServiceCollectionExtensions.cs
@@ -23,8 +23,13 @@ namespace Microsoft.Extensions.DependencyInjection
 {
     public static class ServiceCollectionExtensions
     {
-        public static IFunctionsWorkerApplicationBuilder AddFunctionsWorker(this IServiceCollection services, Action<WorkerOptions> configure)
+        public static IFunctionsWorkerApplicationBuilder AddFunctionsWorker(this IServiceCollection services, Action<WorkerOptions>? configure = null)
         {
+            if (services == null)
+            {
+                throw new ArgumentNullException(nameof(services));
+            }
+
             // Converters
             services.RegisterDefaultConverters();
 
--- a/src/DotNetWorker/Configuration/ServiceCollectionExtensions.cs
+++ b/src/DotNetWorker/Configuration/ServiceCollectionExtensions.cs
@@ -76,6 +76,11 @@ namespace Microsoft.Extensions.DependencyInjection
                     config.Bind(arguments);
                 });
 
+            if (configure != null)
+            {
+                services.Configure(configure);
+            }
+
             return new FunctionsWorkerApplicationBuilder(services);
         }
 
--- a/src/DotNetWorker/Configuration/WorkerOptions.cs
+++ b/src/DotNetWorker/Configuration/WorkerOptions.cs
@@ -1,29 +1,27 @@
-// Copyright (c) .NET Foundation. All rights reserved.
+﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
-using System.Collections.Generic;
+using System.Text.Json;
+using Azure.Core.Serialization;
 
-namespace Microsoft.Azure.Functions.Worker.Configuration
+namespace Microsoft.Azure.Functions.Worker
 {
-    // TODO: Slim this down.
     public class WorkerOptions
     {
         public WorkerOptions()
         {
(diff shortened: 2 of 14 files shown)


src/DotNetWorker/Configuration/ServiceCollectionExtensions.cs
```diff
@@ -68,6 +68,11 @@ public static IFunctionsWorkerApplicationBuilder AddFunctionsWorker(this IServic
                     config.Bind(arguments);
                 });
 
+            if (configure != null)
+            {
+                services.Configure(configure);
```
fabiocav, line 73: nit: `services?.Configure(configure)`

src/DotNetWorker/Configuration/WorkerOptions.cs
```diff
@@ -1,29 +1,27 @@
...
-namespace Microsoft.Azure.Functions.Worker.Configuration
+namespace Microsoft.Azure.Functions.Worker
 {
-    // TODO: Slim this down.
     public class WorkerOptions
     {
         public WorkerOptions()
         {
-            Capabilities = new List<string>();
+            var serializerOptions = new JsonSerializerOptions
```
fabiocav, line 13: Is there a reason we're setting this? Expectations from our internal protocol?

src/DotNetWorker/Converters/JsonPocoConverter.cs
```diff
@@ -1,19 +1,23 @@
...
     internal class JsonPocoConverter : IConverter
     {
-        private readonly IOptions<JsonSerializerOptions> _options;
+        private readonly ObjectSerializer _serializer;
 
-        public JsonPocoConverter(IOptions<JsonSerializerOptions> options)
+        public JsonPocoConverter(IOptions<WorkerOptions> options)
         {
-            _options = options ?? throw new ArgumentNullException(nameof(options));
+            _serializer = options?.Value?.Serializer ?? throw new ArgumentNullException(nameof(options));
```
fabiocav, line 20: The exception only applies if the parameter is null, for the properties, we should throw a more descriptive exception.

src/DotNetWorker/Configuration/WorkerOptions.cs
```diff
@@ -1,29 +1,27 @@
...
+            {
+                PropertyNameCaseInsensitive = true
+            };
 
-            // Default value for HeartBeatRateFactor is 6
-            HeartBeatRateFactor = 6;
+            Serializer = new JsonObjectSerializer(serializerOptions);
```
fabiocav, line 18: Should this be deferred? If a different serializer is used, we shouldn't need to create this instance, right?

src/DotNetWorker/Configuration/WorkerOptions.cs
brettsam: I debated this one -- I don't think anything internally matters here. But I was trying to pick the biggest difference between Json.NET and System.Text.Json. In retrospect, it makes sense (and it's easier to explain) if we just use the default here.

src/DotNetWorker/Configuration/ServiceCollectionExtensions.cs
fabiocav: Assuming this didn't spark joy...

src/DotNetWorker/Configuration/WorkerOptions.cs
```diff
@@ -1,29 +1,23 @@
...
-            // Default value for HeartBeatRateFactor is 6
-            HeartBeatRateFactor = 6;
+        /// <summary>
+        /// The <see cref="ObjectSerializer"/> to use for all JSON serialization and deserialization. By default,
+        /// this is a default <see cref="JsonObjectSerializer"/> with default <see cref="JsonSerializerOptions"/>.
+        /// </summary>
+        public ObjectSerializer Serializer
+        {
+            get => _serializer ??= new JsonObjectSerializer();
+            set => _serializer = value;
```
fabiocav, line 20: Should we perform a `null` check on the value? Do we ever want to allow this to be set to `null`?

fabiocav (approved):
Two small comments here, but looks good, so flagging as approved.

PR #139, opened 2021-02-17
Add new extensions, and samples

ankitkumarr:
I have added the following extensions --
1. EventGrid
2. ServiceBus
3. EventHubs
4. SignalRService
5. RabbitMQ
6. Kafka
7. Warmup

In addition, I have created a `SampleApp` that uses each of these extension.
I do need to test these end to end to verify each of these extensions work as expected.


samples/SampleApp/Properties/serviceDependencies.local.json
```diff
@@ -0,0 +1,8 @@
+{
+  "dependencies": {
+    "storage1": {
+      "type": "storage.emulator",
+      "connectionId": "AzureWebJobsStorage"
+    }
+  }
+}
```
brettsam, line 8: What are these? Did these just get added by VS?

brettsam (approved):
Just one minor comment...

samples/SampleApp/Properties/serviceDependencies.local.json
ankitkumarr: I am not sure tbh. Even our `FunctionApp` sample has that. It might have been because I copied from there. Let me remove them from here and see.

PR #147, opened 2021-02-19
adding blob and storage e2e; fixing related bugs

### LINKED ISSUE #87: E2E tests: Blob

### LINKED ISSUE #88: E2E tests: Queue

brettsam:
Fixes #87 
Fixes #88 

Notable changes to product:
- `BlobTrigger` passes us `ByteString`, so we now support conversions from `ReadOnlyMemory<byte>` to `string`, `byte[]` or POCO.
- E2E tests in other languages support `byte[]` as a supported parameter type, so now support conversions from `string` to `byte[]` (explicitly needed for blob tests)
- Trigger metadata now available as a source for conversions (some e2e tests use this).


src/DotNetWorker/Context/GrpcValueProvider.cs
```diff
@@ -1,42 +1,60 @@
...
     {
-        private readonly IDictionary<string, ParameterBinding> _inputData;
+        private readonly IDictionary<string, TypedData> _inputData;
+        private readonly IDictionary<string, TypedData> _triggerMetadata;
 
-        public GrpcValueProvider(IEnumerable<ParameterBinding> inputData)
+        public GrpcValueProvider(IEnumerable<ParameterBinding> inputData, MapField<string, TypedData> triggerMetadata)
         {
-            if (inputData is null)
+            if (inputData == null)
```
ankitkumarr, line 19: Did something prompt this change? Curious, as I have been using `is` around for the null check.

ankitkumarr (approved):

src/DotNetWorker/Context/GrpcValueProvider.cs
brettsam: no, is is fine... i just saw an is and an == right next to each other.... I believe Fabio meant that we can start using is.

PR #150, opened 2021-02-20
HTTP Request and Response updates

fabiocav:


src/DotNetWorker/Grpc/GrpcHttpResponseData.cs
```diff
@@ -0,0 +1,80 @@
...
+            }
+
+            _rpcHttpConsumed = true;
+
+            foreach (var header in _headers)
+            {
+                _rpcHttp.Headers.Add(header.Key, string.Join(",", header.Value));
+            }
+
+            _rpcHttp.StatusCode = StatusCode.ToString("d");
```
mhoeger, line 64: I'm not sure what this is doing here so just flagging to make sure it's intentional!

src/DotNetWorker/Http/HttpCookie.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        public string Name { get; set; }
+
+        public string? Path { get; set; }
+
+        public SameSite SameSite { get; set; }
+
+        public bool? Secure { get; set; }
+
+        public string Value { get; set; }
```
mhoeger, line 32: Same as earlier comment: `string? Value`?

src/DotNetWorker/Http/HttpResponseDataExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+            WriteString(response, value, Encoding.UTF8);
+        }
+
+        /// <summary>
+        /// Writes the provided string to the response body using the specified encoding.
+        /// </summary>
+        /// <param name="response">The response to write the string to.</param>
+        /// <param name="value">The string content to write to the request body.</param>
+        /// <param name="encoding">The encoding to use when writing the string.</param>
+        public static void WriteString(this HttpResponseData response, string value, Encoding encoding)
```
mhoeger, line 29: Wondering why reading the body through HttpRequestDataExtensions has an Async flavor but the write doesn't?

src/DotNetWorker/Http/HttpResponseData.cs
```diff
@@ -1,25 +1,34 @@
...
+        /// <summary>
+        /// Gets or sets the response body stream.
+        /// </summary>
+        public abstract Stream Body { get; set; }
 
-        public Dictionary<string, string> Headers { get; set; }
+        /// <summary>
+        /// Gets an <see cref="HttpCookies"/> instance containing the request cookies.
+        /// </summary>
+        public abstract HttpCookies Cookies { get; }
```
mhoeger, line 32: I think I remember talks about having some sort of static "CreateResponse" factory method attached here too in case folks tried to instantiate directly from HttpResponseData. Should that be included here?

src/DotNetWorker/Converters/ConverterMiddleware.cs
```diff
@@ -19,6 +19,9 @@ public ConverterMiddleware(IEnumerable<IConverter> converters)
 
         public Task Invoke(FunctionContext context, FunctionExecutionDelegate next)
         {
+            // TODO: The value needs to be moved to the context
+            // parameters values should be properly associated with the context
+            // and support disposal.
```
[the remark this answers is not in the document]
fabiocav: It's a bit odd... if you look at the remarks for that documentation, it states that the value cannot be `null` (should be an empty string if not set). The exceptions documentation there also states that a null ref exception will be thrown if the value is null 🤷🏼‍♂️

src/DotNetWorker/Grpc/GrpcHttpCookies.cs
```diff
@@ -0,0 +1,55 @@
...
+    internal class GrpcHttpCookies : HttpCookies
+    {
+        private RepeatedField<RpcHttpCookie> _cookies;
+
+        public GrpcHttpCookies(RepeatedField<RpcHttpCookie> cookies)
+        {
+            _cookies = cookies;
+        }
+
+        public override void Append(string name, string value)
```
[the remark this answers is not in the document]
fabiocav: It's a bit odd... if you look at the remarks for that documentation, it states that the value cannot be `null` (should be an empty string if not set). The exceptions documentation there also states that a null ref exception will be thrown if the value is null 🤷🏼‍♂️

src/DotNetWorker/Grpc/GrpcHttpResponseData.cs
fabiocav: Intentional, but is there a problem I might be missing?

src/DotNetWorker/Grpc/GrpcHttpCookies.cs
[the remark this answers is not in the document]
mhoeger: @fabiocav - I think from the code that null will become empty string by default: https://github.com/dotnet/runtime/blob/4893645536af64b01884ce5f3c392ad63513d04b/src/libraries/System.Net.Primitives/src/System/Net/Cookie.cs#L669

Found a related PR here: https://github.com/dotnet/corefx/pull/8206 (which resolves [this issue](https://github.com/dotnet/runtime/issues/17156))

And an issue on [the docs](https://github.com/dotnet/runtime/issues/16891) saying it will throw for null, which appears to be resolved but the docs look the same? :|

src/DotNetWorker/Http/HttpResponseData.cs
fabiocav: Yeah, I wanted to add that, so thanks for the reminder. Done.

src/DotNetWorker/Http/HttpResponseDataExtensions.cs
fabiocav: Just hadn't added :). It's there now

src/DotNetWorker/Http/HttpCookie.cs
fabiocav: Same comment as above..... not sure we should add...

src/DotNetWorker/Grpc/GrpcHttpResponseData.cs
mhoeger: Nvm - Sorry was getting mixed up with which "StatusCode" named properties :) this makes sense!

src/DotNetWorker/Grpc/GrpcHttpCookies.cs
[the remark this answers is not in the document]
fabiocav: Yeah, looks like a doc issue.

PR #160, opened 2021-02-24
Adding binding feature. Fixing issue with value binding scope

fabiocav:


src/DotNetWorker/Context/Features/IBindingFeature.cs
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Functions.Worker.Context.Features
+{
+    internal interface IBindingFeature
```
brettsam, line 11: We just talked about an `IFunctionBindingsFeature` yesterday. Should we call this `IModelBindingFeature` or something else to disambiguate? I know you don't like "Converters" :-)

src/DotNetWorker/Definition/FunctionParameter.cs
```diff
@@ -5,19 +5,30 @@
...
         public string Name { get; }
 
+        /// <summary>
+        /// Gets the parameter <see cref="System.Type"/>.
+        /// </summary>
         public Type Type { get; }
-
-        // TODO: Pop out to Context (or Invocation)
-        public object? Value { get; set; }
     }
```
brettsam, line 33: Do you want to add Properties to this now? I just remembered one use for this -- we can stick the GrpcType in here (i.e. JSON/Byte[]) so conversion is quicker.

src/DotNetWorker/Invocation/DefaultFunctionExecutor.Log.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        private static class Log
+        {
+            private static readonly Action<ILogger, string, string, Exception?> _functionBindingFeatureUnavailable =
+                WorkerMessage.Define<string, string>(LogLevel.Warning, new EventId(2, nameof(FunctionBindingFeatureUnavailable)),
+                    "The feature " + nameof(IBindingFeature) + " was not available for invocation '{invocationId}' of function '{functionName}'. Unable to process input bindings.");
+
+            public static void FunctionBindingFeatureUnavailable(ILogger<DefaultFunctionExecutor> logger, FunctionContext context)
+            {
+                _functionBindingFeatureUnavailable(logger, context.Invocation.InvocationId, context.Invocation.FunctionId, null);
```
brettsam, line 20: You're passing the `FunctionId` (which is a guid from the host) here. `Name` is available on `FunctionMetadata`. I also now think that `FunctionId` should move there -- or even directly on `FunctionDefinition` b/c this is something that does not change per-invocation and would reduce some confusion with `InvocationId`

src/DotNetWorker/Invocation/DefaultFunctionExecutor.cs
```diff
@@ -2,18 +2,39 @@
...
+        {
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
         public async Task ExecuteAsync(FunctionContext context)
         {
             var invoker = context.FunctionDefinition.Invoker;
             object? instance = invoker.CreateInstance(context.InstanceServices);
-            object? result = await invoker.InvokeAsync(instance, context.FunctionDefinition.Parameters.Select(p => p.Value).ToArray());
+            var bindingFeature = context.Features.Get<IBindingFeature>();
```
brettsam, line 24: Okay yeah, I like this better.

brettsam (approved):
I've got a few comments -- some can be follow-ups though.

src/DotNetWorker/Invocation/DefaultFunctionExecutor.cs
```diff
@@ -2,18 +2,39 @@
...
+
+            object?[] inputArtuments;
+            if (bindingFeature is null)
+            {
+                Log.FunctionBindingFeatureUnavailable(_logger, context);
+                inputArtuments = new object?[context.FunctionDefinition.Parameters.Length];
+            }
+            else
+            {
+                inputArtuments = bindingFeature.BindFunctionInput(context);
```
ankitkumarr, line 34: `inputArtuments`- `inputArguments`

src/DotNetWorker/Context/Features/IBindingFeature.cs
fabiocav: I'll make the change now to avoid confusion. Still keeping those types internal, so we can go through another API and naming review at a later time.

src/DotNetWorker/Definition/FunctionParameter.cs
fabiocav: Would this require the converters to be aware of that type? Or are we planning on having gRPC specific converters (which would make sense)?

fabiocav: Adding the properties. Will just have that dictionary to begin with, we can talk about potential optimizations in the future

PR #161, opened 2021-02-24
Update proto code generation. Access modifier and namespace changes

fabiocav:
Addressing one of the API review items.

### DIFF
--- a/src/DotNetWorker/DotNetWorker.csproj
+++ b/src/DotNetWorker/DotNetWorker.csproj
@@ -27,4 +27,10 @@
   <ItemGroup>
      <Protobuf Include=".\protobuf\src\proto\**\*.proto" ProtoRoot=".\protobuf\src\proto" />
   </ItemGroup>
+
+  <ItemGroup>
+    <Protobuf Update=".\protobuf\src\proto\FunctionRpc.proto" Access="Internal" />
+    <Protobuf Update=".\protobuf\src\proto\identity\ClaimsIdentityRpc.proto" Access="Internal" />
+    <Protobuf Update=".\protobuf\src\proto\shared\NullableTypes.proto" Access="Internal" />
+  </ItemGroup>
 </Project>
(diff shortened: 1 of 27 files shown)


src/DotNetWorker/DotNetWorker.csproj
```diff
@@ -27,4 +27,10 @@
   <ItemGroup>
      <Protobuf Include=".\protobuf\src\proto\**\*.proto" ProtoRoot=".\protobuf\src\proto" />
   </ItemGroup>
+
+  <ItemGroup>
+    <Protobuf Update=".\protobuf\src\proto\FunctionRpc.proto" Access="Internal" />
```
brettsam, line 32: If we use `GrpcServices="Client"` does that get us less code-gen? Looks like by default it generates Server and Client...

src/DotNetWorker/protobuf/src/proto/FunctionRpc.proto
```diff
@@ -4,7 +4,7 @@ syntax = "proto3";
 option java_multiple_files = true;
 option java_package = "com.microsoft.azure.functions.rpc.messages";
 option java_outer_classname = "FunctionProto";
-option csharp_namespace = "Microsoft.Azure.WebJobs.Script.Grpc.Messages";
+option csharp_namespace = "Microsoft.Azure.Functions.Worker.Grpc.Messages";
```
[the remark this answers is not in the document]
fabiocav: Unfortunately, manual, yes. The task didn't seem to support this and the fact that this is explicitly defined in the proto file. Will need to investigate some more, but worse case scenario, we can automate our proto file updates and make sure this is handled.

src/DotNetWorker/DotNetWorker.csproj
fabiocav: I debated changing, but didn't want to increase the scope of the changes. Figured we could have this as a follow up change after some additional validation.

PR #170, opened 2021-02-25
fixing some build warnings

brettsam:
Cleaning up some low-hanging fruit to get our warnings down a little. Also fixing the documentation file issue.

### DIFF
--- a/src/DotNetWorker/DotNetWorker.csproj
+++ b/src/DotNetWorker/DotNetWorker.csproj
@@ -8,7 +8,7 @@
     <AssemblyName>Microsoft.Azure.Functions.Worker</AssemblyName>
     <RootNamespace>Microsoft.Azure.Functions.Worker</RootNamespace>
     <SignAssembly>true</SignAssembly>
-    <DocumentationFile>$(OutputPath)\$(AssemblyName).xml</DocumentationFile>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <AssemblyOriginatorKeyFile>..\..\key.snk</AssemblyOriginatorKeyFile>
   </PropertyGroup>
   <ItemGroup>
--- a/src/DotNetWorker/Http/HttpRequestDataExtensions.cs
+++ b/src/DotNetWorker/Http/HttpRequestDataExtensions.cs
@@ -19,7 +19,7 @@ namespace Microsoft.Azure.Functions.Worker.Http
         /// </summary>
         /// <param name="request">The request from which to read.</param>
         /// <param name="encoding">The encoding to use when reading the string. Defaults to UTF-8</param>
-        /// <returns>A <see cref="Task{string?}"/> that represents the asynchronous read operation.</returns>
+        /// <returns>A <see cref="Task{String}"/> that represents the asynchronous read operation.</returns>
         public static async Task<string?> ReadAsStringAsync(this HttpRequestData request, Encoding? encoding = null)
         {
             if (request is null)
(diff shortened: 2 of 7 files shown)


src/DotNetWorker/DotNetWorker.csproj
```diff
@@ -8,7 +8,7 @@
     <AssemblyName>Microsoft.Azure.Functions.Worker</AssemblyName>
     <RootNamespace>Microsoft.Azure.Functions.Worker</RootNamespace>
     <SignAssembly>true</SignAssembly>
-    <DocumentationFile>$(OutputPath)\$(AssemblyName).xml</DocumentationFile>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
```
brettsam, line 11: Following the advice here -- https://docs.microsoft.com/en-us/dotnet/csharp/codedoc, which I *think* should work for us.

Turns out `$(OutputPath)` isn't always set here -- so you end up with `\....xml`, which resolves to c:\

fabiocav: Have you checked the result? It does get packaged, right?

src/DotNetWorker/Http/HttpRequestDataExtensions.cs
```diff
@@ -19,7 +19,7 @@ public static class HttpRequestDataExtensions
         /// </summary>
         /// <param name="request">The request from which to read.</param>
         /// <param name="encoding">The encoding to use when reading the string. Defaults to UTF-8</param>
-        /// <returns>A <see cref="Task{string?}"/> that represents the asynchronous read operation.</returns>
+        /// <returns>A task that represents the asynchronous read operation.</returns>
```
fabiocav, line 22: We should be able to have a generic type reference here, correct?

src/DotNetWorker/IFunctionActivator.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Microsoft.Azure.Functions.Worker
 {
-    public interface IFunctionActivator
+    internal interface IFunctionActivator
```
fabiocav, line 8: Do we expect this to change? Should it stay public?

src/DotNetWorker/Http/HttpRequestDataExtensions.cs
brettsam: I think you can do Task<T> -- but not Task<specifictype> -- I checked some other official docs and they just say "A task..." -- but let me look for other examples.

src/DotNetWorker/IFunctionActivator.cs
brettsam: I'd rather keep it internal by default and open it up in a follow-up.

src/DotNetWorker/DotNetWorker.csproj
brettsam: Yes, just verified.

fabiocav: 👍🏼

src/DotNetWorker/Http/HttpRequestDataExtensions.cs
fabiocav: Was this maybe just an issue with using the alias? (`string` vs `String`)?

src/DotNetWorker/IFunctionActivator.cs
fabiocav: Not opposed to that, but this is one of the key interfaces that I can see people needing to use

fabiocav (approved):
Approving, but if you can check on the Task comment, that would be good

src/DotNetWorker/Http/HttpRequestDataExtensions.cs
brettsam: oh... it was the `?` throwing it off. fixed it up.

PR #189, opened 2021-02-26
incorporating feedback from FunctionDefintion and Invoker API review

brettsam:


src/DotNetWorker/Definition/DefaultFunctionDefinition.cs
```diff
@@ -1,29 +1,32 @@
...
             OutputBindingsInfo = outputBindings;
         }
 
-        public override FunctionMetadata Metadata { get; set; }
+        public override FunctionMetadata Metadata { get; }
 
-        public override ImmutableArray<FunctionParameter> Parameters { get; set; }
+        public override ImmutableArray<FunctionParameter> Parameters { get; }
 
-        public override IFunctionInvoker Invoker { get; set; }
```
brettsam, line 25: The comments from the review said:
> Remove invoker. The middleware should create this if reached.

But creating the invoker requires the `MethodInfo` today. Without a major reshuffling, I created an `Items` dictionary here where we can store it without exposing the type. Thoughts? 

I could also refactor the IFunctionInvokerFactory so that it doesn't use a `MethodInfo` directly... I'll try that as well now and see what it looks like.

fabiocav: We can have the more elaborate change with a later update. Let's land the initial set of changes first. We definitely want to remove all reflection types from the abstractions (goal should be to have 100% non-reflection paths), but that can come in a  later change.

src/DotNetWorker/Invocation/DefaultFunctionExecutor.cs
```diff
@@ -19,7 +20,12 @@ public DefaultFunctionExecutor(ILogger<DefaultFunctionExecutor> logger)
 
         public async Task ExecuteAsync(FunctionContext context)
         {
-            var invoker = context.FunctionDefinition.Invoker;
+            if (!context.FunctionDefinition.Items.TryGetValue(DefaultFunctionDefinition.InvokerKey, out object? invokerObject) ||
+                invokerObject is not IFunctionInvoker invoker)
+            {
+                throw new InvalidOperationException("No function invoker found.");
```
fabiocav, line 26: Should we be more detailed in this exception? Maybe describe the item/key expectation here? For more involved things we can also link to a doc (wiki/other) if needed.

fabiocav (approved):

src/DotNetWorker/Definition/DefaultFunctionDefinition.cs
brettsam: I didn't like the fact that I had to add `Items` (when we may not need it later) -- so I have another approach in my next commit.

src/DotNetWorker/Invocation/DefaultFunctionExecutor.cs
```diff
@@ -10,16 +11,22 @@ namespace Microsoft.Azure.Functions.Worker.Invocation
...
         {
+            _invokerFactory = invokerFactory ?? throw new ArgumentNullException(nameof(invokerFactory));
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
         }
 
         public async Task ExecuteAsync(FunctionContext context)
         {
-            var invoker = context.FunctionDefinition.Invoker;
+            var invoker = _invokerCache.GetOrAdd(context.Invocation.FunctionId,
+                _ => _invokerFactory.Create(context.FunctionDefinition.Metadata));
```
brettsam, line 28: We want to make sure these exist 1:1 with functions we're running.

fabiocav (approved):

PR #190, opened 2021-02-26
removing FunctionMetadata and adding props to FunctionDefinition

brettsam:


test/DotNetWorkerTests/DotNetWorkerTests.csproj
```diff
@@ -22,6 +22,8 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\extensions\Worker.Extensions.Abstractions\Worker.Extensions.Abstractions.csproj" />
+    <ProjectReference Include="..\..\extensions\Worker.Extensions.Http\Worker.Extensions.Http.csproj" />
```
ankitkumarr, line 26: Did you need this reference? I may have missed it, but didn't see any types from here used in the test.

ankitkumarr (approved):
Looks good! One tiny comment.

test/DotNetWorkerTests/DotNetWorkerTests.csproj
brettsam: oh, good catch! we don't need it -- I had originally added it to make the test function, then realized i didn't even need it since we don't use it at runtime.

PR #197, opened 2021-03-01
Adding JSON request and response extension methods

fabiocav:


src/DotNetWorker/Http/HttpResponseDataExtensions.cs
```diff
@@ -63,6 +66,92 @@ public static Task WriteStringAsync(this HttpResponseData response, string value
...
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> used to cancel the operation.</param>
+        /// <returns>A <see cref="ValueTask"/> that represents the asynchronous operation.</returns>
+        public static ValueTask WriteAsJsonAsync<T>(this HttpResponseData response, T instance, CancellationToken cancellationToken = default)
+        {
+            return WriteAsJsonAsync(response, instance, "application/json; charset=utf-8", cancellationToken);
+        }
+
+        /// <summary>
+        /// Write the specified value as JSON to the response body using the default <see cref="ObjectSerializer"/> configured for this worker.
+        /// The response content-type will be set to <code>application/json; charset=utf-8</code> and the status code set to 200.
```
brettsam, line 85: This content-type note doesn't seem right here (since it can be set in a param).

brettsam (approved):
Just one small note on the doc comments.

src/DotNetWorker/Http/HttpResponseDataExtensions.cs
fabiocav: Copy and paste error :(

Issue #127, opened 2021-02-16
Add support for batch/cardinality


ankitkumarr:
We need to add `cardinality` to our function metadata (for bindings) based on the request parameter type.

mhoeger:
related: https://github.com/Azure/azure-functions-dotnet-worker/issues/115

Issue #115, opened 2021-02-10
Support array types as input data (string[], byte[][], long[], double[])


mhoeger:
Add support for rich types in arrays (ex: event hubs, cardinality many, binary data type array). Similar to: https://github.com/Azure/azure-functions-java-worker/issues/297

See: https://github.com/Azure/azure-functions-host/wiki/Out-of-Proc-Worker-Capabilities#typeddatacollection

PR #205, opened 2021-03-02
Getting batched events

mhoeger:
Resolves https://github.com/Azure/azure-functions-dotnet-worker/issues/127
Resolves https://github.com/Azure/azure-functions-dotnet-worker/issues/115

### SDK Changes
- For inputs, try to set "dataType" if we know the requested payload is `string` or `byte[]`
- For trigger attributes that implement `IBatchedOutput`, convert IsBatched property to add "cardinality": "many" or "cardinality": "one". 
  - Fail if we detect "cardinality": "many" but the requested payload DOES NOT implement IEnumerable 
  - Set "dataType" if we know that the requested payload inside of the collection is `string` or `byte[]`
- _Note that not all IEnumerables will be converted properly by .NET worker side (example: "Queue"). SDK shouldn't need to be changed though for future converter additions._ 

### .NET Worker Changes
- Enable capability to receive TypedData of:
  - collection bytes
  - collection string
  - collection double
  - collection long
- New converter which converts `RepeatedCollection<string>`, `RepeatedCollection<byte[]>`, `RepeatedCollection<ReadOnlyMemory<byte>>`, `RepeatedCollection<double>`, and `RepeatedCollection<long>` (from above capability) to corresponding Array types.
  - RepeatedCollection implements IEnumerable<T>, IList<T>, and ICollection<T>

### E2E App Changes
- Added some EventHubs functions I was using to test - not hooked up to actual tests yet since we can't emulate and the PR is already big...


fabiocav:
Was this replaced by the other PR? Should it be closed?

extensions/Worker.Extensions.EventHubs/EventHubTriggerAttribute.cs
```diff
@@ -5,15 +5,16 @@
...
 {
-    public sealed class EventHubTriggerAttribute : TriggerBindingAttribute
+    public sealed class EventHubTriggerAttribute : TriggerBindingAttribute, IBatchedInput
     {
         /// <summary>
         /// Create an instance of this attribute.
         /// </summary>
         /// <param name="eventHubName">Event hub to listen on for messages. </param>
-        public EventHubTriggerAttribute(string eventHubName)
+        public EventHubTriggerAttribute(string eventHubName, bool isBatched = true)
```
mhoeger, line 14: This is required for default - constructor name must match property name

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -378,14 +390,217 @@ private static ExpandoObject BuildBindingMetadataFromAttribute(CustomAttribute a
...
+            {
+                bindingDict["DataType"] = "Binary";
+            }
+
             foreach (var property in attribute.GetAllDefinedProperties())
             {
                 bindingDict.Add(property.Key, property.Value);
             }
 
+            // TODO: do not rely on property alone
```
mhoeger, line 409: In this approach, I do the conversion from public-facing "isBatched" to "cardinality": "many". This is mainly due to the mono.cecil limitations on available information. I don't know if there's a cleaner way without large changes

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -378,14 +390,217 @@ private static ExpandoObject BuildBindingMetadataFromAttribute(CustomAttribute a
...
+                        }
+                        else if (dataType.Equals(DataType.Binary))
+                        {
+                            bindingDict["DataType"] = "Binary";
+                        }
+                    }
+                    else
+                    {
+                        throw new Exception("Function is configured to process events in batches but parameter type is not iterable. " +
+                            $"Change parameter { parameterName ?? "type" } to be an IEnumerable type or set 'IsBatched' to false on your {attribute.AttributeType.Name.Replace("Attribute", "")} attribute.");
```
mhoeger, line 432: I will collect this before throwing

src/DotNetWorker/Converters/EnumerableConverter.cs
```diff
@@ -0,0 +1,95 @@
...
+                {
+                    target = GetTarget(enumerableDouble, targetType);
+                    return true;
+                }
+                else if (context.Source is IEnumerable<long> enumerableLong)
+                {
+                    target = GetTarget(enumerableLong, targetType);
+                    return true;
+                }
+            }
```
fabiocav, line 63:
```suggestion
            if (targetType is not null)
            {
                // Valid options from FunctionRpc.proto are string, byte, double and long collection
                target = context.Source switch
                {
                    IEnumerable<string> source => GetTarget(source, targetType),
                    IEnumerable<byte[]> source => GetTarget(source, targetType),
                    IEnumerable<double> source => GetTarget(source, targetType),
                    IEnumerable<long> source => GetTarget(source, targetType),
                    _ => null
                };
            }
```

src/DotNetWorker/Converters/EnumerableConverter.cs
```diff
@@ -0,0 +1,95 @@
...
+            {
+                case EnumerableTargetType.Array:
+                    return source.ToArray();
+                case EnumerableTargetType.HashSet:
+                    return source.ToHashSet();
+                case EnumerableTargetType.List:
+                    return source.ToList();
+                default:
+                    return null;
+            }
```
fabiocav, line 83:
```suggestion
       return targetType switch
            {
                EnumerableTargetType.Array => source.ToArray(),
                EnumerableTargetType.HashSet => source.ToHashSet(),
                EnumerableTargetType.List => source.ToList(),
                _ => null,
            };
```

src/DotNetWorker/Converters/EnumerableConverter.cs
fabiocav: In addition to the suggestion, assign the value of target to null at the beginning, return target != null at the end

src/DotNetWorker/Converters/EnumerableConverter.cs
```diff
@@ -0,0 +1,95 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Azure.Functions.Worker.Converters
+{
+    // Converting IEnumerable<> to Array
+    internal class EnumerableConverter : IConverter
```
fabiocav, line 11: Adding some suggestions to the class, but given the fact this is pretty specific to gRPC, I'm wondering if this shouldn't be a converted injected by the gRPC integration (mostly thinking about when we break it off) and operating on the gRPC types, which would make this a bit easier and rely less on type checking. We can chat, but wanted to share this to get your thoughts.

fabiocav (commented):
This looks great! Awesome set of tests. 

Leaving an initial set of comments, but some things may require a chat

extensions/Worker.Extensions.EventHubs/EventHubTriggerAttribute.cs
ankitkumarr: If I understand correctly, I am guessing you had to add this as right now we only get values that were defined during the instantiation (constructor args, and props) through `CustomAttributes`, and so if there is a static default value not declared here or during instantiation, it's not recorded.

I wonder if instead of doing it here, we should special case this one in sorts that -- if `IBatchedInput` is implemented, we go and look for the property `IsBatched` from the FunctionMetadataGenerator to get the default value set in the property.
I was looking through, out of curiosity (when I needed that), and looks like we could get constructor instructions to see the default value of static types. There might be a much simpler way. But here's how it looked like -- 
```csharp
CustomAttribute.AttributeType.Resolve().Methods.FirstOrDefault(m => m.Name == '.ctor').Body.Instructions.FirstOrDefault(i.Operand == <arg-type>).Previous.Operand
```

sdk/Sdk/FunctionMetadataGenerator.cs
mhoeger: ^exactly, and i was also having trouble getting a property that is set by default outside of the constructor, but i might need to double check on that!

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -378,14 +390,217 @@ private static ExpandoObject BuildBindingMetadataFromAttribute(CustomAttribute a
...
+
+                bindingDict.Remove("isBatched");
+            }
+
             return binding;
         }
 
+        private static bool IsIterableCollection(TypeReference type, out DataType dataType)
+        {
+            // Array and not byte array
```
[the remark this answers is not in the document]
mhoeger: Unfortunately no :( when we "Resolve" the type reference to a type definition, it just resolves the non-array element type. I tried removing this too and the `byte[][]` test fails. Interestingly, `string[]` doesn't fail, but only because string is an ienumerable<char>

extensions/Worker.Extensions.EventHubs/EventHubTriggerAttribute.cs
mhoeger: We dug through this and ran into some trouble with looking through the instructions for "body". Instead, for Attributes that require a default value, we're adding a new attribute and checking for that!

extensions/Worker.Extensions.EventHubs/EventHubTriggerAttribute.cs
```diff
@@ -30,5 +30,11 @@ public EventHubTriggerAttribute(string eventHubName)
         /// Gets or sets the optional app setting name that contains the Event Hub connection string. If missing, tries to use a registered event hub receiver.
         /// </summary>
         public string? Connection { get; set; }
+
+        /// <summary>
+        /// Gets or sets the configuration to enable batch processing of events. Default value is "true".
+        /// </summary>
+        [DefaultValue(true)]
+        public bool IsBatched { get; set; }
```
mhoeger, line 38: include default value of "true" here too for backwards compat when we remove the `[DefaultValue]` attribute

extensions/Worker.Extensions.Abstractions/IBatchedInput.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+﻿using System;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Abstractions
+{
+    public interface IBatchedInput
```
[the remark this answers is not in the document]
mhoeger: Will change this to ISupportCardinality which has a property named "Cardinality" with type enum (many/one)

src/DotNetWorker/Converters/EnumerableConverter.cs
```diff
@@ -0,0 +1,87 @@
...
+        {
+            EnumerableTargetType? targetType = null;
+            target = null;
+            // Array
+            if (context.Parameter.Type.IsArray)
+            {
+                targetType = EnumerableTargetType.Array;
+            }
+            // List or HashSet
+            else if (context.Parameter.Type.IsGenericType)
```
mhoeger, line 28: Don't support concrete list and hashset type => make this a an `ArrayConverter`

src/DotNetWorker/Converters/EnumerableConverter.cs
```diff
@@ -0,0 +1,87 @@
...
+        // Convert IEnumerable from common types. IEnumerable types will
+        // be converted by TypeConverter.
+        public bool TryConvert(ConverterContext context, out object? target)
+        {
+            EnumerableTargetType? targetType = null;
+            target = null;
+            // Array
+            if (context.Parameter.Type.IsArray)
+            {
+                targetType = EnumerableTargetType.Array;
```
mhoeger, line 25: Are array element types compatible? need to check.

src/DotNetWorker/Converters/EnumerableConverter.cs
```diff
@@ -0,0 +1,87 @@
...
+        private static Type HashSetType = typeof(HashSet<>);
+
+        // Convert IEnumerable from common types. IEnumerable types will
+        // be converted by TypeConverter.
+        public bool TryConvert(ConverterContext context, out object? target)
+        {
+            EnumerableTargetType? targetType = null;
+            target = null;
+            // Array
+            if (context.Parameter.Type.IsArray)
```
mhoeger, line 23: Use `ElementType` or `GetElementType` API

mhoeger: Add cache on function id and parameter (or add a TODO)

src/DotNetWorker/Converters/EnumerableConverter.cs
mhoeger: Check if source is IReadOnlyMemory vs. is byte[]

src/DotNetWorker/Context/Features/GrpcFunctionBindingsFeature.cs
```diff
@@ -99,6 +99,12 @@ public void SetOutputBinding(string name, object value)
                 // This is guaranteed to be Json here -- we can use that.
                 TypedData.DataOneofCase.Json => typedData.Json,
                 TypedData.DataOneofCase.Bytes => typedData.Bytes.Memory,
+                TypedData.DataOneofCase.CollectionBytes => typedData.CollectionBytes.Bytes.Select(element => {
+                    return element.Memory.ToArray();
```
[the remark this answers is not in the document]
mhoeger: ^yes, adding as followup per discussion!

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -378,14 +372,223 @@ private static ExpandoObject BuildBindingMetadataFromAttribute(CustomAttribute a
...
+        }
+
+        private static bool IsStringType(string fullName)
+        {
+            return string.Equals(fullName, Constants.StringType, StringComparison.Ordinal);
+        }
+
+        private static bool IsBinaryType(string fullName)
+        {
+            return string.Equals(fullName, Constants.ByteArrayType, StringComparison.Ordinal);
```
mhoeger, line 589: also check if type is IReadOnlyMemory<T>

mhoeger: Note: type is ReadOnlyMemory<T> 
proposal for IReadOnlyMemory<T>: https://github.com/dotnet/runtime/issues/32706

extensions/Worker.Extensions.Abstractions/DefaultValueAttribute.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    [AttributeUsage(AttributeTargets.Property)]
+    public class DefaultValueAttribute : Attribute
```
mhoeger, line 9: This is temporary to set default while we do not load attributes.

sdk/Sdk/CustomAttributeExtensions.cs
```diff
@@ -14,16 +14,40 @@ public static class CustomAttributeExtensions
...
             // argument names are equal to property names.
+            LoadDefaultProperties(properties, attribute);
             LoadConstructorArguments(properties, attribute);
             LoadDefinedProperties(properties, attribute);
 
             return properties;
         }
 
-        private static void LoadConstructorArguments(IDictionary<string, object> properties, CustomAttribute attribute)
+        private static IEnumerable<(string, CustomAttributeArgument?)> GetDefaultValues(this CustomAttribute attribute)
```
mhoeger, line 26: this is all should be removed later... after above todo is done: `avoid needing to instantiate any types, assume that the constructor argument names are equal to property names.`

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -378,14 +372,233 @@ private static ExpandoObject BuildBindingMetadataFromAttribute(CustomAttribute a
...
+            {
+                // Batching set to true
+                if (isBatched)
+                {
+                    bindingDict["Cardinality"] = "Many";
+                    // Throw if parameter type is *definitely* not a collection type.
+                    // Note that this logic doesn't dictate what we can/can't do, and
+                    // we can be more restrictive in the future because today some 
+                    // scenarios result in runtime failures.
+                    if (IsIterableCollection(parameterType, out DataType dataType))
```
mhoeger, line 415: Note that this should be changed, but likely for a future iteration it will be a runtime failure anyways, so no worries about introducing breaking behavior (it'll just fail faster)

src/DotNetWorker/Converters/ArrayConverter.cs
```diff
@@ -0,0 +1,63 @@
...
+            {
+                Type? elementType = context.Parameter.Type.GetElementType();
+                if (elementType is not null)
+                {
+                    // Ensure that we can assign from source to parameter type
+                    if (elementType.IsAssignableFrom(typeof(string))
+                        || elementType.IsAssignableFrom(typeof(byte[]))
+                        || elementType.IsAssignableFrom(typeof(ReadOnlyMemory<byte>))
+                        || elementType.IsAssignableFrom(typeof(long))
+                        || elementType.IsAssignableFrom(typeof(double)))
```
mhoeger, line 27: Note: should improve efficiency here... if anyone has quick suggestions they would be very welcome but otherwise i think i'll flag this as a future improvement

sdk/Sdk/FunctionMetadataGenerator.cs
mhoeger: Actually - something that is derived from `IEnumerable<Poco>` (ex: `List<Poco>`) is valid for cardinality: many and **does** have an appropriate converter. Would not want to throw in these cases.

The only failure cases are when element type is `string`, `byte[]`, `ReadOnlyMemory<byte>` `int`, `double`, `long` AND collection type is not equal to IEnumerable, IList, or ICollection (or corresponding generic).

We could go one step further and say if not derived from IEnumerable assume it's not batched... in this case we want to know if it was set explicitly or if it's just from the user... for event hubs we definitely want to discourage non-batched anyways

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -378,14 +372,233 @@ private static ExpandoObject BuildBindingMetadataFromAttribute(CustomAttribute a
...
+                || IsOrDerivedFrom(type, Constants.LookupGenericType)
+                || IsOrDerivedFrom(type, Constants.DictionaryGenericType);
+            if (isMappingEnumerable)
+            {
+                dataType = DataType.Undefined;
+                return false;
+            }
+
+            // IEnumerable and not string or dictionary
+            bool isEnumerableOfT = IsOrDerivedFrom(type, Constants.IEnumerableOfT);
```
fabiocav, line 468: nit: Since this may not be used until you reach the last check, could we defer this check?

src/DotNetWorker/Converters/ArrayConverter.cs
```diff
@@ -0,0 +1,63 @@
...
+                    {
+                        target = null;
```
fabiocav, line 29: nit: this assignment shouldn't be required

src/DotNetWorker/Converters/ArrayConverter.cs
```diff
@@ -0,0 +1,63 @@
...
+                        target = context.Source switch
+                        {
+                            IEnumerable<string> source => source.ToArray(),
+                            IEnumerable<ReadOnlyMemory<byte>> source => GetBinaryData(source, elementType!),
+                            IEnumerable<double> source => source.ToArray(),
+                            IEnumerable<long> source => source.ToArray(),
+                            _ => null
+                        };
+
+                        if (target is not null)
```
fabiocav, line 39: You can remove this check and let if fall through...

src/DotNetWorker/Converters/ArrayConverter.cs
```diff
@@ -0,0 +1,63 @@
...
+
+namespace Microsoft.Azure.Functions.Worker.Converters
+{
+    // Converting IEnumerable<> to Array
+    internal class ArrayConverter : IConverter
+    {
+        // Convert IEnumerable to array
+        public bool TryConvert(ConverterContext context, out object? target)
+        {
+            // Ensure requested type is an array
```
fabiocav, line 16: You can assign `target` to null at the beginning, which will help avoid the checks below

src/DotNetWorker/Converters/ArrayConverter.cs
```diff
@@ -0,0 +1,63 @@
...
+
+                        if (target is not null)
+                        {
+                            return true;
+                        }
+                    }
+                }
+            }
+
+            target = default;
```
fabiocav, line 47: once the assignment is done above, you can just `return target is not null`

src/DotNetWorker/Converters/ArrayConverter.cs
```diff
@@ -0,0 +1,63 @@
...
+        public bool TryConvert(ConverterContext context, out object? target)
+        {
+            // Ensure requested type is an array
+            if (context.Parameter.Type.IsArray)
+            {
+                Type? elementType = context.Parameter.Type.GetElementType();
+                if (elementType is not null)
+                {
+                    // Ensure that we can assign from source to parameter type
+                    if (elementType.IsAssignableFrom(typeof(string))
```
fabiocav, line 23: All of the comparisons here should be for equality, no need for `IsAssignableFrom`. Types like ReadOnlyMemory<byte>, for example, would only apply in exact matches.

test/DotNetWorkerTests/Converters/ArrayConverterTests.cs
```diff
@@ -0,0 +1,65 @@
...
+        private static readonly ReadOnlyMemory<byte> _sourceMemory = new ReadOnlyMemory<byte>(_sourceBytes);
+        private ArrayConverter _converter = new ArrayConverter();
+
+        private static readonly IEnumerable<ReadOnlyMemory<byte>> _sourceMemoryEnumerable = new RepeatedField<ReadOnlyMemory<byte>>() { _sourceMemory };
+        private static readonly RepeatedField<string> _sourceStringEnumerable = new RepeatedField<string>() { _sourceString };
+        private static readonly RepeatedField<double> _sourceDoubleEnumerable = new RepeatedField<double>() { 1.0 };
+        private static readonly RepeatedField<long> _sourceLongEnumerable = new RepeatedField<long>() { 2000 };
+
+        [Fact]
+        public void ConvertCollectionBytesToByteDoubleArray()
```
fabiocav, line 26: nit: `ConvertCollectionBytesToJaggedByteArray`?

test/DotNetWorkerTests/Converters/ArrayConverterTests.cs
```diff
@@ -0,0 +1,65 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Google.Protobuf.Collections;
+using Microsoft.Azure.Functions.Worker.Converters;
+using Xunit;
+
+namespace Microsoft.Azure.Functions.Worker.Tests.Converters
+{
+    public class ArrayConverterTests
```
fabiocav, line 13: Nice test coverage!

fabiocav (commented):
Added some comments, but nothing major. We should be able to get this in as I don't see major blockers, but some of the comments would be good to address.

sdk/Sdk/FunctionMetadataGenerator.cs
mhoeger: is there a fancy .net thing i can do? I use `isEnumerableOfT` a bit earlier in line 473 too

mhoeger:
Looks like `long[]` and `double[]` aren't used by host - updated release notes to reflect this and filed issue: https://github.com/Azure/azure-functions-host/issues/7209

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -378,14 +372,228 @@ private static ExpandoObject BuildBindingMetadataFromAttribute(CustomAttribute a
...
+                else
+                {
+                    return null;
+                }
+            }
+
+            TypeDefinition definition = type.Resolve();
+            if (definition.HasGenericParameters && type is GenericInstanceType genericType)
+            {
+                for (int i = 0; i < genericType.GenericArguments.Count(); i++)
```
ankitkumarr, line 547: For later -- Would it ever make sense to just look at the first argument, as we know `IEnumerable<T>` would always only have at most one generic arg?

Also this can be made `foreach`. As apparently, `ElementAt()` could go over the whole ienumerable every lookup depending on the implementation it says. Not that it matters much here :)

extensions/Worker.Extensions.Abstractions/DefaultValueAttribute.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.Functions.Worker.Annotations
```
ankitkumarr, line 6: I think dotnet doesn't use `Annotations` word. Maybe attributes? Or honestly, even if this is at `Microsoft.Azure.Functions.Worker.Extensions.Abstractions`, that's probably better than `Microsoft.Azure.Functions.Worker`, as users wouldn't ever have to use `...Extensions.Abstractions` namespace.

sdk/Sdk/CustomAttributeExtensions.cs
```diff
@@ -34,13 +57,13 @@ private static void LoadConstructorArguments(IDictionary<string, object> propert
...
 
-                if (paramName == null || paramValue == null)
+                if (paramName is null || paramValue is null)
                 {
                     continue;
                 }
 
                 paramValue = GetEnrichedValue(param!.ParameterType, paramValue);
-
+
```
ankitkumarr, line 66: accidental space

ankitkumarr (approved):
Looks great! All these tests make it so much easier. Thank you!!

I added very minor final nit comments, that don't need to be addressed in this PR. The only thing I will leave it to you is to change the namespace `Microsoft.Azure.Functions.Worker.Annotations`

extensions/Worker.Extensions.Abstractions/DefaultValueAttribute.cs
mhoeger: Good call - i also realized that the namespace `Microsoft.Azure.Functions.Worker` was a complete copy paste accident!

sdk/Sdk/FunctionMetadataGenerator.cs
mhoeger: ^This was done because you might have some Foo : Bar<K, V> : IEnumerable<V> 

So in this case the recursive search would want to match V from Bar (the second generic) with V from IEnumerable (the only generic)

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -378,14 +372,233 @@ private static ExpandoObject BuildBindingMetadataFromAttribute(CustomAttribute a
...
+            if (definition.BaseType is null)
+            {
+                return false;
+            }
+
+            TypeDefinition baseType = definition.BaseType.Resolve();
+            return IsDerivedFrom(baseType, interfaceFullName);
+        }
+
+        private static string? ResolveIEnumerableOfTType(TypeReference type, Dictionary<string, string> foundMapping)
```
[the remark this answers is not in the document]
mhoeger: Definitely a good call!!

sdk/Sdk/FunctionMetadataGenerator.cs
fabiocav: Flagging this as resolved for now, but as we discussed offline, we'll follow up with enhancements.

sdk/Sdk/FunctionMetadataGenerator.cs
fabiocav: Let's follow up on this case.

src/DotNetWorker/Converters/ArrayConverter.cs
```diff
@@ -0,0 +1,57 @@
...
+                    }
+                }
+            }
+
+            return target is not null;
+        }
+
+        private static object? GetBinaryData(IEnumerable<ReadOnlyMemory<byte>> source, Type targetType)
+        {
+            if (targetType.IsAssignableFrom(typeof(ReadOnlyMemory<byte>)))
```
fabiocav, line 47: Had left an earlier comment about the comparison using IsAssignableFrom, but can't find it. This check here isn't required, are any of the tests specifically testing this code path? Aside from `ReadOnlyMemory<byte>` this shouldn't evaluate to true, in practice.

sdk/Sdk/FunctionMetadataGenerator.cs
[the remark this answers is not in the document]
fabiocav: We should flag some of the logic using mono cecil for review. I noticed a few other things we may want to reevaluate, but these are much simpler to walk through with the tests.

PR #209, opened 2021-03-03
Cardinality many from array

mhoeger:
Related to: https://github.com/Azure/azure-functions-dotnet-worker/pull/205 , which implements cardinality many for anything that is an IEnumerable. This is scoped down to array types, which is a more simple rule to follow on when data type is batched vs. isn't.

Also appears consistent with the current WebJobs contract for "cardinality": "many"
https://github.com/Azure/azure-functions-host/blob/v3.0.15417/src/WebJobs.Script/Binding/GeneralScriptBindingProvider.cs#L72-L77

Resolves https://github.com/Azure/azure-functions-dotnet-worker/issues/127
Resolves https://github.com/Azure/azure-functions-dotnet-worker/issues/115

### SDK Changes
- For inputs, try to set "dataType" if we know the requested payload is `string` or `byte[]`
- Try set "cardinality": "many" if we know that the requested payload implements IEnumerable or IEnumerable Generic
  - Set "dataType" if we know that the requested payload inside of the collection is `string` or `byte[]`

### .NET Worker Changes
- Enable capability to receive TypedData of:
  - collection bytes
  - collection string
  - collection double
  - collection long
- New converter which converts `IEnumerable<string>`, `IEnumerable<byte[]>`, `IEnumerable<double>`, and `IEnumerable<long>` (from above capability) to Array type
- _Observation: not all types the user requests (ex: "Queue" which implements IEnumerable) will be properly converted, although "cardinality": "many" will be set. Potentially worth surfacing some sort of warning or error if we are unable to convert. From user perspective seems difficult to debug._

### DIFF
--- a/sdk/Sdk/FunctionMetadataGenerator.cs
+++ b/sdk/Sdk/FunctionMetadataGenerator.cs
@@ -23,6 +23,8 @@ namespace Microsoft.Azure.Functions.Worker.Sdk
         private const string VoidType = "System.Void";
         private const string ReturnBindingName = "$return";
         private const string HttpTriggerBindingType = "HttpTrigger";
+        private const string StringType = "System.String";
+        private const string ByteArrayType = "System.Byte[]";
 
         private readonly IndentableLogger _logger;
 
@@ -491,7 +493,7 @@ namespace Microsoft.Azure.Functions.Worker.Sdk
         private static bool IsIterableCollection(TypeReference type, out DataType dataType)
         {
             // Array and not byte array 
-            bool isArray = type.IsArray && !string.Equals(type.FullName, typeof(byte[]).FullName, StringComparison.Ordinal);
+            bool isArray = type.IsArray && !string.Equals(type.FullName, ByteArrayType, StringComparison.Ordinal);
             if (isArray)
             {
                 TypeSpecification? typeSpecification = type as TypeSpecification;
@@ -522,12 +524,12 @@ namespace Microsoft.Azure.Functions.Worker.Sdk
 
         private static bool IsStringType(string fullName)
         {
-            return string.Equals(fullName, typeof(string).FullName, StringComparison.Ordinal);
+            return string.Equals(fullName, StringType, StringComparison.Ordinal);
         }
 
         private static bool IsBinaryType(string fullName)
         {
-            return string.Equals(fullName, typeof(byte[]).FullName, StringComparison.Ordinal);
+            return string.Equals(fullName, ByteArrayType, StringComparison.Ordinal);
         }
 
         private enum DataType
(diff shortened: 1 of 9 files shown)


sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -462,5 +487,55 @@ private static bool TryGetBaseAttributeType(CustomAttribute attribute, string ba
...
+            {
+                return DataType.Binary;
+            }
+
+            return DataType.Undefined;
+        }
+
+        private static bool IsStringType(string fullName)
+        {
+            return string.Equals(fullName, typeof(string).FullName, StringComparison.Ordinal);
```
fabiocav, line 525: nit: but as an optimization, the typeof(<type>) name can be resolved and reused.

src/DotNetWorker/Converters/EnumerableConverter.cs
```diff
@@ -0,0 +1,74 @@
...
+            if (context.Parameter.Type.IsArray)
+            {
+                targetType = EnumerableTargetType.Array;
+            }
+
+            // Only apply if user is requesting an array, list, or hashset
+            if (targetType is not null)
+            {
+                // Valid options from FunctionRpc.proto are string, byte, double and long collection
+                if (context.Source is IEnumerable<string> enumerableString)
```
fabiocav, line 26: Would this be a little simpler using pattern matching on the source?

src/DotNetWorker/Converters/EnumerableConverter.cs
```diff
@@ -0,0 +1,74 @@
...
+                    return true;
+                }
+            }
+
+            target = default;
+            return false;
+        }
+
+        // Dictionary and Lookup not handled because we don't know 
+        // what they keySelector and elementSelector should be.
```
fabiocav, line 53: Can you clarify this?

src/DotNetWorker/Converters/EnumerableConverter.cs
```diff
@@ -0,0 +1,74 @@
...
+            switch (targetType)
+            {
+                case EnumerableTargetType.Array:
+                    return source.ToArray();
+                default:
+                    return null;
+            }
+        }
+
+        private enum EnumerableTargetType
```
ankitkumarr, line 65: Wondering if this is needed, as I see after the recent change to move to only Array type, this file only ever uses the `Array` enum value.

ankitkumarr (approved):
Looks great to me! Just a couple of optional nits.

mhoeger:
Closing in favor of - https://github.com/Azure/azure-functions-dotnet-worker/pull/205
We won't be doing any magic sniffing because array type != batched events

PR #213, opened 2021-03-03
moving all grpc refs out of IFunctionsApplication

brettsam:


src/DotNetWorker/GrpcWorker.cs
```diff
@@ -53,7 +115,6 @@ private Task ProcessRequestCoreAsync(StreamingMessage request)
                 MsgType.WorkerInitRequest => WorkerInitRequestHandlerAsync(request),
                 MsgType.FunctionLoadRequest => FunctionLoadRequestHandlerAsync(request),
                 MsgType.InvocationRequest => InvocationRequestHandlerAsync(request),
-                MsgType.FunctionEnvironmentReloadRequest => FunctionEnvironmentReloadRequestHandlerAsync(request),
```
mhoeger, line 56: I think we still want `FunctionEnvironmentReloadRequest` and its handler?

src/DotNetWorker/FunctionsApplication.cs
```diff
@@ -40,20 +38,18 @@ public FunctionContext CreateContext(IInvocationFeatures features)
...
+            var response = new InitializationResponse
             {
-                Result = new StatusResult { Status = Status.Success }
+                WorkerVersion = typeof(FunctionsApplication).Assembly.GetName().Version?.ToString()
             };
 
             response.Capabilities.Add("RpcHttpBodyOnly", bool.TrueString);
             response.Capabilities.Add("RawHttpBodyBytes", bool.TrueString);
             response.Capabilities.Add("RpcHttpTriggerMetadataRemoved", bool.TrueString);
             response.Capabilities.Add("UseNullableValueDictionaryForHttp", bool.TrueString);
```
mhoeger, line 52: All should be moved to GrpcWorker? Maybe instead of "InitializeAsync" there could be a WorkerVersion property that's used by GrpcWorker?

mhoeger (commented):
few comments

src/DotNetWorker/GrpcWorker.cs
brettsam: We weren't doing anything with it today. I believe @pragnagopa said that we didn't need to support it? I can re-add it though...

src/DotNetWorker/FunctionsApplication.cs
brettsam: That's a great point -- these capabilities are specifically for Grpc. I think moving this all to the worker would make sense (and not even exposing the WorkerVersion property anywhere... that seems like a Grpc implementation detail as well).

src/DotNetWorker/GrpcWorker.cs
brettsam: I added it back. Didn't want to get too crazy.

src/DotNetWorker/FunctionsApplication.cs
brettsam: Done.

fabiocav (approved):

src/DotNetWorker/FunctionsApplication.cs
```diff
@@ -40,20 +38,18 @@ public FunctionContext CreateContext(IInvocationFeatures features)
             return _functionContextFactory.Create(features);
         }
 
-        public Task<WorkerInitResponse> InitializeWorkerAsync(WorkerInitRequest request)
+        public Task<InitializationResponse> InitializeAsync()
```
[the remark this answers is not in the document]
brettsam: turns out i didn't even need this -- i deleted it

PR #216, opened 2021-03-03
Throw if no converter

mhoeger:
Instead of keeping the input as null if we are unable to convert it, we should throw (or at least warn) why the input is null.

Example:
```
[2021-03-03T07:32:17.585Z] Executed 'Functions.HelloFromQuery' (Failed, Id=e2f3400d-23a8-4a1e-8bee-b3268a652376, Duration=195ms)
[2021-03-03T07:32:17.590Z] System.Private.CoreLib: Exception while executing function: Functions.HelloFromQuery. System.Private.CoreLib: Result: Failure
Exception: System.InvalidCastException: Cannot convert input parameter 'req' for Function 'HelloFromQuery' to type 'System.String' from type 'Microsoft.Azure.Functions.Worker.GrpcHttpRequestData'.
```

### DIFF
--- a/src/DotNetWorker/Context/Features/DefaultModelBindingFeature.cs
+++ b/src/DotNetWorker/Context/Features/DefaultModelBindingFeature.cs
@@ -1,4 +1,4 @@
-﻿g// Copyright (c) .NET Foundation. All rights reserved.
+﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
--- a/src/DotNetWorker/Context/Features/DefaultModelBindingFeature.cs
+++ b/src/DotNetWorker/Context/Features/DefaultModelBindingFeature.cs
@@ -1,9 +1,12 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+﻿g// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using Microsoft.Azure.Functions.Worker.Converters;
+using Microsoft.Azure.Functions.Worker.Definition;
+using Microsoft.Azure.Functions.Worker.Diagnostics.Exceptions;
 
 namespace Microsoft.Azure.Functions.Worker.Context.Features
 {
@@ -31,6 +34,7 @@ namespace Microsoft.Azure.Functions.Worker.Context.Features
             _parameterValues = new object?[context.FunctionDefinition.Parameters.Length];
             _inputBound = true;
 
+            List<string>? errors = null;
             for (int i = 0; i < _parameterValues.Length; i++)
             {
                 FunctionParameter param = context.FunctionDefinition.Parameters[i];
@@ -51,8 +55,20 @@ namespace Microsoft.Azure.Functions.Worker.Context.Features
                 }
                 else if (source is not null)
                 {
-                    throw new InvalidCastException($"Cannot convert input parameter '{param.Name}' for Function '{context.FunctionDefinition.Name}' to type '{param.Type.FullName}' from type '{source.GetType().FullName}'.");
+                    // Don't initialize this list unless we have to
+                    if (errors is null)
+                    {
+                        errors = new List<string>();
+                    }
+
+                    errors.Add($"Cannot convert input parameter '{param.Name}' to type '{param.Type.FullName}' from type '{source.GetType().FullName}'.");
                 }
+            }
+            
+            // found errors
+            if (errors is not null)
+            {
+                throw new FunctionInputConverterException($"Error converting {errors.Count} input parameters for Function '{context.FunctionDefinition.Name}': {string.Join(" ", errors)}");
             }
 
             return _parameterValues;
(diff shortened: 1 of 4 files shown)


src/DotNetWorker/Context/Features/DefaultModelBindingFeature.cs
```diff
@@ -38,12 +38,17 @@ public DefaultModelBindingFeature(IEnumerable<IConverter> converters)
...
-                    continue;
+                    var converterContext = new DefaultConverterContext(param, source, context);
+                    if (TryConvert(converterContext, out object? target))
+                    {
+                        _parameterValues[i] = target;
+                    }
+                    else
+                    {
+                        throw new InvalidCastException($"Cannot convert input parameter '{param.Name}' for Function '{context.FunctionDefinition.Name}' to type '{param.Type.FullName}' from type '{source.GetType().FullName}'.");
+                    }
```
brettsam, line 51: I'd like to err on the side of warning here rather than throwing. I can't think of any right now, but there may be valid scenarios to let null flow through to the function. And that way, a function author can handle null in whatever way they want (maybe they just ignore it).

Be sure to use `WorkerMessage.Define` so we can see how often this happens in backend logs.

src/DotNetWorker/Context/Features/DefaultModelBindingFeature.cs
```diff
@@ -22,7 +22,7 @@ public DefaultModelBindingFeature(IEnumerable<IConverter> converters)
 
         public object?[]? InputArguments => _parameterValues;
 
-        public object?[] BindFunctionInput(FunctionContext context)
+        public object?[] TryBindFunctionInput(FunctionContext context)
```
brettsam, line 25: The pattern for `Try...` methods is that they return bool and "return" via out params. If you want this to be a `Try` (which I think makes sense), can you swap that signature around?

src/DotNetWorker/Invocation/DefaultFunctionExecutor.cs
```diff
@@ -38,7 +38,7 @@ public async Task ExecuteAsync(FunctionContext context)
             }
             else
             {
-                inputArguments = bindingFeature.BindFunctionInput(context);
+                inputArguments = bindingFeature.TryBindFunctionInput(context);
```
brettsam, line 41: Then, if this is returning a bool, you can log from here if it returns false.

src/DotNetWorker/Context/Features/DefaultModelBindingFeature.cs
```diff
@@ -54,7 +54,10 @@ public DefaultModelBindingFeature(IEnumerable<IConverter> converters)
                 if (TryConvert(converterContext, out object? target))
                 {
                     _parameterValues[i] = target;
-                    continue;
+                }
+                else if (source is not null)
+                {
+                    throw new InvalidCastException($"Cannot convert input parameter '{param.Name}' for Function '{context.FunctionDefinition.Name}' to type '{param.Type.FullName}' from type '{source.GetType().FullName}'.");
```
mhoeger, line 60: Per feedback:
- Iterate through all parameters and collect error messages - throw as one
- Introduce model binding exception

mhoeger:
Increasing test coverage tracked here: https://github.com/Azure/azure-functions-dotnet-worker/issues/230

src/DotNetWorker/Context/Features/DefaultModelBindingFeature.cs
```diff
@@ -48,8 +52,23 @@ public DefaultModelBindingFeature(IEnumerable<IConverter> converters)
...
+                    }
+
+                    errors.Add($"Cannot convert input parameter '{param.Name}' to type '{param.Type.FullName}' from type '{source.GetType().FullName}'.");
+                }
+            }
+            
+            // found errors
+            if (errors is not null)
+            {
+                throw new FunctionInputConverterException($"Error converting {errors.Count} input parameters for Function '{context.FunctionDefinition.Name}': {string.Join(" ", errors)}");
```
fabiocav, line 71: nit: should the join be with `Environment.NewLine` so we're providing each error in a new line?

src/DotNetWorker/Diagnostics/Exceptions/FunctionInputConverterException.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Functions.Worker.Diagnostics.Exceptions
+{
+    internal class FunctionInputConverterException : FunctionWorkerException
+    {
+        internal FunctionInputConverterException(string message) : base(message) { }
+
```
fabiocav, line 12: Should this have a property for the function name?

fabiocav (approved):

src/DotNetWorker/Diagnostics/Exceptions/FunctionWorkerException.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Functions.Worker.Diagnostics
+{
+    /// <summary>
+    /// Internal exception that is surfaced to the user
+    /// </summary>
+    internal class FunctionWorkerException : Exception
```
fabiocav, line 12: Are you defining this to serve as the base for worker exceptions?

PR #218, opened 2021-03-03
Fix issue with worker config on Linux

ankitkumarr:
I tested portable build in Linux and that seemed to work.

I wasn't able to test self contained build in Windows or Linux, but I think that might be due to other issues? I will validate after checking with @brettsam.

### DIFF
--- a/samples/FunctionApp/FunctionApp.csproj
+++ b/samples/FunctionApp/FunctionApp.csproj
@@ -13,7 +13,6 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.0.1-preview5" />
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
     <PackageReference Include="System.Net.NameResolution" Version="4.3.0" />
   </ItemGroup>
--- a/samples/FunctionApp/FunctionApp.csproj
+++ b/samples/FunctionApp/FunctionApp.csproj
@@ -13,6 +13,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.0.1-preview5" />
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
     <PackageReference Include="System.Net.NameResolution" Version="4.3.0" />
   </ItemGroup>
--- a/sdk/FunctionMetadataLoaderExtension/Startup.cs
+++ b/sdk/FunctionMetadataLoaderExtension/Startup.cs
@@ -18,8 +18,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.FunctionMetadataLoader
 {
     internal class Startup : IWebJobsStartup2, IWebJobsConfigurationStartup
     {
-        private const string _workerConfigFile = "worker.config.json";
-        private const string _exePathPropertyName = "defaultExecutablePath";
+        private const string WorkerConfigFile = "worker.config.json";
+        private const string ExePathPropertyName = "defaultExecutablePath";
+        private const string ExeSelfContainedReserved = "{WorkerRoot}";
+
         private static readonly string _dotnetIsolatedWorkerConfigPath = ConfigurationPath.Combine("languageWorkers", "dotnet-isolated", "workerDirectory");
         private static readonly string _dotnetIsolatedWorkerExePath = ConfigurationPath.Combine("languageWorkers", "dotnet-isolated", "defaultExecutablePath");
 
@@ -27,8 +29,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.FunctionMetadataLoader
         {
             string appRootPath = context.ApplicationRootPath;
 
-            // We need to adjust the path to the worker exe based on the root.
-            string exePath = GetPathToExe(appRootPath);
+            // We need to adjust the path to the worker exe based on the root, if running a self contained build.
+            string exePath = GetUpdatedExeIfSelfContained(appRootPath);
 
             builder.ConfigurationBuilder.AddInMemoryCollection(new Dictionary<string, string>
             {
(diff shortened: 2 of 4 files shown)


sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
```diff
@@ -44,9 +44,25 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
       OverwriteReadOnlyFiles="true" />
 
     <WriteLinesToFile
+        Condition="$(SelfContained)"
         File="$(OutputFile)"
-        Lines="$([System.IO.File]::ReadAllText($(_FunctionsWorkerConfigInputFile)).Replace('$functionExe$', '$(TargetName).exe'))"
+        Lines="$([System.IO.File]::ReadAllText($(_FunctionsWorkerConfigInputFile))
```
ankitkumarr, line 49: @brettsam, I think getting rid of `.exe` doesn't work in Windows. I started seeing this error -- `Microsoft.Azure.WebJobs.Script: File DefaultWorkerPath: <redacted>\selfcontained does not exist.`

Wondering if in here we should check RID and based on that, add `.exe` when in Windows.

sdk/FunctionMetadataLoaderExtension/Startup.cs
```diff
@@ -18,17 +18,19 @@ namespace Microsoft.Azure.WebJobs.Extensions.FunctionMetadataLoader
 {
     internal class Startup : IWebJobsStartup2, IWebJobsConfigurationStartup
     {
-        private const string _workerConfigFile = "worker.config.json";
-        private const string _exePathPropertyName = "defaultExecutablePath";
+        private const string WorkerConfigFile = "worker.config.json";
+        private const string ExePathPropertyName = "defaultExecutablePath";
+        private const string ExeSelfContainedReserved = "{WorkerRoot}";
```
brettsam, line 23: This is a nit -- but let's remove reference to `SelfContained` here -- call it `WorkerRootToken` or something. In the loader we don't know about self-contained -- all we know is that we need to replace this if we see it.

sdk/FunctionMetadataLoaderExtension/Startup.cs
```diff
@@ -18,17 +18,19 @@ namespace Microsoft.Azure.WebJobs.Extensions.FunctionMetadataLoader
...
         private static readonly string _dotnetIsolatedWorkerConfigPath = ConfigurationPath.Combine("languageWorkers", "dotnet-isolated", "workerDirectory");
         private static readonly string _dotnetIsolatedWorkerExePath = ConfigurationPath.Combine("languageWorkers", "dotnet-isolated", "defaultExecutablePath");
 
         public void Configure(WebJobsBuilderContext context, IWebJobsConfigurationBuilder builder)
         {
             string appRootPath = context.ApplicationRootPath;
 
-            // We need to adjust the path to the worker exe based on the root.
-            string exePath = GetPathToExe(appRootPath);
+            // We need to adjust the path to the worker exe based on the root, if running a self contained build.
```
brettsam, line 32: Update comment to be...`if WorkerRootToken is found.`

sdk/FunctionMetadataLoaderExtension/Startup.cs
```diff
@@ -50,42 +52,63 @@ public void Configure(IWebJobsBuilder builder)
...
 
-        internal static string GetPathToExe(string appRootPath)
+        internal static string GetUpdatedExeIfSelfContained(string appRootPath)
         {
-            string fullPathToWorkerConfig = Path.Combine(appRootPath, _workerConfigFile);
+            string fullPathToWorkerConfig = Path.Combine(appRootPath, WorkerConfigFile);
+            string? exeValue = GetExeValueFromWorkerConfig(fullPathToWorkerConfig);
 
-            if (!File.Exists(fullPathToWorkerConfig))
+            if (HasSelfContainedIdentifier(exeValue))
```
brettsam, line 60: Another nit -- but could you do `string.Replace()` here? If there's no `{WorkerRoot}`, it'll be a no-op.

sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
brettsam: Have the loader do this:
```
if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
{
    fileName += ".exe";
}
```

It's the same thing we do in the functions host to find dotnet. And it looks like we took it from here: https://github.com/dotnet/extensions/blob/04e2b857352b74e3edad5f36abe1ec3abd114a08/shared/Microsoft.Extensions.CommandLineUtils.Sources/Utilities/DotNetMuxer.cs#L42-L45

samples/FunctionApp/FunctionApp.csproj
```diff
@@ -13,7 +13,6 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.0.1-preview5" />
```
fabiocav, line 16: can you share more information here? Don't we expect this app to behave like any function? Are we switching to a project reference?

fabiocav (approved):

samples/FunctionApp/FunctionApp.csproj
ankitkumarr: Whoops! This was supposed to not be staged for commit. Undoing.

brettsam (approved):

sdk/FunctionMetadataLoaderExtension/Startup.cs
```diff
@@ -50,42 +54,99 @@ public void Configure(IWebJobsBuilder builder)
...
             {
-                throw new FileNotFoundException($"The file '{exePathString}' was not found.");
+                file += ".exe";
             }
 
-            return exePathString;
+            return file;
+        }
+
+        private class WorkerConfigDescription
```
ankitkumarr, line 143: So I had created this when I was starting to update both properties "defaultExcutablePath" and "defaultWorkerPath". But with switching to ".dll" for "defaultWorkerPath" as discussed offline for self-contained, this is no longer required.

I could remove this type and go back to single string. But it doesn't hurt to have this specific type. Thoughts?

brettsam: yeah just keep it.

fabiocav:
Is this ready to be merged?

PR #222, opened 2021-03-03
adding a minimal ubuntu e2e test run to CI

brettsam:


tools/devpack.ps1
```diff
@@ -8,21 +8,26 @@
 # Specify --E2E to instead target the E2E test app.
 
 $rootPath = Split-Path -Parent $PSScriptRoot
-$project = "$rootPath\samples\FunctionApp\FunctionApp.csproj"
+$project = "$rootPath/samples/FunctionApp/FunctionApp.csproj"
```
fabiocav, line 11: We should create a solution (sln) for the SDK with the analyzers, that would make this a bit simpler.

fabiocav (approved):

tools/devpack.ps1
brettsam: Yes, it'd track a few other little nits too. Will follow up on those in the coming weeks.

PR #227, opened 2021-03-04
moving result and outputs from context to feature

brettsam:


src/DotNetWorker/Context/Features/DefaultModelBindingFeature.cs
```diff
@@ -36,17 +35,13 @@ public DefaultModelBindingFeature(IEnumerable<IConverter> converters)
             {
                 FunctionParameter param = context.FunctionDefinition.Parameters[i];
 
-                var functionBindings = context.Features.Get<IFunctionBindingsFeature>();
-
+                var functionBindings = context.GetBindings();
```
fabiocav, line 38: nit: `var`-> type

src/DotNetWorker/Context/Features/GrpcFunctionBindingsFeature.cs
```diff
@@ -16,6 +16,7 @@ internal class GrpcFunctionBindingsFeature : IFunctionBindingsFeature
 
         private IReadOnlyDictionary<string, object?>? _triggerMetadata;
         private IReadOnlyDictionary<string, object?>? _inputData;
+        private IDictionary<string, object?>? _outputData;
```
fabiocav, line 19: Some of the logic here should probably be pulled to the base type (we may need to introduce one, though, doesn't have to be now) to lower the implementation cost.

src/DotNetWorker/Invocation/DefaultFunctionExecutor.cs
```diff
@@ -28,22 +28,20 @@ public async Task ExecuteAsync(FunctionContext context)
...
             else
             {
-                inputArguments = bindingFeature.BindFunctionInput(context);
+                inputArguments = modelBindingFeature.BindFunctionInput(context);
             }
 
-            object? result = await invoker.InvokeAsync(instance, inputArguments);
-
-            context.InvocationResult = result;
+            context.GetBindings().InvocationResult = await invoker.InvokeAsync(instance, inputArguments);
```
fabiocav, line 44: nit: We should do the same thing we're doing for model bindings: pull the function bindings feature, validate it was present and use the feature directly.

fabiocav (approved):
A few nit comments, nothing blocking

src/DotNetWorker/Invocation/DefaultFunctionExecutor.cs
brettsam: `GetBindings()` will throw if it's not registered. I really wasn't sure what we'd do if it wasn't there...

PR #231, opened 2021-03-05
Middleware registration API, refactoring and XML docs

fabiocav:


src/DotNetWorker/Hosting/WorkerHostBuilderExtensions.cs
```diff
@@ -0,0 +1,80 @@
...
+using System;
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Extensions.Hosting
+{
+    /// <summary>
+    /// Provides extension methods to work with a <see cref="IHostBuilder"/>.
+    /// </summary>
+    public static class WorkerHostBuilderExtensions
```
fabiocav, line 13: Moved to a more specific file name

src/DotNetWorker/Hosting/WorkerMiddlewareWorkerApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,104 @@
...
+        ///         <description><see cref="OutputBindingsMiddleware"/></description>
+        ///     </item>
+        ///     <item>
+        ///         <description><see cref="FunctionExecutionMiddleware"/></description>
+        ///     </item>
+        /// </list>
+        /// </summary>
+        /// <param name="builder">The <see cref="IFunctionsWorkerApplicationBuilder"/> to configure.</param>
+        /// <returns>The configured <see cref="IFunctionsWorkerApplicationBuilder"/>.</returns>
+        public static IFunctionsWorkerApplicationBuilder UseDefaultWorkerMiddleware(this IFunctionsWorkerApplicationBuilder builder)
```
fabiocav, line 31: new method to be explicit about the default. Previously, the call to the execution middleware was registering the output bindings one. Will update PR with the call to this instead.

src/DotNetWorker/Hosting/WorkerMiddlewareWorkerApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,104 @@
...
+
+            return builder;
+        }
+
+        /// <summary>
+        /// Configures the <see cref="IFunctionsWorkerApplicationBuilder"/> to use the default <see cref="OutputBindingsMiddleware"/>.
+        /// </summary>
+        /// <param name="builder">The <see cref="IFunctionsWorkerApplicationBuilder"/> to configure.</param>
+        /// <returns>The configured <see cref="IFunctionsWorkerApplicationBuilder"/>.</returns>
+        internal static IFunctionsWorkerApplicationBuilder UseMiddleware<T>(this IFunctionsWorkerApplicationBuilder builder)
```
fabiocav, line 86: New extension method to make it easier to register custom middleware

src/DotNetWorker/Pipeline/IFunctionsWorkerMiddleware.cs
```diff
@@ -0,0 +1,22 @@
...
+
+using System.Threading.Tasks;
+using Microsoft.Azure.Functions.Worker.Pipeline;
+
+namespace Microsoft.Azure.Functions.Worker.Middleware
+{
+    /// <summary>
+    /// Represents a middleware to be used in the worker execution pipeline.
+    /// </summary>
+    public interface IFunctionsWorkerMiddleware
```
fabiocav, line 12: Interface for custom middleware to work with the new extension method

src/DotNetWorker/RpcExtensions.cs
```diff
@@ -10,12 +10,12 @@
 using Microsoft.Azure.Functions.Worker.Http;
 using Microsoft.Azure.Functions.Worker.Invocation;
 
-namespace Microsoft.Azure.Functions.Worker
+namespace Microsoft.Azure.Functions.Worker.Rpc
```
fabiocav, line 13: Moved this class to a more specific namespace as the `**Rpc` methods were polluting the entire project.

src/DotNetWorker/Hosting/WorkerMiddlewareWorkerApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,104 @@
...
+            return builder.UseOutputBindingsMiddleware()
+                .UseFunctionExecutionMiddleware();
+        }
+
+        /// <summary>
+        /// Configures the <see cref="IFunctionsWorkerApplicationBuilder"/> to use the default <see cref="FunctionExecutionMiddleware"/>.
+        /// </summary>
+        /// <param name="builder">The <see cref="IFunctionsWorkerApplicationBuilder"/> to configure.</param>
+        /// <returns>The configured <see cref="IFunctionsWorkerApplicationBuilder"/>.</returns>
+        public static IFunctionsWorkerApplicationBuilder UseFunctionExecutionMiddleware(this IFunctionsWorkerApplicationBuilder builder)
```
mhoeger, line 42: Might be confusing if both `UseFunctionExecutionMiddleware` and `UseDefaultWorkerMiddleware` are public, thoughts on just exposing the default? "FunctionExecutionMiddleware" sounds important enough that I may try to use both.

fabiocav: I tried to make sure the documentation is pretty explicit about what gets registered with the default. The main reason I want to keep this is because it gives people the ability to compose with their own middleware. If they want to register theirs before the execution but after the output binding, the individual registration calls gives them that flexibility.

mhoeger: is UseOutputBindingsMiddleware (and UseMiddleware) meant to be internal?

src/DotNetWorker/Definition/BindingDirection.cs
```diff
@@ -3,9 +3,18 @@
 
 namespace Microsoft.Azure.Functions.Worker
 {
+    /// <summary>
+    /// Specifies the direction of the binding.
+    /// </summary>
     public enum BindingDirection
     {
+        /// <summary>
+        /// Identifieds an input binding; a binding that provides data to the function.
```
ankitkumarr, line 12: Identifieds -> Identifies

src/DotNetWorker/Hosting/WorkerMiddlewareWorkerApplicationBuilderExtensions.cs
ankitkumarr: @fabiocav, the concern I have here is that, if people update their existing project references to the new worker version, nothing would through, and their output bindings would just stop working. As this function now doesn't register that, which makes sense.

I am not sure what the alternative should be though. This may be Ok moving from preview, but wanted to bring it up, if you had any thoughts.

src/DotNetWorker/Hosting/WorkerMiddlewareWorkerApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,104 @@
...
+
+            return builder;
+        }
+
+        /// <summary>
+        /// Configures the <see cref="IFunctionsWorkerApplicationBuilder"/> to use the default <see cref="OutputBindingsMiddleware"/>.
+        /// </summary>
+        /// <param name="builder">The <see cref="IFunctionsWorkerApplicationBuilder"/> to configure.</param>
+        /// <returns>The configured <see cref="IFunctionsWorkerApplicationBuilder"/>.</returns>
+        internal static IFunctionsWorkerApplicationBuilder UseOutputBindingsMiddleware(this IFunctionsWorkerApplicationBuilder builder)
```
ankitkumarr, line 64: Could this use the `UseMiddleware` function below? That way, we also end up indirectly testing that :D

src/DotNetWorker/Hosting/WorkerMiddlewareWorkerApplicationBuilderExtensions.cs
fabiocav: These kinds of breaking changes need to be made prior to GA. Unfortunately, the previous middleware registration was overloaded, which shouldn't have been. One of the options is to cause a compilation break by renaming the method. We also need to update the templates and, as discussed, the registration should be further simplified with a simple call that adds all the defaults.

For now, since we do expect the output binding one to be registered, we can also cause an error if that isn't present. A few options here, but we shouldn't keep an API bug introduced in preview. We can have detailed information in the release notes.

src/DotNetWorker/Hosting/WorkerMiddlewareWorkerApplicationBuilderExtensions.cs
fabiocav: I knew the question was coming :) 

We could, but didn't really want to have to implement the interface when we already have that plumbing in place for the middleware as is. It's also an example of the pattern you can follow for that style of registration. We can have a test that uses the new model.

src/DotNetWorker/Hosting/WorkerMiddlewareWorkerApplicationBuilderExtensions.cs
ankitkumarr: Discussed offline for the breaking change issue, and I agree with Fabio that it makes most sense to update this, and callout in release notes.

Sorry @mhoeger, I accidentally used your comment thread for my question on a separate topic. But to answer your question, I don't think it should be internal as it gives people ability to setup differently than how do we by default. For example, if they want to add a middleware between ExecutionMiddleware and OutputBindingMiddleware. They can call these in some different order.

src/DotNetWorker/Hosting/WorkerMiddlewareWorkerApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,104 @@
...
+
+                    return middleware.Invoke(context, next);
+                };
+            });
+
+            return builder;
+        }
+
+        /// <summary>
+        /// Configures the <see cref="IFunctionsWorkerApplicationBuilder"/> to use the default <see cref="OutputBindingsMiddleware"/>.
```
brettsam, line 82: Update doc comment...

src/DotNetWorker/Hosting/WorkerMiddlewareWorkerApplicationBuilderExtensions.cs
brettsam: If this is internal, should IFunctionsWorkerMiddleware also be internal?

src/DotNetWorker/RpcExtensions.cs
brettsam: Good call.

brettsam (approved):
Couple small comments!

src/DotNetWorker/Hosting/WorkerMiddlewareWorkerApplicationBuilderExtensions.cs
fabiocav: The intention was for this to be public....

PR #239, opened 2021-03-06
Updates to the build process and solutions

fabiocav:


sdk/Sdk/Sdk.csproj
```diff
@@ -9,8 +9,8 @@
     <RootNamespace>Microsoft.Azure.Functions.Worker.Sdk</RootNamespace>
     <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
     <IncludeBuildOutput>false</IncludeBuildOutput>
-    <FunctionsExecutionModel>isolated</FunctionsExecutionModel>
```
fabiocav, line 12: Moving this to the props file as it should be present in the application's project system

tools/devpack.ps1
```diff
@@ -1,45 +1,74 @@
...
 {
   New-Item -Path $localPack -ItemType directory | Out-Null
 }
 Write-Host
 Write-Host "---Updating project with local SDK pack---"
-Write-Host "Packing SDK to $localPack"
-& "dotnet" "pack" $sdkProject "-o" "$localPack" "-nologo"
-Write-Host "Packing Analyzers to $localPack"
-& "dotnet" "pack" $analyzerProject "-o" "$localPack" "-nologo"
+Write-Host "Packing Core .NET Worker projects to $localPack"
```
fabiocav, line 37: This should eventually be broken up into reusable functions. Keeping as is for now

PR #243, opened 2021-03-08
Hosting extensions updates

fabiocav:
Still needs test updates. A bit of a WIP, but wanted to get this out for discussion.

### DIFF
--- a/samples/FunctionApp/Program.cs
+++ b/samples/FunctionApp/Program.cs
@@ -26,10 +26,7 @@ namespace FunctionApp
                 })
                 //</docsnippet_configure_app>
                 //<docsnippet_middleware>
-                .ConfigureFunctionsWorker((c, b) =>
-                {
-                    b.UseDefaultWorkerMiddleware();
-                })
+                .ConfigureFunctionsWorkerDefaults()
                 //</docsnippet_middleware>
                 //<docsnippet_dependency_injection>
                 .ConfigureServices(s =>
--- a/src/DotNetWorker/Hosting/FunctionsWorkerHost.cs
+++ b/src/DotNetWorker/Hosting/FunctionsWorkerHost.cs
@@ -1,0 +1,44 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Extensions.Hosting
+{
+    /// <summary>
+    /// Hosing helpers to work with Azure Functions Workers
+    /// </summary>
+    public static class FunctionsWorkerHost
+    {
+        /// <summary>
+        /// Creates a default Azure Functions Worker configured host, runs and 
+        /// blocks the calling thread until the host shuts down.
+        /// </summary>
+        /// <param name="configureServices">An optional delegate to configure host services.</param>
+        public static void CreateDefaultAndRun(Action<IServiceCollection>? configureServices = null)
+        {
+            CreateDefaultAndRunAsync(configureServices).GetAwaiter().GetResult();
+        }
+
+        /// <summary>
+        /// Creates a default Azure FUnctions Worker configured host, runs and
+        /// returns a <see cref="Task"/> that will complete when the host shuts down.
+        /// </summary>
+        /// <param name="configureService">An option delegate to configure host services.</param>
+        /// <returns>A <see cref="Task"/> that will complete when the host shuts down.</returns>
+        public async static Task CreateDefaultAndRunAsync(Action<IServiceCollection>? configureService = null)
+        {
+           var builder = Host.CreateDefaultBuilder()
+                .ConfigureFunctionsWorkerDefaults();
+
+            if (configureService is not null)
+            {
+                builder.ConfigureServices(configureService);
+            }
+            
+            await builder.Build().RunAsync();
+        }
+    }
+}
--- a/src/DotNetWorker/Hosting/GrpcServiceCollectionExtensions.cs
+++ b/src/DotNetWorker/Hosting/GrpcServiceCollectionExtensions.cs
@@ -1,0 +1,70 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Threading.Channels;
+using Grpc.Core;
+using Grpc.Net.Client;
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Azure.Functions.Worker.Diagnostics;
+using Microsoft.Azure.Functions.Worker.Grpc.Messages;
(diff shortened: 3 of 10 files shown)


samples/FunctionApp/Program.cs
```diff
@@ -26,10 +26,7 @@ static async Task Main(string[] args)
                 })
                 //</docsnippet_configure_app>
                 //<docsnippet_middleware>
-                .ConfigureFunctionsWorker((c, b) =>
-                {
-                    b.UseDefaultWorkerMiddleware();
-                })
+                .ConfigureFunctionsWorkerDefaults()
```
fabiocav, line 29: @ggailey777 this would have documentation impact.

We need to decide if a new sample is required, as this uses the simpler model, but doesn't add middleware.

src/DotNetWorker/Hosting/FunctionsWorkerHost.cs
```diff
@@ -0,0 +1,44 @@
...
+using System;
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Extensions.Hosting
+{
+    /// <summary>
+    /// Hosing helpers to work with Azure Functiosn Workers
+    /// </summary>
+    public static class FunctionsWorkerHost
```
fabiocav, line 13: This is primarily for some scenarios that won't impact some of the mainline samples, but will allow for quick configuration and execution of the host with reduced ceremony:
``` csharp
FunctionsWorkerHost.CreateDefaultAndRunAsync();
```

src/DotNetWorker/Hosting/GrpcServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+                    SingleWriter = false,
+                    SingleReader = true,
+                    AllowSynchronousContinuations = true
+                };
+
+                return new GrpcHostChannel(System.Threading.Channels.Channel.CreateUnbounded<StreamingMessage>(outputOptions));
+            });
+        }
+
+        public static IServiceCollection AddGrpc(this IServiceCollection services)
```
fabiocav, line 35: The type is internal as additional work is required to expose this. Just flagging as we want this to be an option.

src/DotNetWorker/Hosting/GrpcWorkerApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.Functions.Worker;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    internal static class GrpcWorkerApplicationBuilderExtensions
+    {
+        public static IFunctionsWorkerApplicationBuilder UseGrpc(this IFunctionsWorkerApplicationBuilder builder)
```
fabiocav, line 10: Same as above about visibility

src/DotNetWorker/Hosting/FunctionsWorkerHost.cs
fabiocav: I think I actually like 
```csharp
FunctionsWorkerHost.RunDefaultAsync();
```

samples/FunctionApp/Program.cs
ggailey777: This snippet is currently called out in the [middleware section](https://review.docs.microsoft.com/en-us/azure/azure-functions/dotnet-isolated-process-guide?branch=pr-en-us-149017#middleware). I assume this new version isn't as useful in that context.

src/DotNetWorker/Hosting/FunctionsWorkerHost.cs
```diff
@@ -0,0 +1,44 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Extensions.Hosting
+{
+    /// <summary>
+    /// Hosing helpers to work with Azure Functiosn Workers
```
brettsam, line 11:
```suggestion
    /// Hosing helpers to work with Azure Functions Workers
```

src/DotNetWorker/Hosting/FunctionsWorkerHost.cs
brettsam: The use of `Host` here seems confusing since there's already a concept of a functions `Host`. Can we drop the `Host`? Or use `Application`?

src/DotNetWorker/Hosting/ServiceCollectionExtensions.cs
```diff
@@ -74,33 +53,41 @@ public static IFunctionsWorkerApplicationBuilder AddFunctionsWorker(this IServic
...
+        }
+
+        /// <summary>
+        /// Adds the core set of services for the Azure Functions worker.
+        /// This call also adds the default set of binding converters and gRPC support.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection"/>.</param>
+        /// <param name="configure">The action used to configure <see cref="WorkerOptions"/>.</param>
+        /// <returns>The same <see cref="IFunctionsWorkerApplicationBuilder"/> for chaining.</returns>
+        public static IFunctionsWorkerApplicationBuilder AddFunctionsWorkerDefaults(this IServiceCollection services, Action<WorkerOptions>? configure = null)
```
brettsam, line 74: Could this be something like `AddDefaultFunctionsWorker`? I'm not sure why, but ending it with `Defaults` was hard for me to parse :-)

brettsam (commented):
Couple of api naming questions...

samples/FunctionApp/Program.cs
fabiocav: Yeah... I can leave it with middleware registration, but I think a sample with custom middleware would actually better drive the point.

src/DotNetWorker/Hosting/FunctionsWorkerHost.cs
fabiocav: We could, but was afraid that would be confuses with `FunctionApplication`.... I'm also not a huge fan, but `Host` would be consistent with the terminology used in ASP.NET, Generic Host and others

src/DotNetWorker/Hosting/ServiceCollectionExtensions.cs
fabiocav: Yeah, at first, it felt  odd, but it's consistent with `ConfigureWebHostDefaults` for ASP.NET, and it does make sense when you think about what it does; it gives you the default services and you can configure them (this is the `Add` version against the service collection)

brettsam: Ah okay -- if this naming exists already, I'm good.

brettsam (approved):

PR #248, opened 2021-03-08
Remove Linux support from known issues

ankitkumarr:
We now support Linux everywhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ We're working with the Visual Studio team to provide an integrated debugging exp
     ```
 ## Known issues
 
-* Optimizations are not in place in the consumption plan and you may experience longer cold starts.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -86,7 +86,7 @@ We're working with the Visual Studio team to provide an integrated debugging exp
     ```
 ## Known issues
 
-* Deployment to Azure is currently limited to Windows plans. Note that some optimizations are not in place in the consumption plan and you may experience longer cold starts.
+* Optimizations are not in place in the consumption plan and you may experience longer cold starts.
```
mhoeger, line 89: nit: "Not all performance optimizations are in place in the consumption plan"? Just so it doesn't sound like we don't have like any optimizations?

mhoeger (approved):
🚀 !

PR #261, opened 2021-03-09
Ignore casing on property names by default

mhoeger:
By default, property name casing is ignored.


src/DotNetWorker/Hosting/ServiceCollectionExtensions.cs
```diff
@@ -81,6 +84,15 @@ public static IFunctionsWorkerApplicationBuilder AddFunctionsWorkerDefaults(this
             // Converters
             services.RegisterDefaultConverters();
 
+            // Default Json serialization should ignore casing on property names
+            services.Configure<WorkerOptions>(options =>
+            {
+                options.Serializer = new JsonObjectSerializer(new JsonSerializerOptions
```
fabiocav, line 90: Instead of creating the serializer here (which we defer in the options to avoid creating an instance if the user intends to replace with another serializer), can we just setup a configure call for JsonSerializerOptions and make sure we use that in the WorkerOptions?

e.g.: `services.Configure<WorkerOptions>(...)`
then update WorkerOptions to pull the options if an instance is created.

brettsam: I think you mean... `services.Configure<JsonSerializerOptions>(...)`?

src/DotNetWorker/Hosting/ServiceCollectionExtensions.cs
```diff
@@ -66,7 +68,8 @@ public static IFunctionsWorkerApplicationBuilder AddFunctionsWorkerCore(this ISe
 
         /// <summary>
         /// Adds the core set of services for the Azure Functions worker.
-        /// This call also adds the default set of binding converters and gRPC support.
+        /// This call adds the default set of binding converters and gRPC support.
+        /// This call also adds a default ObjectSerializer that treats property names as case insensitive.
```
brettsam, line 72: Maybe add this to the lists of things we do by default in `ConfigureFunctionsWorkerDefaults`? That's the API most people will interact with.

src/DotNetWorker/Hosting/ServiceCollectionExtensions.cs
fabiocav: Yes, that's what I meant. I stated the right thing in the feedback, but typed the wrong options in the example.

fabiocav: Will update the example for clarity.

fabiocav (approved):

PR #276, opened 2021-03-10
Update comments

mhoeger:


src/DotNetWorker/Hosting/WorkerHostBuilderExtensions.cs
```diff
@@ -38,6 +38,7 @@ public static IHostBuilder ConfigureFunctionsWorkerDefaults(this IHostBuilder bu
         /// The following defaults are configured:
         /// <list type="bullet">
         ///     <item><description>A default set of converters.</description></item>
+        ///     <item><description>Set default serializer to ignore casing on property names.</description></item>
```
ankitkumarr, line 41: Should we say `Set default JSON serializer`?

fabiocav: To be clearer, should this be `Configures the default <see cref="JsonObjectSerializerOptions(or whatever the name was)"/> to ignore casing on property names.`?

mhoeger: good call! done :)

fabiocav (approved):

PR #298, opened 2021-03-13
Creating custom middleware sample and reorganizing samples

### LINKED ISSUE #280: Create middleware samples
Create a comprehensive set of middleware samples

/cc @ggailey777

fabiocav:
Resolves #280


fabiocav:
@ggailey777 for awareness

samples/CustomMiddleware/HttpFunction.cs
```diff
@@ -0,0 +1,36 @@
...
+    public class HttpFunction
+    {
+        [Function("HttpFunction")]
+        public static HttpResponseData Run([HttpTrigger(AuthorizationLevel.Anonymous, "get", "post", Route = null)] HttpRequestData req,
+            FunctionContext context)
+        {
+            var logger = context.GetLogger<HttpFunction>();
+
+            // Get the item set by the middleware
+            if (context.Items.TryGetValue("middlewareitem", out object value) && value is string message)
```
SeanFeldman, line 20: Could you do `context.Items.TryGetValue("middlewareitem", out string value)` right away?

samples/CustomMiddleware/MyCustomMiddleware.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Threading.Tasks;
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Azure.Functions.Worker.Middleware;
+using Microsoft.Extensions.Logging;
+
+namespace CustomMiddleware
+{
+    public class MyCustomMiddleware : IFunctionsWorkerMiddleware
+    {
+        public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next)
```
SeanFeldman, line 14: Would be nice if instead of a Delegate it would be `Func<FunctionContext, Task>`

SeanFeldman (commented):
Great sample 👍

samples/CustomMiddleware/HttpFunction.cs
fabiocav: Absolutely. This is what you end up with when you decide last minute on a Friday; “let me write this sample really quick” and end up refactoring some code as you’re cleaning up for PR😆

samples/CustomMiddleware/MyCustomMiddleware.cs
fabiocav: That’s essentially what this delegate is, but a specialized version specific to this context, which is helpful in a number of scenarios.

samples/CustomMiddleware/HttpFunction.cs
fabiocav: Actually, I take that back 😄 

I remember this actually being the syntax I though it would be nice, but we're working with an `IDictionary<object, object>` here, that that won't work. When I read the comment I immediately thought I had spaced and this was working on string values (same mistake as originally described, just on the opposite end, where I decided to respond to GH reviews too early in the morning :) )

The sample shows how you'd ensure you're working with a valid string that is not `null` in a simple statement here.

samples/CustomMiddleware/MyCustomMiddleware.cs
fabiocav: This delegate is also part of the public API. I do have plans to introduce UseMiddleware APIs that will take a simple lambda (or a `Func` with the appropriate signature), but this is the basic delegate that ends up being used and is safe to use in some contexts where it would otherwise become ambiguous without extra hoops.

Can you expand on concerns you had with this? Is this mostly about the usability? And if so, does the plan to support the APIs I mentioned above help with that?

ggailey777:
@fabiocav please warn me after you merge this PR. The SampleApp rename will break some of the existing snippets unless I also update the ref links in docs. Thanks!

samples/CustomMiddleware/MyCustomMiddleware.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            // This is added pre-function execution, function will have access to this information
+            // in the context.Items dictionary
+            context.Items.Add("middlewareitem", "Hello, from middleware");
+            
+            await next(context);
```
Romfos, line 20: how can I get\override http response here? for ex. authorization, global exception handler, e.t.c ?

fabiocav: @Romfos since those are more general questions about the middleware model, as opposed to this PR, it might be simpler to iterate on an issue if you can open one.

For the worker middleware, it's important to understand that the middleware pipeline in the worker is **not** HTTP centric. The context and abstractions expose information about the various triggers and bindings your functions use, one of which would be HTTP. You can use a the various properties in the context to inspect the payload and (depending on where your middleware is injected) the values for the bindings. This would be a topic that would be good to document and follow up in a separate issue.

For the exception handler, though (and other cross-cutting concerns not specific to HTTP), the pattern is the same; your middleware can handle and deal with exceptions bubbling back up after invoking the next middleware in the chain.

SeanFeldman: @fabiocav, I wonder if the sample would be a bit more concrete it would help to understand that the middleware pipeline is a Russian doll concept, allowing to add something before Function invocation and after, that would also be able to exchange information from/to middleware and the function. For instance, measuring Function execution time. That works great before and after next()/Function invocation but doesn't demonstrate the middleware and Function interaction, passed back and forth data that is useful in a real-world scenario. Must admit, I understand what the sample is doing technically but can't think of how where it'd be useful.

ankitkumarr (approved):

samples/CustomMiddleware/MyCustomMiddleware.cs
fabiocav: The sample tries to illustrate that flow, passing and sharing data between middleware and the function. Some more advanced features build on this to enable things like model binding in the worker. Collecting some core scenarios will be helpful so we can expand the sample, but my key goal here was to keep this simple and primarily demonstrate the registration process, as we were getting questions about this and didn't have an example we could share.

Some of the ideas for samples in this space are:
 - Cross cutting concerns
   - Error handling
   - Logging scopes
   - Authorization/payload inspection and short-circuiting
- Data enrichment
  - Additional state
  - Influence conversion logic

That's a small list, and we can expand quite a bit on this. But I think having something simple, to highlight the basic registration process, helps a lot of people get started here.

If we take the sample here, we could envision a scenario where "customer profile" information is retrieved and added to the context, we can further dive into this to understand how middleware can enrich the context with custom features (another topic that requires more documentation) to make the same scenario more efficient and drive interaction with other pieces of middleware as well (as opposed to just the function). This would also help enable scenarios like custom model binding to things a middleware is adding to the context, for example...

fabiocav:
@ggailey777 waiting for an ack from you as I know this has docs impact.

ggailey777:
@fabiocav I think it's fine for docs. If the middleware sample gets too extensive, we can just show registration in the docs and link out to the repo for the full example. I'm ready to update the docs after these changes get merged.

PR #302, opened 2021-03-15
adding TimerInfo and e2e Timer test

### LINKED ISSUE #292: Is TimerInfo gone?
https://github.com/Azure/azure-functions-dotnet-worker/blob/main/samples/SampleApp/Timer/TimerFunction.cs#L21

I can't find a built in TimerInfo type? This seems very weird to have a custom CLIMutable custom type.

Any pointers to the built in one?

brettsam:
Fixes #292


fabiocav (approved):

forki:
Is this already available?

fabiocav:
Not yet on NuGet, but we’ll be publishing an update next week with a batch of changes. If it would help, we can make it available on a private feed before then.

Issue #300, opened 2021-03-14
HttpResponseData missing writeBytesAsync method


allanlw:
`HttpResponseData` has a `writeString` and `writeStringAsync`, and it has `writeBytes` but not `writeByteAsync`.

Workaround is to call to `response.Body.writeAsync` instead, but this seems like an oversight.

oscarvantol:
@fabiocav Would you consider/appreciate a PR for something like this?

fabiocav:
Sure! I was planning on taking this, but if you want to submit a PR, that would be more than welcome!

PR #315, opened 2021-03-17
Adding WriteBytesAsync to HttpResponseDataExtensions.cs

oscarvantol:
Related to issue #300 

- Added missing WriteBytesAsync method on HttpResponseData. 
- Also modified encoding comment on WriteString to make it consistent with WriteStringAsync.


fabiocav (approved):
@brettsam this type is missing some tests. Given the nature of the change and the fact we don't have an example, not blocking on tests for this, but something we should track.

fabiocav:
Thank you!

PR #333, opened 2021-03-20
Registering environment variables configuration provider by default

### LINKED ISSUE #288: ConfigureFunctionsWorkerDefaults should also call AddEnvironmentVariables
Env. variables is default means by which a function receives it's parameters when run on Azure, aren't they? Things like connection strings, various secrets and stuff. One would expect those are loaded to IConfiguration by default. And that seem to be the in-proc worker behavior everyone is used to. Am I correct?

fabiocav:
Resolves #288

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -2,16 +2,6 @@
 <!-- Please add your release notes in the following format:
 - My change description (#PR/#issue)
 -->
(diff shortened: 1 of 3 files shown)


mhoeger (approved):
✅

PR #334, opened 2021-03-20
API for inline registration of middleware

fabiocav:


mhoeger (approved):
Nit: add custom middleware sample for @ggailey777 / docs?

fabiocav:
Yeah, I wanted to add a sample, but we need to figure out the proper flow here. Changes here would be reflected in the docs right away, but this capability won't be available until the packages are released, so anyone following the docs would have broken code... just need to define how the samples get updated. Currently, they reference the public packages, which helps avoid those issues.

PR #349, opened 2021-03-23
Exposing IFunctionActivator API

### LINKED ISSUE #290: Registering a function class in .ConfigureServices(...) doesn't work.
This
```
.ConfigureServices((context, services) =>
{
    services.TryAddTransient(x => new MyFunction(...));
})
```
doesn't work. I can't find a way to provide my own functions factory. It makes it impossible to use third party containers like Autofac, for example.
I found [IFunctionActivator.cs](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/src/DotNetWorker.Core/IFunctionActivator.cs) but whatever `IServiceProvider` it uses is not the one I have any control over. And I can't provide my own `IFunctionActivator` either because it's `internal`.

fabiocav:
Resolves #290


src/DotNetWorker.Core/Invocation/FunctionActivatorExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    /// <summary>
+    /// Contains extension methods to work with <see cref="IFunctionActivator"/> instances.
+    /// </summary>
+    public static class FunctionActivatorExtensions
+    {
+        /// <summary>
+        /// Creates an instance of the specified generic type argument, <typeparamref name="T"/>, to be used as the target of the invocation.
+        /// </summary>
+        /// <typeparam name="T">Tye type of the instance to create.</typeparam>
```
ankitkumarr, line 16: `Tye type` -> `The type`

ankitkumarr (approved):

PR #351, opened 2021-03-24
Use guard over if statements

SeanFeldman:


SeanFeldman:
@fabiocav don't know if this is something you'd like to take in or not. Removes the chattiness of the guarding `if` clauses for `null` checks.

fabiocav:
@SeanFeldman we discussed the changes with the team, and I think, at least for now, we'd like to avoid this change as we want to keep this consistent with the other repos (in addition to some other concerns, including the muscle memory many have on using the IDE to add those checks). As always, we do appreciate the contribution, though!

PR #357, opened 2021-03-25
Exposing RetryContext

fabiocav:


src/DotNetWorker.Grpc/GrpcFunctionInvocation.cs
```diff
@@ -20,5 +21,7 @@ public GrpcFunctionInvocation(InvocationRequest invocationRequest)
         public override string FunctionId => _invocationRequest.FunctionId;
 
         public override TraceContext TraceContext { get; }
+
+        public RetryContext Context => _retryContext ??= new GrpcRetryContext(_invocationRequest.RetryContext);
```
mhoeger, line 25: I think `_invocationRequest.RetryContext` can be null. From Alexey's PR to the node repo, there's a [comment here](https://github.com/Azure/azure-functions-nodejs-worker/blob/7005a1012e9ba88f6e402db36d3b20052b6a0a9a/src/public/Interfaces.ts#L186) that "The retry context of the current funciton execution. The retry context of the current function execution. Equals null if retry policy is not defined or it's the first function execution." I think maybe the property itself should be nullable or there should be some defaults?

src/DotNetWorker.Core/Context/FunctionContext.cs
```diff
@@ -34,6 +34,12 @@ public abstract class FunctionContext
         /// </summary>
         public abstract BindingContext BindingContext { get; }
 
+        /// <summary>
+        /// Gets the retry context containing information about retry acvitity for the event that triggered
+        /// the current function invocation.
+        /// </summary>
+        public abstract RetryContext RetryContext { get; }
```
mhoeger, line 41: Maybe nullable?

src/DotNetWorker.Core/Context/RetryContext.cs
```diff
@@ -0,0 +1,24 @@
...
+        /// The the retry count for the current event.
+        /// </summary>
+        public abstract int RetryCount { get; }
+
+        /// <summary>
+        /// The maximum number of retry attempts that will be made by the trigger/host
+        /// before the event is considered undeliverable.
+        /// </summary>
+        public abstract int MaxRetryCount { get; }
+
```
mhoeger, line 22: Should we surface the retry exception here too?

mhoeger (commented):
A few comments!

src/DotNetWorker.Grpc/GrpcFunctionInvocation.cs
fabiocav: I'll apply defaults. It makes if for a pretty difficult API to work with if this expectation changes depending on which execution we're working with. Wish we had caught that in the original implementation as this should become available if requested with information that indicates that we're dealing with the first execution.

src/DotNetWorker.Core/Context/FunctionContext.cs
fabiocav: See above. I think we should make sure the defaults are applied and avoid a world where people have to perform `null` checks when working against the context all over the place (or worse yet, failing to do so and running into `null` ref exceptions)

src/DotNetWorker.Core/Context/RetryContext.cs
fabiocav: Thought about this, need to look at what exception details are actually passed in (if it's truly the original exception, with an appropriate stack trace and something actionable), otherwise, we might be better off simply surfacing the error message. 

I had this flagged for a future PR, but would be open to adding now if we feel it's critical to have that.

src/DotNetWorker.Grpc/GrpcFunctionInvocation.cs
mhoeger: Yeah - makes sense!! Retry count of 0 makes sense as a default but I'm having a hard time thinking of what the default would be for MaxRetryCount :/

src/DotNetWorker.Core/Context/RetryContext.cs
mhoeger: Makes sense!

src/DotNetWorker.Grpc/GrpcFunctionInvocation.cs
fabiocav: Yes, that will be an issue. I would really want to avoid releasing this with the null behavior, but without getting that information from the host, it will be difficult to expose the max count. I'll follow up to see what we can change in the host.

fabiocav: Followed on the SDK item. I think we'll be going with a change there in order to properly enable this here, so this work will be blocked until we make that change (unfortunately). This capability is being positioned as preview in the SDK, but there isn't a great way to communicate that here beyond the approaches used in the SDK, which are not great.

mhoeger: Would it make sense to check in this change with just the RetryCount property since we know we can default to "0"?

fabiocav: We could, and I think there's some value even with the limited information. We made the change in the SDK last Friday, so though the behavior has been fixed there, it will take a few weeks until we can unblock this work (when it makes its way through deployment).

I'm on the fence. I'm all in favor of delivering incremental updates, but wonder if not having the max count for now will lead to some frustration with the initial implementation.

src/DotNetWorker.Core/Context/RetryContext.cs
fabiocav: I don't think we need to block on this. Let's flag this for a follow up.

onionhammer:
What is the utility of throwing an exception when trying to access this context? The property itself is throwing an NRE

image

PR #368, opened 2021-03-27
Minor cleanup and optimizations

fabiocav:


src/DotNetWorker.Grpc/GrpcFunctionsHostLogger.cs
```diff
@@ -77,24 +77,16 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
 
         private static Level ToRpcLogLevel(LogLevel logLevel)
         {
-            switch (logLevel)
+            return logLevel switch
```
SeanFeldman, line 80: 😍

src/DotNetWorker.Core/Hosting/WorkerMiddlewareWorkerApplicationBuilderExtensions.cs
```diff
@@ -75,15 +75,13 @@ public static IFunctionsWorkerApplicationBuilder UseOutputBindingsMiddleware(thi
                 return builder;
             }
 
-            builder.Services.AddSingleton<OutputBindingsMiddleware>();
+            builder.Services.AddSingleton<OutputBindingsMiddleware>(p => throw new InvalidOperationException($"Type {nameof(OutputBindingsMiddleware)} cannot be resolved using the container."));
```
ankitkumarr, line 78: I understand that if this type is requested, it will thrown an exception now. But I am wondering, do we even need to make this call?

ankitkumarr (approved):

src/DotNetWorker.Core/Hosting/WorkerMiddlewareWorkerApplicationBuilderExtensions.cs
ankitkumarr: Talked to Fabio, and this change makes sense. Leaving this registration call and the check right above prevents duplicate registration of this same middleware.

fabiocav: Followed up offline.

Issue #420, opened 2021-04-28
"dotnet publish --no-build" fails


ankitkumarr:
This fails with `error NETSDK1085: The 'NoBuild' property was set to true but the 'Build' target was invoked.` This happens because we perform a build for `WorkerExtensions`. We need to handle the `dotnet publish --no-build` scenario for WorkerExtensions.

If we change WorkerExtensions to do a publish, we need to do either of these two choices--
1. Remove "NoBuild" property on publishing WorkerExtensions, and do a full publish of that. Because publish will not have access to the built obj as WorkerExtensions is built dynamically in a temp directory. 
2. Handle "NoBuild" -- don't do a publish if NoBuild, and make sure we copy `.azurefunctions` from the build directory to the publish directory using a target if NoBuild is set.

PR #421, opened 2021-04-28
Fix sdk targets to support nobuild publish

ankitkumarr:
Switching to "publish" for WorkerExtensions isn't trivial as when a user passes `--no-build`, the WorkerExtensions can't use the last build to generate the publish content. This is because WorkerExtensions build runs dynamically in a temporary directory.

So, we had two choices --
1. Ignore NoBuild and do a full publish even when "--no-build" is passed. But this goes against the argument semantics.
2. When NoBuild is passed, manually copy the binaries from user's built contents.

I went with 2. as it makes most sense to honor the `--no-build` flag. This behavior can be overwritten by passing a property `_FunctionsExtensionsFullPublish`. 

Fixes https://github.com/Azure/azure-functions-dotnet-worker/issues/420


fabiocav (approved):
The changes, overall, look good. I'm wondering if

brettsam:
> The changes, overall, look good. I'm wondering if

@fabiocav -- I can't wait for you to finish this sentence!

Issue #425, opened 2021-04-29
Error after adding the Microsoft.AspNetCore.DataProtection to a project


stukalin:
Most likely I am facing the issue described in https://github.com/Azure/azure-functions-host/issues/5447. They claimed to fix it but apparently it reappeared again in Azure Functions for .NET 5.

Steps: 
1. Create a fresh function app project (dotnet-isolated of course)
2. Add a new function, HTTP triggered is fine
3. Add the `Microsoft.AspNetCore.DataProtection` package (at the time of writing it was 5.0.5)
4. Configure data protection:
```
        public static void Main()
        {
            var host = new HostBuilder()
                .ConfigureFunctionsWorkerDefaults()
                .ConfigureServices(services =>
                {
                    services.AddDataProtection();
                })
                .Build();

            host.Run();
        }
```
5. Run the function

I then receive an error message, not very detailed though, because apparently I face the #417 or #359 or something similar and instead of real error message I get the following.
![image](https://user-images.githubusercontent.com/9863365/116520347-44c50980-a8db-11eb-8324-2ff2ec0ab085.png)

I applied the suggested workaround and the data protection seems to work fine if I configure it like this:
```
                    services.AddDataProtection();
                    services.Remove(services.First(x => x.ImplementationType?.Name == "DataProtectionHostedService"));
```

The workaround seems to work on Azure side as well however I haven't tried any sophisticated scenarios/configurations.

For reference
```
Core Tools Version:       3.0.3442 Commit hash: 6bfab24b2743f8421475d996402c398d2fe4a9e0  (64-bit)
Function Runtime Version: 3.0.15417.0
```

stukalin:
Any news on this one @ankitkumarr ?

stukalin:
@ankitkumarr errr... logging error is nice, but that's not the error causing problems for me (I am kinda reporting issues 2 simultaneously). The most relevant issue for me is the one with `DataProtection`, logging issue just adds some troubles realising what happens.

ankitkumarr:
@stukalin, the `DataProtection` issue is caused because of the logging issue I described above. Every log event is sent to Azure Functions Host (running on a different process) from the worker process running your code via gRPC. Due to the race condition explained above, the log message gets set to Azure Functions Host before `StartStream` message which causes the gRPC channel in host to fail with message such as `Object reference not set to an instance` and the process crashes.

[...]

PR #445, opened 2021-05-13
Fix ordering of grpc stream registration

ankitkumarr:
Fixes https://github.com/Azure/azure-functions-dotnet-worker/issues/425


fabiocav (approved):

brettsam (approved):
This works for this incident -- but we may want to tighten up the Logger (or maybe the grpc comunication itself) to make sure it doesn't try to send anything until StartStream is called.

Can you file a follow-up item to look into that?

ankitkumarr:
@brettsam, that makes sense. But even if the loggers are writing the messages before Start Stream, it should ok as it will be queued up in the `System.Threading.Channel` that will be piped into the grpc channel later.

brettsam:
Yes, that's what I was thinking -- at the grpc level, we can make sure we don't send anything until that start message is sent. That way the logger doesn't even have to know about this.

PR #460, opened 2021-05-24
Tweak extension samples

oscarvantol:
While testing out some of the samples I found a few things in that might be confusing to folks. 

- cleared up naming confusion (queue/table/blob/servicebus)
- fix format of http response headers
- update to the current version of the worker packages
- removed no longer needed custom timerinfo class (MyInfo)

I hope it is of any value.

### DIFF
--- a/samples/Extensions/Table/TableFunction.cs
+++ b/samples/Extensions/Table/TableFunction.cs
@@ -4,28 +4,26 @@
 using System;
 using Microsoft.Azure.Functions.Worker;
 using Microsoft.Extensions.Logging;
-using Newtonsoft.Json.Linq;
 
 namespace SampleApp
 {
     public static class TableFunction
     {
         [Function("TableFunction")]
-        [TableOutput("outputQueue", Connection = "ServiceBusConnection")]
+        [TableOutput("OutputTable", Connection = "AzureWebJobsStorage")]
         public static MyTableData Run([QueueTrigger("table-items")] string input,
-            [TableInput("MyTable", "MyPartition", "{queueTrigger}")] JObject tableItem,
+            [TableInput("MyTable", "MyPartition", "{queueTrigger}")] MyTableData tableInput,
             FunctionContext context)
         {
             var logger = context.GetLogger("TableFunction");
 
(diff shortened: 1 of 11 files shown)


samples/Extensions/Table/TableFunction.cs
```diff
@@ -4,28 +4,26 @@
...
 using Microsoft.Extensions.Logging;
-using Newtonsoft.Json.Linq;
 
 namespace SampleApp
 {
     public static class TableFunction
     {
         [Function("TableFunction")]
-        [TableOutput("outputQueue", Connection = "ServiceBusConnection")]
+        [TableOutput("OutputTable", Connection = "AzureWebJobsStorage")]
```
fabiocav, line 13: Is there a specific reason we need to change the connection string to the default?

@ankitkumarr was the original name something you were trying to have as a sample?

samples/Extensions/Table/TableFunction.cs
```diff
@@ -4,28 +4,26 @@
...
         public static MyTableData Run([QueueTrigger("table-items")] string input,
-            [TableInput("MyTable", "MyPartition", "{queueTrigger}")] JObject tableItem,
```
fabiocav, line 16: @ankitkumarr was this here to demonstrate the ability to use a `JObject` in this context?

fabiocav (commented):
Thank you!

A couple of questions/comments, but this looks good!

samples/Extensions/Table/TableFunction.cs
oscarvantol: For me this did look not intended but more like reused code from a service bus example. It had **"ServiceBusConnection"** as the configuration key name and a ``TableOutput`` attribute with the table name that is **"queue"**. 

_There is a choice not specifying the Connection at all but doing so does show that it can be pointed to any other storage connection than the default._

fabiocav: Indeed! Had overlooked the name/trigger mismatch.

fabiocav (approved):

PR #469, opened 2021-06-02
Applying AZURE_FUNCTIONS_ prefixed variables to host configuration

### LINKED ISSUE #385: Enviroment is not set on entering ConfigureAppConfiguration (.Net Isolated)
The environment is still 'Production' when I enter the ConfigureAppConfiguration((context, builder) => in the HostBuilder, even when I run this locally and I am sure all environment variables are set up properly.

So the following pattern to get Development settings in doesn't work:
```csharp
builder
  .AddCommandLine(args)
  .AddJsonFile(Path.Combine(context.HostingEnvironment.ContentRootPath, "appsettings.json"), optional: true, reloadOnChange: false)
  .AddJsonFile(Path.Combine(context.HostingEnvironment.ContentRootPath, $"appsettings.{context.HostingEnvironment.EnvironmentName}.json"), optional: true, reloadOnChange: false)
  .AddEnvironmentVariables();
```

Printing out the environment variables gives "AZURE_FUNCTIONS_ENVIRONMENT Development" but the value is ignored apparently.
Adding "ASPNETCORE_ENVIRONMENT" and "DOTNET_ENVIRONMENT" to local.settings.json Values also seems to be ignored.

The documentation is not that good on it either. https://docs.microsoft.com/en-us/azure/azure-functions/functions-dotnet-dependency-injection#customizing-configuration-sources references the old (non isolated) way of doing this, and if I try to implement this Startup : FunctionsStartup the IConfigurationBuilder isn't available anymore there (only IConfiguration) so it probably runs too late in the startup process.

fabiocav:
Resolves #385

### DIFF
--- a/test/DotNetWorkerTests/WorkerHostBuilderExtensionsTests.cs
+++ b/test/DotNetWorkerTests/WorkerHostBuilderExtensionsTests.cs
@@ -8,6 +8,7 @@ using Microsoft.Extensions.DependencyInjection;
 using Xunit;
 using System.Linq;
 using Microsoft.Extensions.Configuration.EnvironmentVariables;
+using System;
 
 namespace Microsoft.Azure.Functions.Worker.Tests
 {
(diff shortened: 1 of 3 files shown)


test/DotNetWorkerTests/WorkerHostBuilderExtensionsTests.cs
```diff
@@ -48,5 +49,25 @@ public void EnvironmentVariablesAreRegistered()
...
+                    .ConfigureFunctionsWorkerDefaults()
+                    .Build();
+
+                var environment = host.Services.GetService<IHostEnvironment>();
+
+                Assert.Equal("Development" , environment.EnvironmentName);
+            }
+            finally
+            {
+                Environment.SetEnvironmentVariable("AZURE_FUNCTIONS_ENVIRONMENT", functionsEnvironment);            }
```
brettsam, line 70: nit: the spacing/line-breaks look a little wonky here.

brettsam (approved):
One nit, then good.

PR #507, opened 2021-06-18
WriteAsJsonAsync always return status code OK bug fixes 

### LINKED ISSUE #344: HttpResponseDataExtensions.WriteAsJsonAsync overwrites response status code
The `WriteAsJsonAsync` method sets the status code to OK ([in this line](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/src/DotNetWorker.Core/Http/HttpResponseDataExtensions.cs#L150)).

That behavior is incorrect if the status code is already set. For example, here's a common usage pattern:

```csharp
[Function("MyFunction")]
public async Task<HttpResponseData> Run(
            [HttpTrigger(AuthorizationLevel.Function, "GET", "POST", Route = null)]
            HttpRequestData request) {
  var response = request.CreateResponse(HttpStatusCode.InternalServerError);
  await response.WriteAsJsonAsync(new { message = "Errors occurred!" }, "application/json");
  return response;
}
```

Expected: The response status is 500.
Actual: The response status is 200.

emrekara37:
Hi @fabiocav , @ankitkumarr 
Can you review it when you have time ? 
Resolves  #344

### DIFF
--- a/src/DotNetWorker.Core/Http/HttpResponseDataExtensions.cs
+++ b/src/DotNetWorker.Core/Http/HttpResponseDataExtensions.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Net;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
@@ -77,7 +78,23 @@ namespace Microsoft.Azure.Functions.Worker.Http
         /// <returns>A <see cref="ValueTask"/> that represents the asynchronous operation.</returns>
         public static ValueTask WriteAsJsonAsync<T>(this HttpResponseData response, T instance, CancellationToken cancellationToken = default)
         {
-            return WriteAsJsonAsync(response, instance, "application/json; charset=utf-8", cancellationToken);
+            return WriteAsJsonAsync(response, instance, "application/json; charset=utf-8", HttpStatusCode.OK, cancellationToken);
+        }
+        
+        /// <summary>
+        /// Write the specified value as JSON to the response body using the provided <see cref="ObjectSerializer"/>.
+        /// The response content-type will be set to <code>application/json; charset=utf-8</code> and the status code set to 200.
+        /// </summary>
+        /// <typeparam name="T">The type of object to write.</typeparam>
+        /// <param name="response">The response to write JSON to.</param>
+        /// <param name="instance">The instance to serialize and write as JSON.</param>
+        /// <param name="statusCode">The status code to set on the response.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> used to cancel the operation.</param>
+        /// <returns>A <see cref="ValueTask"/> that represents the asynchronous operation.</returns>
+        public static ValueTask WriteAsJsonAsync<T>(this HttpResponseData response, T instance, HttpStatusCode statusCode,
+            CancellationToken cancellationToken = default)
+        {
+            return WriteAsJsonAsync(response, instance, "application/json; charset=utf-8", statusCode, cancellationToken);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


fabiocav:
@emrekara37 thank you for the PR!

As is, this PR introduces a behavior change that could be breaking for customer relying on the existing functionality (typically, these kinds of behavior changes are not made in a minor version).

An option to address this is to introduce an overload that takes the status code as an additional argument, keeping the default as-is for backwards compatibility. Thoughts?

emrekara37:
> @emrekara37 thank you for the PR!
> 
> As is, this PR introduces a behavior change that could be breaking for customer relying on the existing functionality (typically, these kinds of behavior changes are not made in a minor version).
> 
> An option to address this is to introduce an overload that takes the status code as an additional argument, keeping the default as-is for backwards compatibility. Thoughts?

Hi @fabiocav 

i think you're right,  i can implement overload version if you want

fabiocav:
@emrekara37 if you're open to that, absolutely! That would be great!

emrekara37:
> @emrekara37 if you're open to that, absolutely! That would be great!

I updated the PR

src/DotNetWorker.Core/Http/HttpResponseDataExtensions.cs
```diff
@@ -100,9 +117,35 @@ public static ValueTask WriteAsJsonAsync<T>(this HttpResponseData response, T in
             ObjectSerializer serializer = response.FunctionContext.InstanceServices.GetService<IOptions<WorkerOptions>>()?.Value?.Serializer
                  ?? throw new InvalidOperationException("A serializer is not configured for the worker.");
 
-            return WriteAsJsonAsync(response, instance, serializer, contentType, cancellationToken);
+            return WriteAsJsonAsync(response, instance, serializer, contentType, HttpStatusCode.OK, cancellationToken);
+        }
+      
+        /// <summary>
+        /// Write the specified value as JSON to the response body using the default <see cref="ObjectSerializer"/> configured for this worker.
+        /// The response content-type will be set to the provided <paramref name="contentType"/> and the status code set to 200.
```
fabiocav, line 125: nit: Looks like this needs to be updated, as this will set to the status provided as the parameter.

src/DotNetWorker.Core/Http/HttpResponseDataExtensions.cs
```diff
@@ -100,9 +117,35 @@ public static ValueTask WriteAsJsonAsync<T>(this HttpResponseData response, T in
...
+        /// <returns>A <see cref="ValueTask"/> that represents the asynchronous operation.</returns>
+        public static ValueTask WriteAsJsonAsync<T>(this HttpResponseData response, T instance, string contentType, HttpStatusCode statusCode,
+            CancellationToken cancellationToken = default)
+        {
+            if (response is null)
+            {
+                throw new ArgumentNullException(nameof(response));
+            }
+
+            ObjectSerializer serializer = response.FunctionContext.InstanceServices.GetService<IOptions<WorkerOptions>>()?.Value?.Serializer
```
fabiocav, line 142: Introduce a helper to avoid the code duplication with the method above?

src/DotNetWorker.Core/Http/HttpResponseDataExtensions.cs
```diff
@@ -115,7 +158,42 @@ public static ValueTask WriteAsJsonAsync<T>(this HttpResponseData response, T in
         /// <returns>A <see cref="ValueTask"/> that represents the asynchronous operation.</returns>
         public static ValueTask WriteAsJsonAsync<T>(this HttpResponseData response, T instance, ObjectSerializer serializer, CancellationToken cancellationToken = default)
         {
-            return WriteAsJsonAsync(response, instance, serializer, "application/json; charset=utf-8", cancellationToken);
+            return WriteAsJsonAsync(response, instance, serializer, "application/json; charset=utf-8", HttpStatusCode.OK, cancellationToken);
+        }
+
+        /// <summary>
+        /// Write the specified value as JSON to the response body using the provided <see cref="ObjectSerializer"/>.
+        /// The response content-type will be set to <code>application/json; charset=utf-8</code> and the status code set to 200.
```
fabiocav, line 166: nit: same comment as above about the status code.

src/DotNetWorker.Core/Http/HttpResponseDataExtensions.cs
```diff
@@ -127,9 +205,10 @@ public static ValueTask WriteAsJsonAsync<T>(this HttpResponseData response, T in
         /// <param name="instance">The instance to serialize and write as JSON.</param>
         /// <param name="serializer">The serializer used to serialize the instance.</param>
         /// <param name="contentType">The content-type to set on the response.</param>
+        /// <param name="statusCode">The status code to set on the response.</param>
         /// <param name="cancellationToken">A <see cref="CancellationToken"/> used to cancel the operation.</param>
         /// <returns>A <see cref="ValueTask"/> that represents the asynchronous operation.</returns>
-        public static ValueTask WriteAsJsonAsync<T>(this HttpResponseData response, T instance, ObjectSerializer serializer, string contentType, CancellationToken cancellationToken = default)
+        public static ValueTask WriteAsJsonAsync<T>(this HttpResponseData response, T instance, ObjectSerializer serializer, string contentType, HttpStatusCode statusCode, CancellationToken cancellationToken = default)
```
fabiocav, line 211: Update the comment here to match the status code behavior (same as above)

src/DotNetWorker.Core/Http/HttpResponseDataExtensions.cs
```diff
@@ -77,7 +78,23 @@ public static Task WriteStringAsync(this HttpResponseData response, string value
         /// <returns>A <see cref="ValueTask"/> that represents the asynchronous operation.</returns>
         public static ValueTask WriteAsJsonAsync<T>(this HttpResponseData response, T instance, CancellationToken cancellationToken = default)
         {
-            return WriteAsJsonAsync(response, instance, "application/json; charset=utf-8", cancellationToken);
+            return WriteAsJsonAsync(response, instance, "application/json; charset=utf-8", HttpStatusCode.OK, cancellationToken);
+        }
+        
+        /// <summary>
+        /// Write the specified value as JSON to the response body using the provided <see cref="ObjectSerializer"/>.
+        /// The response content-type will be set to <code>application/json; charset=utf-8</code> and the status code set to 200.
```
fabiocav, line 86: nit: Looks like this needs to be updated, as this will set to the status provided as the parameter.

fabiocav (changes requested):
Thank you!

Small nit comments.

src/DotNetWorker.Core/Http/HttpResponseDataExtensions.cs
emrekara37: > Thank you!
> 
> Small nit comments.

Hi @fabiocav , 
i applied your suggestions

emrekara37:
> Thank you!
> 
> Small nit comments.

Hi @fabiocav,
i updated the PR

fabiocav (approved):

fabiocav:
Thank you!

haacked:
Thanks for resolving #344! ✨

Issue #473, opened 2021-06-03
OpenAPI support for .NET 5


mabvanaartrijk:
Azure Functions in .NET Core 3.1 has OpenAPI support: https://docs.microsoft.com/en-us/azure/azure-functions/openapi-apim-integrate-visual-studio. Is OpenAPI support on the roadmap for .NET 5? Thanks!

cosminstirbu:
I'm also interested in this functionality.

h3x4d3c1m4l:
Maybe this can be helpful: https://github.com/hajekj/azure-functions-openapi-demo

cosminstirbu:
It does, only that it uses `<PackageReference Include="Microsoft.OpenApi.CSharpAnnotations.DocumentGeneration" Version="2.0.0-beta02" />` which hasn't been updated in 2 years and it doesn't look like it's [actively maintained](https://github.com/microsoft/OpenAPI.NET.CSharpAnnotations).

h3x4d3c1m4l:
Mmm haven't paid attention to that. It does work though, just implemented it for a few of our http triggers.

fabiocav:
We're discussing this with the team working on the OpenAPI extension. Will keep this open for tracking, but this work is not currently prioritized/planned, so I can't provide an ETA.

/cc @bradygaster @justinyoo @spboyer

bradygaster:
I do think it would be good for @justinyoo and @spboyer and I to triage and understand what would be needed to support it. I also think with 6 being right around the corner, it would be better for us to pivot to just making sure that works. Frankly, that's probably the best investment choice atm.

spboyer:
@justinyoo - Can you look into the .NET 5/6 story and provide some feedback here?

related issue in extension repo: https://github.com/Azure/azure-functions-openapi-extension/issues/22

justinyoo:
@spboyer I’ve been on it. My challenge is that I have to decouple some WebJobs packages.

justinyoo:
Related: https://github.com/Azure/azure-functions-openapi-extension/pull/142

pnda489:
Maybe
https://github.com/Azure/azure-functions-openapi-extension/tree/main/samples/Microsoft.Azure.Functions.Worker.Extensions.OpenApi.FunctionApp.V3Net5
is what you are Looking for?

soninaren:
[...]

PR #509, opened 2021-06-21
Add extra HttpTriggerAttribute constructors for OpenAPI extension

justinyoo:
This PR is to allow `HttpTriggerAttribute` to set the `AuthorizationLevel` by reading an environment variable, like `%AuthLevel%`.

By accepting this PR, it will enable Azure Functions OpenAPI extension to dynamically set the Swagger UI and OpenAPI document access level in the isolated worker environment.

Reference: https://github.com/Azure/azure-functions-openapi-extension/issues/22

It follows the existing convention to read the connection string for Azure Service Bus trigger, Cosmos DB trigger, etc.

As a fallback, if invalid environment key is supplied, the `AuthorizationLevel` is set to `Function` as a default.

Example:

```csharp
[Function("MyFunction1")]
public HttpResponseData Run(
    [HttpTrigger("%AuthLevel%", "GET", "POST")] HttpRequestData req,
    FunctionContext context)
{
    ...
}
```

### DIFF
--- a/extensions/Worker.Extensions.Http/HttpTriggerAttribute.cs
+++ b/extensions/Worker.Extensions.Http/HttpTriggerAttribute.cs
@@ -92,32 +92,27 @@ namespace Microsoft.Azure.Functions.Worker
         /// </summary>
         public AuthorizationLevel AuthLevel { get; private set; }
 
-        private static AuthorizationLevel GetAuthorizationLevel(string authLevel)
+        private static AuthorizationLevel GetAuthorizationLevel(string authLevelKey)
         {
-            if (string.IsNullOrWhiteSpace(authLevel))
+            if (string.IsNullOrWhiteSpace(authLevelKey))
             {
                 return AuthorizationLevel.Function;
             }
 
-            var trimmed = authLevel.Trim();
+            var trimmed = authLevelKey.Trim();
             if (string.IsNullOrWhiteSpace(trimmed))
             {
                 return AuthorizationLevel.Function;
             }
 
-            if (!trimmed.StartsWith("%"))
-            {
-                return AuthorizationLevel.Function;
-            }
-
-            if (!trimmed.EndsWith("%"))
+            if ((trimmed.StartsWith("%") && trimmed.EndsWith("%")) == false)
             {
                 return AuthorizationLevel.Function;
             }
 
             trimmed = trimmed.Trim('%');
-            var authLevelKey = Environment.GetEnvironmentVariable(trimmed);
-            var authorizationLevel = Enum.TryParse<AuthorizationLevel>(authLevelKey, ignoreCase: true, out var result)
+            var authLevel = Environment.GetEnvironmentVariable(trimmed);
+            var authorizationLevel = Enum.TryParse<AuthorizationLevel>(authLevel, ignoreCase: true, out var result)
                                      ? result
                                      : AuthorizationLevel.Function;
 
--- a/extensions/Worker.Extensions.Http/HttpTriggerAttribute.cs
+++ b/extensions/Worker.Extensions.Http/HttpTriggerAttribute.cs
@@ -41,6 +41,17 @@ namespace Microsoft.Azure.Functions.Worker
 
         /// <summary>
         /// Creates an instance of the <see cref="HttpTriggerAttribute"/>, specifying the
+        /// required <see cref="AuthorizationLevel"/> in string.
+        /// </summary>
+        /// <param name="authLevelKey">The environment variable key to get the string representation of the <see cref="AuthorizationLevel"/> value to apply.</param>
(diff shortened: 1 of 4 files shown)


justinyoo:
Related: #473

extensions/Worker.Extensions.Http/HttpTriggerAttribute.cs
```diff
@@ -67,5 +91,37 @@ public HttpTriggerAttribute(AuthorizationLevel authLevel, params string[] method
...
+
+            if (!trimmed.StartsWith("%"))
+            {
+                return AuthorizationLevel.Function;
+            }
+
+            if (!trimmed.EndsWith("%"))
+            {
+                return AuthorizationLevel.Function;
+            }
```
SeanFeldman, line 116: These two conditional are a single prerequisite - if it's a parameterized variable, `%` should appear at the beginning and at the end. Collapsing the two checks into a single if statement would keep them together.

```suggestion
            if ((trimmed.StartsWith("%") && trimmed.EndsWith("%")) is false)
            {
                return AuthorizationLevel.Function;
            }
```

justinyoo: Thanks for the comment. It's updated.

fabiocav (changes requested):
@justinyoo it would be good to discuss the changes, as functionality like this is handled by the underlying binding, and not the worker extension. 

This type of logic would need to be added to the Functions HTTP trigger, using the proper name resolver. In addition, we need to ensure other components that require use of the function key are able to identify the proper authorization level required to invoke the function, so that needs to be properly resolved in the metadata that is managed by the host.

Also, did you validate the functionality with a runtime change (changing the value of the environment variable with a compiled payload, rather than changing the value prior to compilation)? You'll notice that, since this doesn't rely on just forwarding the reference to the host, runtime resolution will not happen as expected.

justinyoo:
> @justinyoo it would be good to discuss the changes, as functionality like this is handled by the underlying binding, and not the worker extension.

Yes, it's only applied to the `HttpTrigger` binding, which is defined in the `Microsoft.Azure.Functions.Worker.Extensions.Http` extension. Is that what you meant?

> This type of logic would need to be added to the Functions HTTP trigger, using the proper name resolver. In addition, we need to ensure other components that require use of the function key are able to identify the proper authorization level required to invoke the function, so that needs to be properly resolved in the metadata that is managed by the host.

Good point! Please let me know where to look at the metadata resolution part.

> Also, did you validate the functionality with a runtime change (changing the value of the environment variable with a compiled payload, rather than changing the value prior to compilation)? You'll notice that, since this doesn't rely on just forwarding the reference to the host, runtime resolution will not happen as expected.

You're right. Although the authLevel was read from the environment (I set it `Anonymous`), the actual bounded value was different from it (it still recognised as `Function`) when it was deployed to Azure.

To sum up, changing the `HttpTriggerAttribute` itself is not sufficient. Where should I look for more? It would be great if you let me know the point.

Just note that, to make the OpenAPI extension barely working, this change doesn't need. However, for devs to control the extension dynamically, this PR should be applied, once our discussion is over.

fabiocav:
> Yes, it's only applied to the HttpTrigger binding, which is defined in the Microsoft.Azure.Functions.Worker.Extensions.Http extension. Is that what you meant?

No. What I mean is that behavior changes like this one must be made to the actual HTTP Trigger packaged with `Microsoft.Azure.WebJobs.Extensions.Http` so no change in logic here would be required.

> Good point! Please let me know where to look at the metadata resolution part.

No changes to the metadata resolution process would be required, this binding would continue to be a façade and just forward the metadata to the underlying binding. The changes would need to take place in https://github.com/Azure/azure-webjobs-sdk-extensions

> Just note that, to make the OpenAPI extension barely working, this change doesn't need. However, for devs to control the extension dynamically, this PR should be applied, once our discussion is over.

Makes sense, but once the change is made to the HTTP binding, the functionality of the trigger should work as-is, as the metadata parsing and configuration resolution would ultimately be handled by the trigger itself.

justinyoo:
> > Yes, it's only applied to the HttpTrigger binding, which is defined in the Microsoft.Azure.Functions.Worker.Extensions.Http extension. Is that what you meant?
> 
> No. What I mean is that behavior changes like this one must be made to the actual HTTP Trigger packaged with `Microsoft.Azure.WebJobs.Extensions.Http` so no change in logic here would be required.

This part is unclear to me. Let me rephrase.

Although we use this isolated worker extension, `Microsoft.Azure.Functions.Worker.Extensions.Http`, we still need to make the change on `Microsoft.Azure.WebJobs.Extensions.Http` because this isolated worker extension only forwards metadata to that "existing" binding, `Microsoft.Azure.WebJobs.Extensions.Http`.

Is that correct?

justinyoo:
I'm closing this PR for now.

PR #548, opened 2021-07-23
Adding support for custom metric logging via LogMetric

### LINKED ISSUE #543: ILogger missing LogMetric extension method
In Functions v2 (in proc) we had [an extension method defined in `Microsoft.Azure.Webjobs.Host` that allowed logging of AppInsights `metric` types](https://github.com/Azure/azure-webjobs-sdk/blob/206e194e03d82a101a6b5dffb7b4c3904730801d/src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/LoggerExtensions.cs) but this doesn't exist in the new `Worker.*` namespaces. We should bring it forward.

brandonh-msft:
Adds a new `LogMetric()` extension method which sends a new `RpcMetric` type through the gRPC stream to the host. Contains all the information needed to log the metric as a custom metric on the host side.

Changes should be merged only after https://github.com/Azure/azure-functions-host/pull/7593

Fixes #543


brandonh-msft:
This at least provides the extension, but we still need something [like this](https://github.com/Azure/azure-webjobs-sdk/commit/e6babfd91fc03bbe9e534a90c61fab290b85a9ab) to actually log things _as metrics_ in AppInsights.  I couldn't find where in the codebase we're actually sending stuff to `TelemetryClient`

brandonh-msft:
Changes should be merged after https://github.com/Azure/azure-functions-host/pull/7593

brandonh-msft:
😊 worth a shot

src/DotNetWorker.Core/Context/FunctionContextLoggerExtensions.cs
```diff
@@ -31,7 +31,7 @@ public static ILogger GetLogger(this FunctionContext context, string categoryNam
         {
             return context.InstanceServices
                    .GetService<ILoggerFactory>()!
-                   .CreateLogger(categoryName);
+                   .CreateLogger(Logging.LogCategories.CreateFunctionUserCategory(categoryName));
```
liliankasem, line 34: Just wanna understand why we're doing this:

`CreateFunctionUserCategory` -> The category for logs written from within user functions. Are all the logs here user logs? Curious why we need to wrap the categoryName this way

liliankasem (commented):
@brandonh-msft some questions for you!

src/DotNetWorker.Core/Logging/LogConstants.cs
```diff
@@ -0,0 +1,189 @@
...
+namespace Microsoft.Azure.Functions.Worker.Logging.ApplicationInsights
+{
+    // Copyright (c) .NET Foundation. All rights reserved.
+    // Licensed under the MIT License. See License.txt in the project root for license information.
+
+
+    /// <summary>
+    /// Keys used by the <see cref="ILogger"/> infrastructure.
+    /// </summary>
+    public static class LogConstants
```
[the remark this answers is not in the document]
brandonh-msft: This is a bring-forward of work already present in the other Functions Host code; I've a feeling the rest of these are going to start being needed in the future but if you want to trim just for this PR it's your call :)

src/DotNetWorker.Core/Context/FunctionContextLoggerExtensions.cs
brandonh-msft: This was a difference I spotted in the previous .NET Worker code and so made it match our previous logging behavior.

brandonh-msft: Also yes. When you call `.GetLogger()` it's because you're the user and you're going to do some logging.

samples/Configuration/Configuration.csproj
```diff
@@ -9,7 +9,7 @@
     <PackageReference Include="Microsoft.Azure.Core.NewtonsoftJson" Version="1.0.0" />
     <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.0.0" />
     <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http" Version="3.0.12" />
-    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.0.1" OutputItemType="Analyzer" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.0.3" OutputItemType="Analyzer" />
```
liliankasem, line 12: @brettsam Should we bump this to 1.3.0?

brettsam: Yeah, might as well get it up on the latest whenever we see it falling behind.

src/DotNetWorker.Core/Context/FunctionContextLoggerExtensions.cs
```diff
@@ -30,8 +30,8 @@ public static ILogger<T> GetLogger<T>(this FunctionContext context)
         public static ILogger GetLogger(this FunctionContext context, string categoryName)
         {
             return context.InstanceServices
-                   .GetService<ILoggerFactory>()!
-                   .CreateLogger(categoryName);
+                    .GetService<ILoggerFactory>()!
+                    .CreateLogger(Logging.LogCategories.CreateFunctionUserCategory(categoryName));
```
brettsam, line 34: This doesn't seem like something we'd want to do... it'd change the category that someone explicitly passed into us.

src/DotNetWorker.Grpc/GrpcFunctionsHostLogger.cs
```diff
@@ -17,6 +19,8 @@ namespace Microsoft.Azure.Functions.Worker.Logging
     /// </summary>
     internal class GrpcFunctionsHostLogger : ILogger
     {
+        private static readonly ObjectSerializer _serializer = new JsonObjectSerializer();
```
brettsam, line 22: Instead of taking the serializer here -- have it injected via the constructor. That allows a customer to change it (see the Configuration sample in the repo for how they'd do that). They can change from Newtonsoft to System.Text.Json, for example.

liliankasem: @brettsam Done! I followed your pattern of doing it through a set function instead of the constructor in `GrpcFunctionsHostLoggerProvider`, the same way `scopeProvider` is currently being set. Is that okay or should I change that?

Also just curious why `_scopeProvider` wasn't being set in the constructor the same way `_channelWriter` is when it is also a required parameter for `GrpcFunctionsHostLogger` and cannot be null.

liliankasem: answered offline

fabiocav:
Flagging this as blocked for clarity

brandonh-msft:
Need anything from me?

liliankasem:
> Need anything from me?

I don't think so, just waiting for a new host release

src/DotNetWorker.Core/Logging/ILoggerExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Microsoft.Extensions.Logging
+{
+    /// <summary>
+    /// Extensions for <see cref="ILogger"/>.
+    /// </summary>
+    public static class ILoggerExtensions
+    {
+        /// <summary>
+        /// Logs a metric value.
```
kshyju, line 16: Should we explicitly specify that we will be logging as Information level message? Either in the method comment/documentation or the method name?

src/DotNetWorker.Grpc/GrpcFunctionsHostLogger.cs
```diff
@@ -1,101 +1,148 @@
...
+
+                response.RpcLog = rpcLog;
+
+                _channelWriter.TryWrite(response);
+            }
+        }
+
+        private void LogMetric(IDictionary<string, object> state)
+        {
+            if (state != null)
```
kshyju, line 91: nit: invert `if` to reduce nesting.

src/DotNetWorker.Grpc/GrpcFunctionsHostLoggerProvider.cs
```diff
@@ -5,20 +5,33 @@
...
         {
             _channelWriter = outputChannel.Channel.Writer;
+
+            if (workerOptions?.Value?.Serializer != null)
+            {
+                _serializer = workerOptions.Value.Serializer;
+            }
+            else
+            {
+                _serializer = new JsonObjectSerializer();
```
kshyju, line 29: [In other places, we throw](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/src/DotNetWorker.Core/Converters/JsonPocoConverter.cs#L26) when workerOptions.Serializer is `null`.

test/DotNetWorkerTests/Diagnostics/GrpcHostLoggerTests.cs
```diff
@@ -63,7 +106,7 @@ public async Task SystemLog_WithException_AndScope()
                 catch (Exception ex)
                 {
                     // The only way to log a system log.
-                    var log = WorkerMessage.Define<string>(LogLevel.Trace, new EventId(1, "One"), "system log with {param}");
+                    var log = WorkerMessage.Define<string>(LogLevel.Trace, new EventId(-1, "One"), "system log with {param}");
```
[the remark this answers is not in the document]
brandonh-msft: My best guess 6mo later is because the Metric log event's ID is `1` now (see `LogConstants`) and so system log id values needed to be different.

src/DotNetWorker.Grpc/GrpcFunctionsHostLogger.cs
```diff
@@ -1,101 +1,150 @@
...
+                LogCategory = RpcLogCategory.CustomMetric,
+            };
+
+            foreach (var kvp in state)
+            {
+                rpcMetric.PropertiesMap.Add(kvp.Key, kvp.Value.ToRpc(_serializer));
+            }
+
+            // Grab the invocation id from the current scope, if present.
+            _scopeProvider.ForEachScope((scope, log) =>
```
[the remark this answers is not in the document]
brandonh-msft: this code is simply mimicking what's already in `Log` but for metrics; if we could reuse the code in `Log` somehow that's cool

src/DotNetWorker.Core/Logging/LogCategories.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.Functions.Worker.Logging
```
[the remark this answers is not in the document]
brandonh-msft: From what I recall this was used by something else in the pipeline and served to create system vs user log categories.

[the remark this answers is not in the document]
brandonh-msft: found it
https://github.com/Azure/azure-functions-host/blob/852eed9ceef6ef56b431428a8eb31f1bd9c97f3b/src/WebJobs.Script.WebHost/Diagnostics/SystemLoggerProvider.cs#L52

src/DotNetWorker.Grpc/GrpcFunctionsHostLoggerProvider.cs
```diff
@@ -5,20 +5,25 @@
...
     {
         private readonly ChannelWriter<StreamingMessage> _channelWriter;
+        private readonly ObjectSerializer _serializer;
         private IExternalScopeProvider? _scopeProvider;
 
-        public GrpcFunctionsHostLoggerProvider(GrpcHostChannel outputChannel)
+        public GrpcFunctionsHostLoggerProvider(GrpcHostChannel outputChannel, IOptions<WorkerOptions> workerOptions)
         {
             _channelWriter = outputChannel.Channel.Writer;
+            _serializer = workerOptions?.Value?.Serializer ?? throw new InvalidOperationException(nameof(workerOptions.Value.Serializer));
```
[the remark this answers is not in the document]
brandonh-msft: who do we envision "you" is addressing now that this PR is the better part of 1yr old? :)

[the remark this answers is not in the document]
liliankasem: I'll be addressing PR changes, just needed you to answer some of the why questions :)

src/DotNetWorker.Grpc/GrpcFunctionsHostLogger.cs
```diff
@@ -1,101 +1,143 @@
...
+        }
+
+        public bool IsEnabled(LogLevel logLevel)
+        {
+            return logLevel != LogLevel.None;
+        }
+
+        public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
+        {
+            if (eventId == LogConstants.MetricEventId)
```
brettsam, line 48: Can we check the event *name* here instead of the *id*? Then we can remove that `-1` change in the test. Otherwise, I worry this is a breaking change and that anyone currently using an event with id of `1` will now go down this path.

brettsam (approved):

PR #552, opened 2021-07-26
Change queue output sample to multiple messages

ggailey777:


ggailey777:
Whoever reviews and merges this PR, can you please squash all of my crazy commits. I didn't mean for the PR to be this messy for a one file change 😦  

cc. @fabiocav

ggailey777:
I should also mention that the updated code does work 😉

PR #571, opened 2021-08-09
Update SignalR to 1.6.0

### LINKED ISSUE #566: Multiple Azure SignalR Service Instances Support
I would like to use multiples **AzureSignalRConnectionString** to connect with Azure SignalR Service. But I can't do it on Azure Function with **.NET 5**.
I was trying to follow this documentation [Multiple Azure SignalR Service Instances Support in Azure Functions](https://github.com/Azure/azure-functions-signalrservice-extension/blob/dev/docs/sharding.md), but it looks like that works only in a **.NET Core App 3.1** version that uses **Microsoft.Azure.WebJobs.Extensions.SignalRService** package.

Y-Sindo:
- Update SignalR to 1.5.0
- Update SignalR to 1.6.0
close #566


extensions/Worker.Extensions.SignalRService/SignalREndpointsAttribute.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    public sealed class SignalREndpointsAttribute : InputBindingAttribute
```
phantomcosmonaut, line 8: Will this work as a class attribute when applied to this model?
https://docs.microsoft.com/en-us/azure/azure-signalr/signalr-concept-serverless-development-config#negotiate-experience-in-class-based-model

extensions/Worker.Extensions.SignalRService/SignalRNegotiationAttribute.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    public sealed class SignalRNegotiationAttribute : SignalRNegotiationBaseAttribute
```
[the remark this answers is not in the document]
Y-Sindo: This is because `SignalRNegotiateAttribute` shares same properties with `SignalRConnectionInfoAttribute`.

extensions/Worker.Extensions.SignalRService/SignalREndpointsAttribute.cs
Y-Sindo: No.

extensions/Worker.Extensions.SignalRService/SignalRNegotiationAttribute.cs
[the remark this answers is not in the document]
phantomcosmonaut: I realize these two attributes already exist but they appear to be identical just with different names.

Y-Sindo:
@phantomcosmonaut Any update on this?

phantomcosmonaut:
@Y-Sindo I was just snooping around, I don't have anything useful to offer on this issue

fabiocav (commented):
The changes here look good to me, but I wanted to ask a couple of questions:

- Can we take the opportunity to add XML Doc information to the public surface?
- Was this validated with the matching extension

@anthonychu for awareness/review if needed.

extensions/Worker.Extensions.SignalRService/Worker.Extensions.SignalRService.csproj
```diff
@@ -6,7 +6,7 @@
     <Description>Azure SignalR Service extensions for .NET isolated functions</Description>
 
     <!--Version information-->
-    <VersionPrefix>1.2.2</VersionPrefix>
+    <VersionPrefix>1.6.0</VersionPrefix>
 
     <!--Temporarily opting out of documentation. Pending documentation-->
     <GenerateDocumentationFile>false</GenerateDocumentationFile>
```
fabiocav, line 12: One minor change would be this property. With the XML docs in place, this should be enabled.

Y-Sindo, line 12:
```suggestion
    <GenerateDocumentationFile>true</GenerateDocumentationFile>
```

liliankasem:
@fabiocav I think this is ready for a second pass :)

extensions/Worker.Extensions.SignalRService/src/SignalRConnectionInfoInputAttribute.cs
```diff
@@ -1,16 +1,22 @@
...
+    /// public class MyConnectionInfo 
+    /// {
+    /// public string Url { get; set; }
+    /// public string AccessToken { get; set; }
+    /// } 
+    /// </code>
+    /// </para>
+    /// </summary>
+    public sealed class SignalRConnectionInfoInputAttribute : SignalRNegotiationBaseAttribute
     {
```
fabiocav, line 20: This is introducing a breaking change in a public type in a minor version. Intentional? Seems like a local refactor based on the other attribute that was introduced.

fabiocav:
One initial question. Wanted to pause the review until we have a chance to discuss

extensions/Worker.Extensions.SignalRService/src/SignalRConnectionInfoInputAttribute.cs
Y-Sindo: According to doc [Types of breaking changes - .NET](https://docs.microsoft.com/en-us/dotnet/core/compatibility/), this seems not to be a breaking change. But anyway, I recovered it. 
> ALLOWED: Moving a member into a class higher in the hierarchy than the type from which it was removed

fabiocav (approved):

PR #584, opened 2021-08-20
fixing assembly attributes; improving test stability; incrementing version

### LINKED ISSUE #547: Need to fix AssemblyInformationalVersion
Current releases have `AssemblyInformationalVersion` looking like:  `1.3.0-local Commit hash: N/A+fd673155eca2c7654ea69c61e4fa8ee862cda1a8`. The `local` and `N/A` should not be there.

brettsam:
Fixes #547


build/Common.props
```diff
@@ -14,8 +14,6 @@
     <Version Condition=" '$(Version)' == '' ">$(VersionPrefix)$(VersionSuffix)</Version>
     <AssemblyVersion Condition="$(AssemblyVersion) == ''">$(VersionPrefix).0</AssemblyVersion>
     <FileVersion  Condition="$(FileVersion) == ''">$(AssemblyVersion)</FileVersion>
-    <CommitHash Condition="$(CommitHash) == ''">N/A</CommitHash>
-    <InformationalVersion  Condition="$(InformationalVersion) == ''">$(Version) Commit hash: $(CommitHash)</InformationalVersion>
```
brettsam, line 18: This happens automatically now!

fabiocav: What is the exact informational version you see with the changes? The above is consistent with what we do for the other projects.

fabiocav (approved):
One question, but changes look good.

brettsam:
Previous InformationalVersion: `1.4.0-local Commit hash: N/A+dee65e78cbd22ce884dca653f594f0e509c40837`
New InformationalVersion: `1.5.0-20210820.11+f21362df7b9444d81ba93f8433d3e6a988c604a5` (the `-20210820.11` won't be there for an official build)

build/Common.props
brettsam: It's now `1.5.0-20210820.11+f21362df7b9444d81ba93f8433d3e6a988c604a5`, which is auto-set by the .NET SDK if `$(SourceRevisionId)` is present. This seems to come from the SourceLink package we're using. I'm double-checking this now just so I fully understand.

Looking at all .NET packages, this seems to be the standard format now (adding `+`, then the revision).

https://github.com/dotnet/sdk/blob/main/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.GenerateAssemblyInfo.targets#L58-L69

fabiocav:
I see. I think the change is OK, but we should consider the change to the other projects for consistency (though, that could potentially impact consumers as the host actually uses this information in the status response)

build/Common.props
brettsam: Yep -- `SourceRevisionId` gets set here, which comes from the SourceLink package -- https://github.com/dotnet/sourcelink/blob/main/src/Microsoft.Build.Tasks.Git/build/Microsoft.Build.Tasks.Git.targets#L25-L34.

test/SdkE2ETests/xunit.runner.json
```diff
@@ -0,0 +1,5 @@
+{
+  "$schema": "https://xunit.net/schema/current/xunit.runner.schema.json",
+  "parallelizeAssembly": false,
+  "parallelizeTestCollections": false
```
brettsam, line 4: I noticed that locally there were multiple tests trying to run simultaneously, which would explain some of the weird errors we were getting. Tests were trying to build/rebuild the same project at the same time. I think this will fix it.

PR #589, opened 2021-08-20
Change first character of Type property value of metadata bindings entries to lower case

### LINKED ISSUE #555: Metadata generation must convert values to camel casing
The metadata generation build process is not handling the required casing correctly. Currently, the infrastructure expects things like the trigger type to be camel cased (e.g. `httpTrigger` vs `HttpTrigger`).

kshyju:
Fixes #555 

The scaling infrastructure expects the `Type` property value in the `bindings` entries of `functions.metadata` to be in camelCase, for it to work correctly. This PR does that by converting that property value to camel case when generating the JSON content for the metadata file.


sdk/Sdk/Extensions/CustomAttributeExtensions.cs
```diff
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
 using Mono.Cecil;
 
 namespace Microsoft.Azure.Functions.Worker.Sdk
```
kshyju, line 9: Moved 2 existing extension method classes under the new "Extensions" directory.They were in the root of the project earlier.

sdk/Sdk/Extensions/StringExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.Functions.Worker.Sdk
+{
+    internal static class StringExtensions
+    {
+        /// <summary>
+        /// Returns a copy of the string in camelCase.
+        /// </summary>
+        public static string ToCamelCase(this string value)
```
brettsam, line 11: I was looking at how the in-proc SDK does this.... and all it does is lowercase the first letter. I think we should just copy that behavior so that they behave the same.

https://github.com/Azure/azure-functions-vs-build-sdk/blob/v3.x/src/Microsoft.NET.Sdk.Functions.Generator/AttributeExtensions.cs#L32

kshyju: Thanks Brett. Switched to that implementation.

brettsam (approved):

Issue #607, opened 2021-08-30
Async azure functions will be always succes and does not have extra deliveries.


RikShoeby:
With our current azure functions (.net core 5) we have had a big production problem. The azure functions stopped running at the 23-08 for days until we looked into the app and it was starting again. I still have to look into that. Lets focus on the first problem I found.

Luckily all messages where still on the "Dead-letter". We put them again on the "Active" list and the messages where processed. Not all the messages where successful due timeouts of a API. For some reason all the messages where handled and gone instead of doing the 5 retries and there after put it on the dead letter.

Some of the RpcExceptions where unhandled but where not shown in the failed list on the left..
![image](https://user-images.githubusercontent.com/89585999/131315260-d5348b37-8604-4290-b6af-3cf3492cd962.png)

After a lot of testing and creating a new small azure function I found something i did not expect..
The function Test.Test1-async and Test.Test1 are the same with only the async keyword as difference.

```
public class Test1Message
    {
        [JsonPropertyName("id")]
        [Required]
        public int? Id { get; set; }
    }

    public class Test1
    {
        [Function("Test.Test1-async")]
        public async void RunAsync([ServiceBusTrigger("test-test1", "shoeby-order-service", Connection = "ServiceBus")] string mySbMsg, FunctionContext context)
        {
            var logger = context.GetLogger("OrderUpdated");
            logger.LogInformation($"C# ServiceBus topic trigger (test-test1) function processed message: {mySbMsg}");

            var messageModel = JsonSerializer.Deserialize<Test1Message>(mySbMsg);

            if (messageModel?.Id == null)
            {
                throw new ArgumentNullException(nameof(messageModel));
            }


            if (messageModel.Id < 10)
            {
                throw new NotSupportedException("Id must be larger then 10!!");
            }
        }

        [Function("Test.Test1")]
        public void Run([ServiceBusTrigger("test-test1", "shoeby-order-service", Connection = "ServiceBus")] string mySbMsg, FunctionContext context)
        {
            var logger = context.GetLogger("OrderUpdated");
            logger.LogInformation($"C# ServiceBus topic trigger (test-test1) function processed message: {mySbMsg}");

            var messageModel = JsonSerializer.Deserialize<Test1Message>(mySbMsg);

            if (messageModel?.Id == null)
            {
                throw new ArgumentNullException(nameof(messageModel));
            }


            if (messageModel.Id < 10)
            {
[...]

PR #620, opened 2021-09-08
Adding an analyzer & code fixer provider to prevent async void return types

### LINKED ISSUE #610: Add analyzer to catch usages of 'async void'
Functions do not behave as expected when `async void` is used -- we need to catch this at build time and warn via an Analyzer.

kshyju:
Fixes #610 

Adding an analyzer & code fixer to prevent `async void` return types.

async-void-analyzer-demo

- Added Analyzer & code fix provider & tests for these.
- Added the analyzer documentation files.
- Updated minor version number of Worker.Sdk.Analyzers and Worker.Sdk to 1.

Few issues where users were seeing unexpected behavior due to `async void` usage: 

- #607 
- #624

### DIFF
--- a/docs/analyzer-rules/AZFW0002.md
+++ b/docs/analyzer-rules/AZFW0002.md
@@ -1,0 +1,51 @@
+# AZFW0002: Rule title
+
+| | Value |
+|-|-|
+| **Rule ID** |AZFW0002|
+| **Category** |[Usage]|
+| **Severity** |Error|
+
+
+## Cause
+
+This rule is triggered when an async method uses `void` as return type.
+
+## Rule description
+
+When an exception is thrown out of an `async Task` or `async Task<T>` method, that exception is captured and placed on the `Task` object. With `async void` methods, there is no `Task` object, so any exceptions thrown out of an async void method will be posted to thread pool. An unhandled exception on thread pool will cause the whole process to terminate.
+
+Please refer [Avoid Async Void section in Best Practices in Asynchronous Programming](https://docs.microsoft.com/en-us/archive/msdn-magazine/2013/march/async-await-best-practices-in-asynchronous-programming#avoid-async-void) for more info.
+
+## How to fix violations
+
+To fix violations, change the method return type from `async void` to `async Task`
+
+Sample code with violation, before the fix
+
+    [Function("Function1")]
+    public static async void Run([QueueTrigger("myqueue-items",
+                                  Connection = "connstr")] string myQueueItem,
+        FunctionContext context)
+    {
+        await Task.Delay(1);
+        var logger = context.GetLogger("Function1");
+        logger.LogInformation($"Queue trigger function processed: {myQueueItem}");
+    }
+
+Sample code, after the fix
+
+    [Function("Function1")]
+    public static async Task Run([QueueTrigger("myqueue-items",
+                                  Connection = "connstr")] string myQueueItem,
+        FunctionContext context)
+    {
+        await Task.Delay(1);
+        var logger = context.GetLogger("Function1");
+        logger.LogInformation($"Queue trigger function processed: {myQueueItem}");
+    }
+
+
+## When to suppress warnings
+
+This rule should not be suppressed, as the an exception thrown from the async void method would cause the application process to crash.
--- a/docs/analyzer-rules/AZFW0002.md
+++ b/docs/analyzer-rules/AZFW0002.md
@@ -1,4 +1,4 @@
-# AZFW0002: Rule title
+# AZFW0002: Avoid async void
 
 | | Value |
 |-|-|
@@ -13,7 +13,7 @@ This rule is triggered when an async method uses `void` as return type.
 
 ## Rule description
 
-When an exception is thrown out of an `async Task` or `async Task<T>` method, that exception is captured and placed on the `Task` object. With `async void` methods, there is no `Task` object, so any exceptions thrown out of an async void method will be posted to thread pool. An unhandled exception on thread pool will cause the whole process to terminate.
+When an exception is thrown out of an `async Task` or `async Task<T>` method, that exception is captured and placed on the `Task` object. With `async void` methods, there is no `Task` object, so any exceptions thrown out of an async void method will be posted to thread pool. An unhandled exception on the thread pool will cause the whole process to terminate.
 
 Please refer [Avoid Async Void section in Best Practices in Asynchronous Programming](https://docs.microsoft.com/en-us/archive/msdn-magazine/2013/march/async-await-best-practices-in-asynchronous-programming#avoid-async-void) for more info.
 
--- a/docs/analyzer-rules/AZFW0002.md
+++ b/docs/analyzer-rules/AZFW0002.md
@@ -24,8 +24,8 @@ To fix violations, change the method return type from `async void` to `async Tas
 Sample code with violation, before the fix
 
     [Function("Function1")]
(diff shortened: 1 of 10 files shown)


sdk/Sdk/Sdk.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   
   <PropertyGroup>
-    <PatchProductVersion>4</PatchProductVersion>
+    <MinorProductVersion>1</MinorProductVersion>
```
kshyju, line 4: [Current Worker sdk](https://www.nuget.org/packages/Microsoft.Azure.Functions.Worker.Sdk/) version is 1.0.4 . This change will create 1.1.0.

sdk/Sdk.Analyzers/Sdk.Analyzers.csproj
```diff
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <MinorProductVersion>1</MinorProductVersion>
```
kshyju, line 4: [Current version of Worker.Sdk.Analyzers ](https://www.nuget.org/packages/Microsoft.Azure.Functions.Worker.Sdk.Analyzers/)is 1.0.0. This change will create 1.1.0

test/FunctionMetadataGeneratorTests/SdkTests.csproj
```diff
@@ -25,6 +25,8 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.25" />
     <PackageReference Include="Microsoft.Azure.WebJobs.Script.Abstractions" Version="1.0.0-preview" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.11.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.Common" Version="3.11.0" />
```
kshyju, line 29: I had to add these package reference entries with specific version to resolve the version conflict issue.

`SdkTests` depends on `FunctionsMetadataLoaderExtension` which depends on `WebJobs.Script.Abstractions 1.0.0-preview` which pulls in an older version of `Microsoft.CodeAnalysis.CSharp.Workspaces`. This caused the version conflict (with the more recent version of `Microsoft.CodeAnalysis.CSharp.Workspaces` I used in the Analyzer project. (This is a new dependency to analyzer project as we did not have any code fix providers so far).

image

docs/analyzer-rules/AZFW0002.md
```diff
@@ -0,0 +1,51 @@
...
+| **Severity** |Error|
+
+
+## Cause
+
+This rule is triggered when an async method uses `void` as return type.
+
+## Rule description
+
+When an exception is thrown out of an `async Task` or `async Task<T>` method, that exception is captured and placed on the `Task` object. With `async void` methods, there is no `Task` object, so any exceptions thrown out of an async void method will be posted to thread pool. An unhandled exception on the thread pool will cause the whole process to terminate.
```
fabiocav, line 16: This verbiage seems to position exceptions as the main issue associated with a failure to return a `Task`, but the absence of a `Task` impacts more than the ability to handle and deal with exceptions. Without a `Task`, the infrastructure has nothing that provides a representation of the ongoing function execution and completion status, so a number of features may be impacted, causing reliability issues even for functions executing without exceptions.

fabiocav (commented):
Minor  comment

docs/analyzer-rules/AZFW0002.md
fabiocav: How about something like:

> An `async void` method does not provide a `Task` to represent the operation in progress, preventing the Azure Functions Worker and Host from tracking function completion and errors, leading to potential reliability.

Thoughts?

kshyju: Yea, This language is much cleaner. Switched to that. Thank you!

docs/analyzer-rules/AZFW0002.md
```diff
@@ -0,0 +1,51 @@
...
+## How to fix violations
+
+To fix violations, change the method return type from `async void` to `async Task`
+
+Sample code with violation, before the fix
+
+    [Function("Function1")]
+    public static async void Run([QueueTrigger("myqueue-items",
+                                  Connection = "connstr")] string myQueueItem,
+        FunctionContext context)
```
liliankasem, line 29: super nit: can we line this up with the other parameters?

sdk/Sdk.Analyzers/AsyncVoidAnalyzer.cs
```diff
@@ -0,0 +1,34 @@
...
+            context.RegisterSymbolAction(AnalyzeMethod, SymbolKind.Method);
+        }
+
+        private static void AnalyzeMethod(SymbolAnalysisContext symbolAnalysisContext)
+        {
+            var symbol = (IMethodSymbol)symbolAnalysisContext.Symbol;
+
+            if (symbol.IsAsync && symbol.ReturnsVoid)
+            {
+                var diagnostic = Diagnostic.Create(DiagnosticDescriptors.AsyncVoidReturnType, symbol.Locations[0]);
```
liliankasem, line 29: Definitely a nit pick here, but I think it might be nicer to avoid a magic number such as `symbol.Locations[0]` in code; is there a reason this will always be `0`? would it make sense to make this a variable or add a comment around it?

kshyju: Good point. Based on the [docs](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.isymbol.locations?view=roslyn-dotnet-3.11.0), the `ISymbol.Locations` property is an array and it will have more than one entry in cases like partial classes where you will have more than one place where the symbol is located. 

The Location property will have at-least one item in the array all the time (where the symbol we are analyzing is located). In our case, we already narrowed it down to a method and it is guaranteed to be of size 1. I could not find a use case where it will have more than one locations (I am curious about the local function use case. I will investigate that). I will do a bit more research to see there is a more readable API/Extension method available to use. I left a comment in the file for time being, as you suggested.

test/Sdk.Analyzers.Tests/Sdk.Analyzers.Tests/AsyncVoidAnalyzerTests.cs
```diff
@@ -0,0 +1,179 @@
...
+{
+    public sealed class AsyncVoidAnalyzerTests
+    {
+        private const string DiagnosticId = "AZFW0002";
+
+        [Fact]
+        public async Task AnalyzerReportsErrorForAsyncVoid()
+        {
+            string inputCode = @"
+using System.Threading.Tasks;
```
liliankasem, line 21: nit: can we fix the indententation of these so it lines up with the variable they its part of

liliankasem (approved):

fabiocav (approved):
Looks good to me.

test/Sdk.Analyzers.Tests/Sdk.Analyzers.Tests/AsyncVoidAnalyzerTests.cs
kshyju: I see other analyzers use this formatting (existing one in worker and the ones in host). So will stick to that to be consistent.

PR #636, opened 2021-09-23
Making FunctionsToolingSuffix conditional on AzureFunctionsVersion and TargetFramework

### LINKED ISSUE #517: Worker SDK's tooling suffix property needs to be dependent on TFM
Instead of setting [this](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.props#L160) to `net5-isolated` for all versions, we need to make `FunctionsToolingSuffix` property dependent on target framework to reflect the right value (eg- "net6-isolated" for net6.0).

cc: @phil-allen-msft

soninaren:
closes #517

### DIFF
--- a/sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
+++ b/sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
@@ -47,7 +47,7 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
         <Error Condition="$(AzureFunctionsVersion.StartsWith('v2',StringComparison.OrdinalIgnoreCase))" Text="AzureFunctionVersion is set to an incompatible version, Please set it to v3"/>
         <Error Condition="!$(AzureFunctionsVersion.StartsWith('v3',StringComparison.OrdinalIgnoreCase)) And !$(AzureFunctionsVersion.StartsWith('v4',StringComparison.OrdinalIgnoreCase))" Text="AzureFunctionVersion is set to an incompatible version"/>
         <Error Text="TargetFrameWork is invalid. Change the TargetFrameWork to net5.0 or net6.0" Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp' Or ('$(TargetFrameworkVersion)' != 'v6.0' And '$(TargetFrameworkVersion)' != 'v5.0')"/>
-        <Error Condition="'$(_ToolingSuffix)' == ''" Text="Invalid combination of TargetFrameWork and AzureFunctionsVersion is set. Go to fwlink for more information"/>
+        <Error Condition="'$(_ToolingSuffix)' == ''" Text="Invalid combination of TargetFrameWork and AzureFunctionsVersion is set."/>
     </Target>
 
     <Target Name="_FunctionsPostBuild" AfterTargets="Build">
--- a/sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
+++ b/sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
@@ -42,16 +42,16 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
   <UsingTask TaskName="ZipDeployTask"
             AssemblyFile="$(_FunctionsTaskAssemblyFullPath)"/>
 
-    <Target Name="_FunctionsPreBuild" BeforeTargets="Build">
-        <Message Condition="'$(_AzureFunctionsNotSet)' == 'true'" Importance="high" Text="AzureFunctionsVersion not configured in the project. Setting AzureFunctionsVersion to v3"/>
-        <Error Condition="$(AzureFunctionsVersion.StartsWith('v1',StringComparison.OrdinalIgnoreCase))" Text="AzureFunctionVersion is set to an incompatible version, Please set it to v4"/>
-        <Error Condition="$(AzureFunctionsVersion.StartsWith('v2',StringComparison.OrdinalIgnoreCase))" Text="AzureFunctionVersion is set to an incompatible version, Please set it to v3"/>
-        <Error Condition="!$(AzureFunctionsVersion.StartsWith('v3',StringComparison.OrdinalIgnoreCase)) And !$(AzureFunctionsVersion.StartsWith('v4',StringComparison.OrdinalIgnoreCase))" Text="AzureFunctionVersion is set to an incompatible version"/>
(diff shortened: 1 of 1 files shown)


sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
```diff
@@ -12,6 +12,11 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
          xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <PropertyGroup>
+    <_AzureFunctionsNotSet Condition="'$(AzureFunctionsVersion)' == ''">true</_AzureFunctionsNotSet>
+    <AzureFunctionsVersion Condition="'$(AzureFunctionsVersion)' == ''">v3</AzureFunctionsVersion>
+    <_ToolingSuffix Condition="($(AzureFunctionsVersion.StartsWith('v3',StringComparison.OrdinalIgnoreCase)) Or $(AzureFunctionsVersion.StartsWith('v4',StringComparison.OrdinalIgnoreCase))) And '$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(TargetFrameworkVersion)' == 'v5.0'">net5-isolated</_ToolingSuffix>
```
phil-allen-msft, line 17: You should probably set _ToolingSuffix to blank before conditionally setting it.

sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
```diff
@@ -36,7 +41,16 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
...
 
-  <Target Name="_FunctionsPostBuild" AfterTargets="Build">
+    <Target Name="_FunctionsPreBuild" BeforeTargets="Build">
+        <Message Condition="'$(_AzureFunctionsNotSet)' == 'true'" Importance="high" Text="AzureFunctionsVersion not configured in the project. Setting AzureFunctionsVersion to v3"/>
+        <Error Condition="$(AzureFunctionsVersion.StartsWith('v1',StringComparison.OrdinalIgnoreCase))" Text="AzureFunctionVersion is set to an incompatible version, Please set it to v4"/>
+        <Error Condition="$(AzureFunctionsVersion.StartsWith('v2',StringComparison.OrdinalIgnoreCase))" Text="AzureFunctionVersion is set to an incompatible version, Please set it to v3"/>
+        <Error Condition="!$(AzureFunctionsVersion.StartsWith('v3',StringComparison.OrdinalIgnoreCase)) And !$(AzureFunctionsVersion.StartsWith('v4',StringComparison.OrdinalIgnoreCase))" Text="AzureFunctionVersion is set to an incompatible version"/>
+        <Error Text="TargetFrameWork is invalid. Change the TargetFrameWork to net5.0 or net6.0" Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp' Or ('$(TargetFrameworkVersion)' != 'v6.0' And '$(TargetFrameworkVersion)' != 'v5.0')"/>
```
phil-allen-msft, line 49: I *think* MSBuild is case insensitive, but I usually see the property as TargetFramework.
https://docs.microsoft.com/en-us/dotnet/standard/frameworks#how-to-specify-a-target-framework

sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
```diff
@@ -36,7 +41,16 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
...
+        <Error Condition="'$(_ToolingSuffix)' == ''" Text="Invalid combination of TargetFrameWork and AzureFunctionsVersion is set."/>
```
phil-allen-msft, line 50: I think the only case for this (with all of the above checks) is v3 + net6.  Anything else is either valid or caught by the above checks.  Does that match your expectation?

sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
```diff
@@ -36,7 +41,16 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
   <UsingTask TaskName="ZipDeployTask"
             AssemblyFile="$(_FunctionsTaskAssemblyFullPath)"/>
 
-  <Target Name="_FunctionsPostBuild" AfterTargets="Build">
+    <Target Name="_FunctionsPreBuild" BeforeTargets="Build">
+        <Message Condition="'$(_AzureFunctionsNotSet)' == 'true'" Importance="high" Text="AzureFunctionsVersion not configured in the project. Setting AzureFunctionsVersion to v3"/>
+        <Error Condition="$(AzureFunctionsVersion.StartsWith('v1',StringComparison.OrdinalIgnoreCase))" Text="AzureFunctionVersion is set to an incompatible version, Please set it to v4"/>
```
phil-allen-msft, line 46: Should it be AzureFunctionsVersion instead of AzureFunctionVersion? (all error messages)

sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
soninaren: I thought it was set to blank by default

sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
```diff
@@ -36,7 +41,16 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
   <UsingTask TaskName="ZipDeployTask"
             AssemblyFile="$(_FunctionsTaskAssemblyFullPath)"/>
 
-  <Target Name="_FunctionsPostBuild" AfterTargets="Build">
+    <Target Name="_FunctionsPreBuild" BeforeTargets="Build">
```
phil-allen-msft, line 44: It looks like the new line is indented differently than the old line; is it what you expect in the real file?

sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
soninaren: yes that is correct. It is also there for any new version that might pop up later like `v3 + net 7` or `v5 + net 6`

sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
soninaren: `AzureFunctionsVersion` is the property name in the csproj file. Wanted to match the property name in the error message.

sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
soninaren: nope that was an unexpected indentation from working in different editors. Have fixed this.

liliankasem (approved):

PR #616, opened 2021-09-07
updating FunctionApp for local debugging

brettsam:
The project FunctionApp wasn't set up correctly to do local debugging anymore. Now that the F5 experience for the worker includes injecting the STARTUPHOOK environment variable, the launchsettings.json was no longer necessary (and didn't seem to work).

PR #642, opened 2021-09-27
Populating Cookies using requestData headers

satvu:
First iteration of populating cookies in the HttpRequestData using the headers. 

- Lazy Initialization 
- Tested with no cookies, single cookie, and multiple (3) cookies
- Sample apps edited for testing the issue and will be reverted to original state before merging 
- Edited .csproj and launch files to set up a working debugging environment (will revert these changes back before merging)


src/DotNetWorker.Core/Http/HttpCookie.cs
```diff
@@ -10,6 +11,25 @@ namespace Microsoft.Azure.Functions.Worker.Http
     /// </summary>
     public sealed class HttpCookie : IHttpCookie
     {
+        private KeyValuePair<string, string> c;
```
satvu, line 14: Changes in this file were added/populated automatically - I didn't touch this file.

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
```diff
@@ -20,11 +20,50 @@ internal class GrpcHttpRequestData : HttpRequestData, IAsyncDisposable, IDisposa
...
+
+                // this produces either an empty list (no cookie) or a list with a single KeyValuePair
+                // in the format of ("Cookie", "cookie_1=value;cookie_2=value...")
+                var cookieList = _headers.ToLookup(item => item.Key)["Cookie"].ToList();
+
+                if (cookieList.Count > 0)
+                {
+                    // cookieList should have only one KeyValuePair, where the Value is all of the cookies concatenated
+                    // together as a string (ex. "Cookie_1=value;Cookie_2=value;Cookie_3=value")
+                    return cookieList.ConvertAll(new Converter<KeyValuePair<string, IEnumerable<string>>, IReadOnlyCollection<IHttpCookie>>(CookieStringsToHttpCookie)).First();
```
satvu, line 44: Is there a better way to handle the format the cookies come in from the headers? They are all sent in a long string with all the cookies concatenated together.

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
```diff
@@ -20,11 +20,50 @@ internal class GrpcHttpRequestData : HttpRequestData, IAsyncDisposable, IDisposa
...
+            _cookies = new Lazy<IReadOnlyCollection<IHttpCookie>>(() =>
+            {
+                if(_headers is null)
+                {
+                    return new List<IHttpCookie>();
+                }
+
+                // this produces either an empty list (no cookie) or a list with a single KeyValuePair
+                // in the format of ("Cookie", "cookie_1=value;cookie_2=value...")
+                var cookieList = _headers.ToLookup(item => item.Key)["Cookie"].ToList();
```
fabiocav, line 38: We should be able to use `FirstOrDefault` here to get the cookie header.

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
fabiocav: With the proposed change above, you'll end up with the actual string at this point.

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
```diff
@@ -20,11 +20,50 @@ internal class GrpcHttpRequestData : HttpRequestData, IAsyncDisposable, IDisposa
...
+
+                return new List<IHttpCookie>();
+               
+            });
+        }
+
+        public static IReadOnlyCollection<IHttpCookie> CookieStringsToHttpCookie(KeyValuePair<string, IEnumerable<string>> cookieStrings)
+        {
+            // cookieStrings.Value comes in the format "Cookie_1=value;Cookie_2=value;Cookie_3=value"
+            var separateCookies = cookieStrings.Value.First().ToString().Split(";");
```
fabiocav, line 55: With the changes proposed above, you should be able to modify this method to work directly with the string.

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
```diff
@@ -20,11 +20,50 @@ internal class GrpcHttpRequestData : HttpRequestData, IAsyncDisposable, IDisposa
         private HttpHeadersCollection? _headers;
         private Stream? _bodyStream;
         private bool _disposed;
+        private Lazy<IReadOnlyCollection<IHttpCookie>> _cookies;
```
fabiocav, line 23: We should be able to make this `readonly`

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
```diff
@@ -20,11 +20,50 @@ internal class GrpcHttpRequestData : HttpRequestData, IAsyncDisposable, IDisposa
...
 
         public GrpcHttpRequestData(RpcHttp httpData, FunctionContext functionContext)
             : base(functionContext)
         {
             _httpData = httpData ?? throw new ArgumentNullException(nameof(httpData));
+            _cookies = new Lazy<IReadOnlyCollection<IHttpCookie>>(() =>
```
fabiocav, line 29: Nice to have this deferred!

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
```diff
@@ -20,11 +20,50 @@ internal class GrpcHttpRequestData : HttpRequestData, IAsyncDisposable, IDisposa
...
+               
+            });
+        }
+
+        public static IReadOnlyCollection<IHttpCookie> CookieStringsToHttpCookie(KeyValuePair<string, IEnumerable<string>> cookieStrings)
+        {
+            // cookieStrings.Value comes in the format "Cookie_1=value;Cookie_2=value;Cookie_3=value"
+            var separateCookies = cookieStrings.Value.First().ToString().Split(";");
+
+            List<IHttpCookie> httpCookiesList = new List<IHttpCookie>();
```
kshyju, line 57: There is an overload which takes the capacity of the list. If you know how many items you are going to add to the list before hand, you can use that overload. This will prevent some array allocation & copying (_List internally allocates a small array first and when you add items more than that capacity, List will create a new array, double the size of current capacity and copy the elements from first array to the new array. This is how pretty much all dynamic collections auto grows._)

`new List<IHttpCookie>(separateCookies.Length);`

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
```diff
@@ -20,11 +20,50 @@ internal class GrpcHttpRequestData : HttpRequestData, IAsyncDisposable, IDisposa
...
+
+            for(int c = 0; c < separateCookies.Length; c++)
+            {
+                var name = separateCookies[c].Split("=")[0];
+                var value = separateCookies[c].Split("=")[1];
```
kshyju, line 62: Do not call `Split` on `separateCookies[c]` twice(single call to `Split` allocates an array). Call it once and store the result to a variable and access the items. 

```
var splitArray = separateCookies[c].Split("=");
var name = splitArray[0];
var value = splitArray[1];
```

samples/FunctionApp/Function3/Function3.cs
```diff
@@ -1,34 +1,33 @@
...
 using Microsoft.Azure.Functions.Worker.Http;
+using Microsoft.Extensions.Logging;
 
 namespace FunctionApp
 {
     public static class Function3
     {
         [Function("Function3")]
-        public static MyOutputType Run([HttpTrigger(AuthorizationLevel.Anonymous, "get", "post", Route = null)] HttpRequestData req,
+        public static string Run([HttpTrigger(AuthorizationLevel.Anonymous, "get", "post", Route = null)] HttpRequestData req,
```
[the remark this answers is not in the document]
satvu: I was going to revert this back since I was using it to test, but I can do that + make a new sample.

samples/FunctionApp/Properties/serviceDependencies.json
```diff
@@ -1,8 +0,0 @@
-{
```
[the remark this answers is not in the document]
satvu: https://github.com/Azure/azure-functions-dotnet-worker/pull/616/files#diff-5ed6fe086b82963a1fc843c092daf5ee6e9c266a2c270cdfdd864380568d5f4a

I was making changes from this PR to allow local debugging - will revert when I no longer need to debug it.

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
```diff
@@ -20,11 +20,48 @@ internal class GrpcHttpRequestData : HttpRequestData, IAsyncDisposable, IDisposa
...
         public GrpcHttpRequestData(RpcHttp httpData, FunctionContext functionContext)
             : base(functionContext)
         {
             _httpData = httpData ?? throw new ArgumentNullException(nameof(httpData));
+            _cookies = new Lazy<IReadOnlyCollection<IHttpCookie>>(() =>
+            {
+                if(_headers is null)
+                {
+                    return Array.Empty<IHttpCookie>();
+                }
```
kshyju, line 34: Wouldn't this cause it to always return an empty array because the `_headers` property will be null until the `Headers` (public) property is accessed ? Did you mean to use `Headers` here instead of `_headers` ?

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
```diff
@@ -20,11 +20,48 @@ internal class GrpcHttpRequestData : HttpRequestData, IAsyncDisposable, IDisposa
...
+
+                var cookieString = _headers.FirstOrDefault(item => item.Key == "Cookie").Value;
```
kshyju, line 36: Should this be a case insensitive check so that it works for all cases (ex: COOKIE, cookie, Cookie) ? In ASP.NET core, you can read Cookies from HttpContext irrespective of the case of the request header .

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
```diff
@@ -20,11 +20,48 @@ internal class GrpcHttpRequestData : HttpRequestData, IAsyncDisposable, IDisposa
...
+
+                if (cookieString != null && cookieString.Count() > 0)
+                {
```
kshyju, line 39: Use `.Any()` instead of `.Count() > 0`

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
kshyju: Or the host is guaranteed to send this it in this case(`Cookie`) irrespective of what the client request header casing is ?

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
```diff
@@ -20,11 +20,48 @@ internal class GrpcHttpRequestData : HttpRequestData, IAsyncDisposable, IDisposa
...
+                    return ToHttpCookies(cookieString);
+                }
+
+                return Array.Empty<IHttpCookie>();
+
+
+            });
+        }
+
+        private IReadOnlyCollection<IHttpCookie> ToHttpCookies(IEnumerable<string> cookieString)
```
kshyju, line 50: nit: arg name sounds singular while the type is a collection. May be you can change the type to string and pass the first item in line 41 ?

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
```diff
@@ -20,11 +20,48 @@ internal class GrpcHttpRequestData : HttpRequestData, IAsyncDisposable, IDisposa
...
+                var cookieString = _headers.FirstOrDefault(item => item.Key == "Cookie").Value;
+
+                if (cookieString != null && cookieString.Count() > 0)
+                {
+
+                    return ToHttpCookies(cookieString);
+                }
+
+                return Array.Empty<IHttpCookie>();
+
```
kshyju, line 45: nit: Extra line break.

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
```diff
@@ -20,11 +20,47 @@ internal class GrpcHttpRequestData : HttpRequestData, IAsyncDisposable, IDisposa
...
+                {
+                    return Array.Empty<IHttpCookie>();
+                }
+
+                var cookieString = Headers.FirstOrDefault(item => String.Equals(item.Key, "Cookie")).Value;
+
+                if (cookieString != null && cookieString.Any())
+                {
+
+                    return ToHttpCookies(cookieString.First().ToString());
```
kshyju, line 41: Is the `ToString()` call really needed ? `cookieString.First()` is returning a string.

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
```diff
@@ -20,11 +20,47 @@ internal class GrpcHttpRequestData : HttpRequestData, IAsyncDisposable, IDisposa
...
         private bool _disposed;
+        private readonly Lazy<IReadOnlyCollection<IHttpCookie>> _cookies;
 
         public GrpcHttpRequestData(RpcHttp httpData, FunctionContext functionContext)
             : base(functionContext)
         {
             _httpData = httpData ?? throw new ArgumentNullException(nameof(httpData));
+            _cookies = new Lazy<IReadOnlyCollection<IHttpCookie>>(() =>
+            {
+                if(Headers is null)
```
kshyju, line 31: nit: Add a whitespace after `if`. (Visual studio has a"_fix formatting_" quick action item which will do these kind of things for you. Keep cursor here and press Ctrl + . )

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
kshyju: FYI: I did a bit of debugging and I see that the cookie header is always received as "Cookie" irrespective of the case sent from the client for the header name.

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
```diff
@@ -66,7 +102,13 @@ public override Stream Body
 
         public override HttpHeadersCollection Headers => _headers ??= new HttpHeadersCollection(_httpData.NullableHeaders.Select(h => new KeyValuePair<string, string>(h.Key, h.Value.Value)));
```
kshyju, line 103: Unrelated question: Should we consider brining the lazy loading behavior to Headers as well ?

kshyju (approved):

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
```diff
@@ -20,11 +20,47 @@ internal class GrpcHttpRequestData : HttpRequestData, IAsyncDisposable, IDisposa
...
+
+                    return ToHttpCookies(cookieString.First().ToString());
+                }
+
+                return Array.Empty<IHttpCookie>();
+
+            });
+        }
+
+        private IReadOnlyCollection<IHttpCookie> ToHttpCookies(string cookieString)
```
kshyju, line 49: nit: Consider moving the private method down, after all public members.

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
satvu: @fabiocav @brettsam

PR #643, opened 2021-09-27
Adding support for custom converters

### LINKED ISSUE #612: Add support for wiring up custom IConverter implementations
Add support for wiring up custom `IConverter` implementations. We currently ships 7 built-in converters to handle majority of parameter binding use cases. With this change, users will be able to write their own `IConverter` implementations as needed. 

Currently `IConverter` is an internal type. We will expose this as public along with other related types to support this feature.

kshyju:
resolves #612

Changes for custom input converter support.  Allows customers to plugin their converter implementation to the pipeline to customize how function input parameters are populated.

Major changes:

1. Exposed an `InputConverters` property to WorkerOptions. Customers will use this property to add their custom converters when bootstrapping the app.

2. Added a new `InputConversionFeature`. This feature is responsible for finding the appropriate converter to be used and does the conversion. This feature uses the 7 built-in converters along with any other custom converters customer has defined.

3. The existing `DefaultBindingFeature` does not directly work with the converters now. Instead, it uses the new `InputConversionFeature` implementation to do the conversion.

#### How does function app developers use this?
Customers can create a new custom input converter by implementing the `IInputConverter` interface.

    public interface IInputConverter
    {
        ValueTask<ConversionResult> ConvertAsync(ConverterContext context);
    }
The `ConverterContext` instance will have information such as target type and source data (input data/ trigger data).

The custom converter can be hooked up to the function app in any of the following 3 approaches.

### 1) App level
Add the new converter to `WorkerOptions.InputConverters` while bootstrapping the app.

```
var host = new HostBuilder()
            .ConfigureFunctionsWorkerDefaults((workerOptions) =>
            {
                // Add MyCustomConverter to the pipeline
                workeroptions.InputConverters.Register<MyCustomConverter>();
            })
            .Build();

await host.RunAsync();
```
This option also allows the customers to insert this converter in a specific position or clear all existing built-in converters etc.

##### Adding at a specific index:
````
// Adds MyCustomConverter to the set of available (built-in) converters, but at index 1. 
workeroptions.InputConverters.RegisterAt<MyCustomConverter>(1);
````
##### Clearing existing registered(7 built-in) converters and adding custom converters
````
// Adds MyCustomConverter to the set of available (built-in) converters, but at index 1. 
workeroptions.InputConverters.Clear();
workeroptions.InputConverters.Register<MyCustomConverter>();
workeroptions.InputConverters.Register<MyOtherCustomConverter>();
````

### 2) Type level
Decorate the type used as the parameter of the function with `InputConverter` attribute where you will define the custom converter type to be used.

```
[InputConverter(typeof(MyCustomConverter))]
public sealed class Product
{
    public int Id { set; get; }
    public string Name { set; get; }
}
```
and

```
[Function("products")]
public HttpResponseData Run(
                 [HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = null)] HttpRequestData req,
                 Product product)
{
   ....    
}
```

### 3) Function parameter level
Decorate the function parameter using the `InputConverter` attribute

```
[Function("products")]
public HttpResponseData Run(
                 [HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = null)] HttpRequestData req,
                 [InputConverter(typeof(MyCustomConverter))] ProductViewModel productvm)
{
   ....    
}
```

### CODE REVIEW

File: src/DotNetWorker.Core/Context/Features/DefaultInputConversionFeature.cs [hunk 1]
```diff
@@ -0,0 +1,116 @@
...
+using Microsoft.Azure.Functions.Worker.Converters;
+using Microsoft.Azure.Functions.Worker.Core.Converters;
+using Microsoft.Azure.Functions.Worker.Core.Converters.Converter;
+
+namespace Microsoft.Azure.Functions.Worker.Context.Features
+{
+    /// <summary>
+    /// Default implementation of <see cref="IInputConversionFeature"/>
+    /// </summary>
+    internal sealed class DefaultInputConversionFeature : IInputConversionFeature
```
kshyju, line 16: This is the new feature added. This is being called from DefaultModelBindingFeature (for each parameter conversion)

src/DotNetWorker.Core/Context/Features/DefaultInputConversionFeature.cs
```diff
@@ -0,0 +1,116 @@
...
+        /// </summary>
+        /// <param name="context">The converter context.</param>
+        /// <returns>An IInputConverter instance or null</returns>
+        private IInputConverter? GetConverterFromContext(ConverterContext context)
+        {
+            Type? converterType = default;
+
+            // Check a converter is specified on the conversionContext.Properties. If yes,use that.
+            if (context.Properties != null
+                && context.Properties.TryGetValue(PropertyBagKeys.ConverterType, out var converterTypeFullAssemblyNameObj))
```
kshyju, line 93: DefaultModelBinderFeature sets this value.

src/DotNetWorker.Core/Context/Features/DefaultInputConversionFeature.cs
```diff
@@ -0,0 +1,116 @@
...
+            {
+                converterType = Type.GetType(converterTypeFullAssemblyNameObj.ToString());
```
kshyju, line 95: A potential optimization is a cache for this fully qualified assembly name to the type so that we do not need to make `GetType` every time.

test/DotNetWorkerTests/Converters/DefaultBinderTests.cs
```diff
@@ -1,56 +1,56 @@
...
-
-            var targetEnum = TestUtility.AssertIsTypeAndConvert<IReadOnlyList<Book>>(target);
-            Assert.Collection(targetEnum,
-                p => Assert.True(p.Id == "1" && p.Author == "a"),
-                p => Assert.True(p.Id == "2" && p.Author == "c"),
-                p => Assert.True(p.Id == "3" && p.Author == "e"));
-        }
-    }
-}
+﻿//// Copyright (c) .NET Foundation. All rights reserved.
```
kshyju, line 1: I will add tests for DefaultBinder in a future iteration.

src/DotNetWorker.Grpc/Definition/GrpcFunctionDefinition.cs
```diff
@@ -51,5 +54,23 @@ public GrpcFunctionDefinition(FunctionLoadRequest loadRequest, IMethodInfoLocato
...
+            Dictionary<string, object> properties = new();
+
+            // Get the input converter attribute information, if present on the parameter.
+            var inputConverterAttribute = parameterInfo?.GetCustomAttribute<InputConverterAttribute>();
+
+            if (inputConverterAttribute != null)
+            {
+                properties = new Dictionary<string, object>()
+                {                     
+                    { PropertyBagKeys.ConverterType, inputConverterAttribute.ConverterType.AssemblyQualifiedName }
```
kshyju, line 69: Can't pass the `System.Type` (inputConverterAttribute.ConverterType) here in the properties dictionary. This is because we serialize the function definition in the [OnFunctionLoaded method](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/src/DotNetWorker.Grpc/Http/GrpcWorkerDiagnostics.cs#L68). An object with a property of type `System.Type` cannot be serialized because accessing the `DeclaringMethod` prop of a System.Type instance just throws.

The below code will throw.
```
var myInstance = new ClasWithTypeProperty();
myInstance.TheData = typeof(Customer);

var str = System.Text.Json.JsonSerializer.Serialize(myInstance);
```


This is why I have to recreate the type using `Type.GetType` from the fully qualified name inside the DefaultInputConversionFeature.

src/DotNetWorker.Core/Context/Features/DefaultInputConversionFeature.cs
```diff
@@ -0,0 +1,118 @@
...
+                    converterType = ((InputConverterAttribute)converterAttribute).ConverterType;
+                }
+            }
+
+            if (converterType != null)
+            {
+                return this._inputConverterProvider.GetOrCreateConverterInstance(converterType);
+            }
+
+            return null;
```
liliankasem, line 115: Just wondering why we return null here but default in other instances? (I'm assuming that default is the keyword and not a defined variable that I missed?)

src/DotNetWorker.Core/Converters/Converter/ConversionResult.cs
```diff
@@ -0,0 +1,50 @@
...
+        public ConversionResult(bool isSuccess, object? model)
+        {
+            IsSuccess = isSuccess;
+            Model = model;
+        }
+
+        /// <summary>
+        /// Creates a new <see cref="ConversionResult"/> instance.
+        /// </summary>
+        public ConversionResult(object model)
```
liliankasem, line 32: Curious when this instance is used and why the default for success is true?

src/DotNetWorker.Core/Converters/Converter/DefaultConverterContext.cs
```diff
@@ -1,23 +1,48 @@
...
+    public sealed class DefaultConverterContext : ConverterContext
     {
-        public DefaultConverterContext(FunctionParameter parameter, object? source, FunctionContext context)
+        /// <summary>
+        /// Creates a new instance of <see cref="DefaultConverterContext"/>
+        /// </summary>
+        /// <param name="targetType">The target type.</param>
+        /// <param name="source">The source.</param>
+        /// <param name="context">The function context.</param>
+        public DefaultConverterContext(Type targetType, object? source, FunctionContext context)
```
liliankasem, line 20: nit: is it better practice to put the optional parameter at the end, or it doesn't matter?

src/DotNetWorker.Core/Converters/Converter/DefaultInputConverterProvider.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        /// <summary>
+        /// Stores the default converter instances.
+        /// This is an ordered sub set of what is present in _converterCache.
+        /// </summary>
+        private IReadOnlyList<IInputConverter> _defaultConverters;
+
+        public DefaultInputConverterProvider(IOptions<WorkerOptions> workerOptions, IServiceProvider serviceProvider)
+        {
+            _workerOptions = workerOptions.Value ?? throw new ArgumentNullException(nameof(workerOptions));
```
liliankasem, line 38: Would this work here?

```suggestion
            _workerOptions = workerOptions?.Value ?? throw new ArgumentNullException(nameof(workerOptions));
```

src/DotNetWorker.Core/Converters/Converter/InputConverterAttribute.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// <exception cref="InvalidOperationException">Thrown when the converterType parameter value is a type which has not implemented Microsoft.Azure.Functions.Worker.Converters.</exception>
+        public InputConverterAttribute(Type converterType)
+        {
+            if (converterType == null)
+            {
+                throw new ArgumentNullException(nameof(converterType));
+            }
+
+            var interfaceType = typeof(IInputConverter);
+            if (interfaceType.IsAssignableFrom(converterType) == false)
```
liliankasem, line 37: If `IsAssignableFrom` returns a boolean, do we need to add the `== false` part or can we just do `!interfaceType.IsAssignableFrom(converterType)`

src/DotNetWorker.Core/Converters/JsonPocoConverter.cs
```diff
@@ -28,13 +29,11 @@ public JsonPocoConverter(IOptions<WorkerOptions> options)
...
-        public bool TryConvert(ConverterContext context, out object? target)
+        public async ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
         {
-            target = default;
-
-            if (context.Parameter.Type == typeof(string))
+            if (context.TargetType == typeof(string))
             {
-                return false;
+                return ConversionResult.Failed();
```
liliankasem, line 36: Just a lack of context question: Why don't we want to convert items of type `string`? Is it just that this is specifically an object converter and if for some reason we get a string we're like "no thanks, wrong type!"

liliankasem (commented):
Mostly just questions, but happy to do another review after building a bit more context on the feature

src/DotNetWorker.Core/Context/Features/DefaultInputConversionFeature.cs
kshyju: Yes, it is the `default` keyword. `return default` will also work here because `return default` will be lowered to `return null` in this case. I chose `return null` as I felt that is more readable in the case of this method.

I updated the other method which was doing `return default` so that it is more readable what it returns.

src/DotNetWorker.Core/Converters/Converter/ConversionResult.cs
kshyju: Good catch. 😄 This is unused and now removed.

src/DotNetWorker.Core/Converters/Converter/DefaultConverterContext.cs
kshyju: If you meant the `source` parameter, it is not an optional parameter. It is nullable, So the caller must provide a value(null or non null) for this parameter.

src/DotNetWorker.Core/Converters/Converter/DefaultInputConverterProvider.cs
kshyju: Good catch. Fixed in the latest iteration.

src/DotNetWorker.Core/Converters/Converter/InputConverterAttribute.cs
kshyju: Compiler will lower `==false` to `!`. I fixed here as well now though.

src/DotNetWorker.Core/Converters/JsonPocoConverter.cs
kshyju: We are checking the `TargetType`, not source type. If the TargetType is string, we already have another converter (TypeConverter) which will do that conversion. That converter is registered in the order to be used before this one.

test/DotNetWorkerTests/Converters/DefaultBinderTests.cs
kshyju: I moved the tests from DefaultBinderTests to DefaultInputConversionTests.

src/DotNetWorker.Core/Converters/JsonPocoConverter.cs
liliankasem: Ahh that makes sense, thank you :)

src/DotNetWorker.Core/Converters/Converter/DefaultInputConverterProvider.cs
```diff
@@ -0,0 +1,108 @@
...
+        /// Gets an instance of the converter for the type requested.
+        /// </summary>
+        /// <param name="converterType">The type of IConverter implementation to return.</param>
+        /// <returns>IConverter instance of the requested type.</returns>
+        public IInputConverter GetOrCreateConverterInstance(Type converterType)
+        {
+            if (converterType == null)
+            {
+                throw new ArgumentNullException((nameof(converterType)));
+            }
```
brettsam, line 57: Should we check that this implements `IInputConverter`?

src/DotNetWorker.Core/Converters/Converter/IConverterContextFactory.cs
```diff
@@ -0,0 +1,22 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.Functions.Worker.Converters
+{
+    /// <summary>
+    /// An abstraction to create <see cref="ConverterContext"/> instances.
+    /// </summary>
+    public interface IConverterContextFactory
```
brettsam, line 11: Does this need to be public in this pass?

brettsam:
First round of comments!

brettsam (commented):
First round of comments!

src/DotNetWorker.Core/Converters/Converter/ConversionResult.cs
```diff
@@ -0,0 +1,83 @@
...
+        
+        /// <summary>
+        /// Gets a value indicating whether the converter acted on the input to execute a conversion operation.
+        /// </summary>
+        public bool IsHandled { get; }
+
+        /// <summary>
+        /// Gets a value indicating whether the conversion operation was successful or not.
+        /// </summary>
+        public bool? IsSuccessful { get; }
```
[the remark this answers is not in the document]
kshyju: Yes. Those 3 are the valid states. 

Currently we expose the 3 static methods ( `Unhandled`, `Success` and `Failure`) on this type to create those 3 variations. The fourth option, one can create is using the default constructor (`new ConversionResult()`), But the properties of this instance cannot be set as they can be only set using the above mentioned methods. So essentially that newed up instance will be same as the result of `ConversionResult.Unhandled()`

Another option to reduce the confusion is to move the `IsHandled` to `IInputConverter`. So the caller first needs to explicitly check the converter can handle before calling `ConvertAsync`. In this way the `ConversionResult` only represents the real result, not a capability of the converter. feels more clean.

    if (converter.CanHandle(converterContext))
    {
        var conversionResultTask = converter.ConvertAsync(converterContext);
        ....
    }

src/DotNetWorker.Core/Converters/Converter/DefaultInputConverterProvider.cs
kshyju: Good call. Added that check now.

src/DotNetWorker.Core/Context/Features/DefaultInputConversionFeature.cs
```diff
@@ -0,0 +1,115 @@
...
+        private IInputConverter? GetConverterFromContext(ConverterContext context)
+        {
+            Type? converterType = default;
+
+            // Check a converter is specified on the conversionContext.Properties. If yes,use that.
+            if (context.Properties != null
+                && context.Properties.TryGetValue(PropertyBagKeys.ConverterType, out var converterTypeAssemblyQualifiedNameObj)
+                && converterTypeAssemblyQualifiedNameObj is string converterTypeAssemblyQualifiedName)
+            {
+                converterType = Type.GetType(converterTypeAssemblyQualifiedName);
```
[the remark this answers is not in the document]
kshyju: Made changes so that this type name string is now passed to `DefaultInputConverterProvider` and it's internal cache is used.

src/DotNetWorker.Core/Context/Features/DefaultInputConversionFeature.cs
```diff
@@ -0,0 +1,115 @@
...
+            }
+            else
+            {
+                // check the type used as "TargetType" has an "InputConverter" attribute decoration.
+                var converterAttribute = context.TargetType.GetCustomAttributes(_inputConverterAttributeType, inherit: true)
```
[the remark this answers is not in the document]
kshyju: Good call. Added a cache look up in the latest iteration.

src/DotNetWorker.Core/Context/Features/DefaultModelBindingFeature.cs
```diff
@@ -46,24 +46,35 @@ public DefaultModelBindingFeature(IEnumerable<IConverter> converters)
...
 
-                var converterContext = new DefaultConverterContext(param, source, context);
+                var converterContext = new DefaultConverterContext(param.Type, source, context);
+
+                if (param.Properties != null)
+                {
+                    // Pass info about specific input converter type defined for this parameter if present.
+                    if (param.Properties.TryGetValue(PropertyBagKeys.ConverterType, out var converterTypeAssemblyFullName))
+                    {
+                        converterContext.Properties = new Dictionary<string, object>()
```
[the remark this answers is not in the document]
kshyju: I had the `Properties` as nullable type in the `ConverterContext` definition.
Now I made it non-nullable property and `DefaultConverterContext` is immutable now. A constructor overload which takes the property bag also been added.  

But, this class still has to create a new dictionary instance if there is additional props to be passed. Is there a better way to handle it ? Currently the properties type is `IReadOnlyDictionary` in `ConverterContext`

it so that it is not nullable and the . Latest iteration should reflect those changes.

src/DotNetWorker.Core/Converters/Converter/ConverterContext.cs
```diff
@@ -1,16 +1,34 @@
...
 
+        /// <summary>
+        /// The function context.
+        /// </summary>
         public abstract FunctionContext FunctionContext { get; set; }
+                
+        /// <summary>
+        /// Dictionary for additional meta information used for conversion.
+        /// </summary>
+        public abstract IReadOnlyDictionary<string, object>? Properties { get; set;}
```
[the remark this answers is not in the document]
kshyju: Made this type immutable now. Let me know if you did not mean to suggest that. 😭

src/DotNetWorker.Core/Converters/Converter/DefaultInputConverterProvider.cs
```diff
@@ -0,0 +1,108 @@
...
+
+            IInputConverter converterInstance;
+            
+            // Get the IConverter instance for converterType from cache if present.
+            if (_converterCache.TryGetValue(converterType, out var converterFromCache))
```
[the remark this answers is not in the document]
kshyju: Done.

src/DotNetWorker.Core/Converters/Converter/IConverterContextFactory.cs
kshyju: Good catch. No, this can be `internal` for this version. Changed.

satvu (approved):
i think it looks good - just need the csproj files conflicts resolved

src/DotNetWorker.Core/Context/Features/DefaultInputConversionFeature.cs
```diff
@@ -0,0 +1,137 @@
...
+            }
+
+            return ConversionResult.Unhandled();
+        }
+
+        private ValueTask<ConversionResult> ConvertAsyncUsingConverter(IInputConverter converter, ConverterContext context)
+        {
+            var conversionResultTask = converter.ConvertAsync(context);
+
+            if (conversionResultTask.IsCompletedSuccessfully)
```
liliankasem, line 70: Do we need to check if `conversionResultTask` can be null or empty here?

src/DotNetWorker.Core/Context/Features/DefaultInputConversionFeature.cs
```diff
@@ -0,0 +1,137 @@
...
+            {
+                return new ValueTask<ConversionResult>(conversionResultTask.Result);
```
liliankasem, line 72: What happens if `Result` doesn't exist?

src/DotNetWorker.Core/Context/Features/DefaultInputConversionFeature.cs
```diff
@@ -0,0 +1,137 @@
...
+        /// <summary>
+        /// Gets an <see cref="IInputConverter"/> instance if converter context has information about what converter to be used.
+        /// </summary>
+        /// <param name="context">The converter context.</param>
+        /// <returns>An IInputConverter instance or null</returns>
+        private IInputConverter? GetConverterFromContext(ConverterContext context)
+        {
+            string? converterTypeFullName;
+
+            // Check a converter is specified on the conversionContext.Properties. If yes,use that.
```
liliankasem, line 94: nit

```suggestion
            // Check a converter is specified on the conversionContext.Properties. If yes, use that.
```

src/DotNetWorker.Core/Context/Features/DefaultInputConversionFeature.cs
```diff
@@ -0,0 +1,137 @@
...
+        /// Checks a type has an "InputConverter" attribute decoration present
+        /// and if present, return the assembly qualified name of the "ConverterType" property.
+        /// else return null.
+        /// </summary>
+        private static string? GetConverterTypeNameFromAttributeOnType(Type targetType)
+        {            
+            return _typeToConverterCache.GetOrAdd(targetType.AssemblyQualifiedName!, (key, type) =>
+            {
+                var converterAttribute = type.GetCustomAttributes(_inputConverterAttributeType, inherit: true)
+                                                   .FirstOrDefault();
```
liliankasem, line 124: nit: line up periods

src/DotNetWorker.Core/Converters/Converter/ConversionResult.cs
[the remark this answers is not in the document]
liliankasem: I like the idea of checking if the converter can handle before calling convert, that seems like a cleaner way to use this. However I don't think that's the same as what `IsHandled` is about? My understanding is that IsHandled is used to check if the converter has handled the conversion, at this stage we already know if we "can handle" the request and are past that.

As for the states, I don't think we should ever be in a state where IsHandled is false and IsSuccessful is true - essentially the 3 states that Brett pointed out above are the only ones that should be valid. Do you remember why we added IsHandled into the mix? I remember we touched on this during one of our design meetings. Was it related to tracking which conversions are queued to be handled but are not yet? I'd like to better understand the scenario where we need to know if a request was handled or not

src/DotNetWorker.Core/Converters/Converter/DefaultInputConverterProvider.cs
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        /// <summary>
+        /// Initializes the converter cache from worker options.
+        /// </summary>
+        private void InitializeConverterCacheWithDefaultConverters()
+        {
+            if (_workerOptions.InputConverters is null || _workerOptions.InputConverters.Count == 0)
+            {
+                throw new InvalidOperationException("No input converters found in worker options!");
```
liliankasem, line 87: nit: if we don't use the `!` in other error messages, we should avoid doing so here :D

src/DotNetWorker.Core/Converters/ArrayConverter.cs
```diff
@@ -4,20 +4,21 @@
...
     {
         // Convert IEnumerable to array
-        public bool TryConvert(ConverterContext context, out object? target)
+        public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
         {
-            target = null;
+            object? target = null;
             // Ensure requested type is an array
-            if (context.Parameter.Type.IsArray)
+            if (context.TargetType.IsArray)
```
liliankasem, line 19: nit: would it be useful here to do `context?.TargetType?.IsArray` ?

src/DotNetWorker.Core/Converters/MemoryConverter.cs
```diff
@@ -3,33 +3,33 @@
...
     {
-        public bool TryConvert(ConverterContext context, out object? target)
+        public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
         {
-            target = default;
-
             if (context.Source is not ReadOnlyMemory<byte> sourceMemory)
             {
-                return false;
+                return new ValueTask<ConversionResult>(ConversionResult.Unhandled());
```
liliankasem, line 17: Do we return unhandled here because it can't be handled? I feel like `Unhandled` is being used for two different scenarios and it's a little confusing:

1. We can't handle the request
1. We can handle the request but this request is currently unhandled?

If it's #1 that we mean here, it might be worth renaming that so it's easier to understand i.e. "unhandleable", alternativetly (if this is causing so much confusion) we remove the `handled` property respond with ConverstionResult.Error instead?

I think this kind of goes back to the other thread we have in this PR about states and IsHandled & IsSuccessful

liliankasem:
I feel like we need to clear up the "handled" discussion before we approve this.

Also please make sure you create a design document for this new feature in Antares Docs

src/DotNetWorker.Core/Context/Features/DefaultInputConversionFeature.cs
kshyju: The latest iteration is not directly accessing the `Result` property.

src/DotNetWorker.Core/Context/Features/DefaultInputConversionFeature.cs
kshyju: it cannot be null (value type). empty struct is same as a result which is representing an unhandled input conversion. I will revisit this part after we land a solution on the confusion related to the "unhandled" property on ConversionResult.

src/DotNetWorker.Core/Converters/Converter/ConversionResult.cs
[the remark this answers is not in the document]
kshyju: Alternate proposal.

```
public interface IInputConverter
{
    /// <summary>
    /// Gets a value indicating whether this converter can handle the conversion request.
    /// </summary>
    /// <returns></returns>
    bool CanConvert(ConverterContext context);

    /// <summary>
    /// Executes an input conversion operation using the converter context provided.
    /// </summary>
    ValueTask<ConversionResult> ConvertAsync(ConverterContext context);
}
```

and `IsHandled` prop will be removed from [ConversionResult](https://github.com/Azure/azure-functions-dotnet-worker/blob/9f2e6dda5ff177e69f80d471750f5ab2d0f7f6fc/src/DotNetWorker.Core/Converters/Converter/ConversionResult.cs#L9) and it will be a type reprenseting only the result of the conversion operation.

The infrastructure code which is calling the ConvertAsync first checks whether this is the right converter to use.
```
if (converter.CanConvert(converterContext))
{
    var conversionResultTask = converter.ConvertAsync(converterContext);
    ....
}
```

src/DotNetWorker.Core/Converters/Converter/ConverterContextFactory.cs
```diff
@@ -0,0 +1,28 @@
...
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+
+namespace Microsoft.Azure.Functions.Worker.Converters
+{
+    /// <summary>
+    /// A factory for creating <see cref="ConverterContext"/> instances.
+    /// </summary>
+    internal sealed class DefaultConverterContextFactory : IConverterContextFactory
```
brettsam, line 13: nit -- file name doesn't match class name?

src/DotNetWorker.Core/Context/Features/DefaultModelBindingFeature.cs
```diff
@@ -46,50 +47,43 @@ public DefaultModelBindingFeature(IEnumerable<IConverter> converters)
...
+                // Pass info about specific input converter type defined for this parameter, if present.
+                if (param.Properties.TryGetValue(PropertyBagKeys.ConverterType, out var converterTypeAssemblyFullName))
+                {
+                    properties = new Dictionary<string, object>()
+                    {
+                        { PropertyBagKeys.ConverterType, converterTypeAssemblyFullName }
+                    };
+                }
+
+                var converterContext = new DefaultConverterContext(param.Type, source, context, properties);
```
brettsam, line 61: Should we be using the `IConverterContextFactory` here?

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -164,6 +170,11 @@ private async Task ProcessRequestCoreAsync(StreamingMessage request)
                 context = application.CreateContext(invocationFeatures);
                 invocationFeatures.Set<IFunctionBindingsFeature>(new GrpcFunctionBindingsFeature(context, request, outputBindingsInfoProvider));
 
+                if (functionInputConversionFeatureProvider.TryCreate(typeof(DefaultInputConversionFeature), out var conversion))
```
brettsam, line 173: I don't think this should be happening here -- we don't want to create one of these for every invocation, do we? Going to think through this a bit...

brettsam (changes requested):

brettsam (approved):

PR #644, opened 2021-09-28
Updating worker sdk Extensions csproj generator to support v4

### LINKED ISSUE #617: Update extensions csproj generator to support Functions 4.0
We need to update the generator to remove the dependency on .NET Core 3.1 in Functions 4.0 scenarios.

soninaren:
closes #617

### DIFF
--- a/sdk/Sdk/ExtensionsCsprojGenerator.cs
+++ b/sdk/Sdk/ExtensionsCsprojGenerator.cs
@@ -14,13 +14,13 @@ namespace Microsoft.Azure.Functions.Worker.Sdk
 
         private readonly IDictionary<string, string> _extensions;
         private readonly string _outputPath;
-        private readonly string _targetFramework;
+        private readonly string _azureFunctionsVersion;
 
-        public ExtensionsCsprojGenerator(IDictionary<string, string> extensions, string outputPath, string targetFramework)
+        public ExtensionsCsprojGenerator(IDictionary<string, string> extensions, string outputPath, string azureFunctionsVersion)
         {
             _extensions = extensions ?? throw new ArgumentNullException(nameof(extensions));
             _outputPath = outputPath ?? throw new ArgumentNullException(nameof(outputPath));
-            _targetFramework = targetFramework ?? throw new ArgumentNullException(nameof(targetFramework));
+            _azureFunctionsVersion = azureFunctionsVersion ?? throw new ArgumentNullException(nameof(azureFunctionsVersion));
         }
 
         public void Generate()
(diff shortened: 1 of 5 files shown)


sdk/Sdk/ExtensionsCsprojGenerator.cs
```diff
@@ -49,14 +49,21 @@ private void WriteExtensionsCsProj(string filePath)
             File.WriteAllText(filePath, csprojContent);
         }
 
+
```
liliankasem, line 52: nit: were the extra lines here by accident?

sdk/Sdk/ExtensionsCsprojGenerator.cs
```diff
@@ -49,14 +49,21 @@ private void WriteExtensionsCsProj(string filePath)
...
 
+            string targetFrameWork = GetTargetFrameWork();
+
+            string netSdkVersion = GetNetSdkVersion();
+
             return $@"
 <Project Sdk=""Microsoft.NET.Sdk"">
     <PropertyGroup>
-        <TargetFramework>{_targetFramework}</TargetFramework>
+        <TargetFramework>{targetFrameWork}</TargetFramework>
```
liliankasem, line 66: I think framework is one word so the original camelcase of the variable makes more sense, or is this something out of our control?

sdk/Sdk/ExtensionsCsprojGenerator.cs
```diff
@@ -69,13 +76,33 @@ internal string GetCsProjContent()
...
+        {
+            if (_azureFunctionsVersion.StartsWith("v3", StringComparison.OrdinalIgnoreCase))
+            {
+                return "3.0.11";
+            }
+
+            return "4.0.0-preview2";
+        }
+
+        private string GetTargetFrameWork()
```
liliankasem, line 96: nit: same here about using `Framework` over `FrameWork`

sdk/Sdk/ExtensionsCsprojGenerator.cs
```diff
@@ -69,13 +76,33 @@ internal string GetCsProjContent()
...
+        {
+            if (_azureFunctionsVersion.StartsWith("v3", StringComparison.OrdinalIgnoreCase))
+            {
+                return "netcoreapp3.1";
+            }
+
+            return "net6.0";
```
liliankasem, line 103: Just wondering, would it be worth us using constant strings for these version variables so they're defined in one place or nah?

sdk/Sdk/ExtensionsCsprojGenerator.cs
soninaren: yeah, let me fix that.

sdk/Sdk/ExtensionsCsprojGenerator.cs
soninaren: Should definitely use a constant. Will change it

liliankasem (approved):

PR #656, opened 2021-10-04
Print E2E Tests Host Startup logs

satvu:
Printing testing logs for E2E builds


brettsam:
We figured out this was related to a host change...

brettsam:
Actually, we want this for better log output...

kshyju (approved):

PR #662, opened 2021-10-06
adding all referenced extensions to metadata

brettsam:


brettsam:
@kshyju / @satvu -- didn't realize that @fabiocav had already talked to you about this scenario. I'm going to update this to add a property on the attribute (this current approach just considers any assembly with `[ExtensionInformation]` as an extension to reference.

brettsam:
+ @gzuber / @anthonychu as FYI too.

samples/FunctionApp/FunctionApp.csproj
```diff
@@ -14,12 +14,13 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.ApplicationInsights" Version="1.0.0-preview3" />
```
brettsam, line 17: Once this has flowed to this extension, I'll update the tests to ensure it all keeps working e2e.

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -632,20 +639,36 @@ private static void AddHttpOutputBinding(IList<ExpandoObject> bindingMetadata, s
...
             {
                 if (string.Equals(assemblyAttribute.AttributeType.FullName, Constants.ExtensionsInformationType, StringComparison.Ordinal))
                 {
                     string extensionName = assemblyAttribute.ConstructorArguments[0].Value.ToString();
-                    string extensionVersion = assemblyAttribute.ConstructorArguments[1].Value.ToString();
+                    string extensionVersion = assemblyAttribute.ConstructorArguments[1].Value.ToString();                    
+                    bool implicitlyRegister = false;
 
-                    extensions[extensionName] = extensionVersion;
+                    if (assemblyAttribute.ConstructorArguments.Count >= 3)
```
satvu, line 655: Question: Why 3?

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -98,17 +98,24 @@ public IEnumerable<SdkFunctionMetadata> GenerateFunctionMetadata(string assembly
...
                 if (functionsResult.Any())
                 {
+                    functionFound = true;
                     _logger.LogMessage($"Found {functionsResult.Length} functions in '{type.GetReflectionFullName()}'.");
                 }
 
                 functions.AddRange(functionsResult);
             }
+            
+            if (!functionFound && TryAddExtensionInfo(_extensions, module.Assembly, usedByFunction: false))
```
kshyju, line 115: If `functionFound` is `true` , we are not adding the implicit registration thing ?

extensions/Worker.Extensions.Abstractions/ExtensionInformationAttribute.cs
```diff
@@ -14,10 +14,18 @@ public class ExtensionInformationAttribute : Attribute
 
         public string ExtensionVersion { get; }
 
+        public bool EnableImplicitRegistration { get; } = false;
```
kshyju, line 17: nit: default value can be removed as it is implicit.

sdk/Sdk/FunctionMetadataGenerator.cs
brettsam: That would force us to load the assembly during the build -- which can be problematic with versioning, etc. That's why we're trying to only figure out the values by using the constructor params, which are able to be determined via assembly metadata. Because we own that piece, we know it's the value we're looking for.

kshyju (approved):

satvu (approved):

sdk/Sdk/FunctionMetadataGenerator.cs
fabiocav: Wonder if it would be clearer (if it would have avoided Shyju's question) if the variable was named something like `moduleExtensionsRegistered` and perhaps a comment down where we evaluate describing that modules with defined functions will have their extension references registered.

PR #676, opened 2021-10-13
Add worker-driven function indexing

### LINKED ISSUE #677: Implement support for function indexing
Issue tracking the work related to worker driven function indexing for .NET

satvu:

resolves #677 

### Pull request checklist

* [] My changes **should not** be added to the release notes for the next release
* [x ] I have added all required tests (Unit tests, E2E tests)

**Note on Tests** 
Manual E2E validation was performed in this PR for the opt-in Stein features.

The E2E tests are using the current function-indexing implementation fallback and do not opt into Stein. The E2E App does not add the `<FunctionsEnableWorkerIndexing>` property in Function App `.csproj` file or set `AzureWebJobsFeatureFlags: true` in `local.settings.json`. 

The SDK changes are equivalent to other code (writing to `worker.config.json`, registering services to hostbuilder) that are not tested in our repo but fall under the E2E tests umbrella too.

### Additional information

This is an implementation change for how functions are loaded/provided to the host. Currently, functions are loaded through a custom provider passed to the host by the `FunctionMetadataLoaderExtension`.

- Add response handling for worker-indexing requests in `GrpcWorker.cs`.
- Implement a class in the Grpc project that reads from `functions.metadata` and provides function metadata to the GrpcWorker to send back to the host.
- Adds an MSBuild Property `<FunctionsEnableWorkerIndexing>` that is read by the worker SDK and used to:
  1. Set the value of `workerIndexing` in `worker.config.json`, which is used by the host to decide whether or not to make a worker-driven function-indexing request.
  2. Decide whether or not to use the `FunctionMetadataLoaderExtension` to load functions. If `workerIndexing` is set to `true`, the extension will not load a custom provider so that duplicate functions are not loaded.


satvu:
[Setting the worker indexing property to true in Python worker](https://github.com/Azure/azure-functions-host/wiki/Worker-Indexing-Changes#how-to-set-up-local-worker-indexing-environment-feat-python-worker)

Currently, to validate that this implementation is working against the Project Stein branch of the host, I'm manually adding the worker indexing property to worker.config.json. Not sure what the path forward for this is, so I haven't implemented anything that would automatically add the property or set it to any specific value.

src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
```diff
@@ -22,8 +22,10 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="Microsoft.Azure.WebJobs.Script.Abstractions" Version="1.0.0-preview" />
     <PackageReference Include="Microsoft.Extensions.Hosting" Version="5.0.0" />
     <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="5.0.0" />
+    <PackageReference Include="Mono.Cecil" Version="0.11.3" />
```
kshyju, line 28: Where/how do we use this package?

kshyju:
Few comments.

samples/FunctionApp/Properties/launchSettings.json
```diff
@@ -1,10 +0,0 @@
-{
```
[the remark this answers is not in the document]
satvu: The changes in samples/FunctionApp/Properties/* should be the same as [this PR that hasn't been merged](https://github.com/Azure/azure-functions-dotnet-worker/pull/616) b/c they're needed for local debugging. If that one doesn't go in before this one does, I'll remove the changes to keep this PR focused on one topic.

samples/FunctionApp/FunctionApp.csproj
```diff
@@ -14,8 +14,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.ApplicationInsights" Version="1.0.0-preview3" />
```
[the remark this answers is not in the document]
satvu: Not sure why this was changed but agreed - will remove in next iteration

src/DotNetWorker.Grpc/protobuf/src/proto/FunctionRpc.proto
```diff
@@ -75,6 +75,10 @@ message StreamingMessage {
     // Ask the worker to close any open shared memory resources for a given invocation
     CloseSharedMemoryResourcesRequest close_shared_memory_resources_request = 27;
     CloseSharedMemoryResourcesResponse close_shared_memory_resources_response = 28;
+
+    // Worker indexing message types
+    FunctionsMetadataRequest functions_metadata_request = 29;
```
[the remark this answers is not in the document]
satvu: I used this process for pulling updates from the other repo: https://github.com/Azure/azure-functions-language-worker-protobuf#pulling-updates

I don't change it myself locally, I just pull updates and commit them but I think it does look like a local change from this side? Here's another [PR that consumes protobuf changes](https://github.com/Azure/azure-functions-dotnet-worker/commit/9eb2d590b6058788713aa957b3f86f3b09148c47) from this repo and it looks similar.

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -228,6 +233,84 @@ internal static WorkerInitResponse WorkerInitRequestHandler(WorkerInitRequest re
...
+            var functionRequests = new List<FunctionLoadRequest>(functionsMetadata.Length);
+
+            foreach (var metadata in functionsMetadata)
+            {
+                var funcId = Guid.NewGuid().ToString(); 
+
+                FunctionLoadRequest request = new FunctionLoadRequest()
+                {
+                    FunctionId = funcId, 
+                    Metadata = new RpcFunctionMetadata()
```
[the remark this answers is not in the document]
satvu: [Python version](https://github.com/Azure/azure-functions-python-worker/blob/d3f02af81e080bf76808b467732765c55a067c88/azure_functions_worker/dispatcher.py#L304) did not, but I'll ask about this in the next Stein sync

src/DotNetWorker.Grpc/FunctionMetadata/FunctionMetadataJsonReader.cs
```diff
@@ -0,0 +1,67 @@
...
+                    {
+                        JArray functionMetadataJson = (JArray)await JToken.ReadFromAsync(js);
+
+                        var functionList = new List<FunctionMetadata>();
+
+                        foreach (JObject function in functionMetadataJson)
+                        {
+                            FunctionMetadata metadata = function.ToObject<FunctionMetadata>();
+
+                            // We need to re-add these by going through the BindingMetadata factory
```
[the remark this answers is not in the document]
satvu: Took the logic from the `FunctionMetadataLoaderExtension` project at this file [here](https://github.com/Azure/azure-functions-dotnet-worker/blob/66c6380a311b800bb301323693b4ce1cf69dd682/sdk/FunctionMetadataLoaderExtension/FunctionMetadataJsonReader.cs#L45). I don't have the background context for this but @brettsam or @fabiocav might.

src/DotNetWorker.Grpc/FunctionMetadata/FunctionMetadataJsonReader.cs
```diff
@@ -0,0 +1,67 @@
...
+                            metadata.Bindings.Clear();
+
+                            JArray bindingArray = (JArray)function["bindings"];
+                            if (bindingArray == null || bindingArray.Count == 0)
+                            {
+                                throw new FormatException("At least one binding must be declared.");
```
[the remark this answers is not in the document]
satvu: Took the logic from [here](https://github.com/Azure/azure-functions-dotnet-worker/blob/66c6380a311b800bb301323693b4ce1cf69dd682/sdk/FunctionMetadataLoaderExtension/FunctionMetadataJsonReader.cs#L50) in the `FunctionMetadataLoaderExtension` project. FormatException doesn't have an overload where I could add something like ` , nameof(funcName)` but agreed that it should be added. If we stick with FormatException I can manipulate the message string.

samples/FunctionApp/FunctionApp.csproj
```diff
@@ -15,7 +15,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.ApplicationInsights" Version="1.0.0-preview3" />
-    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.0.1" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.0.3" OutputItemType="Analyzer" />
```
surgupta-msft, line 18: Just curious, what is the purpose of adding - OutputItemType="Analyzer" here?

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -228,6 +234,86 @@ internal static WorkerInitResponse WorkerInitRequestHandler(WorkerInitRequest re
             return response;
         }
 
+        internal static async Task<FunctionMetadataResponses> FunctionsMetadataRequestHandler(FunctionsMetadataRequest request)
```
surgupta-msft, line 237: Can we rename the method as FunctionsMetadataRequestHandler**Async**? Generally we do this for Async methods.

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -228,6 +234,86 @@ internal static WorkerInitResponse WorkerInitRequestHandler(WorkerInitRequest re
...
+                {
+                    Status = StatusResult.Types.Status.Failure,
+                    Exception = ex.ToRpcException()
+                };
+            }
+
+            return response;
+        }
+
+        internal static async Task<IReadOnlyList<FunctionLoadRequest>> GetFunctionLoadRequests(string directory)
```
surgupta-msft, line 267: nit: Same here. GetFunctionLoadRequests**Async**.

samples/FunctionApp/FunctionApp.csproj
satvu: Changes from [this PR for Debugging](https://github.com/Azure/azure-functions-dotnet-worker/pull/616) - leaving it for now because I may need to locally debug this branch still, but will remove before merging. Hopefully that other PR will be merged soon so I can rebase this branch.

src/DotNetWorker.Grpc/GrpcWorker.cs
satvu: yes good catch!

samples/FunctionApp/FunctionApp.csproj
surgupta-msft: Got it. Thanks.

src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
```diff
@@ -22,8 +22,10 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="Microsoft.Azure.WebJobs.Script.Abstractions" Version="1.0.0-preview" />
```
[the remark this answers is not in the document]
brettsam: I think there are some other subtle issues with bringing this package in here that we want to avoid if possible. Do you know exactly what classes we'd need to bring in manually? We may have a different approach we can take...

src/DotNetWorker.Grpc/protobuf/src/proto/FunctionRpc.proto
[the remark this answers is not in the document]
brettsam: I do think this is the correct way to handle this.

src/DotNetWorker.Grpc/FunctionMetadata/FunctionMetadataJsonReader.cs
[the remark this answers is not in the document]
brettsam: I don't remember exactly, but the bindings must be created using `BindingMetadata.Create()` -- and reading straight from the json didn't work for some reason.

src/DotNetWorker.Grpc/FunctionMetadata/FunctionMetadataJsonReader.cs
```diff
@@ -0,0 +1,68 @@
...
+    {
+        private readonly string _directory;
+        private const string FileName = "functions.metadata";
+
+        public FunctionMetadataJsonReader(string directory)
+        {
+            _directory = directory ?? throw new ArgumentNullException(nameof(directory));
+        }
+
+        public virtual async Task<ImmutableArray<FunctionMetadata>> ReadMetadataAsync()
```
brettsam, line 25: I think that you should use modify this class to return the FunctionLoadRequest itself and drop the need for all of these extra types. Right now you're pulling it into the FunctionMetadata type, then converting it -- but it's not too complex; we can just update this to create the FunctionLoadRequest directly. It'd reduce dependencies and simplify things (and make them faster)...

I'd also recommend using System.Text.Json here rather than Newtonsoft.Json -- again, to reduce dependencies.

brettsam (changes requested):

src/DotNetWorker.Grpc/FunctionMetadata/FunctionMetadataJsonReader.cs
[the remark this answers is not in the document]
kshyju: Could we not concatenate the string and use as the message param value of the overload?
````
string functionName = "to do: Read this from somewhere";
throw new InvalidOperationException($"At least one binding must be declared in function '{functionName}'");;
````

File: LICENSE [hunk 13]
```diff
@@ -18,4 +18,4 @@
     AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
     LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
     OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
-    SOFTWARE
```
fabiocav, line 21: As discussed, let's revert this.

src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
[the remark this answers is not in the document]
fabiocav: Indeed. We don't want to have any WebJobs dependencies coming into those projects. 

Can you clarify what is required from this?

src/DotNetWorker.Grpc/FunctionMetadata/FunctionMetadataJsonReader.cs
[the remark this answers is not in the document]
fabiocav: Yeah, do you know how the behavior differs if you don't do this?

src/DotNetWorker.Grpc/FunctionMetadata/FunctionMetadataJsonReader.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            string metadataFile = Path.Combine(_directory, FileName);
+
+            if (File.Exists(metadataFile))
+            {
+                using (var fs = File.OpenText(metadataFile))
+                {
+                    using (var js = new JsonTextReader(fs))
+                    {
+                        JArray functionMetadataJson = (JArray)await JToken.ReadFromAsync(js);
```
fabiocav, line 35: I think @brettsam mentioned this above, but it would be good to avoid a dependency on JSON.NET here.

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -144,7 +150,7 @@ private async Task ProcessRequestCoreAsync(StreamingMessage request)
             await _outputWriter.WriteAsync(responseMessage);
         }
 
-        internal static async Task<InvocationResponse> InvocationRequestHandlerAsync(InvocationRequest request, IFunctionsApplication application,
+    internal static async Task<InvocationResponse> InvocationRequestHandlerAsync(InvocationRequest request, IFunctionsApplication application,
```
fabiocav, line 153: nit: this change seems off.

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -15,6 +16,7 @@
 using Microsoft.Azure.Functions.Worker.OutputBindings;
 using Microsoft.Azure.Functions.Worker.Rpc;
 using Microsoft.Extensions.Options;
+using Newtonsoft.Json;
```
fabiocav, line 19: Remove dependency on JSON.NET

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -228,6 +234,86 @@ internal static WorkerInitResponse WorkerInitRequestHandler(WorkerInitRequest re
             return response;
         }
 
+        internal static async Task<FunctionMetadataResponses> FunctionsMetadataRequestHandlerAsync(FunctionsMetadataRequest request)
+        {
+            var directory = request.FunctionAppDirectory;
```
fabiocav, line 239: Is this local needed? We should be able to pass the directory directly here

src/DotNetWorker.Grpc/protobuf/src/proto/FunctionRpc.proto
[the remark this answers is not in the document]
fabiocav: You can also just copy the appropriate proto files over. That was the approach I've used in the past, but this works as well (though, it does bring additional files/changes)

src/DotNetWorker.Grpc/FunctionMetadata/FunctionMetadataJsonReader.cs
fabiocav: Agreed. I think that will cleanup some additional things below as well.

fabiocav (commented):
Some comments, but one question...

With the changes, my assumption is that we're expected to update the worker config to indicate that we now support worker indexing, correct? What is the plan for that? Should we make that change now? Should this be driven by a user option (e.g. enabling worker config when bootstrapping). It would be good to make this behavior opt-in for now so we can validate but continue to support the previous behavior.

src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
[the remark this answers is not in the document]
satvu: FunctionMetadata and BindingMetadata + all the other types they use. The sdk uses this package and it simplified deserializing the json and simplified the code b/c of built in functions like .Create( raw JSON) for bindings. 

It's not necessary though - removed it in the next iteration and focused on just the information we needed for the grpc message and to pass the host-side validation checks.

src/DotNetWorker.Grpc/FunctionMetadata/FunctionMetadataJsonReader.cs
```diff
@@ -0,0 +1,123 @@
...
+            {
+                using (var fs = File.OpenRead(metadataFile))
+                {
+                    var jsonMetadataList = await JsonSerializer.DeserializeAsync<JsonElement>(fs);
+
+                    var functionRequests = new List<FunctionLoadRequest>(jsonMetadataList.GetArrayLength());
+
+                    foreach (var jsonMetadata in jsonMetadataList.EnumerateArray())
+                    {
+                        var options = new JsonSerializerOptions();
```
kshyju, line 38: Could we use a single instance of `JsonSerializerOptions` and use that? No need to create a new instance for each iteration of the loop.

src/DotNetWorker.Grpc/FunctionMetadata/FunctionMetadataJsonReader.cs
```diff
@@ -0,0 +1,123 @@
...
+                        options.PropertyNameCaseInsensitive = true;
+
+                        var functionMetadata = JsonSerializer.Deserialize<RpcFunctionMetadata>(jsonMetadata.GetRawText(), options);
```
kshyju, line 41: There is an [async version of this method](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializer.deserializeasync?view=net-5.0#System_Text_Json_JsonSerializer_DeserializeAsync__1_System_IO_Stream_System_Text_Json_JsonSerializerOptions_System_Threading_CancellationToken_) (`DeserializeAsync`) method you could consider switching to.

kshyju: Also, I wonder whether we really need to convert the `JsonElement` to `string`, then deserialize it again. Could we not read the prop values from the `JsonElement` instance and set the properties of `RpcFunctionMetadata` instance manually?

src/DotNetWorker.Grpc/protobuf/src/proto/FunctionRpc.proto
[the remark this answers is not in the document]
liliankasem: I was talking to Pragna about this today and I think is makes sense to keep the old approach. Ideally profobuf changes are in their own PR so it's easier to trace and revert if necessary. (Just had to go through this process for the ILogger PR I took over from Brandon). Unless there's some missing context here? Happy to discuss in parking lot or team meeting :)

src/DotNetWorker.Grpc/FunctionMetadata/FunctionMetadataJsonReader.cs
satvu: I tried it with GetProperty() and TryGetProperty() first, but that leads to null-checking every property when we grab them and call ToString().

Didn't use the async version b/c it throws an error: `'RpcFunctionMetadata' does not contain a definition for 'GetAwaiter' ... `

sdk/Sdk/worker.config.json
```diff
@@ -3,6 +3,7 @@
     "language": "dotnet-isolated",
     "extensions": [ ".dll" ],
     "defaultExecutablePath": "$functionExe$",
-    "defaultWorkerPath": "$functionWorker$"
+    "defaultWorkerPath": "$functionWorker$",
+    "workerIndexing":  "true"
```
satvu, line 7: **General idea of new design:** When a worker.config file has "workerIndexing" set to true, it means that the worker knows how to respond to a metadata request from the host. The worker may decide for itself after getting the request whether it will index the functions and return the metadata or force the indexing responsibility back onto the host. 

The design hasn't been outlined yet or implemented in the host (still under discussion) but wanted to provide context for this.

cc: @fabiocav @surgupta-msft

sdk/FunctionMetadataLoaderExtension/Startup.cs
```diff
@@ -48,9 +49,15 @@ public void Configure(WebJobsBuilderContext context, IWebJobsConfigurationBuilde
         public void Configure(WebJobsBuilderContext context, IWebJobsBuilder builder)
         {
             string appRootPath = context.ApplicationRootPath;
-            builder.Services.AddOptions<FunctionMetadataJsonReaderOptions>().Configure(o => o.FunctionMetadataFileDrectory = appRootPath);
-            builder.Services.AddSingleton<FunctionMetadataJsonReader>();
-            builder.Services.AddSingleton<IFunctionProvider, JsonFunctionProvider>();
+            WorkerConfigDescription newWorkerDescription = GetUpdatedWorkerDescription(appRootPath);
+
+            if (string.IsNullOrEmpty(newWorkerDescription.EnableWorkerIndexing) ||
```
satvu, line 54: This was something I was working on before the new design change (workerIndexing set to true to handle request messages but not necessarily the indexing task) but has no impact on the way indexing would work for Stein b/c we are hardcoding the worker.config file's `workerIndexing` to true.

I did test it w/ `workerIndexing` set to false or not set at all to make sure we could preserve the current function loading behavior w/ the extension if we wanted to.

sdk/FunctionMetadataLoaderExtension/Startup.cs
```diff
@@ -7,17 +7,15 @@
...
 using Microsoft.Extensions.Configuration;
-using Microsoft.Extensions.DependencyInjection;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
 [assembly: WebJobsStartup(typeof(Startup))]
 
 namespace Microsoft.Azure.WebJobs.Extensions.FunctionMetadataLoader
 {
-    internal class Startup : IWebJobsStartup2, IWebJobsConfigurationStartup
```
kshyju, line 20: I do not see any usage of [IWebJobsStartup2 ](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Hosting/IWebJobsStartup2.cs#L10)in host repo.  So, I guess this would not break anything 😄

Is `IWebJobsStartup2` obsolete type now in favor of `IWebJobsConfigurationStartup`? I think we need to fix this in the custom handler startup as well.

sdk/FunctionMetadataLoaderExtension/Startup.cs
```diff
@@ -45,14 +43,6 @@ public void Configure(WebJobsBuilderContext context, IWebJobsConfigurationBuilde
...
 
-        public void Configure(WebJobsBuilderContext context, IWebJobsBuilder builder)
-        {
-            string appRootPath = context.ApplicationRootPath;
-            builder.Services.AddOptions<FunctionMetadataJsonReaderOptions>().Configure(o => o.FunctionMetadataFileDrectory = appRootPath);
-            builder.Services.AddSingleton<FunctionMetadataJsonReader>();
-            builder.Services.AddSingleton<IFunctionProvider, JsonFunctionProvider>();
-        }
-
         public void Configure(IWebJobsBuilder builder)
```
kshyju, line 46: I think this empty Configure method can be removed safely.

src/DotNetWorker.Grpc/FunctionMetadata/FunctionMetadataJsonReader.cs
kshyju: Were you awaiting it? Line 32 uses that overload. So should work here as well.

src/DotNetWorker.Grpc/FunctionMetadata/FunctionMetadataJsonReader.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.IO;
+using System.Threading.Tasks;
+using System.Text.Json;
+using Microsoft.Azure.Functions.Worker.Grpc.Messages;
+
+namespace Microsoft.Azure.Functions.Worker
+{ 
+    internal class FunctionMetadataJsonReader
```
kshyju, line 14: Should we consider renaming this type? This class is more of a Function load response generator, than a JSON reader.

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -228,6 +233,36 @@ internal static WorkerInitResponse WorkerInitRequestHandler(WorkerInitRequest re
...
+        internal static async Task<FunctionMetadataResponses> FunctionsMetadataRequestHandlerAsync(FunctionsMetadataRequest request)
+        {
+            var response = new FunctionMetadataResponses
+            {
+                Result = StatusResult.Success
+            };
+
+            try
+            {
+                var functionReader = new FunctionMetadataJsonReader(request.FunctionAppDirectory);
```
kshyju, line 245: Should we consider abstracting this behind an interface ( ex: `IFunctionMetadataProvider`) so that writing unit tests for this class might be easier (you can `IFunctionMetadataProvider.ReadMetadataAsync` method. When we switch to the source gen implementation of this interface in future, you do not need to make any major changes to UTs.

src/DotNetWorker.Grpc/FunctionMetadata/FunctionMetadataJsonReader.cs
kshyju: Also, the method prefix could be `Get` (Ex: `GetFunctionLoadResponse` or so).

kshyju (commented):
Question: Where do we control the feature flag to enable/disable this new feature? Would that be an environment variable/App setting, which host code reads and use? I am trying to understand how we are planning to ship this change without breaking existing apps. It is possible for a customer to upgrade to a new version of worker NuGet packages with these changes, but still have the old version of host (and vice versa). Are our changes compatible to support that case? Perhaps, we could update the PR description with those details to help other reviewers.

sdk/FunctionMetadataLoaderExtension/Startup.cs
satvu: The method I removed that added the `IFunctionProvider` was part of the `IWebJobsStartup2` interface.

src/DotNetWorker.Grpc/FunctionMetadata/FunctionMetadataJsonReader.cs
satvu: The naming convention was kept from a similar class that read from the `functions.metadata` file. It does not generate any function metadata but "reads" it from a file.

Shyju and I discussed this in a call, and we are deciding between `FunctionMetadataProvider` or `FunctionMetadataReader`, and depending on how this class will be used in later iterations, whether or not to keep "Json" in the name.

src/DotNetWorker.Grpc/FunctionMetadata/FunctionMetadataJsonReader.cs
satvu: Shyju and I just discussed this in a call, but since we have already read from the file stream to get the raw json (and preserve the raw bindings) we would need to create a new stream for the jsonMetadata. This class and its methods are used only once at initialization to provide function metadata to the host, so there are not many advantages to switch to an async version at the moment.

src/DotNetWorker.Grpc/GrpcWorker.cs
brettsam: Yeah I agree with this one. In general, whenever you find yourself calling `new` on something, you should think if it'd be better to do it via DI. In this case, it does seem useful to be able to replace it for testing this method itself and for future replacement.

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -228,6 +233,37 @@ internal static WorkerInitResponse WorkerInitRequestHandler(WorkerInitRequest re
...
+            {
+                Result = StatusResult.Success,
+                UseDefaultMetadataIndexing = false
+            };
+
+            try
+            {
+                var functionReader = new FunctionMetadataJsonReader(request.FunctionAppDirectory);
+
+                var functionMetadataList = await functionReader.ReadMetadataAsync();
```
brettsam, line 248: To improve startup perf, there's likely something else we can do here:
- Have an `IFunctionMetadataProvider.ReadMetadataAsync()`, as @kshyju suggested.
- Instead of calling it here -- call it in `StartAsync()` and store the `Task<>` (i.e. don't `await` it). 
- In this method, you can `await` that `Task<>` that you called earlier.

We know we're going to need to load that metadata, so there's no reason to wait until the host asks for it; we can start that work sooner and do it in parallel.

brettsam:
Agree with @kyshju on his comments, and one other improvement we can make...

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -56,12 +59,22 @@ internal class GrpcWorker : IWorker
             _startupOptions = startupOptions ?? throw new ArgumentNullException(nameof(startupOptions));
             _serializer = workerOptions.Value.Serializer ?? throw new InvalidOperationException(nameof(workerOptions.Value.Serializer));
             _inputConversionFeatureProvider = inputConversionFeatureProvider ?? throw new ArgumentNullException(nameof(inputConversionFeatureProvider));
+            _appLocation = Environment.GetEnvironmentVariable("AzureWebJobsScriptRoot")!;
```
satvu, line 62: This was the easiest way I could think of to get the path I needed to find and read from functions.metadata -- I wasn't able to find it in any of the objects passed in (double checked by debugging here in VS). In previous iterations, this info was provided from the host gRPC message, but we want to gather the function metadata asap before that message since we know we'll need it anyways.

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -34,6 +35,8 @@ internal class GrpcWorker : IWorker
         private readonly IMethodInfoLocator _methodInfoLocator;
         private readonly IOptions<GrpcWorkerStartupOptions> _startupOptions;
         private readonly ObjectSerializer _serializer;
+        private readonly string _appLocation;
+        private Task<FunctionMetadataResponse> _functionMetadataResponse;
```
kshyju, line 39: naming nit: IMHO, Adding the "Task" suffix improves readability. This variable holds the task. ex : `_functionMetadataResponseLoadingTask`

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -56,12 +59,22 @@ internal class GrpcWorker : IWorker
...
             _serializer = workerOptions.Value.Serializer ?? throw new InvalidOperationException(nameof(workerOptions.Value.Serializer));
             _inputConversionFeatureProvider = inputConversionFeatureProvider ?? throw new ArgumentNullException(nameof(inputConversionFeatureProvider));
+            _appLocation = Environment.GetEnvironmentVariable("AzureWebJobsScriptRoot")!;
+            _functionMetadataResponse = Task.FromResult(new FunctionMetadataResponse
+            {
+                Result = new StatusResult
+                {
+                    Status = StatusResult.Types.Status.Failure
+                }
+            });
```
kshyju, line 69: Do we need this? I think you can remove this.

src/DotNetWorker.Grpc/FunctionMetadata/IFunctionMetadataProvider.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Immutable;
+using System.Threading.Tasks;
+using Microsoft.Azure.Functions.Worker.Grpc.Messages;
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    internal interface IFunctionMetadataProvider
```
kshyju, line 10: I do not see this dependency being registered in the DI container.

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -239,6 +256,37 @@ internal static WorkerInitResponse WorkerInitRequestHandler(WorkerInitRequest re
...
+        {
+            var response = new FunctionMetadataResponse
+            {
+                Result = StatusResult.Success,
+                UseDefaultMetadataIndexing = false
+            };
+
+            try
+            {
+                var functionProvider = new FunctionMetadataProvider(functionAppDirectory);
```
kshyju, line 269: I thought you were going to DI `IFunctionMetadataProvider`

src/DotNetWorker.Grpc/protobuf/src/proto/FunctionRpc.proto
```diff
@@ -282,12 +282,15 @@ message FunctionsMetadataRequest {
...
+message FunctionMetadataResponse {
   // list of function indexing responses
-  repeated FunctionLoadRequest function_load_requests_results = 1;
+  repeated RpcFunctionMetadata function_metadata_results = 1;
 
   // status of overall metadata request
   StatusResult result = 2;
+
+  // if set to true then host will perform indexing
+  bool useDefaultMetadataIndexing = 3;
```
kshyju, line 293: Should we use same format to be consistent? (_ instead of camelCase?)

sdk/Sdk/worker.config.json
kshyju, line 7: Are we still considering moving this to [Worker Capabilities?](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/src/DotNetWorker.Grpc/GrpcWorker.cs#L237)

satvu: No we're moving forward with worker config and a [gRPC message flag](https://github.com/Azure/azure-functions-language-worker-protobuf/blob/73658b78e6dd876af13e50488542f63aac544df6/src/proto/FunctionRpc.proto#L307) to handle legacy apps and worker indexing.

src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs
```diff
@@ -46,6 +46,9 @@ public static IServiceCollection AddGrpc(this IServiceCollection services)
                 logging.Services.AddSingleton<IWorkerDiagnostics, GrpcWorkerDiagnostics>();
             });
 
+            // FunctionMetadataProvider for worker driven function-indexing
+            services.AddSingleton<IFunctionMetadataProvider, DefaultFunctionMetadataProvider>();
```
kshyju, line 50: Should this be in `AddGrpc` method? this DI registration is not something needed to enable GRPC communication. rite? Should this go in `AddFunctionsWorkerCore` ?

src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs
```diff
@@ -75,6 +78,7 @@ public static IServiceCollection AddGrpc(this IServiceCollection services)
                 .Configure<IConfiguration>((arguments, config) =>
                 {
                     config.Bind(arguments);
+                    arguments.ScriptRoot = config["AzureWebJobsScriptRoot"];
```
kshyju, line 81: If we use the same value as the property name (`AzureWebJobsScriptRoot` instead of `ScriptRoot`), we do not need this line. Is that rite?

kshyju (approved):

kshyju:
We need to bump up the package version for this change. rite?

satvu:
> We need to bump up the package version for this change. rite?

Yes - not bumping it yet because we need to wait for a compatible host/core-tools version to roll out and not sure what version of the worker we will be on by the time this can be deployed. We also can't pass the E2E tests or debug them until we get a core-tools build with a compatible host version ... this will take at least 1-3 more weeks.

src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs
satvu: @brettsam and I picked the shorter name when he was explaining to me how to get this value passed to GrpcWorker. I think we kept the shorter name for simplicity, but open to changing it to `AzureWebjobsScriptRoot`. The two names are equally understandable to me, but if anyone thinks the latter is more descriptive/understandable then I'm happy to change it and also clean up this line.

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -56,12 +59,20 @@ internal class GrpcWorker : IWorker
...
         public async Task StartAsync(CancellationToken token)
         {
             var eventStream = _rpcClient.EventStream(cancellationToken: token);
 
+            if(_startupOptions.Value.ScriptRoot is null)
+            {
+                throw new ArgumentNullException(nameof(_startupOptions.Value.ScriptRoot));
+            }
+
+            _functionMetadataResponseTask = GetFunctionMetadataAsync(_startupOptions.Value.ScriptRoot);
```
kshyju, line 74: Should we check workerIndexing is enabled before starting this task?

src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs
satvu: `IFunctionMetadataProvider` and `DefaultFunctionMetadataProvider` are in the Grpc project not the Core project. They are tied to the Grpc project because they're dependent on the proto file (deserializing the metadata file to grpc messages defined in the protofile). The proto-file is needed because we are using it to avoid a webjobs dependency (which the current GA function loader and sdk uses).

Moving this to Core and registering it in `AddFunctionsWorkerCore` would require some refactoring and creating a dependency on the gRPC project if we want to use the protofile. If we don't want to use the protofile or the webjobs dependency, and recreate our own functionmetadata and binding objects, we'd also have to maintain it to be consistent with the host / grpc definitions. 

This is what I remember from previous discussions / design questions I had from earlier iterations, but you are correct that this isn't something needed for grpc communication and would normally be part of the core project.

src/DotNetWorker.Grpc/GrpcWorker.cs
satvu: From the last meetings I've had, the plan for the isolated worker moving forward is going to be 100% worker-indexing. Technically, we already have that with the SDK (we load functions to the host through an extension, and host doesn't actually index functions). 

This PR basically guts that extension and moves that functionality into the worker in preparation towards source generation. We're not checking for `workerIndexing` to be enabled because there's no fallback to host-indexing for this worker (unlike other OOP workers). 

cc: @fabiocav

src/DotNetWorker.Grpc/FunctionMetadata/DefaultFunctionMetadataProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+            if (!hasDirection || !hasType)
+            {
+                throw new FormatException("Bindings must declare a direction and type.");
+            }
+
+            BindingInfo bindingInfo = new BindingInfo
+            {
+                Direction = Enum.Parse<BindingInfo.Types.Direction>(jsonDirection.ToString()!),
+                Type = jsonType.ToString()
+            };
```
fabiocav, line 102:
```suggestion
        internal static BindingInfo CreateBindingInfo(JsonElement binding)
        {
            var hasDirection = binding.TryGetProperty("direction", out JsonElement jsonDirection);
            var hasType = binding.TryGetProperty("type", out JsonElement jsonType);

            if (!hasDirection
                || !hasType
                || !Enum.TryParse<BindingInfo.Types.Direction>(jsonDirection.ToString()!, out BindingInfo.Types.Direction direction)
                || !Enum.TryParse<BindingInfo.Types.DataType>(jsonType.ToString()!, out BindingInfo.Types.DataType dataType))
            {
                throw new FormatException("Bindings must declare a direction and type.");
            }
            
            BindingInfo bindingInfo = new BindingInfo
            {
                Direction = direction,
                Type = jsonType.ToString()
            };

            var hasDataType = binding.TryGetProperty("dataType", out JsonElement jsonDataType);

            if (hasDataType)
            {
                bindingInfo.DataType = dataType;
            }

            return bindingInfo;
        }
```

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -56,12 +59,26 @@ internal class GrpcWorker : IWorker
...
         public async Task StartAsync(CancellationToken token)
         {
             var eventStream = _rpcClient.EventStream(cancellationToken: token);
 
+            if(_startupOptions.Value.ScriptRoot is null)
+            {
+                throw new ArgumentNullException(nameof(_startupOptions.Value.ScriptRoot));
+            }
+
+            if (!string.IsNullOrEmpty(_startupOptions.Value.AzureWebJobsFeatureFlags))
```
satvu, line 74: Adding this here to make sure duplicate worker-indexing isn't being done. Previously, worker-indexing would be performed here no matter the settings on the host/worker, but it wouldn't break anything b/c this info wasn't sent to the host unless asked for. 

Setting a check to only perform this work if we know that the host will try to ask for it (aka it has EnableWorkerIndexing set). We could additionally make this check stricter by checking that worker.config.json had "workerIndexing" set to true as well, but I think this is check is good enough. Any thoughts?

src/DotNetWorker.Grpc/GrpcStartupOptions.cs
```diff
@@ -17,5 +17,7 @@ internal class GrpcWorkerStartupOptions
         public int GrpcMaxMessageLength { get; set; }
 
         public string? ScriptRoot { get; set; }
+
+        public string? AzureWebJobsFeatureFlags { get; set; }
```
kshyju, line 21: Will HOST send this value? Where does host get this value from? Is it a comma separated string of feature names? Can you share how a sample value look like?

satvu: This is set in `local.settings.json` on the worker side, or if you're running from the host solution, you can add it to `launchsettings.json`. For our purposes, you just need `"AzureWebJobsFeatureFlags":"EnableWorkerIndexing"` but you can also have multiple features separated by a semicolon.

fabiocav: Is the reason for the worker to care about this an optimization so we kick off the task early as opposed to deferring to when the request is sent?

sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
```diff
@@ -30,6 +30,8 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
     <_FunctionsExtensionsJsonName>extensions.json</_FunctionsExtensionsJsonName>
     <_FunctionsExtensionsFullPublish Condition="$(NoBuild) == '' And $(_FunctionsExtensionsFullPublish) == ''">True</_FunctionsExtensionsFullPublish>
     <_FunctionsExtensionsFullPublish Condition="$(_FunctionsExtensionsFullPublish) == ''">!$(NoBuild)</_FunctionsExtensionsFullPublish>
+
+    <EnableWorkerIndexing Condition="$(EnableWorkerIndexing) == ''">false</EnableWorkerIndexing>
```
brettsam, line 34: I *think* that we have been prefixing all our publicly available properties with "Functions"... so this should be `FunctionsEnableWorkerIndexing`.

src/DotNetWorker.Grpc/GrpcStartupOptions.cs
```diff
@@ -15,5 +15,9 @@ internal class GrpcWorkerStartupOptions
         public string? RequestId { get; set; }
 
         public int GrpcMaxMessageLength { get; set; }
+
+        public string? ScriptRoot { get; set; }
```
brettsam, line 19: While I like this being short and not having `AzureWebJobs`... let's just make it `AzureWebJobsScriptRoot` so we don't have to do any custom binding later.

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -56,12 +59,26 @@ internal class GrpcWorker : IWorker
...
             var eventStream = _rpcClient.EventStream(cancellationToken: token);
 
+            if(_startupOptions.Value.AzureWebJobsScriptRoot is null)
+            {
+                throw new ArgumentNullException(nameof(_startupOptions.Value.AzureWebJobsScriptRoot));
+            }
+
+            if (!string.IsNullOrEmpty(_startupOptions.Value.EnableWorkerIndexing))
+            {
+                if (_startupOptions.Value.EnableWorkerIndexing.Equals("true", StringComparison.OrdinalIgnoreCase))
```
satvu, line 76: [This host PR](https://github.com/Azure/azure-functions-host/pull/8301) would need to be merged for this value to be propagated correctly to this location.

sdk/FunctionMetadataLoaderExtension/Startup.cs
```diff
@@ -23,17 +23,29 @@ internal class Startup : IWebJobsStartup2, IWebJobsConfigurationStartup
...
         {
             string appRootPath = context.ApplicationRootPath;
 
             // We need to adjust the path to the worker exe based on the root, if WorkerRootToken is found.
             WorkerConfigDescription newWorkerDescription = GetUpdatedWorkerDescription(appRootPath);
 
+
+            if (!string.IsNullOrEmpty(newWorkerDescription.EnableWorkerIndexing))
+            {
+                if (newWorkerDescription.EnableWorkerIndexing!.Equals("true", StringComparison.OrdinalIgnoreCase))
```
kshyju, line 43: Should we use `bool` as the type for `EnableWorkerIndexing` property in `WorkerConfigDescription` class? The serializer will be able to populate the Boolean value. With that, we probably don't need 41-47 ?

sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
```diff
@@ -30,6 +30,8 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
     <_FunctionsExtensionsJsonName>extensions.json</_FunctionsExtensionsJsonName>
     <_FunctionsExtensionsFullPublish Condition="$(NoBuild) == '' And $(_FunctionsExtensionsFullPublish) == ''">True</_FunctionsExtensionsFullPublish>
     <_FunctionsExtensionsFullPublish Condition="$(_FunctionsExtensionsFullPublish) == ''">!$(NoBuild)</_FunctionsExtensionsFullPublish>
+
+    <FunctionsEnableWorkerIndexing Condition="$(FunctionsEnableWorkerIndexing) == ''">false</FunctionsEnableWorkerIndexing>
```
kshyju, line 34: Other functions entries uses an `_` prefix.  Should we follow that to be consistent?

src/DotNetWorker.Grpc/FunctionMetadata/DefaultFunctionMetadataProvider.cs
```diff
@@ -0,0 +1,121 @@
...
+                        }
+
+                        functionMetadataResults.Add(functionMetadata);
+                    }
+
+                    return functionMetadataResults.ToImmutableArray();
+                }
+            }
+
+            return ImmutableArray<RpcFunctionMetadata>.Empty;
```
kshyju, line 72: Should [we throw](https://github.com/Azure/azure-functions-dotnet-worker/pull/825) when the file is missing?

sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
satvu: Those are for private ones that our end uses/sets. Without the "_" prefix should be things that customers can set. cc: @brettsam

src/DotNetWorker.Grpc/FunctionMetadata/DefaultFunctionMetadataProvider.cs
satvu: good call

src/DotNetWorker.Grpc/FunctionMetadata/DefaultFunctionMetadataProvider.cs
```diff
@@ -0,0 +1,122 @@
...
+                    }
+
+                    functionMetadataResults.Add(functionMetadata);
+                }
+
+                return functionMetadataResults.ToImmutableArray();
+            }
+        }
+
+
```
brettsam, line 75: nit: should only have one line break between methods...

sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
brettsam: right -- the intent was this is something customers can use to turn this off (if they ever need to) -- and with those "public" properties, we drop the "_"

sdk/Sdk/worker.config.json
```diff
@@ -3,6 +3,7 @@
     "language": "dotnet-isolated",
     "extensions": [ ".dll" ],
     "defaultExecutablePath": "$functionExe$",
-    "defaultWorkerPath": "$functionWorker$"
+    "defaultWorkerPath": "$functionWorker$",
+    "workerIndexing":  "$enableWorkerIndexing$"
```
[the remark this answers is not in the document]
satvu: Host uses this value and expects a string

brettsam (approved):

satvu:
This preview version will work with host version v4.3.2 and higher. There is currently no core-tools package that uses that host-version (latest one uses v4.1.3 which won't be compatible) so this is pending the next core-tools release.

sdk/FunctionMetadataLoaderExtension/Startup.cs
```diff
@@ -23,16 +23,20 @@ internal class Startup : IWebJobsStartup2, IWebJobsConfigurationStartup
...
 
+        private WorkerConfigDescription? newWorkerDescription;
+
         public void Configure(WebJobsBuilderContext context, IWebJobsConfigurationBuilder builder)
         {
             string appRootPath = context.ApplicationRootPath;
 
             // We need to adjust the path to the worker exe based on the root, if WorkerRootToken is found.
-            WorkerConfigDescription newWorkerDescription = GetUpdatedWorkerDescription(appRootPath);
+            // Both Configure methods uses this and require info from WebJobsBuilderContext, so each will either check if it already exists or create it using GetUpdatedWorkerDescription()
```
madelinegordon, line 38: nit-nit on the comment: "use"

src/DotNetWorker.Grpc/FunctionMetadata/DefaultFunctionMetadataProvider.cs
```diff
@@ -0,0 +1,121 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.IO;
+using System.Threading.Tasks;
```
madelinegordon, line 8: nit: alphabetical ordering switch System.Threading.Tasks & System.Text.Json

PR #696, opened 2021-11-01
Trim whitespace when processing cookies in HttpRequestData

### LINKED ISSUE #694: HttpRequestData.Cookies is Empty
When using an HTTP triggered function, it appears that `HttpRequestData.Cookies` is empty. This appears to be an issue that was previously reported and resolved in #466 though I am still not seeing cookies despite them being present in `Headers`. I am currently using version 1.5.2.

As a workaround I copied [`ToHttpCookie`](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs#L147) locally to my project to keep moving and may have noticed another bug.

When splitting the cookies out of the full cookie string, cookie names are not trimmed and contain leading whitespace if it was included in the string. 
Example:
Browser sends the following cookie header: `another-token=aasdfasdf; some-token=asdfasdf`

The dict key for `some-token` will in fact be `[space]some-token`.
Note that when I set the cookie I did not include a leading whitespace. The browser (Chrome in this case) seems to separate multiple cookies with `; ` rather than simply `;`.

satvu:

resolves #694

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
+++ b/src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
@@ -153,8 +153,8 @@ namespace Microsoft.Azure.Functions.Worker
             for (int c = 0; c < separateCookies.Length; c++)
             {
                 var splitArray = separateCookies[c].Split("=", StringSplitOptions.RemoveEmptyEntries);
-                var name = splitArray[0];
-                var value = splitArray[1];
+                var name = splitArray[0].Trim();
+                var value = splitArray[1].Trim();
                 httpCookiesList.Add(new HttpCookie(name, value));
             }
 


src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
```diff
@@ -153,8 +153,8 @@ private IReadOnlyCollection<IHttpCookie> ToHttpCookies(string cookieString)
             for (int c = 0; c < separateCookies.Length; c++)
             {
                 var splitArray = separateCookies[c].Split("=", StringSplitOptions.RemoveEmptyEntries);
```
kshyju, line 155: Use `.Split(",", StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);` and then you do not need the explicit `Trim` call on each item (and I would imagine that would also save as some string allocation)

release_notes.md
```diff
@@ -4,3 +4,4 @@
 -->
 - Fixes `null` reference exception when accessing request identities (#673)
 - Exposed `RetryContext` to support retry policies (#316)
+- Fixes leading/trailing whitespaces in HttpRequestData cookies (#694)
```
kshyju, line 7: We could also release 1.5.3 of worker since we have a few fixes to go out.

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
kshyju: My bad! As discussed offline, this will cause the cookie value to be also trimmed, which is something we don't want.

kshyju (approved):

PR #711, opened 2021-11-10
Enable integration testing for the Dotnet language worker

Francisco-Gamino:

resolves https://github.com/Azure/azure-functions-dotnet-worker/issues/709

This PR contains the following changes:
* Updated `setup-e2e-tests.ps1` to install the Core Tools to validate the language worker against Functions V3 and V4
* Added `install-dotnet.yml` script to install the .NET version required by the Core Tools for Functions V4
* Added the following two variables to the pipeline to control what version of the Core Tools is installed (to validate the Dotnet worker):
   * `FunctionsRuntimeVersion` - Specifies the major version of the Core Tools to be installed. The default value is `3`. To validate the worker against the Core Tools V4, set this value to `4`
   * `UseCoreToolsBuildFromIntegrationTests` – This is a boolean value to specify whether to use the Core Tools for integration testing. This value will be set by the integration test framework. By default, the value is set to `False`.
* Updated `azure-pipelines.yml` to parse the `UseCoreToolsBuildFromIntegrationTests` pipeline variable and pass the boolean value to the `setup-e2e-tests.ps1` script

 
### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -29,6 +29,7 @@ jobs:
   - pwsh: ./tools/start-emulators.ps1 -NoWait
     displayName: "Start emulators (-NoWait)"
 
+  - template: build/install-dotnet.yml
   - task: DotNetCoreCLI@2
     displayName: 'Build projects'
     inputs:
@@ -36,7 +37,14 @@ jobs:
       arguments: '-c Release -p:BuildNumber=$(buildNumber) -p:IsLocalBuild=False'
       projects: 'DotNetWorker.sln'
 
-  - pwsh: ./setup-e2e-tests.ps1
+  - pwsh: |
+      $useIntegrationTestingCoreTools = $null
+      if (-not ([bool]::TryParse($env:USECORETOOLSBUILDFROMINTEGRATIONTESTS, [ref] $useIntegrationTestingCoreTools)))
+      {
+        throw "UseCoreToolsBuildFromIntegrationTests can only be set to True or False. Current value is set to $env:USECORETOOLSBUILDFROMINTEGRATIONTESTS"
+      }
+      ./setup-e2e-tests.ps1 -FunctionsRuntimeVersion $env:FUNCTIONSRUNTIMEVERSION `
+                            -UseCoreToolsBuildFromIntegrationTests:$useIntegrationTestingCoreTools
     displayName: "Setup E2E tests"
 
   - task: DotNetCoreCLI@2
@@ -272,6 +280,7 @@ jobs:
     vmImage: 'ubuntu-latest'
     
   steps:
+    - template: build/install-dotnet.yml
     - task: DotNetCoreCLI@2
       displayName: 'Build projects'
       inputs:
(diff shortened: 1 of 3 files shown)


Francisco-Gamino:
@brettsam -- Once the PR is merged, I will set the pipelines variables `CORE_TOOLS_URL` and `CORE_TOOLS_URL_LINUX` to null/empty string.

azure-pipelines.yml
```diff
@@ -29,14 +29,25 @@ jobs:
   - pwsh: ./tools/start-emulators.ps1 -NoWait
     displayName: "Start emulators (-NoWait)"
 
+  - pwsh: ./install-dotnet.ps1 -Version '6.0.100-rc.2.21505.57'
```
brettsam, line 32: Could you follow the pattern I'm using in the host for this by using a yaml template? It makes sure there's only one place we need to update this if we have to move versions.

https://github.com/Azure/azure-functions-host/blob/v4.x/azure-pipelines.yml#L45
https://github.com/Azure/azure-functions-host/blob/v4.x/build/install-dotnet.yml

brettsam (commented):
Looks good -- just one comment!

azure-pipelines.yml
Francisco-Gamino: Thank you @brettsam for your review and feedback. 

Currently, https://github.com/Azure/azure-functions-host/blob/v4.x/build/install-dotnet.yml does not cover the Linux scenario, so we would have to add the equivalent of this script `install-dotnet.ps1`  in yml. 

Would you be ok if we set the version to be installed in the script itself. This way, we can call the script without parameters, e.g., 

`- pwsh: ./install-dotnet.ps1`


```
param(
    [String]
    $Version = '6.0'

    [String]
    $Channel = 'release'
)

$obtainUrl = "https://raw.githubusercontent.com/dotnet/cli/master/scripts/obtain"
try
{
    $installScript = if ($IsWindows) { "dotnet-install.ps1" } else { "dotnet-install.sh" }
    Invoke-WebRequest -Uri $obtainUrl/$installScript -OutFile $installScript

    if ($IsWindows)
    {
        & .\$installScript -Channel $Channel -Version $Version -InstallDir "$env:ProgramFiles/dotnet"
    }
    else
    {
        bash ./$installScript -c $Channel -v $Version --install-dir "/usr/share/dotnet"
    }
}
finally
{
    Remove-Item $installScript -Force -ErrorAction SilentlyContinue
}
```

Alternatively, if we decide to go the yml route, this is how the code would look like:

```
- pwsh: |
    Invoke-WebRequest 'https://raw.githubusercontent.com/dotnet/cli/master/scripts/obtain/dotnet-install.ps1' -OutFile 'dotnet-install.ps1'
    ./dotnet-install.ps1 -InstallDir "$env:ProgramFiles/dotnet" -Version "6.0" -Channel 'release'
  displayName: 'Install the .Net version used by the Core Tools for Windows'
  condition: eq( variables['Agent.OS'], 'Windows_NT' )

- bash: |
    curl -sSL https://raw.githubusercontent.com/dotnet/cli/master/scripts/obtain/dotnet-install.sh | bash /dev/stdin -v '6.0' -c 'release' --install-dir /usr/share/dotnet
  displayName: 'Install the .Net version used by the Core Tools for Linux'
  condition: eq( variables['Agent.OS'], 'Linux' )
```

Thoughts?

brettsam: I like the yaml route better -- (i'd like to move more things, like signing, out so it keeps the core yaml smaller)

Francisco-Gamino: Sounds good. I will move the code to yml.

Francisco-Gamino:
@brettsam -- I've the dotnet installation logic to `install-dotnet.yml`. Could you please take another look? 

This is the pipeline execution when validating the worker against Functions V4 https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=41247&view=results

brettsam (approved):

Issue #771, opened 2022-01-08
Signalr Extensions using Class-based model is not currently supported


michaeldaw:
Worker.Extensions.SignalRService does not currently contains classes required to develop Functions using the [Class based model](https://docs.microsoft.com/en-us/azure/azure-signalr/signalr-concept-serverless-development-config#class-based-model).
eg:
 - ServerlessHub
 - InvocationContext
 - SignalRConnectionInfo
  
Are there plans to support the Class Based Model when using Worker.Extensions.SignalRService?

fabiocav:
@Y-Sindo have you had a chance to look at this? Some of them should be very simple to bring in.

Y-Sindo:
I will add a PR to provide some convenient classes such as `SignalRConnectionInfo`, `SignalRInvocationContext` soon. 

About the serverless hub in isolated-process model, we do have a plan to support it, but we still have some problems to overcome.

Y-Sindo:
@fabiocav I have added a PR to address the issue, please take a look at it.

fabiocav:
Fantastic! Thanks @Y-Sindo. We'll review that ASAP.

FYI @anthonychu

michaeldaw:
Sorry to interrupt the discussion. Just wanted to say thanks for the response.

Arash-Sabet:
@fabiocav @anthonychu Any idea when the PR associated with this issue will be approved and released? Thanks.

Y-Sindo:
@fabiocav @kshyju Currently the SignalR class-based model can bind the RPC call parameters to trigger function parameters automatically, I want to know if the bindings in host process can get the parameter info of the trigger functions in worker process, so that we can do similar things in isolated model.

liliankasem:
@kshyju looks like the PR was merged last week, any idea on a potential release date?

michaeldaw:
Thanks to everyone who's worked on this. Will it be part of a release soon? I've started a project that could really use this 😊

duncanthescot:
What is the current status of this?  The last message I see is from a year ago.

Y-Sindo:
Please see this project for an example to integrate with SignalR management SDK https://github.com/aspnet/AzureSignalR-samples/tree/main/samples/DotnetIsolated-ClassBased

Y-Sindo:
[...]

PR #774, opened 2022-01-12
[SignalR] Add convenient POJOs.

Y-Sindo:

resolves partially #771

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/extensions/Worker.Extensions.SignalRService/src/Models/SignalRConnectionInfo.cs
+++ b/extensions/Worker.Extensions.SignalRService/src/Models/SignalRConnectionInfo.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    /// <summary>
+    /// Contains necessary information for a SignalR client to connect to SignalR Service.
+    /// </summary>
+    public sealed class SignalRConnectionInfo
+    {
+        /// <summary>
+        /// Gets the URL for a client to connect to SignalR Service.
+        /// </summary>
+        public string Url { get; set; }
+
+        /// <summary>
+        /// Gets the access token for a client to connect to SignalR service.
+        /// </summary>
+        public string AccessToken { get; set; }
+    }
+}
--- a/extensions/Worker.Extensions.SignalRService/src/Models/SignalRGroupAction.cs
+++ b/extensions/Worker.Extensions.SignalRService/src/Models/SignalRGroupAction.cs
@@ -1,0 +1,68 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    /// <summary>
+    /// An action to manage a group.
+    /// </summary>
+    public sealed class SignalRGroupAction
+    {
+        /// <summary>
+        /// Initializes an instance of a <see cref="SignalRGroupAction"/>.
+        /// </summary>
+        /// <param name="actionType">The action type.</param>
+        public SignalRGroupAction(SignalRGroupActionType actionType)
+        {
+            Action = actionType;
+        }
+
+        /// <summary>
+        /// Sets this field if you want to add a connection to a group or remove it from a group.
+        /// </summary>
+        public string? ConnectionId { get; set; }
+
+        /// <summary>
+        /// Sets this field if you want to add a user to a group or remove it from a group/all groups.
+        /// </summary>
+        public string? UserId { get; set; }
+
+        /// <summary>
+        /// The group to manage. Leave it null if you want to remove a user from all groups.
+        /// </summary>
+        public string? GroupName { get; set; }
+
+        /// <summary>
+        /// The group action type.
+        /// </summary>
+        public SignalRGroupActionType Action { get; set; }
+
+        /// <summary>
+        /// The SignalR endpoints to send to. Leave it null if you want to send to all endpoints.
+        /// You can get the available SignalR endpoints from the <see cref="SignalREndpointsInputAttribute"/> binding.
+        /// </summary>
+        public SignalREndpoint[]? Endpoints { get; set; }
+    }
+
+    /// <summary>
+    /// The type of a group action.
+    /// </summary>
+    [JsonConverter(typeof(JsonStringEnumConverter))]
+    public enum SignalRGroupActionType
+    {
+        /// <summary>
+        /// Adds a user or a connection to a group.
+        /// </summary>
+        Add,
+        /// <summary>
+        /// Removes a user or a connection to a group.
+        /// </summary>
+        Remove,
+        /// <summary>
+        /// Remove a user from all the groups.
+        /// </summary>
+        RemoveAll
+    }
+}
--- a/extensions/Worker.Extensions.SignalRService/src/Models/SignalRInvocationContext.cs
+++ b/extensions/Worker.Extensions.SignalRService/src/Models/SignalRInvocationContext.cs
@@ -1,0 +1,76 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    /// <summary>
+    /// The context for a SignalR invocation.
+    /// </summary>
+    public sealed class SignalRInvocationContext
+    {
+        /// <summary>
+        /// The arguments of the invocation.
+        /// </summary>
+        public object[]? Arguments { get; set; }
+
+        /// <summary>
+        /// The error message of connections disconnected event.
+        /// Only connections disconnected event can have this property, and it can be empty if the connection is closed with no error.
+        /// </summary>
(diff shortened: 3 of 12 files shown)


extensions/Worker.Extensions.SignalRService/src/Models/SignalRConnectionInfo.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// </summary>
+    public sealed class SignalRConnectionInfo
+    {
+        /// <summary>
+        /// Gets the URL for a client to connect to SignalR Service.
+        /// </summary>
+        public string Url { get; set; }
+
+        /// <summary>
+        /// Gets the access token for a client to connect to SignalR service.
```
kshyju, line 17: nit: "gets or sets"

extensions/Worker.Extensions.SignalRService/src/Models/SignalRGroupAction.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// The SignalR endpoints to send to. Leave it null if you want to send to all endpoints.
+        /// You can get the available SignalR endpoints from the <see cref="SignalREndpointsInputAttribute"/> binding.
+        /// </summary>
+        public SignalREndpoint[]? Endpoints { get; set; }
+    }
+
+    /// <summary>
+    /// The type of a group action.
+    /// </summary>
+    [JsonConverter(typeof(JsonStringEnumConverter))]
```
kshyju, line 52: Question: Wonder how it will work if customer chooses to override JSON serializer to be JSON.NET.

extensions/Worker.Extensions.SignalRService/src/Models/SignalRInvocationContext.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// <summary>
+        /// The arguments of the invocation.
+        /// </summary>
+        public object[]? Arguments { get; set; }
+
+        /// <summary>
+        /// The error message of connections disconnected event.
+        /// Only connections disconnected event can have this property, and it can be empty if the connection is closed with no error.
+        /// </summary>
+        public string Error { get; set; }
```
kshyju, line 23: Should this be nullable?

extensions/Worker.Extensions.SignalRService/src/Models/SignalRInvocationContext.cs
```diff
@@ -0,0 +1,76 @@
...
+        public string ConnectionId { get; set; }
+
+        /// <summary>
+        /// The user ID of the client which triggers the invocation.
+        /// </summary>
+        public string UserId { get; set; }
+
+        /// <summary>
+        /// The headers of request.
+        /// Headers with duplicated key will be joined by comma.
```
kshyju, line 53: >will be joined by comma.

Or should we consider using `<string,stringvalues>`  as the type?

extensions/Worker.Extensions.SignalRService/src/Models/SignalRInvocationContext.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// The claims of the client.
+        /// If you multiple claims have the same key, only the first one will be reserved.
+        /// </summary>
+        public IDictionary<string, string> Claims { get; set; }
+    }
+
+    [JsonConverter(typeof(JsonStringEnumConverter))]
+    public enum SignalRInvocationCategory
+    {
+        connections,
```
kshyju, line 73: Switch to PascalCase?

extensions/Worker.Extensions.SignalRService/src/Models/SignalRMessageAction.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        /// <summary>
+        /// The group to send to. Sets this field if you want to send a message to a group.
+        /// </summary>
+        public string? GroupName { get; set; }
+
+        /// <summary>
+        /// Required. The target to send to.
+        /// </summary>
+        public string Target { get; set; }
```
kshyju, line 49: Should we remove `set` since we are setting it inside the constructor?

extensions/Worker.Extensions.SignalRService/src/Models/SignalRNegotiationContext.cs
```diff
@@ -0,0 +1,43 @@
...
+    
+    /// <summary>
+    /// Contains metadata of an SignalR endpoint and connection information for a client to connect to the SignalR Service.
+    /// </summary>
+    public sealed class SignalREndpointConnectionInfo
+    {
+        /// <summary>
+        /// The type of an endpoint.
+        /// </summary>
+        public SignalREndpointType EndpointType { get; set; }
```
kshyju, line 25: nit: Add line break.

kshyju, line 25: We have the other class (`SignalREndpoint`) which has all this info exception the connection info prop. Should we use that type here? (as a prop of this class)

release_notes.md
```diff
@@ -2,4 +2,5 @@
 <!-- Please add your release notes in the following format:
 - My change description (#PR/#issue)
 -->
-- Storage extensions moved to 5.0.0
\ No newline at end of file
+- Storage extensions moved to 5.0.0
+- SignalR extensions added some convenient POJOs (PR #774)
```
kshyju, line 6: We should also consider adding sample usage of this types in the extension samples.

extensions/Worker.Extensions.SignalRService/src/Models/SignalRGroupAction.cs
Y-Sindo: I have tried to change the JSON serializer to Newtonsoft.JSON as the code below and it still works. Maybe the conversion for bindings is not controlled by users. I am not sure about this, maybe we need a dotnet worker expert to confirm this.
```cs
    class Program
    {
        static async Task Main(string[] args)
        {
            var host = new HostBuilder()
                .ConfigureFunctionsWorkerDefaults(o =>
                {
                    var settings = NewtonsoftJsonObjectSerializer.CreateJsonSerializerSettings();
                    settings.ContractResolver = new CamelCasePropertyNamesContractResolver();
                    settings.NullValueHandling = NullValueHandling.Ignore;
                    o.Serializer = new NewtonsoftJsonObjectSerializer(settings);
                })
                .Build();
            await host.RunAsync();
```

Y-Sindo: Is there any way to avoid customers changing the serialization and deserialization of some field of my POJO? For example, I want to make the enum `SignalRGroupAction` serialized to string instead of number no matter how customers set their object serializer. But for the `Arguments` field in `SignalRInvocationContext`, `SignalRMessageAction`, I want to let the customers to control the serialization/deserialization.

@fabiocav

release_notes.md
Y-Sindo: Currently the extension samples use package reference, so we cannot add new usages to the samples until they are released.

extensions/Worker.Extensions.SignalRService/src/Models/SignalRInvocationContext.cs
Y-Sindo: I think it is a good idea, but deserialization is still a problem. If we customize the deserialization, how to avoid the customization changed by customers.

extensions/Worker.Extensions.SignalRService/src/Models/SignalRNegotiationContext.cs
Y-Sindo: The structure of the type is consistent with the original binding. We flatten the structure intentionally so that customers can do things like `return signalRNegotiationContext.Endpoints.Select(e=>e.Online==true).First().ConnectionInfo` in a function. If we change the structure, we need to do extra JSON deserialization customization which still has the problem that might be changed by customers.

extensions/Worker.Extensions.SignalRService/src/Models/SignalRGroupAction.cs
kshyju: This is not a blocker to merge this change. Something we should validate though.

kshyju (approved):

extensions/Worker.Extensions.SignalRService/src/Models/SignalRGroupAction.cs
fabiocav: To add context. No, today, the configured serializer will be used by default. We're adding features that will give binding authors better control of this, but in the meantime, I agree with @kshyju , we don't need to block, but it might be something worth documenting.

extensions/Worker.Extensions.SignalRService/src/Models/SignalRInvocationContext.cs
fabiocav: Would be nice (not critical) to match how `HttpRequestData` exposes this information for consistency, but that uses `HttpHeadersCollection`

fabiocav (approved):

extensions/Worker.Extensions.SignalRService/src/Models/SignalRInvocationContext.cs
Y-Sindo: `HttpHeadersCollection` is in a .NET 5 project, and cannot be referenced by .NET Standard 2.0 project.

Y-Sindo:
I think the PR is ready to be merged now.

Y-Sindo:
> > I think the PR is ready to be merged now. Thanks!
> 
> Feel free to merge.

 I don't have the write access to merge currently.

PR #814, opened 2022-03-03
APIs for reading and updating input binding data, output binding data and invocation result.

kshyju:
Exposing the below extension methods on function context:

````
public static ValueTask<InputBindingData<T>> BindInputAsync<T>(BindingMetadata bindingMetadata);
public static InvocationResult<T> GetInvocationResult<T>();;
public static InvocationResult GetInvocationResult();
public static IEnumerable<OutputBindingData<T>> GetOutputBindings<T>();

// Http trigger specific
public static ValueTask<HttpRequestData?> GetHttpRequestDataAsync();
public static HttpResponseData? GetHttpResponseData();
````

The `BindInputAsync` method internally relies on the input conversion feature and caches the result of it so that it can be used for a second time if needed, within the same invocation.

### Usage samples:

1) Global exception handling middleware for http functions.
````
{
    try
    {
        await next(context);
    }
    catch (Exception ex)
    {
        _logger.LogError(ex, "Error processing invocation");

        var httpReqData = await context.GetHttpRequestDataAsync();

        if (httpReqData != null)
        {
            var newResponse = httpReqData.CreateResponse();
            await newResponse.WriteAsJsonAsync(new { Status = "Failed" });

            // Update invocation result.
            var httpInvoationResult = context.GetInvocationResult();
            httpInvoationResult.Value = newResponse;
        }
    }
}
````

2) To get input data inside a middleware for a specific input binding entry.

````
BindingMetadata blobBindingMetaData = context.FunctionDefinition
                                             .InputBindings.Values
                                             .Where(a => a.Type == "blob")
                                             .FirstOrDefault();

if (blobBindingMetaData != null)
{
    var bindingResult = await context.BindInputAsync<MyBlob>(blobBindingMetaData);
    // Update a property value. 
    // This will be reflected in the function parameter value.
    bindingResult.Value.Name = "Edited name";
   
   // or you could even replace the entire object
   bindingResult.Value = new MyBlob { Name = "Totally different object" };
}

````

3) Middleware using these APIs to stamp http response headers.

````
{
    var logger = context.GetLogger<StampHttpHeadersMiddleware>();
    var requestId = "azf-" + Guid.NewGuid();

    using (logger.BeginScope("azfunc-requestid:{requestId}", requestId))
    {
        await next(context);
    }

    var httpRequestData = await context.GetHttpRequestDataAsync();
    if (httpRequestData != null)
    {
        var httpResponseData = context.GetHttpResponseData();
        if (httpResponseData != null)
        {
            httpResponseData.Headers.Add("x-azfunc-requestid", requestId);
        }
    }
}
````

### DIFF
--- a/src/DotNetWorker.Core/Context/DefaultBindingCache.cs
+++ b/src/DotNetWorker.Core/Context/DefaultBindingCache.cs
@@ -1,0 +1,42 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Concurrent;
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    internal class DefaultBindingCache<T> : IBindingCache<T>
+    {
+        private readonly ConcurrentDictionary<string, T> _cache = new();
+
+        public bool TryGetValue(string key, out T? value)
+        {
+            EnsureKeyNotNull(key);
+
+            if (_cache.TryGetValue(key, out var conversionResult))
+            {
+                value = conversionResult;
+                return true;
+            }
+
+            value = default;
+            return false;
+        }
+
+        public bool TryAdd(string key, T value)
+        {
+            EnsureKeyNotNull(key);
+
+            return _cache.TryAdd(key, value);
+        }
+
+        private static void EnsureKeyNotNull(string key)
+        {
+            if (key == null)
+            {
+                throw new ArgumentNullException(nameof(key));
+            }
+        }
+    }
+}
--- a/src/DotNetWorker.Core/Context/DefaultBindingCache.cs
+++ b/src/DotNetWorker.Core/Context/DefaultBindingCache.cs
@@ -9,10 +9,22 @@ namespace Microsoft.Azure.Functions.Worker
     /// <summary>
     /// In memory cache for binding data (per invocation)
     /// </summary>
-    /// <typeparam name="T"></typeparam>
+    /// <typeparam name="T">The type of cache entry value.</typeparam>
     internal class DefaultBindingCache<T> : IBindingCache<T>
     {
         private readonly ConcurrentDictionary<string, T> _cache = new();
+
+        public T this[string key]
+        {
+            get
+            {
+                return _cache[key];
+            }
+            set
+            {
+                _cache[key] = value;
+            }
+        }
 
         public bool TryGetValue(string key, out T? value)
         {
--- a/src/DotNetWorker.Core/Context/DefaultBindingCache.cs
+++ b/src/DotNetWorker.Core/Context/DefaultBindingCache.cs
@@ -6,6 +6,10 @@ using System.Collections.Concurrent;
 
 namespace Microsoft.Azure.Functions.Worker
 {
+    /// <summary>
+    /// In memory cache for binding data (per invocation)
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
     internal class DefaultBindingCache<T> : IBindingCache<T>
     {
         private readonly ConcurrentDictionary<string, T> _cache = new();
@@ -14,14 +18,7 @@ namespace Microsoft.Azure.Functions.Worker
         {
             EnsureKeyNotNull(key);
 
-            if (_cache.TryGetValue(key, out var conversionResult))
-            {
-                value = conversionResult;
-                return true;
-            }
-
-            value = default;
-            return false;
+            return _cache.TryGetValue(key, out value);
         }
 
         public bool TryAdd(string key, T value)
--- a/src/DotNetWorker.Core/Context/FunctionContextBindingFeatureExtensions.cs
+++ b/src/DotNetWorker.Core/Context/FunctionContextBindingFeatureExtensions.cs
@@ -16,26 +16,6 @@ namespace Microsoft.Azure.Functions.Worker
     public static class FunctionContextBindingFeatureExtensions
     {
         /// <summary>
-        /// Gets the input binding data for the current function invocation.
-        /// </summary>
-        /// <param name="context">The function context instance.</param>
-        /// <returns>The input binding data as a read only dictionary.</returns>
-        public static IReadOnlyDictionary<string, object?> GetInputData(this FunctionContext context)
-        {
-            return context.GetBindings().InputData;
-        }
-
-        /// <summary>
-        /// Gets the trigger meta data for the current function invocation.
-        /// </summary>
-        /// <param name="context">The function context instance.</param>
-        /// <returns>The invocation trigger meta data as a read only dictionary.</returns>
-        public static IReadOnlyDictionary<string, object?> GetTriggerMetadata(this FunctionContext context)
-        {
-            return context.GetBindings().TriggerMetadata;
-        }
-
-        /// <summary>
         /// Binds an input item for the requested type.
         /// </summary>
         /// <typeparam name="T">The type of input item to bind to.</typeparam>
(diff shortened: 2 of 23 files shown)

### CODE REVIEW

File: src/DotNetWorker.Core/Context/DefaultBindingCache.cs [hunk 1]
```diff
@@ -0,0 +1,42 @@
...
+            {
+                value = conversionResult;
+                return true;
+            }
+
+            value = default;
+            return false;
+        }
+
+        public bool TryAdd(string key, T value)
```
kshyju, line 27: We could also add a GetOrAdd which takes the key and a value factory func. But that can come as an enhancement later. This type is internal.

src/DotNetWorker.Core/Context/FunctionContextBindingFeatureExtensions.cs
```diff
@@ -0,0 +1,142 @@
...
+    /// </summary>
+    public static class FunctionContextBindingFeatureExtensions
+    {
+        /// <summary>
+        /// Binds an input binding item for the requested <see cref="BindingMetadata"/> instance.
+        /// </summary>
+        /// <param name="context">The function context.</param>
+        /// <param name="bindingMetadata">The BindingMetadata instance for which input data should bound to.</param>
+        /// <returns>An instance of T if binding was successful, else null</returns>
+        public static async ValueTask<T?> BindInputAsync<T>(this FunctionContext context, BindingMetadata bindingMetadata)
```
kshyju, line 24: See PR description for sample usage of this API.

src/DotNetWorker.Core/Context/FunctionContextHttpRequestExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// <returns>HttpResponseData instance if the invocation is http, else null</returns>
+        public static HttpResponseData? GetHttpResponseData(this FunctionContext context)
+        {
+            var httpInvocationResult = context.GetInvocationResult();
+            if (httpInvocationResult.Value is HttpResponseData responseData)
+            {
+                return responseData;
+            }
+            
+            // see output binding entries have a property of type HttpResponseData;
```
kshyju, line 46: for use case like[ this](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/samples/FunctionApp/Function3/Function3.cs#L27)

src/DotNetWorker.Core/Context/IBindingCache.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    internal interface IBindingCache<T>
```
kshyju, line 6: Feel free to suggest a better name for this type. It is a per invocation cache. `IInvocationDataCache` ?

src/DotNetWorker.Core/Hosting/ServiceCollectionExtensions.cs
```diff
@@ -56,6 +56,9 @@ public static IFunctionsWorkerApplicationBuilder AddFunctionsWorkerCore(this ISe
             services.AddSingleton<IConverterContextFactory, DefaultConverterContextFactory>();
             services.AddSingleton<IInputConversionFeatureProvider, DefaultInputConversionFeatureProvider>();
             services.AddSingleton<IInputConverterProvider, DefaultInputConverterProvider>();
+            
+            // Input binding cache
+            services.AddScoped<IBindingCache<ConversionResult>, DefaultBindingCache<ConversionResult>>();
```
kshyju, line 61: We create a new scope when accessing `InstanceServices` property of DefaultFunctionContext.

rvdginste:
This is a very good and useful change in the api. And it looks like this could become available very soon.
What is not clear to me, is whether this kind of change also requires changes on the hosting infrastructure in the Azure cloud (.net 6 / isolated)?
Can I build this branch, try the new api in my code and deploy my updated Azure Function?

src/DotNetWorker.Core/Context/FunctionContextBindingFeatureExtensions.cs
```diff
@@ -0,0 +1,142 @@
...
+        /// <returns>An instance of <see cref="InvocationResult{T}"/>.</returns>
+        /// <exception cref="InvalidOperationException">Throws when the invocation result is not of the requested type.</exception>
+        public static InvocationResult<T> GetInvocationResult<T>(this FunctionContext context)
+        {
+            if (context.GetBindings().InvocationResult is T resultAsT)
+            {
+                return new InvocationResult<T>(context, resultAsT);
+            }
+
+            throw new InvalidOperationException("Invocation result is not of the requested type. Consider using the overload which does not specify the type.");
```
fabiocav, line 82: nit:
I think leaving the message as `"Invocation result is not of the requested type." is sufficient. One thing that might be helpful is also adding the type of the object assigned to `InvocationResult` here

src/DotNetWorker.Core/Definition/BindingMetadata.cs
```diff
@@ -9,12 +9,17 @@ namespace Microsoft.Azure.Functions.Worker
     public abstract class BindingMetadata
     {
         /// <summary>
-        /// The type of the binding. For example, "httpTrigger".
+        /// Gets the name of the binding metadata entry.
+        /// </summary>
+        public abstract string Name { get; }
+
+        /// <summary>
+        /// Gets the type of the binding. For example, "httpTrigger".
```
fabiocav, line 17: 👍🏼

src/DotNetWorker.Core/OutputBindings/OutputBindingData.cs
```diff
@@ -0,0 +1,46 @@
...
+        internal OutputBindingData(FunctionContext functionContext, string name, T? value, string bindingType)
+        {
+            _functionContext = functionContext;
+            _value = value;
+            Name = name;
+            BindingType = bindingType;
+        }
+
+        private readonly FunctionContext _functionContext;
+        private T? _value;
```
fabiocav, line 20: nit: move these above the constructor

fabiocav (approved):
A couple of comments, but those are small things. Would be good to get additional reviews on those once the PR is updated, but I'm flagging as approved as those are not major and you shouldn't block on me to get the changes in, since I'll be out for the next few days.

src/DotNetWorker.Core/Context/FunctionContextBindingFeatureExtensions.cs
kshyju: Updated the exception message to include the type of invocation result in the latest iteration.

src/DotNetWorker.Core/Context/FunctionContextBindingFeatureExtensions.cs
```diff
@@ -0,0 +1,127 @@
...
+    public static class FunctionContextBindingFeatureExtensions
+    {
+        /// <summary>
+        /// Binds an input binding item for the requested <see cref="BindingMetadata"/> instance.
+        /// </summary>
+        /// <param name="context">The function context.</param>
+        /// <param name="bindingMetadata">The BindingMetadata instance for which input data should bound to.</param>
+        /// <returns>An instance of T if binding was successful, else null</returns>
+        public static async ValueTask<T?> BindInputAsync<T>(this FunctionContext context, BindingMetadata bindingMetadata)
+        {
```
brettsam, line 25: Didn't we talk at some point about this returning a wrapper type like `BindingResult<T>`, which would let you completely replace the value in the cache if you wanted to?

I may be misremembering, though :)

brettsam: I see we have similar concepts for `InvocationResult` and `OutputBindingData`...

kshyju: You are remembering it correctly. We did discuss it. In the latest iteration, I wrapped the result in `InputBindingData<T>` type. 
One thing we may consider doing is replacing the new cache with `IFunctionBindingsFeature` (that is what the other 2 items (InvocationResult and OutputBindingData is relying on to update the value). Will follow up offline to discuss this.

PR #830, opened 2022-03-17
Startup hook for extensions

### LINKED ISSUE #798: Create a startup hook for extensions
Extension authors frequently have the need to register additional services and it is often desirable to have those done implicitly and automatically by the extension.

Today, the worker does not expose a hook (similar to startup) to enable extensions to perform service registrations, and this should be exposed.

Creating placeholder item. Will update with the proposed API/experience here.

kshyju:
Fixes #798

-------------
#### Changes this PR is bringing:

We are exposing the below (new) types for extension authors to consume.

1) `WorkerExtensionStartup` abstract class.
````
public abstract class WorkerExtensionStartup
{
    /// <summary>
    /// Configures the function worker application builder option.
    /// </summary>
    /// <param name="applicationBuilder">The IFunctionsWorkerApplicationBuilder instance. </param>
    public abstract void Configure(IFunctionsWorkerApplicationBuilder applicationBuilder);
}
````

2) `WorkerExtensionStartup` attribute.
The assembly in which the implementation of `WorkerExtensionStartup ` is present, must be decorated with this attribute.

Extension authors who wish to participate in the startup process will need to create an implementation of the `WorkerExtensionStartup`  and may use the [IFunctionsWorkerApplicationBuilder](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/src/DotNetWorker.Core/Hosting/IFunctionsWorkerApplicationBuilder.cs) instance passed in to register middlewares or services to the DI container. 

Example:

````
[assembly: WorkerExtensionStartup(typeof(MyHttpExtensionStartup))]
namespace Microsoft.Azure.Functions.Worker.Extensions.Http
{
    public sealed class MyHttpExtensionStartup : WorkerExtensionStartup
    {
        public override void Configure(IFunctionsWorkerApplicationBuilder applicationBuilder)
        {
            applicationBuilder.UseMiddleware<StampHttpHeadersMiddleware>();
            applicationBuilder.Services.AddSingleton<IMyFooService, MyFooService>();
        }      
    }
}
````
We are also adding a source generator which will generate a class with the code needed to call the `Configure` method on each of the extension startup implementations.  During app initialization, we will call this code (present in the generated source code). Sample code generated below.

````
[assembly: WorkerExtensionStartupCodeExecutorInfo(typeof(WorkerExtensionStartupCodeExecutor))]

internal class WorkerExtensionStartupCodeExecutor : WorkerExtensionStartup
{
    public override void Configure(IFunctionsWorkerApplicationBuilder applicationBuilder)
    {
        try
        {
            new Worker.Extensions.Sample.MySampleExtensionStartup().Configure(applicationBuilder);
        }
        catch(Exception ex)
        {
            Console.WriteLine("Error calling Configure on Worker.Extensions.Sample.MySampleExtensionStartup instance."+ex.ToString());
        }
    }
}
````

Currently there will be one try-catch block for each extension in this method. I chose it instead of for-eaching the implementation collection and calling it in a single try-catch to keep the source gen code less (and to reduce the nested `{` in source gen).

### CODE REVIEW

File: src/DotNetWorker.Core/ExtensionStartup/IWorkerExtensionStartup.cs [hunk 1]
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.Functions.Worker.Core
+{
+    /// <summary>
+    /// A type representing the contract for worker extension startup implementation.
+    /// </summary>
+    public interface IWorkerExtensionStartup
```
kshyju, line 11: Extension authors need to implement this interface if they want to participate in the startup.

src/DotNetWorker.Core/ExtensionStartup/WorkerExtensionStartupAttribute.cs
```diff
@@ -0,0 +1,44 @@
...
+using System;
+
+namespace Microsoft.Azure.Functions.Worker.Core
+{
+    /// <summary>
+    /// An attribute to inform that this assembly contains a worker extension startup type.
+    /// </summary>
+
+    [AttributeUsage(AttributeTargets.Assembly, Inherited = false, AllowMultiple = true)]
+    public class WorkerExtensionStartupAttribute : Attribute
```
kshyju, line 13: Extension authors should use this attribute to decorate the assembly in which they have their implementation of `IWorkerExtensionStartup`.

sdk/Sdk.Generators/Sdk.Generators.csproj
```diff
@@ -0,0 +1,30 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <LangVersion>latest</LangVersion>
+    <PackageId>Microsoft.Azure.Functions.Worker.Sdk.Generators</PackageId>
+    <Description>This package provides source generation for azure function worker.</Description>
```
satvu, line 7: Nit: Can we make this description more like [this one](https://github.com/Azure/azure-functions-dotnet-worker/blob/f6c139f1990b3b94fb61408aa09090bd668c927b/sdk/Sdk.Analyzers/Sdk.Analyzers.csproj#L10)? It could be something like "This package provides source generation for the Azure Functions .NET Worker."

satvu (approved):

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
```diff
@@ -0,0 +1,160 @@
...
+                    }
+                }
+            }
+
+            return typeNames;
+        }
+
+        private void WriteStartupDataProviderAttribute(IndentedTextWriter textWriter)
+        {
+            textWriter.WriteLine("[AttributeUsage(AttributeTargets.Class)]");
```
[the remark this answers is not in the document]
kshyju: Switched to this approach.

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
```diff
@@ -0,0 +1,160 @@
...
+            textWriter.WriteLine("public class WorkerExtensionStartupRunnerAttribute : Attribute");
+            textWriter.WriteLine("{");
+            textWriter.WriteLine("}");
+        }
+
+        private static void WriteStartupRunnerClass(IndentedTextWriter textWriter, IEnumerable<string> typeNames)
+        {
+            textWriter.WriteLine("[WorkerExtensionStartupRunner]");
+            textWriter.WriteLine("internal class WorkerExtensionStartupCodeExecutor");
```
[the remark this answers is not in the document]
kshyju: Switched to the suggested approach.

src/DotNetWorker.Core/ExtensionStartup/ExtensionStartupRunnner.cs
```diff
@@ -0,0 +1,42 @@
...
+                                        .FirstOrDefault(type => type.GetCustomAttributes()
+                                                                    .Any(attr => attr.GetType().Name == StartupAttributeName));
+
+            // Our source generator will not create the file when no extension startup hooks are found.
+            if (startupExecutorType == null)
+            {
+                return;
+            }
+
+            var method = startupExecutorType.GetMethod("RunStartupForExtensions");
```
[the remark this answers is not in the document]
kshyju: Using the interface now as suggested and removed this reflection code.

src/DotNetWorker.Core/Hosting/ServiceCollectionExtensions.cs
```diff
@@ -78,7 +79,11 @@ public static IFunctionsWorkerApplicationBuilder AddFunctionsWorkerCore(this ISe
                 services.Configure(configure);
             }
 
-            return new FunctionsWorkerApplicationBuilder(services);
+            var builder = new FunctionsWorkerApplicationBuilder(services);
+            // Execute startup code from worker extensions if apresent.
```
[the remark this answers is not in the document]
kshyju: Fixed typo in latest iteration.

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
```diff
@@ -0,0 +1,168 @@
...
+    /// Also adds the assembly attribute "WorkerExtensionStartupCodeExecutorInfo"
+    /// and pass the information(the type) about the class we generated.
+    /// We are implementing the IWorkerExtensionStartup interface on the generated class.
+    /// (This is the same interface exetension authors will implement for their extension specific startup code)
+    /// We need the same signature as the extension's implementation as our class is an uber class which internally
+    /// calls each of the extension's implementations.
+    /// </summary>
+
+    // Sample code generated (with one extensions particiapating in startup hook)
+    // There will be one try-catch block for each extension particiapating in startup hook.
```
brettsam, line 25: fix spelling of "participating" (2x here)

src/DotNetWorker.Core/ExtensionStartup/IWorkerExtensionStartup.cs
brettsam: I wonder if we should make this an abstract class `WorkerExtensionStartup` instead of an interface. We bumped into this with `IWebJobsStartup` when we later wanted to add additional methods to it and wound up with `IWebJobsStartup2` and a bunch of plumbing code to handle that.

An abstract class makes this easier as you can add methods with default implementations in the future.

src/DotNetWorker.Core/ExtensionStartup/WorkerExtensionStartupAttribute.cs
```diff
@@ -0,0 +1,44 @@
...
+            }
+
+            var interfaceType = typeof(IWorkerExtensionStartup);
+            if (!interfaceType.IsAssignableFrom(startupType))
+            {
+                throw new InvalidOperationException(
+                    $"startupType value must be a type which implements {interfaceType.FullName}.");
+            }
+
+            StartupType = startupType;
```
brettsam, line 36: We should also check here that it has a public parameterless constructor.

src/DotNetWorker.Core/ExtensionStartup/ExtensionStartupRunnner.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        /// <summary>
+        /// Run extension startup execution code.
+        /// </summary>
+        /// <param name="builder">The <see cref="IFunctionsWorkerApplicationBuilder"/> instance.</param>
+
+        /// Our source generator creates a class(WorkerExtensionStartupCodeExecutor)
+        /// which internally calls the "Configure" method on each of the participating 
+        /// extensions. Here we are calling the uber "Configure" method on the generated class.
+        internal static void RunExtensionStartupCode(IFunctionsWorkerApplicationBuilder builder)
```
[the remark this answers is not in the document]
kshyju: The first iteration had a little more code to get the method. Current version is only couple of lines. Moved to ServiceCollectionExtension as a private method.

src/DotNetWorker.Core/ExtensionStartup/WorkerExtensionStartupAttribute.cs
kshyju: Good call. Added a check for the existence of public parameter less constructor.

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
```diff
@@ -0,0 +1,168 @@
...
+                textWriter.WriteLine("try");
+                textWriter.WriteLine("{");
+                textWriter.Indent++;
+                textWriter.WriteLine($"new {fullyQualifiedTpeName}().Configure(applicationBuilder);");
+                textWriter.Indent--;
+                textWriter.WriteLine("}");
+                textWriter.WriteLine("catch (Exception ex)");
+                textWriter.WriteLine("{");
+                textWriter.Indent++;
+                textWriter.WriteLine($"Console.WriteLine(\"Error calling Configure on {fullyQualifiedTpeName} instance.\"+ex.ToString());");
```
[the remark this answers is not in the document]
kshyju: I created #850 to implement the proper startup logger.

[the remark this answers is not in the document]
kshyju: My thought process for wrapping each call in its own `try-catch` was to prevent one bad extension update/upgrade crashing the entire app. This current approach is more of a degraded experience (the bad extensions startup code won't be executed, but rest of the extension's startup will still work) of the app instead of a crashed app. 

Will discuss this item in the next worker triage/design sync.
Let's

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
kshyju: Fixed type.

src/DotNetWorker.Core/ExtensionStartup/IWorkerExtensionStartup.cs
kshyju: Yea, I always wondered about the `IWebJobsStartup2`. I switched this to an abstract base class now.

src/DotNetWorker.Core/ExtensionStartup/WorkerExtensionStartupAttribute.cs
kshyju: I moved both checks to source gen. If the type does not have the constructor / has not implemented our abstract class, it will generate code to throw an exception with the proper message. Runtime code needs to just run it.

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
```diff
@@ -0,0 +1,212 @@
...
+                else
+                {
+                    textWriter.WriteLine($"new {type.TypeName}().Configure(applicationBuilder);");
+                }
+                textWriter.Indent--;
+                textWriter.WriteLine("}");
+                textWriter.WriteLine("catch (Exception ex)");
+                textWriter.WriteLine("{");
+                textWriter.Indent++;
+                textWriter.WriteLine($"Console.WriteLine(\"Error calling Configure on {type.TypeName} instance.\"+ex.ToString());");
```
kshyju, line 197: #850 will address improved error handling/logging.

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
```diff
@@ -0,0 +1,212 @@
...
+                else
+                {
+                    textWriter.WriteLine($"new {type.TypeName}().Configure(applicationBuilder);");
+                }
+                textWriter.Indent--;
+                textWriter.WriteLine("}");
+                textWriter.WriteLine("catch (Exception ex)");
+                textWriter.WriteLine("{");
+                textWriter.Indent++;
+                textWriter.WriteLine($"Console.Error.WriteLine(\"Error calling Configure on {type.TypeName} instance.\"+ex.ToString());");
```
kshyju, line 197: Since the error message contains "error", the host code will [log this as an error](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs#L123).

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
```diff
@@ -0,0 +1,212 @@
...
+    //    }
+    //}
+
+    [Generator]
+    public class ExtensionStartupRunnerGenerator : ISourceGenerator
+    {
+        /// <summary>
+        /// The attribute which extension authors will apply on an assembly which contains their startup type.
+        /// </summary>
+        private string attributeTypeName = "WorkerExtensionStartupAttribute";
```
fabiocav, line 50: nit: make this a constant

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
```diff
@@ -0,0 +1,212 @@
...
+
+        /// <summary>
+        /// Fully qualified name of the above "WorkerExtensionStartupAttribute" attribute.
+        /// </summary>
+        private string attributeTypeFullName = "Microsoft.Azure.Functions.Worker.Core.WorkerExtensionStartupAttribute";
```
fabiocav, line 55: nit: make this a constant

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
```diff
@@ -0,0 +1,212 @@
...
+        private IEnumerable<(string TypeName, string? Error)> GetExtensionStartupTypes(GeneratorExecutionContext context)
+        {
+            List<(string TypeName, string? Error)>? typeInfoList = null;
+
+            // Extension authors should decorate their assembly with "WorkerExtensionStartup" attribute
+            // if they want to participate in startup.
+            foreach (var assembly in context.Compilation.SourceModule.ReferencedAssemblySymbols)
+            {
+                var extensionStartupAttribute = assembly.GetAttributes()
+                                                        .FirstOrDefault(a => a.AttributeClass?.Name == attributeTypeName &&
```
fabiocav, line 108: nit: use string.Equals with the appropriate comparison in both string checks here

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
```diff
@@ -0,0 +1,212 @@
...
+                var constructorExist = namedTypeSymbol.InstanceConstructors
+                                                  .Any(c => c.Parameters.Length == 0 &&
+                                                            c.DeclaredAccessibility == Accessibility.Public);
+                if (!constructorExist)
+                {
+                    return $"{typeSymbol.ToDisplayString()} class must have a public parameterless constructor.";
+                }
+
+                // Check the extension startup class implements WorkerExtensionStartup abstract class.
+                if (!(namedTypeSymbol.BaseType!.GetFullName().Equals(startupBaseClassName)))
```
fabiocav, line 148: nit: pass the string comparison type here

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
```diff
@@ -0,0 +1,212 @@
...
+                {
+                    return $"{typeSymbol.ToDisplayString()} must be a type which implements {startupBaseClassName}.";
```
fabiocav, line 150: nit: Change to `$"{typeSymbol.ToDisplayString()} must derive from {startupBaseClassName}."`

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
```diff
@@ -0,0 +1,212 @@
...
+            textWriter.Indent++;
+
+            foreach (var type in types)
+            {
+                textWriter.WriteLine("try");
+                textWriter.WriteLine("{");
+                textWriter.Indent++;
+                if (type.Error != null)
+                {
+                    textWriter.WriteLine($"throw new InvalidOperationException(\"{type.Error}\");");
```
fabiocav, line 186: This will cause a runtime exception, which to an end user who may not be familiar with the type, could be confusing.

Should this become a build issue with information about the extension/dependency that introduced the bad startup?

fabiocav (commented):
Partial. Will complete later in the day but wanted to send now in case you have cycles to update by then

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
kshyju: Changed to const.

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
kshyju: Changed to const.

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
kshyju: Switched to the overload which takes StringComparison and passing `StringComparison.Ordinal` . (No need to check case insensitive check here).

Interestingly, [in my microbenchmarks](https://gist.github.com/kshyju/dc78a1f5ef9e8da40c1b4985463f79e6), the `Equals` overload without comparison type parameter does better by a tiny fraction.

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
kshyju: Done. Using `StringComparison.Ordinal`. We want to do a case sensitive comparison.

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
kshyju: Changed the language as suggested.

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
kshyju: True. A build time diagnostic error makes sense. In the new iteration, I made changes to report a diagnostic error when startup implementation is invalid (missing parameter less constructor, incorrect base class).

test/Sdk.Generator.Tests/Sdk.Generator.Tests.csproj
```diff
@@ -0,0 +1,39 @@
...
+    <PackageReference Include="coverlet.collector" Version="3.1.2">
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\extensions\Worker.Extensions.Http\src\Worker.Extensions.Http.csproj" />
+    <ProjectReference Include="..\..\sdk\Sdk.Generators\Sdk.Generators.csproj" />
+    <ProjectReference Include="..\Worker.Extensions.Sample-IncorrectImplementation\Worker.Extensions.Sample-IncorrectImplementation.csproj" />
```
kshyju, line 35: FYI:  The generator scans the referenced assemblies to find the extension startup type implementations. To test passing & failing cases, I had to create 2 sample extension projects under "test", one which has valid implementation of startup type and another one which has invalid implementation of startup type).  I played around with creating these projects in-memory within the test context but could not get it working yet. I will continue that effort and will replace this current approach (additional projects) with that when that works.

kshyju:
@fabiocav / @brettsam This is ready for another round of review :)

brettsam (approved):

src/DotNetWorker.Core/DotNetWorker.Core.csproj
```diff
@@ -26,12 +26,12 @@
...
-        <PackageReference Include="System.Collections.Immutable" Version="6.0.0" />
+        <PackageReference Include="System.Collections.Immutable" Version="5.0.0" />
     </ItemGroup>
 
     <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
-        <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="6.0.0" />
+        <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="5.0.0" />
         <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
-        <PackageReference Include="Microsoft.Extensions.Options" Version="6.0.0" />
+        <PackageReference Include="Microsoft.Extensions.Options" Version="5.0.0" />
```
liliankasem, line 35: Just curious why the downgrade in version?

kshyju: These packages were added in previous PR as part of adding .NET standard support. But while testing changes for this PR, I ran into assembly version conflict issue where the CLR loads the 5.X version of an assembly, and the runtime expects the 6.X version of the assembly.  I synced with @fabiocav to discuss this issue and we decided to downgrade these packages to 5X version.

Here is one specific exception message I ran into:
> "Could not load file or assembly 'Microsoft.Extensions.DependencyInjection.Abstractions, Version=6.0.0.0

and I could see that 5.X version is loaded.

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
```diff
@@ -0,0 +1,221 @@
...
+            if (typeSymbol is INamedTypeSymbol namedTypeSymbol)
+            {
+                // Check public parameterless constructor exist for the type.
+                var constructorExist = namedTypeSymbol.InstanceConstructors
+                                                  .Any(c => c.Parameters.Length == 0 &&
+                                                            c.DeclaredAccessibility == Accessibility.Public);
+                if (!constructorExist)
+                {
+                    context.ReportDiagnostic(Diagnostic.Create(DiagnosticDescriptors.ConstructorMissing, Location.None,
+                        typeSymbol.ToDisplayString()));
```
fabiocav, line 155: 👍🏼

fabiocav (approved):
:shipit:

PR #838, opened 2022-03-24
Handle = character in cookie value and added tests

### LINKED ISSUE #837: HttpRequestData cookie value is not valid
I was trying to read cookies.

I see cookies available in HttpRequestData accessed via context.GetHttpRequestData();
However, when I see HttRequestData.Headers.GetValues("Cookie") I can see that cookie value is correct.

There is also Cookies property (i.e. HttpRequestData.Cookies), when I inspect the value which I need, I see that it is slightly different. 

e.g. 
HttpRequestData.Headers.GetValues("Cookies") gives me 
"p5LaU9a9W3ewieRK7pZtr7XwY0h+yRoTkCvfpQ1R9Jxwkw0Ov8s8mTZZZP/78G5rl/ansmmGo3uYVmTSGAPXGroywUinAAHolBF2G75a2Qo="

HttpRequestData.Cookies[0] gives me "p5LaU9a9W3ewieRK7pZtr7XwY0h+yRoTkCvfpQ1R9Jxwkw0Ov8s8mTZZZP/78G5rl/ansmmGo3uYVmTSGAPXGroywUinAAHolBF2G75a2Qo"

HttpRequestData.Cookies[0] has last character different. 

I tried several times to validate different values, but same results. Any help is appreciated !

Thanks!

kshyju:
Resolves #837

When the cookie value has a "=" character, our logic was returning incorrect cookie value because we relied on `Split("=")` to get name and value of cookie. When the cookie item string is like "token=xyz1=", the Split method returns 3 items and we take the second item as the value. This means we only get "xyz1" and will miss the "=" character. To fix this , I am using the overload of Split which takes the "count" parameter. 

While testing this, I also found another issue. When cookie string is empty, we were returning one item in the Cookies property with name and value properties set to empty string. This is inconsistent with what ASP.NET core behaves (they return 0 items).

Added tests for this class.

### DIFF
--- a/src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
+++ b/src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Azure.Functions.Worker
 {
     internal class GrpcHttpRequestData : HttpRequestData, IDisposable
 #if NET5_0_OR_GREATER
-        ,IAsyncDisposable
+        , IAsyncDisposable
 #endif
     {
         private readonly RpcHttp _httpData;
@@ -169,9 +169,9 @@ namespace Microsoft.Azure.Functions.Worker
             {
                 // count:2 tells the split method to stop splitting after
                 // it finds the first instance of separator character.
-                var splitArray = separateCookies[c].Split(separator, count:2, StringSplitOptions.RemoveEmptyEntries);
+                var splitArray = separateCookies[c].Split(separator, count: 2, StringSplitOptions.RemoveEmptyEntries);
                 var name = splitArray[0].Trim();
-                var value = splitArray.Length>1 ? splitArray[1]: string.Empty;
+                var value = splitArray.Length > 1 ? splitArray[1] : string.Empty;
                 httpCookiesList.Add(new HttpCookie(name, value));
             }
 
--- a/src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
+++ b/src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
@@ -38,7 +38,7 @@ namespace Microsoft.Azure.Functions.Worker
 
                 var cookieString = Headers.FirstOrDefault(item => item.Key.Equals("Cookie", StringComparison.OrdinalIgnoreCase)).Value;
 
-                if (cookieString != null && cookieString.Any())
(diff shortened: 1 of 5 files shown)


src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
```diff
@@ -167,10 +167,11 @@ private IReadOnlyCollection<IHttpCookie> ToHttpCookies(string cookieString)
 
             for (int c = 0; c < separateCookies.Length; c++)
             {
-
-                var splitArray = separateCookies[c].Split(separator, StringSplitOptions.RemoveEmptyEntries);
+                // count:2 tells the split method to stop splitting after
```
kshyju, line 170: From the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.string.split?view=net-5.0#system-string-split(system-string-system-int32-system-stringsplitoptions)),

> If the string has already been split count - 1 times, but the end of the string has not been reached, then the last string in the returned array will contain this instance's remaining trailing substring, untouched.

Using of this overload will handle string like "token=a12=" where token is the cookie name and a12= is the cookie value.

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
```diff
@@ -38,7 +38,7 @@ public GrpcHttpRequestData(RpcHttp httpData, FunctionContext functionContext)
 
                 var cookieString = Headers.FirstOrDefault(item => item.Key.Equals("Cookie", StringComparison.OrdinalIgnoreCase)).Value;
 
-                if (cookieString != null && cookieString.Any())
+                if (cookieString != null && cookieString.Any() && !string.IsNullOrWhiteSpace(cookieString.First()))
```
kshyju, line 41: Currently if the cookie string is empty, we are still returning one item with empty name and value properties. This change fixes that.

test/DotNetWorkerTests/GrpcHttpRequestDataTests.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            var headerDict = new Dictionary<string, string>
+            {
+                { "accept-encoding","gzip, deflate, br" },
+                { "cookie","" }
+            };
+
+            var request = CreateGrpcHttpRequestData(headerDict);
+
+            Assert.Equal(0, request.Cookies.Count);
```
kshyju, line 68: This is consistent with how asp.net core behaves.

satvu (approved):

PR #841, opened 2022-03-25
Throw build error when in-proc SDK is referenced in isolated function apps

### LINKED ISSUE #824: [Feature Request] Error handling for SDK vs nonSDK
The request is to provide compile error message when an isolated app reference in-process SDK.

kshyju:
Fixes #824 

Throw build error when [in-proc SDK](https://www.nuget.org/packages/Microsoft.NET.Sdk.Functions/) is referenced in isolated function apps. We check the [_IsFunctionsSdkBuild ](https://github.com/Azure/azure-functions-vs-build-sdk/blob/32ebc25b282fe08ebb283178150450755f2d1044/src/Microsoft.NET.Sdk.Functions.MSBuild/Targets/Microsoft.NET.Sdk.Functions.props#L15)property to determine this package is referenced or not.

Tested with a locally published version of the `Microsoft.Azure.Functions.Worker.Sdk` package.

### DIFF
--- a/sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.props
+++ b/sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.props
@@ -13,7 +13,6 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
   <PropertyGroup>
     <FunctionsExecutionModel>isolated</FunctionsExecutionModel>
     <_FunctionsSkipCleanOutput>true</_FunctionsSkipCleanOutput>
-    <_IsFunctionsSdkBuild>true</_IsFunctionsSdkBuild>
   </PropertyGroup>
 
   <!--Enable Azure Functions project capability to enable tools-->
--- a/sdk/Sdk/Sdk.csproj
+++ b/sdk/Sdk/Sdk.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   
   <PropertyGroup>
-    <MinorProductVersion>3</MinorProductVersion>
+    <MinorProductVersion>4</MinorProductVersion>
     <PatchProductVersion>0</PatchProductVersion>
     <TargetFrameworks>netstandard2.0;net472</TargetFrameworks>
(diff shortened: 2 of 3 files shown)


sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.props
```diff
@@ -13,7 +13,6 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
   <PropertyGroup>
     <FunctionsExecutionModel>isolated</FunctionsExecutionModel>
     <_FunctionsSkipCleanOutput>true</_FunctionsSkipCleanOutput>
-    <_IsFunctionsSdkBuild>true</_IsFunctionsSdkBuild>
```
kshyju, line 16: I am guessing this hardcoded value here was a copy-paste mistake. I do not see this being used anywhere else (except my new code in this PR).

satvu: cc: @brettsam @fabiocav 

I agree with Shyju. Does this sound right?

liliankasem: @kshyju might be worth pining Fabio via Teams to get an answer

liliankasem (approved):

sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.props
fabiocav: It should be OK to remove this, as we add a reference to the main SDK from here: https://github.com/Azure/azure-functions-dotnet-worker/blob/e7f8f127843f39d87df16a6611bccb54aff9a5a7/sdk/Sdk/ExtensionsCsprojGenerator.cs#L74, which is the project that cares about that.

We just need to make sure we validate the flow.

kshyju: Validated the flow of using a few worker extensions and everything looks good.

kshyju:
This **_check-enforcer_** thingy is really testing my patience! 😩

PR #852, opened 2022-04-11
Adding DateTimeConverter which will handle binding of DateTime/DateTimeOffset type parameters

### LINKED ISSUE #851: .NET Isolated DateTime Route parameter issue
Hi and looks like it is same issue as this one #521 but for DateTime.

```
[Function("Rates")]
public async Task<string> Run(
    [HttpTrigger(AuthorizationLevel.Function, "get", Route = "rates/{onDate:datetime}")] HttpRequestData req,
    DateTime onDate,
    FunctionContext executionContext)
{
  ...
}
```

The error that I'm getting

> Exception: Microsoft.Azure.Functions.Worker.Diagnostics.Exceptions.FunctionInputConverterException: Error converting 1 input parameters for Function 'Rates': Cannot convert input parameter 'onDate' to type 'System.DateTime' from type 'System.String'.

Thanks

kshyju:
Fixes #851 

Adding a new `IConverter` implementation to handle binding the below types.

1. `DateTime`
2. `DateTime?`
3. `DateTimeOffset`
4. `DateTimeOffset?`

### DIFF
--- a/src/DotNetWorker.Core/Converters/DateTimeConverter.cs
+++ b/src/DotNetWorker.Core/Converters/DateTimeConverter.cs
@@ -7,14 +7,34 @@ using System.Threading.Tasks;
 namespace Microsoft.Azure.Functions.Worker.Converters
 {
     /// <summary>
-    /// Converter to bind DateTime/DateTime? type parameters.
+    /// Converter to bind DateTime/DateTime?/DateOnly/TimeOnly type parameters.
     /// </summary>
     internal sealed class DateTimeConverter : IInputConverter
     {
         public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
         {
-            if (context.TargetType == typeof(DateTime) || context.TargetType == typeof(DateTime?))
+            if (
+                context.TargetType == typeof(DateTime) || context.TargetType == typeof(DateTime?)
+#if NET6_0
+                || context.TargetType == typeof(DateOnly) || context.TargetType == typeof(DateOnly?)
+                || context.TargetType == typeof(TimeOnly) || context.TargetType == typeof(TimeOnly?)
+#endif
+                )
             {
+#if NET6_0
+                if (context.Source is string source
+                    && (context.TargetType == typeof(DateOnly) || context.TargetType == typeof(DateOnly?))
+                    && DateOnly.TryParse(source, out DateOnly parsedDateOnly))
+                {
+                    return new ValueTask<ConversionResult>(ConversionResult.Success(parsedDateOnly));
+                }
+                if (context.Source is string sourceStr
+                    && (context.TargetType == typeof(TimeOnly) || context.TargetType == typeof(TimeOnly?))
+                    && TimeOnly.TryParse(sourceStr, out TimeOnly parsedTimeOnly))
+                {
+                    return new ValueTask<ConversionResult>(ConversionResult.Success(parsedTimeOnly));
+                }
+#endif
                 if (context.Source is string sourceString && DateTime.TryParse(sourceString, out DateTime parsedDate))
                 {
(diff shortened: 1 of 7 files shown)


src/DotNetWorker.Core/Converters/DateTimeConverter.cs
```diff
@@ -0,0 +1,47 @@
...
+{
+    /// <summary>
+    /// Converter to bind DateTime/DateTime?/DateOnly/TimeOnly type parameters.
+    /// </summary>
+    internal sealed class DateTimeConverter : IInputConverter
+    {
+        public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
+        {
+            if (
+                context.TargetType == typeof(DateTime) || context.TargetType == typeof(DateTime?)
```
SeanFeldman, line 17: Should `DateTimeOffset` be handled as well?

kshyju: Good catch. Yes, we should support `DateTimeOffset` as well. I pushed a new iteration with that support.

src/DotNetWorker.Core/Converters/DateTimeConverter.cs
```diff
@@ -0,0 +1,43 @@
...
+    internal sealed class DateTimeConverter : IInputConverter
+    {
+        public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
+        {
+            if (!IsValidTargetType(context) || context.Source is not string source)
+            {
+                return new ValueTask<ConversionResult>(ConversionResult.Unhandled());
+            }
+
+            if ((context.TargetType == typeof(DateTimeOffset) || context.TargetType == typeof(DateTimeOffset?))
```
brettsam, line 21: In what scenario would we pass `null` to a `DateTimeOffset?` or `DateTime?`? Is there any?

kshyju: Let's say customer has a function definition like below.

````
public static HttpResponseData Run(
    [HttpTrigger(AuthorizationLevel.Anonymous, "get")] HttpRequestData req,
    DateTime startDate,
    DateTime? endDate)
````

where endDate param is nullable type / optional.  The below 2 requests will work for this function definition. 

````
api/Function4?startdate=2021/10/10&enddate=2021/12/12
api/Function4?startdate=2021/12/12
````
For the second request where `endDate` is not part of the request's input data/ trigger metadata, so, we do not have a source data to use (ie: `source` is null). So, the converter will return **_Unhandled_** result and the default value of the [parameterValues array (of type `object[]`) ](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/src/DotNetWorker.Core/Context/Features/DefaultModelBindingFeature.cs#L15)entry will be used, which is `null`. We will end up passing [null as that argument's value when invoking the function](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/src/DotNetWorker.Core/Invocation/MethodInvokerWithReturnValue.cs#L22).

kshyju: We do similar thing for [Guid converter where `Guid?` is supported](url).

brettsam (approved):

kshyju:
[Microsoft.Azure.Functions.Worker 1.8.0-preview3](https://www.nuget.org/packages/Microsoft.Azure.Functions.Worker/1.8.0-preview3) includes this change.

viktorasmickunas:
I have a question: how come this converter converts timestamps in strings, when the  TargetType isn't even `DateTime` or `DateTimeOffset`?

See issue: [https://github.com/Azure/azure-functions-dotnet-worker/issues/2731](url)

PR #882, opened 2022-05-11
Updating gRPC references for .NET Standard 2.0 targets

fabiocav:
Updating gRPC references to work with NetFX

### Pull request checklist

### DIFF
--- a/src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
+++ b/src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
@@ -22,8 +22,6 @@
     <ItemGroup>
         <PackageReference Include="Azure.Core" Version="1.10.0" />
         <PackageReference Include="Google.Protobuf" Version="3.15.8" />
-        <PackageReference Include="Grpc.Net.Client" Version="2.37.0" />
-        <PackageReference Include="Grpc.Net.ClientFactory" Version="2.37.0" />
         <PackageReference Include="Grpc.Tools" Version="2.37.0">
             <PrivateAssets>all</PrivateAssets>
             <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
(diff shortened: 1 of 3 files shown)


src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
```diff
@@ -32,6 +30,15 @@
         <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="5.0.0" />
     </ItemGroup>
 
+    <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.0'">
```
daviburg, line 33: Supported platforms and dependencies appear the same on the package info at https://www.nuget.org/packages/Grpc.Core

Why is the lower version needed?

fabiocav: Are you talking about the supported platforms for Grpc.Net.Client? If so, support for the new gRPC clients for .NET Framework is still limited, in early stages and doesn't provide some of the functionality we need here (e.g., communication over non-secure channels/without TLS).

You can find more information here: https://docs.microsoft.com/en-us/aspnet/core/grpc/netstandard?view=aspnetcore-6.0#net-framework

kshyju (approved):

src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
daviburg: Ok. I see in the GRPC extensions that you an interacting with the channel differently depending if the target is .NET Core 5+ or not. Here you are using a newer version for .NET Standard 2+. But .NET standard 2.0 is supported for all of .NET Core 2+, .NET Framework 4.6.1+. So which is the non-standard 2 target requiring an older version?
Or is it just that the newer/newest version is only required for .NET framework to somewhat work, but you didn't want to update the dependency for .NET Core while there isn't an explicit need there?

Probably an XML comment block in this project file to explain why we pick different versions will be useful to future maintainers of this code.

fabiocav: I see what is causing the confusion :)

Even though what is being referenced for .NET 5+ has a lower version number (2.37.0 vs 2.45.0), those are newer packages (the new model), and different packages (Grpc.Net.Client/Grpc.Net.ClientFactory). Grpc.Core is the legacy package, but it does have a higher version number.

daviburg: oooooh, wow. Yeah, that makes sense now. One of these 'gotcha' includes. I do reiterate my suggestion to add an XML comment element explaining the subtility.

fabiocav: Added a comment as suggested

daviburg (approved):

PR #892, opened 2022-05-13
Updating DefaultModelBinding feature to checking cache entries before using conversion feature

kshyju:
The default model binding feature is not checking the binding cache before engaging the conversion feature to generate the parameter values. This means, if a middleware is setting a different value for an input binding item, that value is not respected for the parameters.  I some have missed this change in the original PR.


fabiocav (approved):
Very small nit comment

PR #896, opened 2022-05-13
Use current thread for handling invocation result

cgillum:
This PR accomplishes two things:

1. Fixes a compatibility problem with Durable Functions
2. Minor performance improvement for all function invocations in the .NET worker

The fix is simple, use the invoker thread to return the function return value instead of doing so on a worker pool thread. It's probably easier to just look at the code change.

One of the rules for orchestrator functions is that they must never do an `await` that would cause the callback to run on a background thread. In other words, all logic MUST run on a single thread. This includes any code that lives in the middleware pipeline between the Durable middleware and the actual function execution.

When debugging the Durable code, I observed behavior that was consistent with an "illegal await" in the function execution pipeline. I debugged the issue with @kshyju, I realized that the problem was in the DefaultFunctionInvoker. Specifically, it's using a `.ContinueWith(t => t.Result)` to convert a `Task<T>` into a `Task<object>`. It's a little strange to do it this way but there's an unfortunate side effect which is that the `t => t.Result` callback gets queued up on a worker pool thread by default. This is bad because it violates the Durable rules and also because of the unnecessary performance overhead of thread switching for something as simple as casting from one type to another.

My fix forces the callback to run on the main invoker thread, rather than scheduling it on a worker pool thread. I verified that this unblocks the Durable Functions integration work.

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/DotNetWorker.Core/Invocation/DefaultFunctionInvoker.cs
+++ b/src/DotNetWorker.Core/Invocation/DefaultFunctionInvoker.cs
@@ -29,7 +29,7 @@ namespace Microsoft.Azure.Functions.Worker.Invocation
         public Task<object> InvokeAsync(object instance, object[] arguments)
         {
             return _methodInvoker.InvokeAsync((TInstance)instance, arguments)
-                .ContinueWith<object>(t => t.Result);
+                .ContinueWith<object>(t => t.Result, TaskContinuationOptions.ExecuteSynchronously);
         }
     }
 }


fabiocav:
Change looks good, but can you please rebase on mean to clean up the commits?

cgillum:
Won't "Squash and merge" clean up all the commits for me? I can do the rebase, but it's just a lot more work.

fabiocav (approved):
Yeah, this should have been there from the beginning. 

About this:

>My fix forces the callback to run on the main invoker thread, rather than scheduling it on a worker pool thread. I verified that this unblocks the Durable Functions integration work.

This might be what you meant by the above, but for clarity, to make sure that was the intention, and it would meet your requirements, this will not cause the continuation to execute on the main invoker thread, but in the same thread the original task completed. I'm assuming that meets your requirements based on the fact that you've validated, but if the `InvokeAsync` call actually goes async, the continuation would execute on the thread pool thread used for the previous task completion.

If that is the intention, might be good to update the commit description, but either way, it's a good change.

cgillum:
You’re right. My comment is true for orchestration triggers but not necessarily for other trigger types where the function can freely go async.

PR #918, opened 2022-06-07
Handling WorkerTerminate signal for graceful termination

### LINKED ISSUE #192: Signal to .net worker shutdown is about to occur
Need to improve the shutdown scenario.

Currently, it's a proces.kill: https://github.com/Azure/azure-functions-host/blob/69e70a87bad28453bae43715c1e7a5ee8893cf37/src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs#L204

surgupta-msft:

resolves #192 
[Design doc](https://eng.ms/docs/cloud-ai-platform/devdiv/serverless-paas-balam/serverless-paas-benbyrd/app-service-web-apps/antares-team-documents/functionteamdocs/design/host/gracefulshutdownofworker)

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/DotNetWorker.Grpc/GrpcWorker.cs
+++ b/src/DotNetWorker.Grpc/GrpcWorker.cs
@@ -51,8 +51,8 @@ namespace Microsoft.Azure.Functions.Worker
 
             _outputReader = outputChannel.Channel.Reader;
             _outputWriter = outputChannel.Channel.Writer;
-            _hostApplicationLifetime = hostApplicationLifetime;
 
+            _hostApplicationLifetime = hostApplicationLifetime ?? throw new ArgumentNullException(nameof(hostApplicationLifetime));
             _application = application ?? throw new ArgumentNullException(nameof(application));
             _rpcClient = rpcClient ?? throw new ArgumentNullException(nameof(rpcClient));
             _invocationFeaturesFactory = invocationFeaturesFactory ?? throw new ArgumentNullException(nameof(invocationFeaturesFactory));
@@ -251,14 +251,7 @@ namespace Microsoft.Azure.Functions.Worker
 
         internal void WorkerTerminateRequestHandler(WorkerTerminate request)
         {
-            try
-            {
-                _hostApplicationLifetime.StopApplication();
-            }
-            catch (Exception ex)
-            {
-                throw new InvalidOperationException("Error during terminating the worker process", ex);
-            }
+            _hostApplicationLifetime.StopApplication();
         }
 
         internal static FunctionLoadResponse FunctionLoadRequestHandler(FunctionLoadRequest request, IFunctionsApplication application, IMethodInfoLocator methodInfoLocator)
--- a/src/DotNetWorker.Grpc/GrpcWorker.cs
+++ b/src/DotNetWorker.Grpc/GrpcWorker.cs
@@ -247,19 +247,16 @@ namespace Microsoft.Azure.Functions.Worker
 
         internal static void WorkerTerminateRequestHandler(WorkerTerminate request)
         {
-            // ToDo:
-                // Add logs about receiving worker terminate message from Host with GracePeriod
-    
             // Terminate the worker process
             try
             {
-                WorkerInformation workerInformation = WorkerInformation.Instance;
+                var workerInformation = WorkerInformation.Instance;
                 Process workerProcess = Process.GetProcessById(workerInformation.ProcessId);
                 workerProcess.Kill();
             }
-            catch (Exception)
+            catch (Exception ex)
             {
-                // Error in killing process
+                throw new InvalidOperationException("Error during terminating the worker process", ex);
             }
         }
 
--- a/src/DotNetWorker.Grpc/GrpcWorker.cs
+++ b/src/DotNetWorker.Grpc/GrpcWorker.cs
@@ -44,6 +44,7 @@ namespace Microsoft.Azure.Functions.Worker
             IOutputBindingsInfoProvider outputBindingsInfoProvider, IMethodInfoLocator methodInfoLocator, 
             IOptions<GrpcWorkerStartupOptions> startupOptions, IOptions<WorkerOptions> workerOptions,
             IInputConversionFeatureProvider inputConversionFeatureProvider,
+            IFunctionMetadataProvider functionMetadataProvider,
             IHostApplicationLifetime hostApplicationLifetime)
         {
             if (outputChannel == null)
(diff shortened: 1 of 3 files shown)


src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -235,10 +240,26 @@ internal static WorkerInitResponse WorkerInitRequestHandler(WorkerInitRequest re
...
             return response;
         }
 
+        internal static void WorkerTerminateRequestHandler(WorkerTerminate request)
+        {
+            // Terminate the worker process
+            try
+            {
+                var workerInformation = WorkerInformation.Instance;
+                var workerProcess = Process.GetProcessById(workerInformation.ProcessId);
```
kshyju, line 254: Isn't it almost same as what Host was doing? (Process.Kill). I was under the impression that we want to use `IHostApplicationLifetime.StopApplication` method. I know this will internally invoke the other registered callbacks (ex: the event hooked up to `IHostApplicationLifetime.ApplicationStopping` cancellation token.

kshyju, line 254: Isn't it almost same as what Host was doing? (Process.Kill). I was under the impression that we want to use `IHostApplicationLifetime.StopApplication` method. I know this will internally invoke the other registered callbacks (ex: the event hooked up to `IHostApplicationLifetime.ApplicationStopping` cancellation token.

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -235,10 +240,26 @@ internal static WorkerInitResponse WorkerInitRequestHandler(WorkerInitRequest re
...
+                workerProcess.Kill();
```
kshyju, line 255: Don't we also need to expose a Cancellation token instance to user's function code so that they can check the `IsCancellationRequested` property and act accordingly. May be it can be a property on the `FunctionContext` or a method parameter they can inject to the function method.

kshyju, line 255: Don't we also need to expose a Cancellation token instance to user's function code so that they can check the `IsCancellationRequested` property and act accordingly. May be it can be a property on the `FunctionContext` or a method parameter they can inject to the function method.

kshyju: `FunctionContext` is a type which carries data specific to a single invocation. Since the app shutdown (as in graceful shutdown) is not invocation specific, we probably do not need expose the cancellation token via FunctionContext. Users can inject `IApplicationLifeTime` to the their class constructor and listen via `ApplicationStopping` property which is a cancellation token, as needed.

I think we may need the Function level(via FunctionContext) cancellation when for the cancellation token feature @liliankasem  is driving.

surgupta-msft: Updated the code based on our discussion offline.

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -127,6 +132,10 @@ private async Task ProcessRequestCoreAsync(StreamingMessage request)
             {
                 responseMessage.WorkerInitResponse = WorkerInitRequestHandler(request.WorkerInitRequest);
             }
+            else if (request.ContentCase == MsgType.WorkerTerminate)
```
liliankasem, line 135: I think this whole block would refactor nicely into a switch statement. And does order matter here? Do we want to put the terminate statement first?

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -235,10 +244,23 @@ internal static WorkerInitResponse WorkerInitRequestHandler(WorkerInitRequest re
...
 
+        internal void WorkerTerminateRequestHandler(WorkerTerminate request)
+        {
+            try
+            {
+                _hostApplicationLifetime.StopApplication();
+            }
+            catch (Exception ex)
+            {
+                throw new InvalidOperationException("Error during terminating the worker process", ex);
```
kshyju, line 260: What code path will catch this exception? I do not think we have a try-catch on the caller of this method.  I think we should just remove the try-catch here.

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -47,6 +51,7 @@ internal class GrpcWorker : IWorker
 
             _outputReader = outputChannel.Channel.Reader;
             _outputWriter = outputChannel.Channel.Writer;
+            _hostApplicationLifetime = hostApplicationLifetime;
```
kshyju, line 54: Consider adding a null check and throw expresssion.

````
hostApplicationLifetime ?? throw new ArgumentNullException(nameof(hostApplicationLifetime));
````

File: src/DotNetWorker.Grpc/GrpcWorker.cs (hunk 3 above)
Comment (kshyju) [reply src/DotNetWorker.Grpc/GrpcWorker.cs:135]: Order does not matter. Agree on the _switch_ to switch.

File: src/DotNetWorker.Grpc/GrpcWorker.cs (hunk 4 above)
Comment (surgupta-msft) [reply src/DotNetWorker.Grpc/GrpcWorker.cs]: done

File: src/DotNetWorker.Grpc/GrpcWorker.cs (hunk 5 above)
Comment (surgupta-msft) [reply src/DotNetWorker.Grpc/GrpcWorker.cs]: ofc!

**Review by kshyju — APPROVED** [review]

File: src/DotNetWorker.Grpc/GrpcWorker.cs [hunk 6]
```diff
@@ -235,10 +244,16 @@ internal static WorkerInitResponse WorkerInitRequestHandler(WorkerInitRequest re
...
             response.Capabilities.Add("UseNullableValueDictionaryForHttp", bool.TrueString);
             response.Capabilities.Add("TypedDataCollection", bool.TrueString);
+            response.Capabilities.Add("HandlesWorkerTerminateMessage", bool.TrueString);
 
             return response;
         }
 
+        internal void WorkerTerminateRequestHandler(WorkerTerminate request)
+        {
+            _hostApplicationLifetime.StopApplication();
```
kshyju, line 254: Q: Do we write a log entry in host before/after making this GRPC call to record this action (something like "sending worker shutdown message" or so)

src/DotNetWorker.Grpc/GrpcWorker.cs
liliankasem: Ah it's by type!! My bad :D

PR #922, opened 2022-06-10
Supporting dependency injection of types which implements IAsyncDisposable

### LINKED ISSUE #833: Registering and consuming a scoped IAsyncDisposable service blocks requests
## Repro steps: 
- Create a new Functions App project from the template:
    - .Net 6 Isolated
    - HttpTrigger
    - For simplicity anonymous authorization
- Add a new service that implements `IAsyncDisposable`: 
   
```csharp
public class MyService : IAsyncDisposable 
{
    public void Beep()
    {
        Console.Beep();
        Console.WriteLine("BEEP");
    }

    public ValueTask DisposeAsync()
    {
        Console.Beep(); 
        Console.Beep();
        Console.WriteLine("Disposing BEEP");
        return ValueTask.CompletedTask;
    }
}
```
- And register it as scoped:
```csharp
    .ConfigureServices(services =>
    {
        services.AddScoped<MyService>();
    })
```
- Update the function to consume it
```csharp
    public class Function1
    {
        private readonly ILogger _logger;
        private readonly MyService myService;

        public Function1(ILoggerFactory loggerFactory, MyService myService)
        {
            _logger = loggerFactory.CreateLogger<Function1>();
            this.myService = myService;
        }

        [Function("Function1")]
        public HttpResponseData Run([HttpTrigger(AuthorizationLevel.Anonymous, "get", "post")] HttpRequestData req)
        {

            response.Headers.Add("Content-Type", "text/plain; charset=utf-8");

            response.WriteString("Welcome to Azure Functions!");

            myService.Beep();

            return response;
        }
    }
```
- Send a request to the endpoint. 

## Expected behaviour
Same as when the service implements `IDisposable`.
The function is executed, beeps and the client receives the welcome message in the response. 
The service is disposed correctly. 

## Actual behaviour
The function will be executed. Beeps as expected.
But you will not receive a response and your request will timeout (depending on your timeout setting) 
The service is not disposed.

## Possible issue and solution
As far as I can tell the issue comes from [this call to `CreateScope()`](https://github.com/Azure/azure-functions-dotnet-worker/blob/9eb2d590b6058788713aa957b3f86f3b09148c47/src/DotNetWorker.Core/Context/DefaultFunctionContext.cs#L45) and could be fixed by replacing it with [`CreateAsyncScope()`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.serviceproviderserviceextensions.createasyncscope?view=dotnet-plat-ext-6.0), which of course would require `DefaultFunctionContext` to implement `IAsyncDisposable`. This may or may not start a chain reaction, I have not checked it further for the time being.

kshyju:
Fixes #833 

Original issue:
When the function code uses a dependency which implements `IAsyncDisposable`, our code throws when [we call Dispose](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/src/DotNetWorker.Grpc/GrpcWorker.cs#L226) method of the `DefaultFunctionContext`, producing an exception like below (_In this example ServiceBusClient is the dependency which implements IAsyncDisposable_).

> 'Azure.Messaging.ServiceBus.ServiceBusClient' type only implements IAsyncDisposable. Use DisposeAsync to dispose the container.
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngineScope.Dispose()
   at Microsoft.Azure.Functions.Worker.DefaultFunctionContext.Dispose() in C:\Dev\OSS\azure-functions-dotnet-worker\src\DotNetWorker.Core\Context\DefaultFunctionContext.cs:line 65

This is because the Dispose method of DefaultFunctionContext[ is calling](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/src/DotNetWorker.Core/Context/DefaultFunctionContext.cs#L65) `Dispose` of the IServiceScope where we should be calling the DisposeAsync when any of the dependencies this scope holds is IAsyncDisposable type.

Fix:
The `CreateScope` method returns an instance of `ServiceProviderEngineScope` which implements both `IAsyncDisposable` and `IDisposable`. So we will call the `DisposeAsync` after casting the IServiceScope instance to `IAsyncDisposable`

Note: The [new ](https://devblogs.microsoft.com/dotnet/announcing-net-6/#microsoft-extensions-dependencyinjection-createasyncscope-apis) CreateAsyncScope method is [available from Net 6 onwards](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.serviceproviderserviceextensions.createasyncscope?view=dotnet-plat-ext-6.0). This means, we need to use version 6.X of the `Microsoft.Extensions.Hosting.Abstractions` package. We decided to not use this method at this time and is continue to use `CreateScope` method.

### DIFF
--- a/src/DotNetWorker.Core/Context/DefaultFunctionContext.cs
+++ b/src/DotNetWorker.Core/Context/DefaultFunctionContext.cs
@@ -3,11 +3,12 @@
 
 using System;
 using System.Collections.Generic;
+using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace Microsoft.Azure.Functions.Worker
 {
-    internal sealed class DefaultFunctionContext : FunctionContext, IDisposable
+    internal sealed class DefaultFunctionContext : FunctionContext, IAsyncDisposable
     {
         private readonly IServiceScopeFactory _serviceScopeFactory;
         private readonly FunctionInvocation _invocation;
@@ -24,7 +25,6 @@ namespace Microsoft.Azure.Functions.Worker
             _invocation = features.Get<FunctionInvocation>() ?? throw new InvalidOperationException($"The '{nameof(FunctionInvocation)}' feature is required.");
             FunctionDefinition = features.Get<FunctionDefinition>() ?? throw new InvalidOperationException($"The {nameof(Worker.FunctionDefinition)} feature is required.");
         }
-
 
         public override string InvocationId => _invocation.Id;
 
@@ -58,11 +58,18 @@ namespace Microsoft.Azure.Functions.Worker
 
         public override RetryContext RetryContext => Features.GetRequired<IExecutionRetryFeature>().Context;
 
-        public void Dispose()
+        public async ValueTask DisposeAsync()
         {
             if (_instanceServicesScope != null)
             {
-                _instanceServicesScope.Dispose();
+                if (_instanceServicesScope is IAsyncDisposable asyncServiceScope)
+                {
+                    await asyncServiceScope.DisposeAsync();
+                }
+                else
+                {
+                    _instanceServicesScope.Dispose();
+                }
             }
 
             _instanceServicesScope = null;
--- a/src/DotNetWorker.Core/Context/DefaultFunctionContext.cs
+++ b/src/DotNetWorker.Core/Context/DefaultFunctionContext.cs
@@ -60,16 +60,9 @@ namespace Microsoft.Azure.Functions.Worker
 
         public async ValueTask DisposeAsync()
         {
-            if (_instanceServicesScope != null)
+            if (_instanceServicesScope is IAsyncDisposable asyncServiceScope)
             {
-                if (_instanceServicesScope is IAsyncDisposable asyncServiceScope)
-                {
-                    await asyncServiceScope.DisposeAsync();
-                }
-                else
-                {
-                    _instanceServicesScope.Dispose();
-                }
+                await asyncServiceScope.DisposeAsync();
             }
 
             _instanceServicesScope = null;
--- a/src/DotNetWorker.Core/Context/DefaultFunctionContext.cs
+++ b/src/DotNetWorker.Core/Context/DefaultFunctionContext.cs
@@ -65,6 +65,8 @@ namespace Microsoft.Azure.Functions.Worker
                 await asyncServiceScope.DisposeAsync();
             }
 
+            _instanceServicesScope?.Dispose();
+
             _instanceServicesScope = null;
             _instanceServices = null;
         }
--- a/src/DotNetWorker.Grpc/GrpcWorker.cs
+++ b/src/DotNetWorker.Grpc/GrpcWorker.cs
@@ -38,7 +38,7 @@ namespace Microsoft.Azure.Functions.Worker
         private readonly IFunctionMetadataProvider _functionMetadataProvider;
 
         public GrpcWorker(IFunctionsApplication application, FunctionRpcClient rpcClient, GrpcHostChannel outputChannel, IInvocationFeaturesFactory invocationFeaturesFactory,
-            IOutputBindingsInfoProvider outputBindingsInfoProvider, IMethodInfoLocator methodInfoLocator, 
+            IOutputBindingsInfoProvider outputBindingsInfoProvider, IMethodInfoLocator methodInfoLocator,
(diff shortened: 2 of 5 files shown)


src/DotNetWorker.Core/Context/DefaultFunctionContext.cs
```diff
@@ -58,11 +58,11 @@ public override IServiceProvider InstanceServices
 
         public override RetryContext RetryContext => Features.GetRequired<IExecutionRetryFeature>().Context;
 
-        public void Dispose()
+        public async ValueTask DisposeAsync()
         {
-            if (_instanceServicesScope != null)
+            if (_instanceServicesScope is IAsyncDisposable asyncServiceScope)
```
kshyju, line 63: explicit null check is not needed as the "is" expression will handle that as well.

liliankasem (approved):

brettsam (approved):

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -223,7 +223,10 @@ private async Task ProcessRequestCoreAsync(StreamingMessage request)
             }
             finally
             {
-                (context as IDisposable)?.Dispose();
+                if (context is IAsyncDisposable asyncContext)
```
fabiocav, line 226: As discussed, we need to maintain the behavior where we dispose of `IDisposable` only contexts, as this change would prevent that from happening.

test/DotNetWorkerTests/TestFunctionContext.cs
```diff
@@ -4,13 +4,14 @@
...
 using System.Linq;
+using System.Threading.Tasks;
 using Microsoft.Azure.Functions.Worker.Context.Features;
 using Microsoft.Azure.Functions.Worker.OutputBindings;
 using Microsoft.Azure.Functions.Worker.Tests.Features;
 
 namespace Microsoft.Azure.Functions.Worker.Tests
 {
-    internal class TestFunctionContext : FunctionContext, IDisposable
+    internal class TestFunctionContext : FunctionContext, IAsyncDisposable
```
fabiocav, line 14: Let's add a set of similar tests that operate on a context that implements `IDisposable` only to make sure that isn't regressed.

src/DotNetWorker.Core/Context/DefaultFunctionContext.cs
fabiocav, line 63: Similar concern here, as we don't control the scope implementation coming back and that might only implement `IDisposable` which would cause scoped services to no longer be disposed of with the changes.

fabiocav (changes requested):
Flow looks good, we just need to make sure we're still handling the `IDisposable` only scenarios 👍🏼

src/DotNetWorker.Grpc/GrpcWorker.cs
kshyju: Brought back the code which handles the case of FunctionContext being just `IDisposable` type.

src/DotNetWorker.Core/Context/DefaultFunctionContext.cs
kshyju: Done.

kshyju:
Ahoy @fabiocav @brettsam @satvu @liliankasem Ping for another review.

liliankasem (approved):

test/DotNetWorkerTests/TestFunctionContext.cs
```diff
@@ -46,7 +66,7 @@ public TestFunctionContext(FunctionDefinition functionDefinition, FunctionInvoca
             BindingContext = new DefaultBindingContext(this);
         }
 
-        public bool IsDisposed { get; private set; }
+        public bool IsDisposed { get; internal set; }
```
madelinegordon, line 69: just wondering - why was this switched to internal?

kshyju: Good catch. This was a change I made in the previous iteration where I had the `IsDisposed` property being set inside the `DisposeAsync` of `TestAsyncFunctionContext`, for which I made this property's set visibility to internal. Later I decided to use the more explicit property name `IsAsyncDisposed`. Forgot to change the visibility. The latest iteration I pushed took care of this now.

fabiocav (approved):

PR #939, opened 2022-06-23
Added support for surfacing user-thrown exception to App Insights

### LINKED ISSUE #912: Change RpcException message for user code exception scenarios
For scenarios in which user code throws an exception, set the fields added [here ](https://github.com/Azure/azure-functions-language-worker-protobuf/pull/70/files) on the RpcException message to facilitate propagating user exception details to app insights. 

Part of work to fix [#4296](https://github.com/Azure/azure-functions-host/issues/4296)

madelinegordon:

Added support for surfacing user-thrown exception in App Insights. Associated with changes in host [#8314](https://github.com/Azure/azure-functions-host/pull/8314)

resolves #912

### Pull request checklist


madelinegordon:
One open question: should these changes be behind a feature flag in the worker as well as in the host?

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -231,7 +231,7 @@ private async Task ProcessRequestCoreAsync(StreamingMessage request)
             {
                 response.Result = new StatusResult
                 {
-                    Exception = ex.ToRpcException(),
+                    Exception = ex.ToUserRpcException(),
```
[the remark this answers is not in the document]
madelinegordon: From my understanding, in the invocation code path it's always a user code-thrown exception.

src/DotNetWorker.Grpc/RpcExtensions.cs
```diff
@@ -131,5 +131,23 @@ internal static FunctionDefinition ToFunctionDefinition(this FunctionLoadRequest
                 StackTrace = exception.StackTrace ?? string.Empty
             };
         }
+
+        internal static RpcException? ToUserRpcException(this Exception exception)
```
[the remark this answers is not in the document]
madelinegordon: That's possible! Are you thinking that would be a cleaner way to do it as an alternative to naming this method ToUserRpcException?

src/DotNetWorker.Grpc/RpcExtensions.cs
```diff
@@ -117,6 +117,10 @@ internal static FunctionDefinition ToFunctionDefinition(this FunctionLoadRequest
             return new GrpcFunctionDefinition(loadRequest, methodInfoLocator);
         }
 
+        /// <summary>
+        /// Returns an RpcException for scenarios not involving a user-thrown exception.
```
liliankasem, line 121: nit: you could say for system exceptions/scenarios?

fabiocav:
@madelinegordon is this now ready for review? Can you confirm whether there's also a feature flag we need to set to enable this?

madelinegordon:
> @madelinegordon is this now ready for review? Can you confirm whether there's also a feature flag we need to set to enable this?

Yes ready for review, and yes the feature flag is an option `EnableUserCodeException` which needs to be set to true manually as it is by default set to false.

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -258,7 +261,7 @@ internal static WorkerInitResponse WorkerInitRequestHandler(WorkerInitRequest re
             response.Capabilities.Add("TypedDataCollection", bool.TrueString);
             response.Capabilities.Add("WorkerStatus", bool.TrueString);
             response.Capabilities.Add("HandlesWorkerTerminateMessage", bool.TrueString);
-
+            response.Capabilities.Add("EnableUserCodeException", enableUserCodeExceptionOption.ToString());
```
liliankasem, line 264: Since you made _workerOptions available globally in this file, you should be able to just use it here directly for `_workerOptions.Value.EnableUserCodeException` instead of changing the method to take in a new parameter

e.g. something like

```csharp
var enableUserCodeException = _workerOptions.Value.EnableUserCodeException ?? false;
response.Capabilities.Add("EnableUserCodeException", enableUserCodeException.ToString());
```

satvu (approved):
I think this looks good overall! Seems like there's just some nit/cleanup items left.

src/DotNetWorker.Grpc/GrpcWorker.cs
madelinegordon: I think because it's a static method the instance variables can't be accessed directly.

liliankasem (approved):
lgtm

fabiocav:
This must wait until the changes in https://github.com/Azure/azure-functions-host/pull/8314 are fully deployed.

liliankasem (changes requested):
Waiting on host release

src/DotNetWorker.Grpc/GrpcWorker.cs
kaibocai: Want to understand how cx can set this workerOptions? @madelinegordon told me it's from bootstrapping code, but I am not quite aware of that. Wondering do we have document explaining this. Thanks.

madelinegordon:
It looks like the changes this is blocked on are deployed [4.9.1](https://github.com/Azure/azure-functions-host/releases/tag/v4.9.1) - is this clear to merge? @fabiocav @brettsam

kaibocai:
> @madelinegordon is this now ready for review? Can you confirm whether there's also a feature flag we need to set to enable this?

Hi @fabiocav , for this feature flag is it set by worker explicitly or it is somehow set by customer through the worker? Thank you.

src/DotNetWorker.Grpc/RpcExtensions.cs
```diff
@@ -131,5 +135,27 @@ internal static FunctionDefinition ToFunctionDefinition(this FunctionLoadRequest
...
+            {
+                Message = exception.Message,
+                Source = exception.Source ?? string.Empty,
+                StackTrace = exception.StackTrace ?? string.Empty,
+                Type = exception.GetType().FullName ?? string.Empty,
+                IsUserException = true
+            };
+        }
     }
+
```
liliankasem, line 160: nit: extra line

release_notes.md
```diff
@@ -3,6 +3,8 @@
 - My change description (#PR/#issue)
 -->
 
+
```
liliankasem, line 6: nit: extra line

liliankasem (commented):
if you can just validate this one more time e2e with the latest host release, that would be awesome - then we should be good to merge.

Also wondering if its worth adding an E2E test? 🤔

src/DotNetWorker.Core/Hosting/WorkerOptions.cs
```diff
@@ -21,5 +22,11 @@ public class WorkerOptions
         /// Gets the collection of input converters.
         /// </summary>
         public InputConverterCollection InputConverters { get; } = new InputConverterCollection();
+
+        /// <summary>
+        /// Gets and sets the flag for opting in to surfacing user-code-thrown
+        /// exceptions to app insights. 
+        /// </summary>
+        public bool EnableUserCodeException { get; set; } = false;
```
liliankasem, line 30: QQ: who is setting this value? Wondering if we meant to use the feature flag via worker.config like we do for worker indexing?

https://github.com/Azure/azure-functions-host/blob/eb8b9dffdc5a417c886c290033cadde68b403ed4/src/WebJobs.Script/Utility.cs#L913-L934

madelinegordon: From my understanding, this is meant to be set by the customer - presenting the opt-in option to them since there is a possibility that certain customers may have dependencies on/subscribe to specific errors which would change with enabling this feature.

kaibocai: @fabiocav , for other language worker that don't have a bootstrap process like java, is it a good way that we use appsetting to allow cx enable this feature flag? Thanks.

jviau:
Some exception messages may contain sensitive information (tokens, PII). How are exceptions handled on the host side of the gRPC contract? Are we going to be writing these to stdout or anything like that?

@brettsam where does this PR stand with our previous discussion around moving away from gRPC channel for sending logs and having users own it themselves?

madelinegordon:
> 

This change in this PR creates parity with how exceptions are already handled in-proc. In the out-of-proc scenario, exceptions thrown by users' function code were being wrapped in an `RpcException` and logged to app insights as that type rather than the type of the originally-thrown exception. So the way that the exceptions themselves are handled in the host in terms of sensitive information hasn't changed.

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -212,8 +214,7 @@ internal static WorkerInitResponse WorkerInitRequestHandler(WorkerInitRequest re
             response.Capabilities.Add("TypedDataCollection", bool.TrueString);
             response.Capabilities.Add("WorkerStatus", bool.TrueString);
             response.Capabilities.Add("HandlesWorkerTerminateMessage", bool.TrueString);
-            response.Capabilities.Add("HandlesInvocationCancelMessage", bool.TrueString);
-
+            response.Capabilities.Add("EnableUserCodeException", bool.TrueString);  //enableUserCodeExceptionOption.ToString());
```
liliankasem, line 217: Can we remove the unused code comment here?

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
```diff
@@ -107,14 +107,21 @@ public async Task<InvocationResponse> InvokeAsync(InvocationRequest request)
...
+            {   
+                // If this isn't a task canceled exception, it was thrown by user's code. 
 
                 if (ex.InnerException is TaskCanceledException or OperationCanceledException)
                 {
                     response.Result.Status = StatusResult.Types.Status.Cancelled;
                 }
+                else
+                {
+                    response.Result = new StatusResult
```
liliankasem, line 121: Do we need to add this `else` statement here? Can we just update line 111 to be 

`response.Result.Exception = ex.ToUserRpcException()`

src/DotNetWorker.Grpc/RpcExtensions.cs
```diff
@@ -131,5 +135,26 @@ internal static FunctionDefinition ToFunctionDefinition(this FunctionLoadRequest
...
             };
         }
+
+        /// <summary>
+        /// Returns an RpcException for when an exception is thrown by user's code. 
+        /// </summary>
+        /// <returns>An RpcException with IsUserException set to true.</returns>
+        internal static RpcException? ToUserRpcException(this Exception exception)
+        {
+            if (exception == null)
```
liliankasem, line 145: super nit: I've noticed that in a lot of the worker code base we've switched to the [is operator](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/is - might be good to do the same here for consistency 

`if (exception is null)`

And if you want you can update `ToRpcException` too

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
liliankasem: Ah wait, I see you dont want it to be a customer error if cancellation happens

src/DotNetWorker.Grpc/GrpcWorker.cs
madelinegordon: Whoops had added that for testing - will change back. Good catch.

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
madelinegordon: I think in the case that it's a TaskCanceledException, we'd want to avoid flagging it as thrown by the user's code.

liliankasem: In this case we will want to make sure we set `response.Result.Exception = ex.ToRpcException();` for cancellation

liliankasem (approved):

src/DotNetWorker.Grpc/RpcExtensions.cs
fabiocav: Yes, that is. the preference here.

src/DotNetWorker.Core/Hosting/WorkerOptions.cs
fabiocav, line 30: This is a good candidate for removal in 2.x, where this would always be the default behavior. Also update the app insights reference as discussed. 

@kshyju for awareness.

src/DotNetWorker.Grpc/GrpcWorker.cs
fabiocav, line 217: This is not using the parameter. Remove comment and pass the actual parameter value.

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
```diff
@@ -108,12 +108,17 @@ public async Task<InvocationResponse> InvokeAsync(InvocationRequest request)
...
+                {
+                    Exception = ex.ToUserRpcException(),
+                    Status = StatusResult.Types.Status.Failure
+                };
 
+                // If worker received a cancellation request from the host, change the response.  
                 if (ex.InnerException is TaskCanceledException or OperationCanceledException)
                 {
                     response.Result.Status = StatusResult.Types.Status.Cancelled;
+                    response.Result.Exception = ex.ToRpcException();
```
fabiocav, line 121: This should remain a `UserRpcException`

fabiocav:
Submitted some comments, but as @liliankasem pointed out, one of the things that seems to be missing are new tests to cover the new functionality

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
```diff
@@ -108,13 +108,20 @@ public async Task<InvocationResponse> InvokeAsync(InvocationRequest request)
...
+                {
+                    response.Result.Exception = ex.ToRpcException();
+                }
                 response.Result.Status = StatusResult.Types.Status.Failure;
 
                 if (ex.InnerException is TaskCanceledException or OperationCanceledException)
                 {
                     response.Result.Status = StatusResult.Types.Status.Cancelled;
-                }
+                }
```
liliankasem, line 124:
```suggestion
                }
```

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -36,6 +36,7 @@ internal class GrpcWorker : IWorker
         private readonly IInputConversionFeatureProvider _inputConversionFeatureProvider;
         private readonly IMethodInfoLocator _methodInfoLocator;
         private readonly IOptions<GrpcWorkerStartupOptions> _startupOptions;
+        private readonly IOptions<WorkerOptions> _workerOptions;
```
jviau, line 39: Does functions worker support reloading of config? If so, we should consider `IOptionsMonitor<>` here, so users can change this behavior while their app is running (without needing to re-deploy).

madelinegordon: noting we discussed offline - since the worker gets this flag from worker options then sends it as a capability to the host in the worker init request code path, this would require more of a design discussion on general dynamic refreshing of options/capabilities.

src/DotNetWorker.Core/Hosting/WorkerOptions.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using System.Dynamic;
```
jviau, line 4: Can this be removed?

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -212,8 +214,7 @@ internal static WorkerInitResponse WorkerInitRequestHandler(WorkerInitRequest re
             response.Capabilities.Add("TypedDataCollection", bool.TrueString);
             response.Capabilities.Add("WorkerStatus", bool.TrueString);
             response.Capabilities.Add("HandlesWorkerTerminateMessage", bool.TrueString);
-            response.Capabilities.Add("HandlesInvocationCancelMessage", bool.TrueString);
```
jviau, line 215: This looks like an accidental removal from merge resolution.

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
```diff
@@ -108,7 +108,14 @@ public async Task<InvocationResponse> InvokeAsync(InvocationRequest request)
             }
             catch (Exception ex)
             {
-                response.Result.Exception = ex.ToRpcException();
+                if (enableUserException)
```
jviau, line 111: Could use a ternary here if you want:

``` CSharp
response.Result.Exception = enableUserException ? ex.ToUserRpcException() : ex.ToRpcException();
```

src/DotNetWorker.Grpc/RpcExtensions.cs
```diff
@@ -117,9 +117,13 @@ internal static FunctionDefinition ToFunctionDefinition(this FunctionLoadRequest
             return new GrpcFunctionDefinition(loadRequest, methodInfoLocator);
         }
 
+        /// <summary>
+        /// Returns an RpcException for system exception scenarios.
+        /// </summary>
+        /// <returns>An RpcException</returns>
         internal static RpcException? ToRpcException(this Exception exception)
```
jviau, line 124: Are we using explicit nullables? If you are interested, you can add this attribute to this function to improve compiler nullable analysis. Same with `ToUserRpcException`

``` CSharp
[return: NotNullIfNotNull("exception")]
```

test/DotNetWorkerTests/Handlers/InvocationHandlerTests.cs
```diff
@@ -132,5 +134,55 @@ public async Task Cancel_InvocationCompleted_ReturnsFalse()
...
+        /// <summary>
+        /// Test unwrapping user code exception functionality. 
+        /// </summary>
+        [Fact]
+        public async Task InvokeAsync_UserCodeThrowsException_OptionEnabled()
+        {
+            var exceptionMessage = "user code exception";
+
+            _mockApplication
+               .Setup(m => m.InvokeFunctionAsync(It.IsAny<FunctionContext>()))
```
jviau, line 147: nit: indentation

```suggestion
                .Setup(m => m.InvokeFunctionAsync(It.IsAny<FunctionContext>()))
```

test/DotNetWorkerTests/Handlers/InvocationHandlerTests.cs
```diff
@@ -132,5 +134,55 @@ public async Task Cancel_InvocationCompleted_ReturnsFalse()
...
+        /// <summary>
+        /// Test keeping user code exception wrapped as RpcException. 
+        /// </summary>
+        [Fact]
+        public async Task InvokeAsync_UserCodeThrowsException_OptionDisabled()
+        {
+            var exceptionMessage = "user code exception";
+
+            _mockApplication
+               .Setup(m => m.InvokeFunctionAsync(It.IsAny<FunctionContext>()))
```
jviau, line 172: nit: indentation

```suggestion
                .Setup(m => m.InvokeFunctionAsync(It.IsAny<FunctionContext>()))
```

jviau (changes requested):
I _think_ there is a regression with the accidental capability removal.

src/DotNetWorker.Core/Hosting/WorkerOptions.cs
kshyju: Filed #1082 to track this item.

src/DotNetWorker.Grpc/GrpcWorker.cs
madelinegordon: Yes good catch!

src/DotNetWorker.Grpc/RpcExtensions.cs
madelinegordon: hmm seeing "Error CS0122: 'NotNullIfNotNullAttribute' is inaccessible due to its protection level" in the build when that's added - is is meant for public methods?

jviau: We just might not be on a late enough .net or SDK version for it to be visible. That is fine - leave it out for now.

src/DotNetWorker.Grpc/RpcExtensions.cs
```diff
@@ -120,7 +121,7 @@ internal static FunctionDefinition ToFunctionDefinition(this FunctionLoadRequest
         /// <summary>
         /// Returns an RpcException for system exception scenarios.
         /// </summary>
-        /// <returns>An RpcException</returns>
+        /// <returns>An RpcException</returns>
```
jviau, line 124: nit: extra whitespace
```suggestion
        /// <returns>An RpcException</returns>
```

src/DotNetWorker.Grpc/Handlers/IInvocationHandler.cs
```diff
@@ -15,9 +16,9 @@ internal interface IInvocationHandler
         /// an associated cancellation token source for the invocation.
         /// </summary>
         /// <param name="request">Function invocation request</param>
-        /// <param name="enableUserException">flag to enable unwrapping user exception</param>
+        /// <param name="workerOptions"></param>
         /// <returns><see cref="InvocationResponse"/></returns>
-        Task<InvocationResponse> InvokeAsync(InvocationRequest request, bool enableUserException = false);
+        Task<InvocationResponse> InvokeAsync(InvocationRequest request, IOptions<WorkerOptions>? workerOptions = null);
```
jviau, line 21: Not sure if this is a back and forth from a comment someone else left, but consider dropping the `IOptions` wrapper here and on the other similar methods.

```suggestion
        Task<InvocationResponse> InvokeAsync(InvocationRequest request, WorkerOptions? workerOptions = null);
```

The benefit is that there are 3 different forms of options - `IOptions`, `IOptionsSnapshot`, and `IOptionsMonitor` - we wouldn't want to force one of the 3.

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
```diff
@@ -45,8 +46,9 @@ internal class InvocationHandler : IInvocationHandler
             _inflightInvocations = new ConcurrentDictionary<string, CancellationTokenSource>();
         }
 
-        public async Task<InvocationResponse> InvokeAsync(InvocationRequest request, bool enableUserException = false)
+        public async Task<InvocationResponse> InvokeAsync(InvocationRequest request, IOptions<WorkerOptions>? workerOptions = null)
         {
+            bool enableUserCodeException = workerOptions is not null && workerOptions.Value.EnableUserCodeException;
```
jviau, line 51: Can also be done as so:

```suggestion
            bool enableUserCodeException = workerOptions?.Value.EnableUserCodeException ?? false;
```

jviau (approved):

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -213,7 +216,7 @@ internal static WorkerInitResponse WorkerInitRequestHandler(WorkerInitRequest re
             response.Capabilities.Add("WorkerStatus", bool.TrueString);
             response.Capabilities.Add("HandlesWorkerTerminateMessage", bool.TrueString);
             response.Capabilities.Add("HandlesInvocationCancelMessage", bool.TrueString);
-
+            response.Capabilities.Add("EnableUserCodeException", enableUserCodeException.ToString());
```
kshyju, line 219: If the worker option property value is false, should we even send this capability entry? Looks like host code does not check the value of the entry, but checks whether this capability entry is present or not.

https://github.com/Azure/azure-functions-host/blob/ec3c2d1049351870bbf611760520cf1e48075c8f/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L786

kshyju: That pattern is used for all Boolean type capabilities. 
https://github.com/Azure/azure-functions-host/blob/ec3c2d1049351870bbf611760520cf1e48075c8f/src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs#L328-L356

madelinegordon: Ooh that's a good point! Yes in that case we shouldn't send the capability if the option is false.

src/DotNetWorker.Core/Hosting/WorkerOptions.cs
liliankasem: You should be able to use app settings

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -35,7 +35,8 @@ internal class GrpcWorker : IWorker
         private readonly IOutputBindingsInfoProvider _outputBindingsInfoProvider;
         private readonly IInputConversionFeatureProvider _inputConversionFeatureProvider;
         private readonly IMethodInfoLocator _methodInfoLocator;
-        private readonly IOptions<GrpcWorkerStartupOptions> _startupOptions;
```
liliankasem, line 38: I don't know that we had to change this item IOptions<GrpcWorkerStartupOptions> but I'm not the best person to ask xD

cc: @kshyju @jviau

madelinegordon: Discussed with Jacob offline - didn't need to but changing to create consistency with WorkerOptions

liliankasem:
ship it

fabiocav (approved):

inishantmishra:
Even after enabling the EnableUserCodeException = true in dotnet isolated Azure function, I am still receiving RPC exceptions with no call stack.
I thought by this change, if l throw some custom exception in my code and if I handle it and log it, I will not get the RPC exception but my custom exception in appInsights. 
But still, I am receiving the RPC exception.

Let me know what am I missing? Why am i still receiving RPC exceptions?

I am using
dotnet-isolated
~4
.net 7 Isolated Azure function

inishantmishra:
Even after enabling the EnableUserCodeException = true in dotnet isolated Azure function, I am still receiving RPC exceptions with no call stack.
I thought by this change, if l throw some custom exception in my code and if I handle it and log it, I will not get the RPC exception but my custom exception in appInsights.
But still, I am receiving the RPC exception.

Let me know what am I missing? Why am i still receiving RPC exceptions?

I am using
dotnet-isolated
~4
.net 7 Isolated Azure function

PR #943, opened 2022-06-23
Adding ActivitySource for invocations

brettsam:


src/DotNetWorker.Core/Diagnostics/FunctionActivitySource.cs
```diff
@@ -0,0 +1,28 @@
...
+        private static readonly string _processId = Process.GetCurrentProcess().Id.ToString();
+        private static readonly ConcurrentDictionary<string, string> _categoryCache = new ConcurrentDictionary<string, string>();
+
+        public static Activity? StartInvoke(FunctionContext context)
+        {
+            var activity = _activitySource.StartActivity("Invoke", ActivityKind.Internal, context.TraceContext.TraceParent);
+
+            if (activity != null)
+            {
+                activity.AddTag(LogConstants.InvocationIdKey, context.InvocationId);
```
jviau, line 19: Should we start adopting OpenTelemetry semantic conventions? https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/faas.md

Although that is an experimental status still. Something I am not a fan of is attribute names like `faas.invoked_name`. I would prefer `faas.invoked.name`.

src/DotNetWorker.Core/FunctionsApplication.cs
```diff
@@ -59,9 +60,13 @@ public void LoadFunction(FunctionDefinition definition)
         public Task InvokeFunctionAsync(FunctionContext context)
         {
             var scope = new FunctionInvocationScope(context.FunctionDefinition.Name, context.InvocationId);
-            using (_logger.BeginScope(scope))
+
+            using (FunctionActivitySource.StartInvoke(context))
```
jviau, line 64: Consider using newer using statement style:

``` CSharp
using Activity activity = FunctionActivitySource.StartInvoke(context);
using IDisposable scope = _logger.BeginScope(scope);
...
```

src/DotNetWorker.Core/Diagnostics/FunctionActivitySource.cs
brettsam: Yes! This is the guidance that I was looking for -- I wasn't sure what to call all of these things. I recall a previous attempt at conventions has been abandoned, so I'm not sure what we should be doing with these "experimental" standards...

I think we need to discuss this before we remove the preview tag on this package.

brettsam:
Closing this in favor of a new one I have coming.

Issue #760, opened 2021-12-16
Correct way to use Application Insights in v4 and .Net 6


emilssonn:
From reading different issues, discussions and documentation from the past months, it is still unclear to me how to correctly use/install Application Insights in a .Net isolated function running .Net 6 and v4.

I have tested with the nuget Microsoft.Azure.Functions.Worker.Extensions.ApplicationInsights. The absolute most basic features of application insights works with this nuget. But no dependency tracking etc.

Where can I find the correct documentation for using Application Insights with a .Net isolated function running .Net 6 and v4? Including depedency tracking,  log levels filter, using TelemetryClient and TelemetryConfiguration in code.

elsand:
@emilssonn Microsoft.Azure.Functions.Worker.Extensions.ApplicationInsights seems to be non-working. https://github.com/Azure/azure-functions-applicationinsights-extension/tree/main/src/Worker is literally empty except for a csproj. 

I'm not sure quite sure how to parse https://github.com/Azure/app-service-announcements/issues/346#issuecomment-965840479, but it appears that Microsoft.Azure.Functions.Worker.Extensions.ApplicationInsights isn't supposed to be used at this point.

In any case we're experiencing the same issue as you. Rudimentary trace logging is performed for external calls, but they are not tracked as dependencies (without using Microsoft.Azure.Functions.Worker.Extensions.ApplicationInsights).

elsand:
@emilssonn It appears to work if you follow the steps at https://docs.microsoft.com/en-us/azure/azure-monitor/app/worker-service#using-application-insights-sdk-for-worker-services, that is get the `Microsoft.ApplicationInsights.WorkerService` package and add `services.AddApplicationInsightsTelemetryWorkerService();`  to `ConfigureServices` (`services` being `IServiceCollection`).

emilssonn:
`Microsoft.ApplicationInsights.WorkerService` does indeed seem to work, but not fully. Dependencies are logged, but they are not connected to the requests/other telemetry.

I am 99.9% sure I saw some official documentation about `Microsoft.Azure.Functions.Worker.Extensions.ApplicationInsights` and that it was the official/correct way. But I can no longer find it and the app service announcement might explain that.

But I still cannot find any official documentation about .Net isolated functions running .Net 6 and v4.
https://docs.microsoft.com/en-us/azure/azure-monitor/app/worker-service#using-application-insights-sdk-for-worker-services makes no mention of Azure Functions.

[...]

Issue #702, opened 2021-11-08
Logging using an `ILogger<T>` instance does not work when it is not part of a function execution


kshyju:
Logging using a [constructor injected](https://docs.microsoft.com/en-us/azure/azure-functions/dotnet-isolated-process-guide#logging) `ILogger<T>` instance does not work when it is not part of a function execution. This means logging does not work inside an implementation of `IHostedService` or `BackgroundService` which is registered as a hosted service in worker.

````
internal class CacheRefreshService : BackgroundService
{
    private readonly ILogger<CacheRefreshService> _logger;
    public CacheRefreshService(ILogger<CacheRefreshService> logger)
    {
        _logger = logger ?? throw new ArgumentNullException(nameof(logger));
    }

    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
    {
        // Below logging does not work.
        _logger.LogInformation("Inside ExecuteAsync");
        // code to refresh cache.
        await Task.Delay(5000);
    }
}
````
In this case, the code executed in the background service has no invocation id associated with it. So I believe [this line in host](https://github.com/Azure/azure-functions-host/blob/dcfd7fc1458da020de2c3f49b1d9672b1e1bc229/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L696) is dropping the log messages sent from worker to host.

SeanFeldman:
@kshyju I might be missing some context/knowledge here, but a function has to be triggered. And a `BackgroundService` would not be executed if there's no trigger.

ayrton97:
@SeanFeldman Can you please explain why the function has to be triggered? 
In my case I just register a service and background service.
The service is going to consume some data which will be periodically retrieved by the background service which is added as a hosted service and doesn't require any invocation.

SeanFeldman:
@ayrton97, my question is based on understanding how Functions operate. Take a consumption plan where a Function is constrained in the [execution time](https://docs.microsoft.com/en-us/azure/azure-functions/functions-scale#timeout). That means b/g service or not, the function will be shut down and the background service won't be able to run.

ayrton97:
I'm not talking about the consumption plan, I'm talking about the way function is initialized.
For references: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/hosted-services?view=aspnetcore-5.0&tabs=visual-studio
[...]

PR #944, opened 2022-06-23
adding AI project

brettsam:
This PR introduces a new nuget package (in preview): `Microsoft.Azure.Functions.Worker.ApplicationInsights`. This package should help to address issues like #702, #760, and others.

The goal of this new package is to simplify the Application Insights wire-up required in order to have proper correlation between the RequestTelemetry generated in the Functions host and any telemetry generated during the invocation on the isolated worker.

The package uses as much of the default Application Insights setup as possible so that we can utilize the existing documentation and (hopefully) not introduce many new Functions-only concepts. 

### Background
Azure Functions has automatic support for App Insights built into the host (and for in-proc .NET functions). This emits `RequestTelemetry` for every invocation, `TraceTelemetry` for `ILogger` logs, has Live Metrics enabled and auto-tracks Dependencies like outgoing HTTP calls, SQL queries, etc.

In the out-of-proc worker model, a gRPC connection is made between the worker and host. This connection is used to send invocation details from the host to the worker, but it’s also used for a worker to send log messages *back* to the host. This allows a worker to expose some logger class that behind-the-scenes sends log messages to the host via gRPC, which then logs it to the configured Application Insights resource through the host. This is the extent of "Application Insights support" that we offer to all of the out-of-proc workers. Side note that this support only works *during an invocation*. Any logs emitted outside the scope of an invocation are ignored by the host, which results in issues like #702.

For the dotnet-isolated model we want to expand this support to create a richer offering that allows the worker to emit telemetry directly to Application Insights while maintaining the proper telemetry correlation (i.e. Dependencies are auto-tracked and show up as part of the overall Request).

The package introduces two new extension methods available on `IFunctionsWorkerApplicationBuilder` that allow you register Application Insights services (`AddApplicationInsights()`)and the Application Insights logger (`AddApplicationInsightsLogger()`).

```csharp
var host = new HostBuilder()
    {
        builder
            .AddApplicationInsights()
            .AddApplicationInsightsLogger();
    })
    .Build();
```

### AddApplicationInsights()
This method internally calls the Application Insights `IServiceCollection` extension method `AddApplicationInsightsTelemetryWorkerService()`, with a couple of extra services added to provide the correlation needed for Functions. See the [Application Insights documentation for Worker Services](https://docs.microsoft.com/en-us/azure/azure-monitor/app/worker-service) for details. The Functions extension method includes an optional callback parameter to allow you to configure the underlying `ApplicationInsightsServiceOptions` options.

With the default configuration, this will include:
- Automatic dependency tracking as a child of the parent invocation (which is still tracked via the host).
- A new per-invocation tracked dependency -- currently of type `Azure.Functions` with name `Invoke` that tracks the time spent on the isolated worker.
- Live Metrics from the worker (as well as the host).

### AddApplicationInsightsLogger()
This method internally calls the Application Insights `ILoggingBuilder` extension method `AddApplicationInsights()`, with a couple of extra services added to provide the correlation needed for Functions. See the [Application Insights documentation for using ILogger in a Console application](https://docs.microsoft.com/en-us/azure/azure-monitor/app/ilogger#console-application) for details. The Functions extension method includes an optional callback parameter to allow you to configure the underlying `ApplicationInsightsLoggerOptions` options. 

Note that calling this function will disable the gRPC logging that occurs by default; all `ILogger` logs will be emitted directly to Application Insights as `TraceTelemetry`. This should address issues like #702 without changing host behavior and affecting other language workers.

### Configuration
Some notes on configuration:
- Using the app setting `APPLICATIONINSIGHTS_CONNECTION_STRING` will automatically wire up the both the host and the worker to the same Appliation Insights instance.
- Logging and Application Insights configurations in the worker are completely separate from those configured in `host.json`. For example, if you've disabled sampling in the host, you will need to also disable it in the worker. Same with configuring logging levels. Please refer to the documentation links above to configure your worker.


src/DotNetWorker/Hosting/ServiceCollectionExtensions.cs
```diff
@@ -1,47 +1,47 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the MIT License. See License.txt in the project root for license information.
-
-using System;
-using System.Text.Json;
-using Microsoft.Azure.Functions.Worker;
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
fabiocav, line 1: Just changing line endings here?

fabiocav (commented):
Very small  nit comments

src/DotNetWorker.Core/Hosting/WorkerOptions.cs
```diff
@@ -21,5 +22,10 @@ public class WorkerOptions
         /// Gets the collection of input converters.
         /// </summary>
         public InputConverterCollection InputConverters { get; } = new InputConverterCollection();
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to send <see cref="ILogger"/> logs through the Functions host.
+        /// </summary>
+        public bool DisableHostLogger { get; set; } = false;
```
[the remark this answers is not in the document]
fabiocav: One of the things I had previously mentioned in discussion would be to introduce a proper service for user logs. By default, the gRPC implementation would register one (overridable) that would push to the host. If the implementation here registers one as well, that would bypass the gRPC logic without requiring a flag. Thoughts?

liliankasem (approved):
lgtm! Just 1 question about custom metrics and use of a const file :D

src/DotNetWorker.ApplicationInsights/FunctionActivitySource.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    /// Note: This class will eventually move in to the core worker assembly. Including it in the
+    ///       ApplicationInsights package so we can utilize it during preview.
+    internal static class FunctionActivitySource
+    {
+        private const string InvocationIdKey = "InvocationId";
+        private const string NameKey = "Name";
+        private const string ProcessIdKey = "ProcessId";
+
+        private static readonly ActivitySource _activitySource = new("Microsoft.Azure.Functions.Worker");
```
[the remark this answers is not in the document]
brettsam: I was holding off on that until this stuff is in its final form (this ActivitySource will move into Core, for example). Then I was going to constant-ize everything I could. So I'll punt this for now, but good catch.

src/DotNetWorker.Core/Hosting/WorkerOptions.cs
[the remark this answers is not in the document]
brettsam: Let me look at that -- will submit a change here soon with this split out.

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryInitializer.cs
```diff
@@ -0,0 +1,93 @@
...
+            if (string.IsNullOrEmpty(instanceName))
+            {
+                instanceName = Environment.GetEnvironmentVariable(ComputerNameKey);
+                if (string.IsNullOrEmpty(instanceName))
+                {
+                    instanceName = Environment.GetEnvironmentVariable(ContainerNameKey);
+                }
+            }
+
+            return instanceName ?? "Unknown";
```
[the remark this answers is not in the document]
brettsam: Done.

src/DotNetWorker.Grpc/GrpcFunctionsHostLogger.cs
```diff
@@ -51,6 +54,12 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
             }
             else
             {
+                // disable user logs going to host
+                if (_disableHostLogger && !WorkerMessage.IsSystemLog)
```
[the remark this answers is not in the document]
liliankasem: I think we can leave it for now since other languages are also maybe using custom metrics

src/DotNetWorker.ApplicationInsights/FunctionsApplicationInsightsExtensions.cs
```diff
@@ -0,0 +1,94 @@
+﻿using System;
```
madelinegordon, line 1: nit - Are there certain files that need the copyright header and some that don't?

brettsam: No they all should... will go through and double-chedk!

src/DotNetWorker.Core/Logging/NullLogWriter.cs
```diff
@@ -0,0 +1,44 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.Functions.Worker.Logging
+{
+    /// <summary>
+    /// Minimalistic LogWriter that does nothing.
```
satvu, line 11: Could you elaborate on this LogWriter? Trying to understand it. Is it for when a user doesn't create/define other specific loggers (UserLogWriter, SystemLogWriter) and this is the default put in their place? 

This summary caught me off guard 🤣

brettsam: All of our "logging through the host" goes through Grpc, which isn't part of the core packages. So we need some default way to basically "do nothing" without breaking dependency injection if for some reason Grpc isn't wired up. In reality this will be very rare.

src/DotNetWorker.Core/Logging/WorkerLogger.cs
```diff
@@ -0,0 +1,54 @@
...
+        public bool IsEnabled(LogLevel logLevel)
+        {
+            return logLevel != LogLevel.None;
+        }
+
+        public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
+        {
+            if (WorkerMessage.IsSystemLog)
+            {
+                _systemLogWriter.WriteSystemLog(_scopeProvider, _category, logLevel, eventId, state, exception, formatter);
```
kshyju, line 40: Would this log anything ? Because `ISystemLogWriter` implementation registered to DI container is `NullLogWriter` which does nothing.

src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs
```diff
@@ -43,12 +42,12 @@ public static IServiceCollection AddGrpc(this IServiceCollection services)
...
-            {
-                logging.Services.AddSingleton<ILoggerProvider, GrpcFunctionsHostLoggerProvider>();
-                logging.Services.AddSingleton<IWorkerDiagnostics, GrpcWorkerDiagnostics>();
-            });
+            // Internal logging
+            services.AddSingleton<GrpcFunctionsHostLogWriter>();
+            services.AddSingleton<IUserLogWriter>(p => p.GetRequiredService<GrpcFunctionsHostLogWriter>());
+            services.AddSingleton<ISystemLogWriter>(p => p.GetRequiredService<GrpcFunctionsHostLogWriter>());
+            services.AddSingleton<IUserMetricWriter>(p => p.GetRequiredService<GrpcFunctionsHostLogWriter>());
+            services.AddSingleton<IWorkerDiagnostics, GrpcWorkerDiagnostics>();
```
kshyju, line 50: Unrelated to changes from this PR: 
Are we supposed to call ` _outputChannel.TryWrite(message);` [here?](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/src/DotNetWorker.Grpc/Http/GrpcWorkerDiagnostics.cs#L70)

samples/FunctionApp/Program.cs
```diff
@@ -17,7 +18,12 @@ static async Task Main(string[] args)
             //<docsnippet_startup>
             var host = new HostBuilder()
                 //<docsnippet_configure_defaults>
-                .ConfigureFunctionsWorkerDefaults()
+                .ConfigureFunctionsWorkerDefaults(builder =>
+                {
+                    builder
+                        .AddApplicationInsights()
```
kshyju, line 24: Should we consider exposing a single method (which internally calls these 2) for the convenience of user?

src/DotNetWorker.ApplicationInsights/FunctionsApplicationInsightsExtensions.cs
```diff
@@ -0,0 +1,85 @@
...
+            // core worker assembly.
+            if (!builder.Services.Any(p => p.ImplementationType == typeof(FunctionActivitySourceMiddleware)))
+            {
+                builder.Services.AddSingleton<FunctionActivitySourceMiddleware>();
+                builder.Use(next =>
+                {
+                    return async context =>
+                    {
+                        var middleware = context.InstanceServices.GetRequiredService<FunctionActivitySourceMiddleware>();
+                        await middleware.Invoke(context, next);
```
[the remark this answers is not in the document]
brettsam: Yes, that's why we'll move this middleware to the Core bits before releasing. It won't actually be middleware in the end, we'll put it directly into the `FunctionsApplication.InvokeFunctionAsync()` so that it's (effectively) un-overrideable.

src/DotNetWorker.ApplicationInsights/FunctionsApplicationInsightsExtensions.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// <returns>The <see cref="IFunctionsWorkerApplicationBuilder"/></returns>
+        public static IFunctionsWorkerApplicationBuilder AddApplicationInsightsLogger(this IFunctionsWorkerApplicationBuilder builder, Action<ApplicationInsightsLoggerOptions>? configureOptions = null)
+        {
+            builder.AddCommonServices();
+
+            builder.Services.AddLogging(logging =>
+            {
+                logging.AddApplicationInsights(options =>
+                {
+                    options.IncludeScopes = false;
```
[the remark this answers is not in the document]
brettsam: Oh good catch. I didn't mean to leave this as `false` -- I think we should keep things as default as possible so we can just link to App Insights docs directly. This does cause some extra App Insights values to show up in telemetry properties, but I think that's okay.

Example: https://docs.microsoft.com/en-us/azure/azure-monitor/app/ilogger#logging-scopes.

src/DotNetWorker.Core/Logging/WorkerLogger.cs
brettsam: This gets overridden by the call to AddGrpc (it adds the `GrpcFunctionsHostLogWriter` via DI).

System logs are meant to be only for us (in kusto). They have to go back through the host and don't show up in App Insights today. That's why I had to split things out so weirdly -- we needed ways to stop one type of log (user logs) from going through the host, while allowing others (the system logs) to continue flowing.

src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs
brettsam: Certainly seems like it... I can follow up on that with a separate PR since it's unrelated.

samples/FunctionApp/Program.cs
brettsam: Ideally yes... but App Insights has these as two separate concepts. I'd like to try to mirror their usage as much as possible so we can offload the documentation to them. I agree it's kinda odd... but we can review before making this GA.

fabiocav (approved):
Let's get this in so we can start validating the functionality. We'll continue to iterate as you go. :shipit:

michaelelleby:
@brettsam You need to add a dependency on `Microsoft.Azure.Functions.Worker` version 1.8.0 or higher, as this code crashes when using `Microsoft.Azure.Functions.Worker` version 1.6.0 with the following exception
`System.MissingMethodException: 'Method not found: 'Microsoft.Extensions.DependencyInjection.IServiceCollection Microsoft.Extensions.DependencyInjection.ServiceCollectionExtensions.RegisterDefaultConverters(Microsoft.Extensions.DependencyInjection.IServiceCollection)'.'`

pseabury:
I don't see the option to set the option for SqlCommandText like in the other telemetry modules... am I missing something or is this functionality missing from this library currently?

In Micrososft.ApplicationInsights.WorkerService
`builder.Services.ConfigureTelemetryModule<DependencyTrackingTelemetryModule>((module, o) => { module.EnableSqlCommandTextInstrumentation = true; });`


In Microsoft.Azure.Functions.Worker.ApplicationInsights
```
builder
            .AddApplicationInsights(opts =>
            {
                opts.EnableSqlCommandTextInstrumentation = true; // Does not exist
            })
            .AddApplicationInsightsLogger();
```

Done3319:
For me this works in general but I still have difficulties getting logs from the isolated worker with a level lower than Warning.
I understand that this now needs to be controlled separately from host.json and therefore tried to controll via appsettings.json:

```
           var host = new HostBuilder()
                .ConfigureAppConfiguration(c =>
                {
                    c.AddJsonFile("appsettings.json", optional: true, reloadOnChange: true);
                    c.AddJsonFile("local.settings.json", optional: true, reloadOnChange: true);
                    c.AddEnvironmentVariables();
                    c.Build();
                })
                {
                        builder
                        .AddApplicationInsights()
                        .AddApplicationInsightsLogger();

                })
```

appsettings.json has the same settings like host.json which work perfectly for the host.

Inside the worker I tried logging to a ILoggerFactory via DI and also via the FunctionContext.GetLogger().
```
            _logger.LogWarning("LogWarning Test");
            _logger.LogInformation("LogInformation Test");
```

The first one comes through, the second one not.

StMarian:
@Done3319 try https://github.com/Azure/azure-functions-dotnet-worker/issues/1123#issuecomment-1301960722

nicolashemery:
Hello,

I was trying to used isolated function to run Availability test.
I had an issue following [](https://learn.microsoft.com/en-us/azure/azure-monitor/app/availability-azure-functions) as Activity.Current was being null if using packages:
-Microsoft.Extensions.Logging.ApplicationInsights 2.21.0
-Microsoft.ApplicationInsights.WorkerService 2.21.0
I tried to make something with functionContext, but unsuccessfully, and finally went to this package and it worked.

But i already new this package and discarded it in the past for other function as i needed scope logs and its hardcoded in [src/DotNetWorker.ApplicationInsights/FunctionsApplicationInsightsExtensions.cs](https://github.com/Azure/azure-functions-dotnet-worker/blob/f5a232c00eea67a49ca9e2db0aa774d69cba607c/src/DotNetWorker.ApplicationInsights/FunctionsApplicationInsightsExtensions.cs)
` logging.AddApplicationInsights(options =>
 {
    options.IncludeScopes = false;
    configureOptions?.Invoke(options);
}); `

My question is then (probably @brettsam) is there a reason why this is harcoded?

theo-albers:
Hi @brettsam, is there a reason why TelemetryClient isn't registered as singleton? Doing so would allow us to resolve that dependency and start tracking events. The scenario I see is to track retries and single successful runs for Azure Service Bus triggers. 

Right now https://github.com/Azure/azure-functions-dotnet-worker/blob/main/src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs creates a TelemetryClient. I see no reason why this dependency can't be registered as singleton. 

https://github.com/Azure/azure-functions-dotnet-worker/blob/main/src/DotNetWorker.ApplicationInsights/FunctionsApplicationInsightsExtensions.cs alsready registers the module as singleton. 

Or are you worried that someone already registered the TelemetryClient and that's why you would like an instance of "your own"?

theo-albers:
Yes, I discovered that this afternoon: once I include and enable the Functions AI package, I can resolve the TelemetryClient.

SPSCS-Simon:
@brettsam given this is merged, what are the plane to actually release the`Microsoft.Azure.Functions.Worker.ApplicationInsights` package?

BrandonSchreck:
I'm not sure if this is covered in another associated ticket, but after updating my project to use Microsoft.Azure.Functions.Worker.ApplicationInsights v1.0.0-preview4 I am only seeing warning show up in Application Insights traces. I've purposely left our host.json to use the default logging levels (Warning):

host.json
```
{
    "version": "2.0",
    "logging": {
        "applicationInsights": {
            "samplingSettings": {
                "isEnabled": true,
                "excludedTypes": "Request"
            }
        }
    }
}
```

What I was wanting to accomplish was to update the Functions Configuration, set the logging level to Information, and be able to see those logged in Application Insights traces. However, setting either `AzureFunctionsJobHost__logging__logLevel__default` or `AzureFunctionsJobHost__logging__logLevel__Function.{{FunctionName}}` to Information doesn't seem to work. I thought it might be an issue with the configuration not be recognized without a restart, but even after restarting I'm only seeing Warning/Error logs in Application Insights traces.

I think it is worth mentioning that I can watch the Log Stream and see the Information logs being written out.

What would be the correct approach here?

leozin:
Hi @brettsam ,
Is it still possible to use TelemetryClient instead of logging when sending Telemetry data to Azure in isolated Azure Functions?  
Our platform has many submodules that use TelemetryClient without Dependency Injection, which is created through a static initializer, using APPINSIGHTS_INSTRUMENTATIONKEY.   
As our new Azure (Service Bus Listeners) Functions are dotnet-isolated, the `TelemetryClient` doesn't send telemetry data anymore. The content in the `traces` table lacks some important data, for example, `operationName`, `customDimensions` (like Function Invocation Id) and synthetic source, among other things.  
If we use logging, then the data is sent correctly, however, if we want to re-implement our telemetry mechanism using logging and DI, it will require a massive overhaul and we would lose some features like metrics etc.  
Previously, when using TelemetryClient, it would populate everything automatically (in a non-isolated function), but now it seems that I have to set all these parameters manually, or I'm doing something wrong when configuring Logging.

My Program.cs is defined as:
```
var host = Host.CreateDefaultBuilder()
.ConfigureFunctionsWorkerDefaults(builder => builder.AddApplicationInsights().AddApplicationInsightsLogger())
.ConfigureServices() // our Singletons
.Build();

await host.RunAsync();
```

and our packages are:
```
<PackageReference Include="Microsoft.ApplicationInsights.WorkerService" Version="2.21.0" />
<PackageReference Include="Microsoft.Azure.Functions.Worker.ApplicationInsights" Version="1.0.0-preview4" />
<PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.11.0" />
<PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
<PackageReference Include="System.Net.NameResolution" Version="4.3.0" />
<PackageReference Include="Microsoft.Azure.Functions.Extensions" Version="1.1.0" />
<PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.ServiceBus" Version="5.11.0" />
<PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http" Version="3.0.13" />
<PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.16.0" />
<PackageReference Include="Microsoft.NETCore.Targets" Version="5.0.0" />
```

Would you have an idea if I need to add some extra config or something else?

Issue #930, opened 2022-06-15
Refactor GRPCWorker if-else ladder


surgupta-msft:
> Related to code cleanup and refactoring

There is a big if-else ladder in GRPCWorker which can be refactored into a neat switch statement (or we can discuss approaches here if needed).

https://github.com/Azure/azure-functions-dotnet-worker/blob/c522e7be238b5cc716f732f1ea8be3a857c742b3/src/DotNetWorker.Grpc/GrpcWorker.cs#L130-L145

kshyju:
FWIW, There is no perf degradation by switching to switch statement or switch expression. [See the result ](https://gist.github.com/kshyju/d099eacc1dc1558c4382b6afd186d2bb)of a benchmark between the three.

VS also has a quick action which offers converting the **_if else_** block to switch statements or switch expression.

victormarante:
Hi, I can with this issue if it's ok.

kshyju:
Go for it @victormarante

victormarante:
PR is up, please feel free to review. @kshyju

PR #972, opened 2022-07-19
Handle InvocationCancel message and signal cancellation for in-flight invocations

### LINKED ISSUE #976: Handle InvocationCancel message and signal cancellation for in-flight invocations
Related to: 

- https://github.com/Azure/azure-functions-host/issues/2152

Following [this host PR](https://github.com/Azure/azure-functions-host/pull/8556), the host will now have the ability to send a cancellation request to workers. It is up to the workers how they handle the cancellation message (`InvocationCancel`) we receive.

The goal of this issue is to implement a handler for the `InvocationCancel` message and pass a `CancellationToken` to functions, allowing users to handle cancellation tokens in their code.

This issue includes creating a sample and updating the design document with a reference on how to use this new feature and handle the `CancellationToken`. We will also need to update public-facing [documentation](https://docs.microsoft.com/en-us/azure/azure-functions/dotnet-isolated-process-guide#differences-with-net-class-library-functions).

Design document available in the [wiki](https://github.com/Azure/azure-functions-host/wiki/Out-of-proc-Support-for-Cancellation-Tokens).

liliankasem:

resolves #976

- Implemented a converter so that users can bind to a CancellationToken in their function
- If users are binding to a CancellationToken, we create a new CancellationTokenSource when an invocation is created
- Implementation of FunctionInvocationManager to keep track of inflight invocations and their associated CancellationTokenSource
- Handling the InvocationCancel message from the host and cancelling an invocations associated CancellationTokenSource

### Pull request checklist

  * [x] Otherwise: Documentation issue linked to PR: #996
  * E2E tests will be in a separate [PR](https://github.com/Azure/azure-functions-dotnet-worker/compare/liliankasem/ct-e2e-tests)


src/DotNetWorker.Core/Context/DefaultFunctionContext.cs
```diff
@@ -58,6 +59,9 @@ public override IServiceProvider InstanceServices
 
         public override RetryContext RetryContext => Features.GetRequired<IExecutionRetryFeature>().Context;
 
+        // public override CancellationToken CancellationToken => base.CancellationToken;
+        public override CancellationToken CancellationToken => Features.GetRequired<IExecutionCancellationFeature>().CancellationToken;
```
liliankasem, line 63: @fabiocav mentioned that we should not need to use a `Feature` here,  but without the Feature, the cancellation token was not getting cancelled / updated in the context from within the user app. In other words, the cancellation was not getting signaled and we never hit the `OperationCanceledException` in the sample. Looking for advice on the right way to do this.

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -65,6 +66,7 @@ internal class GrpcWorker : IWorker
             _serializer = workerOptions.Value.Serializer ?? throw new InvalidOperationException(nameof(workerOptions.Value.Serializer));
             _inputConversionFeatureProvider = inputConversionFeatureProvider ?? throw new ArgumentNullException(nameof(inputConversionFeatureProvider));
             _functionMetadataProvider = functionMetadataProvider ?? throw new ArgumentNullException(nameof(functionMetadataProvider));
+            _invocationCancelDictionary = new Dictionary<string, CancellationTokenSource>();
```
liliankasem, line 69: I just put this here to get myself unblocked, but again looking for guidance on the right approach here. We need to have a cancellation token source associated with each invocation (via invocationId), and since the context is created in `InvocationRequestHandlerAsync` we do not have access to the cancellation token or cts at all. 

@fabiocav mentioned maybe keeping track of invocations and their context somewhere (instead of the invocationId and cts like I am doing here) - any thoughts on this?

src/DotNetWorker.Core/Context/FunctionContext.cs
```diff
@@ -59,5 +60,10 @@ public abstract class FunctionContext
         /// Gets a collection containing the features supported by this context.
         /// </summary>
         public abstract IInvocationFeatures Features { get; }
+
+        /// <summary>
+        /// Gets the <see cref="CancellationToken"/> that signals a function invocation is being cancelled.
+        /// </summary>
+        public virtual CancellationToken CancellationToken { get; }
```
kshyju, line 67: Wonder we should have this as a property of FunctionContext. Instead, what If we allow customers to just use a parameter of the type `CancellationToken`  in their functions? If they do not use this parameter, we do not need to keep an entry in the dictionary (which tracks invocation to CTS). That also makes the programming experience consistent with the [in-proc model](https://docs.microsoft.com/en-us/azure/azure-functions/functions-dotnet-class-library?tabs=v2%2Ccmd#cancellation-tokens)

kshyju: I hope this should not block your host changes though Lilian.

liliankasem: I actually had this in mind when I first started out as I wanted it to be like the in-proc example, but I just didn't know how to expose a new parameter like this for the users - if this is something you know how to do, would you mind walking be through it? Or you can branch off this branch and make those changes, then we can walk through it together. 

Technically doesn't block host changes since I have validated that we are hitting the Cancellation handler, and how we deal with that in the worker is totally up to us and shouldn't impact the host. But it is probably a good idea to validate E2E with host if we make those changes

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -248,6 +262,16 @@ private async Task ProcessRequestCoreAsync(StreamingMessage request)
             return response;
         }
 
+        internal static void InvocationCancelRequestHandler(InvocationCancel request)
+        {
+            _invocationCancelDictionary.TryGetValue(request.InvocationId, out CancellationTokenSource? cts);
+            if (cts is not null)
+            {
+                cts.Cancel();
```
liliankasem, line 270: Would I also need to dispose the cts here?

src/DotNetWorker.Core/Context/DefaultFunctionContext.cs
liliankasem: Was able to figure his out and removed the cancellation feature

src/DotNetWorker.Core/Context/FunctionContext.cs
liliankasem: Had a sync with Shyju and updated the code to use a CancellationToken parameter

src/DotNetWorker.Core/Converters/CancellationTokenConverter.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Microsoft.Azure.Functions.Worker.Converters
+{
+    internal class CancellationTokenConverter : IInputConverter
+    {
+        public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
+        {
+            if (context.TargetType == typeof(CancellationToken))
+            {
+                return new ValueTask<ConversionResult>(ConversionResult.Success(context.FunctionContext.CancellationToken));
```
liliankasem, line 15: We cannot get the cancellation token from the invocation, so we need to flow it through to the function context as it's public

kshyju: May be we should update the if to support `CancellationToken?` type as well. Customer can add `CancellationToken? token` param to the function. See the GuidConverter.

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -174,15 +179,19 @@ private async Task ProcessRequestCoreAsync(StreamingMessage request)
             IOutputBindingsInfoProvider outputBindingsInfoProvider,
             IInputConversionFeatureProvider functionInputConversionFeatureProvider)
         {
+            var cts = new CancellationTokenSource();
```
kshyju, line 182: Are we always creating a `CancellationTokenSource` instance and adding to dictionary for every invocation. I wonder we should conditionally do this (check the function definition / Input bindings of the function being invoked, to determine Cancellation token is a parameter/input binding for this function)

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -234,6 +240,12 @@ private async Task ProcessRequestCoreAsync(StreamingMessage request)
                     Exception = ex.ToRpcException(),
                     Status = StatusResult.Types.Status.Failure
                 };
+
+                if (ex.InnerException is TaskCanceledException)
+                {
+                    Console.WriteLine("TaskCanceledException withing handle invocation request");
+                    response.Result.Status = StatusResult.Types.Status.Cancelled;
```
kshyju, line 247: How is host handling this Cancelled status? I wonder we should just send Failure as the status here. What if customer function has some code which is making an http call using HttpClient/ Some Azure SDK to talk to Azure resources and a TaskCanceledException was thrown from that. The function failed in that case. Should we not report that as a failure ?

liliankasem: discussed offline

src/DotNetWorker.Core/Context/DefaultFunctionContext.cs
```diff
@@ -12,13 +13,15 @@ internal sealed class DefaultFunctionContext : FunctionContext, IAsyncDisposable
...
         private IServiceScope? _instanceServicesScope;
         private IServiceProvider? _instanceServices;
         private BindingContext? _bindingContext;
 
-        public DefaultFunctionContext(IServiceScopeFactory serviceScopeFactory, IInvocationFeatures features)
+        public DefaultFunctionContext(IServiceScopeFactory serviceScopeFactory, IInvocationFeatures features, CancellationToken cancellationToken)
         {
+            _cancellationToken = cancellationToken;
             _serviceScopeFactory = serviceScopeFactory ?? throw new ArgumentNullException(nameof(serviceScopeFactory));
             Features = features ?? throw new ArgumentNullException(nameof(features));
```
brettsam, line 27: We've established a pattern of pulling properties directly from `IInvocationFeatures`... should se pull `CancellationToken` from there a well and not have it passed in directly?

src/DotNetWorker.Core/FunctionsApplication.cs
```diff
@@ -17,28 +20,52 @@ internal partial class FunctionsApplication : IFunctionsApplication
...
         public FunctionContext CreateContext(IInvocationFeatures features)
         {
+            CancellationTokenSource? cancellationTokenSource = null;
             var invocation = features.Get<FunctionInvocation>() ?? throw new InvalidOperationException($"The {nameof(FunctionInvocation)} feature is required.");
 
             var functionDefinition = _functionMap[invocation.FunctionId];
             features.Set<FunctionDefinition>(functionDefinition);
 
-            return _functionContextFactory.Create(features);
+            if (functionDefinition.Parameters.Where(p => p.Type == typeof(CancellationToken)).Any())
```
brettsam, line 52: I don't love that we have to loop through the Parameters every time. Is there a way we can cache whether or not we need to create a CTS for this function? It's not going to change, and this happens with every invocation.

src/DotNetWorker.Core/FunctionsApplication.cs
```diff
@@ -64,5 +91,19 @@ public Task InvokeFunctionAsync(FunctionContext context)
                 return _functionExecutionDelegate(context);
             }
         }
+
+        public void CancelInvocation(string invocationId)
+        {
+            var invocationDetails = _functionInvocationManager.TryGetInvocationDetails(invocationId);
+            if (invocationDetails?.CancellationTokenSource is not null)
+            {
+                invocationDetails.CancellationTokenSource.Cancel();
```
brettsam, line 100: I think you should do a `TryCancel` here... if there's a race it won't throw. I *think* `Cancel` will throw if it's already canceled.

brettsam (commented):
Some initial questions...

src/DotNetWorker.Core/FunctionsApplication.cs
liliankasem: Great point, will look into caching that information

src/DotNetWorker.Core/Context/DefaultFunctionContext.cs
liliankasem: I'm not super familiar with `IInvocationFeatures`, but I do remember @fabiocav mentioning that we shouldn't use features for this - is this the same thing? If the correct pattern is for us to use a feature for this then happy to do that

src/DotNetWorker.Core/FunctionsApplication.cs
liliankasem: Looks like there's no `TryCancel` or equivalent on CTS. I can wrap this in a try/catch unless there's a better approach here

src/DotNetWorker.Core/FunctionsApplication.cs
```diff
@@ -64,5 +91,27 @@ public Task InvokeFunctionAsync(FunctionContext context)
...
+            var invocationDetails = _functionInvocationManager.TryGetInvocationDetails(invocationId);
+            if (invocationDetails?.CancellationTokenSource is not null)
+            {
+                try
+                {
+                    invocationDetails.CancellationTokenSource.Cancel();
+                }
+                catch (Exception)
+                {
+                    // what should we do here?
```
liliankasem, line 106: Thoughts on if we need to handle this here i.e. if the cancellation token source has already been disposed

src/DotNetWorker.Core/FunctionsApplication.cs
liliankasem: Using lazy init in the function definition to cache if the function is using a cancellation token or not - does this work or is there a better way to do this?

src/DotNetWorker.Core/Hosting/ServiceCollectionExtensions.cs
```diff
@@ -50,6 +50,9 @@ public static IFunctionsWorkerApplicationBuilder AddFunctionsWorkerCore(this ISe
             // Function Execution Contexts
             services.AddSingleton<IFunctionContextFactory, DefaultFunctionContextFactory>();
 
+            // Function Invocation Manager
+            services.AddSingleton<FunctionInvocationManager>();
```
satvu, line 54: nit: The other singletons (see above in this file) follow a pattern of `services.AddSingleton<Interface, DefaultImplementation>();` 

Is there a reason for the deviation?

liliankasem: I didn't create an interface to use for `FunctionInvocationManager`, I can create one for sure but wasn't sure yet if we needed one

test/DotNetWorkerTests/TestFunctionInvocation.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Threading;
```
liliankasem, line 5: todo: remove

src/DotNetWorker.Core/Invocation/IFunctionInvocationManager.cs
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.Functions.Worker.Invocation
+{
+    /// <summary>
+    /// Keeps track of inflight invocations.
+    /// </summary>
+    interface IFunctionInvocationManager
```
kshyju, line 9: naming nit: This type is not actually dealing with the invocation as such (executing the function invocation)

src/DotNetWorker.Core/Definition/FunctionDefinition.cs
```diff
@@ -44,5 +45,10 @@ public abstract class FunctionDefinition
         /// Gets the output binding metadata.
         /// </summary>
         public abstract IImmutableDictionary<string, BindingMetadata> OutputBindings { get; }
+
+        /// <summary>
+        /// Gets if a the function is binding to a cancellation token.
+        /// </summary>
+        public abstract bool BindsToCancellationToken { get; }
```
kshyju, line 52: Would this be a breaking change as this type is public ?

kshyju (commented):
Few more comments/questions.

src/DotNetWorker.Core/Invocation/IFunctionInvocationManager.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        /// <summary>
+        /// Attempts to store the specified key and value, where 'invocationId' is the key.
+        /// </summary>
+        /// <param name="invocationId">Invocation ID</param>
+        /// <param name="details"><see cref="FunctionInvocationDetails"/></param>
+        void TryAddInvocationDetails(string invocationId, FunctionInvocationDetails details);
```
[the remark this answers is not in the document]
liliankasem: Great suggestion, will do!

src/DotNetWorker.Core/Invocation/FunctionInvocationManager.cs
```diff
@@ -0,0 +1,48 @@
...
+        internal ConcurrentDictionary<string, FunctionInvocationDetails> _inflightInvocations;
+
+        public FunctionInvocationManager()
+        {
+          _inflightInvocations = new ConcurrentDictionary<string, FunctionInvocationDetails>();
+        }
+
+        public void TryAddInvocationDetails(string invocationId, FunctionInvocationDetails details)
+        {
+            if (string.IsNullOrEmpty(invocationId) || details is null)
```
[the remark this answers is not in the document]
liliankasem: Yup, happy to throw here if it aligns with the rest of the APIs

src/DotNetWorker.Core/FunctionsApplication.cs
```diff
@@ -64,5 +90,27 @@ public Task InvokeFunctionAsync(FunctionContext context)
...
+            if (invocationDetails?.CancellationTokenSource is not null)
+            {
+                try
+                {
+                    invocationDetails.CancellationTokenSource.Cancel();
+                }
+                catch (Exception)
+                {
+                    // Do we need to do something other than log here?
+                    _logger.LogWarning($"Unable to cancel invocation {invocationId}, the token source may already be disposed.");
```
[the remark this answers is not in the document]
liliankasem: yeah I have the same concern here, I do worry about customer confusion - happy to remove this part if that helps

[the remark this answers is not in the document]
kshyju: May be test this scenario and observe how it is being handled here.
Inside the function code, 
 `cancellationToken.Register(() => DoSomething());`. 

`DoSomething` could throw an exception.

kshyju:
Rebase on main to fix the failing build. @satvu fixed it yesterday.

src/DotNetWorker.Core/Invocation/IFunctionInvocationManager.cs
liliankasem: Yup, that's fair - any ideas? It's mainly for keeping track of inflight invocations so I felt like invocation needed to be in the name 🤔

src/DotNetWorker.Core/Definition/FunctionDefinition.cs
liliankasem: Good question, I'm not sure! What's the best way to double check this?

kshyju: 1 1  A M P S T 🏰

src/DotNetWorker.Core/Invocation/IFunctionInvocationManager.cs
surgupta-msft: May be `FunctionInvocationDictionary` or `FunctionInvocationDetailsDictionary`? I have seen `IRpcWorkerChannelDictionary` in Host. Few other suffixes I have noticed in our repos are dict, map, resolver 😄

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -177,7 +178,8 @@ private async Task ProcessRequestCoreAsync(StreamingMessage request)
             FunctionContext? context = null;
             InvocationResponse response = new()
             {
-                InvocationId = request.InvocationId
+                InvocationId = request.InvocationId,
+                Result       = new StatusResult()
```
surgupta-msft, line 182: very nit: Not sure if there are extra spaces after Result or it is just rendered here like this.

src/DotNetWorker.Core/Definition/FunctionDefinition.cs
```diff
@@ -44,5 +45,10 @@ public abstract class FunctionDefinition
         /// Gets the output binding metadata.
         /// </summary>
         public abstract IImmutableDictionary<string, BindingMetadata> OutputBindings { get; }
+
+        /// <summary>
+        /// Gets if a the function is binding to a cancellation token.
```
madelinegordon, line 50: nit - comment wording: "returns true if"... "and false if not"

src/DotNetWorker.Core/Invocation/IFunctionInvocationManager.cs
liliankasem: I like the sound of that, thanks for the suggestion!

src/DotNetWorker.Core/Definition/FunctionDefinition.cs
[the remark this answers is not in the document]
liliankasem: done!

src/DotNetWorker.Core/Invocation/FunctionInvocationManager.cs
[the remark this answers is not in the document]
liliankasem: Actually on second thought, I'm not sure we want to do that - I don't think its critical enough that we need to throw? Unless we want to handle from the calling method to catch and log that we cannot store an invocation. But I can't imagine ever calling this without a real invocation ID - just more of a safety net

src/DotNetWorker.Core/Invocation/FunctionInvocationManager.cs
```diff
@@ -0,0 +1,48 @@
...
+    internal class FunctionInvocationManager : IFunctionInvocationManager
+    {
+        internal ConcurrentDictionary<string, FunctionInvocationDetails> _inflightInvocations;
+
+        public FunctionInvocationManager()
+        {
+          _inflightInvocations = new ConcurrentDictionary<string, FunctionInvocationDetails>();
+        }
+
+        public void TryAddInvocationDetails(string invocationId, FunctionInvocationDetails details)
```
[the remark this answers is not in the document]
liliankasem: Hmmm I don't think it should be, the whole point it to track this information - maybe the other instances need to not be nullable

src/DotNetWorker.Core/FunctionsApplication.cs
```diff
@@ -17,28 +19,52 @@ internal partial class FunctionsApplication : IFunctionsApplication
...
-            return _functionContextFactory.Create(features);
+            if (functionDefinition.BindsToCancellationToken)
+            {
+                cancellationTokenSource = new CancellationTokenSource();
+            }
+
+            var token = cancellationTokenSource is null ? CancellationToken.None : cancellationTokenSource.Token;
+            var context = _functionContextFactory.Create(features, token);
+
+            var invocationDetails = new FunctionInvocationDetails()
```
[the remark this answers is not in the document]
liliankasem: I believe that the context needs to have the token so that we can expose it as a parameter via the CancellationTokenConverter. FunctionInvocationDetails keeps track of the tsc that we use to cancel the cancellation token and we don't expose that to customers

src/DotNetWorker.Grpc/GrpcWorker.cs
liliankasem: extra spaces are intentional to line up the `=` - easier to read :D

src/DotNetWorker.Core/Invocation/FunctionInvocationManager.cs
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Concurrent;
+
+namespace Microsoft.Azure.Functions.Worker.Invocation
+{
+    internal class FunctionInvocationManager : IFunctionInvocationManager
+    {
+        internal ConcurrentDictionary<string, FunctionInvocationDetails> _inflightInvocations;
```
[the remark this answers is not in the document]
liliankasem: Ah I remember why I did this - for testing - which is dumb, I can just mock the dictionary I imagine? Will update back to private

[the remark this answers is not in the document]
liliankasem: Actually now that return a bool, I can just assert that!

src/DotNetWorker.Core/Context/DefaultFunctionContext.cs
```diff
@@ -54,6 +57,8 @@ public override IServiceProvider InstanceServices
 
         public override TraceContext TraceContext => _invocation.TraceContext;
 
+        public override CancellationToken CancellationToken => _cancellationToken;
```
jviau, line 60: Small nit: consider a descriptive property name around what this cancellation token is cancelling exactly. Is it `InvocationCancelled`? Or is the whole worker being cancelled?

src/DotNetWorker.Core/Context/DefaultFunctionContext.cs
```diff
@@ -12,13 +13,15 @@ internal sealed class DefaultFunctionContext : FunctionContext, IAsyncDisposable
...
         private readonly IServiceScopeFactory _serviceScopeFactory;
         private readonly FunctionInvocation _invocation;
+        private readonly CancellationToken _cancellationToken;
 
         private IServiceScope? _instanceServicesScope;
         private IServiceProvider? _instanceServices;
         private BindingContext? _bindingContext;
 
-        public DefaultFunctionContext(IServiceScopeFactory serviceScopeFactory, IInvocationFeatures features)
+        public DefaultFunctionContext(IServiceScopeFactory serviceScopeFactory, IInvocationFeatures features, CancellationToken cancellationToken)
```
jviau, line 22: I like to always have `= default` for all cancellation arguments. It is nice to omit it when you don't care about cancellation, IE: unit tests.

src/DotNetWorker.Core/FunctionsApplication.cs
```diff
@@ -17,28 +19,52 @@ internal partial class FunctionsApplication : IFunctionsApplication
...
+            CancellationTokenSource? cancellationTokenSource = null;
             var invocation = features.Get<FunctionInvocation>() ?? throw new InvalidOperationException($"The {nameof(FunctionInvocation)} feature is required.");
 
             var functionDefinition = _functionMap[invocation.FunctionId];
             features.Set<FunctionDefinition>(functionDefinition);
 
-            return _functionContextFactory.Create(features);
+            if (functionDefinition.IsCancellable)
+            {
+                cancellationTokenSource = new CancellationTokenSource();
```
jviau, line 53: We need to ensure this CTS is disposed when we are done with it (they _can_ cause memory leaks if not disposed). I didn't notice any disposal calls for it (maybe I missed it?)

src/DotNetWorker.Core/FunctionsApplication.cs
```diff
@@ -17,28 +19,52 @@ internal partial class FunctionsApplication : IFunctionsApplication
...
+            var token = cancellationTokenSource is null ? CancellationToken.None : cancellationTokenSource.Token;
+            var context = _functionContextFactory.Create(features, token);
+
+            var invocationDetails = new FunctionInvocationDetails()
+                                        {
+                                            FunctionContext = context,
+                                            CancellationTokenSource = cancellationTokenSource,
+                                        };
+
+            _functionInvocationDictionary.TryAddInvocationDetails(context.InvocationId, invocationDetails);
```
jviau, line 65: In addition to the disposal comment above, make sure disposal is also done if the `TryAddInvocationDetails` fails.

src/DotNetWorker.Grpc/Definition/GrpcFunctionDefinition.cs
```diff
@@ -69,5 +74,14 @@ public GrpcFunctionDefinition(FunctionLoadRequest loadRequest, IMethodInfoLocato
 
             return ImmutableDictionary<string, object>.Empty;
         }
+
+        private bool ParametersContainCancellationToken(ImmutableArray<FunctionParameter> parameters)
+        {
+            if (parameters.Any(p => p.Type == typeof(CancellationToken) || p.Type == typeof(CancellationToken?)))
```
jviau, line 80: I think you can remove the if/else statement and just return the `.Any` result directly.

``` CSharp
return parameters.Any(p => p.Type == typeof(CancellationToken) || p.Type == typeof(CancellationToken?));
```

src/DotNetWorker.Core/FunctionsApplication.cs
```diff
@@ -64,5 +90,26 @@ public Task InvokeFunctionAsync(FunctionContext context)
                 return _functionExecutionDelegate(context);
             }
         }
+
+        public void CancelInvocation(string invocationId)
+        {
+            var invocationDetails = _functionInvocationDictionary.TryGetInvocationDetails(invocationId);
+            if (invocationDetails?.CancellationTokenSource is not null)
```
jviau, line 97: Just to avoid any potential race condition, consider assigning via the pattern matching:

``` CSharp
if (invocationDetails?.CancellationTokenSource is CancellationTokenSource cts)
{
    cts.Cancel(); // with try/catch
}
```

The race condition here is _technically_ the `invocationDetails.CancellationTokenSource` could become `null` between your null check and accessing (because it is a property access, you don't know if the implementation can switch to null between calls). Using pattern matching assignment will capture that reference to the `cts` variable.

src/DotNetWorker.Core/Invocation/FunctionInvocationManager.cs
```diff
@@ -0,0 +1,48 @@
...
+            }
+
+            return _inflightInvocations.TryAdd(invocationId, details);
+        }
+
+        public bool TryRemoveInvocationDetails(string invocationId)
+        {
+            if (string.IsNullOrEmpty(invocationId))
+            {
+              return false;
```
jviau, line 31: nit: indentation (have 2 spaces here, should be 4). Also the other if statements in this file.

src/DotNetWorker.Core/Invocation/FunctionInvocationManager.cs
```diff
@@ -0,0 +1,48 @@
...
+            }
+
+            return _inflightInvocations.TryRemove(invocationId, out FunctionInvocationDetails? details);
+        }
+
+        public FunctionInvocationDetails? TryGetInvocationDetails(string invocationId)
```
jviau, line 37: `TryGet*` patterns should follow the `bool` return and `out` param. I think `Get*` is fine here if you want to keep this signature, especially with explicit nullables the null potential of this return is obvious to consumers. Alternative name could be `Find*` if you feel `Get*` should be reserved for non-null returning ones.

src/DotNetWorker.Core/Invocation/FunctionInvocationManager.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            if (string.IsNullOrEmpty(invocationId))
```
jviau, line 39: Are there cases where we expect this to be called with null or empty string? If we don't have valid cases for that, I recommend throwing on bad inputs to adhere to the fail-fast principal.

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -160,6 +157,10 @@ private async Task ProcessRequestCoreAsync(StreamingMessage request)
                     Result = new StatusResult { Status = StatusResult.Types.Status.Success }
                 };
             }
+            else if (request.ContentCase == MsgType.InvocationCancel)
```
jviau, line 160: Potential switch statement refactoring? (feel free to ignore if you don't want to).

jviau (commented):
My general thought here is that the storage and lookup by ID pattern feels like a gRPC specific implementation detail. Would it be possible to move those concepts into the gRPC portion only? Can the worker keep track of the instance IDs and cancel-by-id pattern?

The core layer could have a smaller portion of this logic revolving mostly around just containing the CTS and supporting cancelling or accessing it. Can the CTS accessing be rolled into some internal version of `FunctionContext`? So the gRPC service can hold onto that for cancellation?

I say this because consider a different scenario in the future where the gRPC worker, or whatever alternative, has an individual (and cancelable) connection per function invocation: then it won't need to cancel by ID, it could just chain that connections cancelation to cancel the function invocation. I understand this is not the case now, but we already have a concept of Core layer separation. It would be nice to stay vigilant about preventing other layer concepts leaking into the core layer.

src/DotNetWorker.Core/Context/DefaultFunctionContext.cs
liliankasem: I believe the goal here is to have parity with the existing [cancellation parameter ](https://docs.microsoft.com/en-us/azure/azure-functions/functions-dotnet-class-library?tabs=v2%2Ccmd#cancellation-tokens) we have with in-proc. Since this is exposed to customers in the context, I think it makes more sense to keep the name they would expect. I think the documentation should be clear about it being a function invocation cancel

src/DotNetWorker.Grpc/GrpcWorker.cs
liliankasem: We made an [issue](https://github.com/Azure/azure-functions-dotnet-worker/issues/930) for this but I can just address is now since it keeps coming up :)

jviau: Was not aware of the separate issue. Someone is working on it in a PR already, probably best to leave it to that other PR then.

src/DotNetWorker.Core/FunctionsApplication.cs
liliankasem: Meant to dispose on `CancelInvocation` - thanks for catching!

src/DotNetWorker.Core/Context/DefaultFunctionContext.cs
jviau: I don't feel that strongly about this, so I am fine if you leave it as is. However, the concern I have is specifically about naming for property vs parameter cancellation token. Parameter (like in proc) the name of `cancellation` or `cancellationToken` is totally fine - it is obvious what the cancellation applies to: it applies to this call chain. A cancellation token as a property it is not as clear what it applies to. Which is were a more explicit name indicating what it applies to comes in handy.

src/DotNetWorker.Grpc/GrpcWorker.cs
liliankasem: It looks like that PR closed and was doing a major refactor so I think it might be better to just resolve that here

liliankasem:
> My general thought here is that the storage and lookup by ID pattern feels like a gRPC specific implementation detail. Would it be possible to move those concepts into the gRPC portion only? Can the worker keep track of the instance IDs and cancel-by-id pattern?
> 
> The core layer could have a smaller portion of this logic revolving mostly around just containing the CTS and supporting cancelling or accessing it. Can the CTS accessing be rolled into some internal version of `FunctionContext`? So the gRPC service can hold onto that for cancellation?
> 
> I say this because consider a different scenario in the future where the gRPC worker, or whatever alternative, has an individual (and cancelable) connection per function invocation: then it won't need to cancel by ID, it could just chain that connections cancelation to cancel the function invocation. I understand this is not the case now, but we already have a concept of Core layer separation. It would be nice to stay vigilant about preventing other layer concepts leaking into the core layer.

Thanks for the PR feedback! I've addressed all the comments below, but let me think on this one a little bit.

I believe @brettsam or @fabiocav has suggested moving the tracking of cts and cancellation to the be part of FunctionsApplication which lives in Core; and I think you're suggesting we move that logic down to Grpc instead and have FunctionsApplication consume that? Maybe through a factory? Whilst having its own cancellation logic that is not tied to the invocation ID?

src/DotNetWorker.Core/Context/FunctionContext.cs
kshyju, line 67: Should we provide a default value for this virtual property? 

`public virtual CancellationToken CancellationToken { get; } = CancellationToken.None;`

src/DotNetWorker.Core/FunctionsApplication.cs
```diff
@@ -17,28 +19,57 @@ internal partial class FunctionsApplication : IFunctionsApplication
...
             var functionDefinition = _functionMap[invocation.FunctionId];
             features.Set<FunctionDefinition>(functionDefinition);
 
-            return _functionContextFactory.Create(features);
+            if (functionDefinition.IsCancellable)
+            {
+                cancellationTokenSource = new CancellationTokenSource();
+            }
+
+            var token = cancellationTokenSource is null ? CancellationToken.None : cancellationTokenSource.Token;
```
kshyju, line 56: nit: `var token = cancellationTokenSource?.Token ?? CancellationToken.None;`

src/DotNetWorker.Core/Invocation/IFunctionInvocationDictionary.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <summary>
+        /// Attempts to store the specified key and value, where 'invocationId' is the key.
+        /// </summary>
+        /// <param name="invocationId">Invocation ID</param>
+        /// <param name="details"><see cref="FunctionInvocationDetails"/></param>
+        /// <returns><see cref="bool"/> representing operation success</returns>
+        bool TryAddInvocationDetails(string invocationId, FunctionInvocationDetails details);
+
+        /// <summary>
+        /// Attempts to remove the specified key.
```
kshyju, line 20: nit: _value associated with_

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -177,7 +179,8 @@ private async Task ProcessRequestCoreAsync(StreamingMessage request)
             FunctionContext? context = null;
             InvocationResponse response = new()
             {
-                InvocationId = request.InvocationId
+                InvocationId = request.InvocationId,
+                Result       = new StatusResult()
```
kshyju, line 183: Should we do this assignment here? I think the original idea to not do here was to avoid an extra allocation if the invocation causes an exception being thrown. In that case, this `StatusResult` instance won't be used as the catch block is creating a new instance.

src/DotNetWorker.Core/IFunctionsApplication.cs
```diff
@@ -12,5 +14,18 @@ internal interface IFunctionsApplication
...
+        /// </summary>
+        /// <param name="invocationId">Invocation ID</param>
+        void CancelInvocation(string invocationId);
+
+        /// <summary>
+        /// Removes a given invocation from the <see cref="FunctionInvocationDictionary"/>.
+        /// Should be called on completed invocations.
+        /// </summary>
+        /// <param name="invocationId">Invocation ID</param>
+        void RemoveInvocationRecord(string invocationId);
```
kshyju, line 29: naming nit: Wonder we should be more generic here instead of very specific about removing the record. Something like "Complete" ?
This comment could be totally not valid as well.

kshyju (commented):
Few more comments. Yet to take a look at the tests.

src/DotNetWorker.Grpc/GrpcWorker.cs
liliankasem: Ah whoops, meant to _not_ new up a new StatusResult in the catch but modify

src/DotNetWorker.Core/IFunctionsApplication.cs
liliankasem: Something like `CompleteInvocation`? Could we be doing something else in here so that we need to make this more generic?

jviau:
> > My general thought here is that the storage and lookup by ID pattern feels like a gRPC specific implementation detail. Would it be possible to move those concepts into the gRPC portion only? Can the worker keep track of the instance IDs and cancel-by-id pattern?
> > The core layer could have a smaller portion of this logic revolving mostly around just containing the CTS and supporting cancelling or accessing it. Can the CTS accessing be rolled into some internal version of `FunctionContext`? So the gRPC service can hold onto that for cancellation?
> > I say this because consider a different scenario in the future where the gRPC worker, or whatever alternative, has an individual (and cancelable) connection per function invocation: then it won't need to cancel by ID, it could just chain that connections cancelation to cancel the function invocation. I understand this is not the case now, but we already have a concept of Core layer separation. It would be nice to stay vigilant about preventing other layer concepts leaking into the core layer.
> 
> Thanks for the PR feedback! I've addressed all the comments below, but let me think on this one a little bit.
> 
> I believe @brettsam or @fabiocav has suggested moving the tracking of cts and cancellation to the be part of FunctionsApplication which lives in Core; and I think you're suggesting we move that logic down to Grpc instead and have FunctionsApplication consume that? Maybe through a factory? Whilst having its own cancellation logic that is not tied to the invocation ID?

I _think_ this logic is all internal, so we can always go ahead with it as is and then do refactoring later as necessary. If something is public and would require breaking changes to refactor later on, then we should think it through more thoroughly here.

Anyways - my initial idea (with only a little bit of time looking at the code) is that we could refactor `FunctionsApplication` to return some object `FunctionInvocation` instead of `FunctionContext` - this `FunctionInvocation` could have methods like `InvokeAsync()` and `Cancel()` on it - it would basically be a container for all the properties and logic you want for a single function invocation (would contain the context and invocationId properties). Then the logic of invocation and cancellation would be up to the worker. Now the gRPC worker would have a concurrent dictionary of instanceId to `FunctionInvocation` for this cancellation by id logic.

src/DotNetWorker.Core/IFunctionsApplication.cs
jviau: My other comment mentions an idea to remove the need for this logic by shifting the responsibility to the caller. But anyways - for cases like this I like using `IDisposable`. So whatever "Start" method we have would return a disposable object, and the dispose would invoke the pairing `Complete` operation. This makes it easy to use via `using` statements and is also a pattern devs will be familiar with.

liliankasem:
> > > My general thought here is that the storage and lookup by ID pattern feels like a gRPC specific implementation detail. Would it be possible to move those concepts into the gRPC portion only? Can the worker keep track of the instance IDs and cancel-by-id pattern?
> > > The core layer could have a smaller portion of this logic revolving mostly around just containing the CTS and supporting cancelling or accessing it. Can the CTS accessing be rolled into some internal version of `FunctionContext`? So the gRPC service can hold onto that for cancellation?
> > > I say this because consider a different scenario in the future where the gRPC worker, or whatever alternative, has an individual (and cancelable) connection per function invocation: then it won't need to cancel by ID, it could just chain that connections cancelation to cancel the function invocation. I understand this is not the case now, but we already have a concept of Core layer separation. It would be nice to stay vigilant about preventing other layer concepts leaking into the core layer.
> > 
> > 
> > Thanks for the PR feedback! I've addressed all the comments below, but let me think on this one a little bit.
> > I believe @brettsam or @fabiocav has suggested moving the tracking of cts and cancellation to the be part of FunctionsApplication which lives in Core; and I think you're suggesting we move that logic down to Grpc instead and have FunctionsApplication consume that? Maybe through a factory? Whilst having its own cancellation logic that is not tied to the invocation ID?
> 
> I _think_ this logic is all internal, so we can always go ahead with it as is and then do refactoring later as necessary. If something is public and would require breaking changes to refactor later on, then we should think it through more thoroughly here.
> 
> Anyways - my initial idea (with only a little bit of time looking at the code) is that we could refactor `FunctionsApplication` to return some object `FunctionInvocation` instead of `FunctionContext` - this `FunctionInvocation` could have methods like `InvokeAsync()` and `Cancel()` on it - it would basically be a container for all the properties and logic you want for a single function invocation (would contain the context and invocationId properties). Then the logic of invocation and cancellation would be up to the worker. Now the gRPC worker would have a concurrent dictionary of instanceId to `FunctionInvocation` for this cancellation by id logic.

Gotcha, I really like this idea and think the separation of responsibility sounds clean to me! My only concern is my lack of knowledge/context on the worker - so definitely worth getting opinions from the team on this. Just to recap my understanding:

**Current** 

`FunctionsApplication`:

- Loads a function
- Creates a function invocation's context
    - (new) Creation of CTS and tracking invocation (by id)
- Invokes a function invocation
- (new) Cancels an invocation
- (new) Removes invocation from dictionary

**Proposed Refactor**

`FunctionsApplication`:

- Loads a function
- Creates a function invocation

`FunctionInvocation`:

- Invokes a function invocation (InvokeAsync)
- Cancels an invocation (Cancel)
- Removes invocation from dictionary (Complete)

FunctionInvocationDictionary

jviau:
@liliankasem yeah that is a good summary of the proposal. Once we start having the concept of this `FunctionInvocation` container we can include other logic in it as well:

1. Make it disposable, which handles disposing of the `CancellationTokenSource`.
2. Speaking of CTS, if we have an `InvokeAsync` on this object, should that take in a `CancellationToken` itself?
    - If it does, is this a complete replacement of the CTS held by the invocation? Or is it linked to the existing CTS?
    - This opens up many possible ways to implement cancellation from the gRPC worker. It doesn't even need to lookup the FunctionInvocation in its dictionary, it just needs its _own_ CTS which is tracked by ID.
    - This is just a very rough idea for you to consider. Not sure if it is a better or worse design.
3. General idea is once we have this `FunctionInvocation` object all interactions should be on that object for this given invocation - no more interactions with `FunctionsApplication`.

src/DotNetWorker.Core/Context/DefaultFunctionContext.cs
liliankasem: Ah gotcha, makes sense!

fabiocav:
> @liliankasem yeah that is a good summary of the proposal. Once we start having the concept of this `FunctionInvocation` container we can include other logic in it as well:
> 
> 1. Make it disposable, which handles disposing of the `CancellationTokenSource`.
> 2. Speaking of CTS, if we have an `InvokeAsync` on this object, should that take in a `CancellationToken` itself?
>    
>    * If it does, is this a complete replacement of the CTS held by the invocation? Or is it linked to the existing CTS?
>    * This opens up many possible ways to implement cancellation from the gRPC worker. It doesn't even need to lookup the FunctionInvocation in its dictionary, it just needs its _own_ CTS which is tracked by ID.
>    * This is just a very rough idea for you to consider. Not sure if it is a better or worse design.
> 3. Can expose the CTS directly as a property on `FunctionInvocation` instead of a `Cancel()` method.
>    
>    * Again, just ideas. Choose whichever approach you like.
> 4. General idea is once we have this `FunctionInvocation` object all interactions should be on that object for this given invocation - no more interactions with `FunctionsApplication`.

I like the overall idea, but we need to flush out the design and ensure this will flow well. We'd also need to come up with the right naming, as things like `FunctionInvocation`, for example, already exist. 

Although I'm not convinced the need here is unique to gRPC, I agree this logic, as currently implemented, shouldn't live in Core, but I think we can make some smaller changes in the current PR to refactor the code a bit and move the logic where it belongs to avoid significantly increasing the scope of what this PR is trying to accomplish, we can then follow up with more significant changes, if needed. 

I have a sync with @liliankasem tomorrow (would be happy to expand) and will share thoughts the changes, which would essentially result in removal of the new APIs added to `FunctionsApplication` (as they would no longer be needed), removing some of the concerns with that, while moving some of the mapping logic into the worker.

I do want to follow up on some of the ideas you've shared, though!

src/DotNetWorker.Core/Context/DefaultFunctionContext.cs
liliankasem: Had a sync with Fabio and we think we won't get much benefit introducing the cancellation token

src/DotNetWorker.Core/FunctionsApplication.cs
liliankasem: Spoke with Fabio and he mentioned that we should always be creating a cancellation token so this becomes irrelevant

src/DotNetWorker.Core/Definition/FunctionDefinition.cs
liliankasem: Removed this prop so no longer relevant

src/DotNetWorker.Core/FunctionsApplication.cs
```diff
@@ -17,28 +19,57 @@ internal partial class FunctionsApplication : IFunctionsApplication
...
         public FunctionContext CreateContext(IInvocationFeatures features)
         {
+            CancellationTokenSource? cancellationTokenSource = null;
             var invocation = features.Get<FunctionInvocation>() ?? throw new InvalidOperationException($"The {nameof(FunctionInvocation)} feature is required.");
 
             var functionDefinition = _functionMap[invocation.FunctionId];
             features.Set<FunctionDefinition>(functionDefinition);
 
-            return _functionContextFactory.Create(features);
+            if (functionDefinition.IsCancellable)
```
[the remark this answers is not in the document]
liliankasem: Removed check as discussed!

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
```diff
@@ -0,0 +1,154 @@
...
+            InvocationResponse response = new()
+            {
+                InvocationId = request.InvocationId,
+                Result       = new StatusResult()
+            };
+
+            try
+            {
+                var invocation = new GrpcFunctionInvocation(request);
+                var cancellationTokenSource = new CancellationTokenSource();
```
jviau, line 59: By the time this function (InvokeAsync) finishes, is it safe to say that the function invocation is completely done? If so, can we add a `using` statement in front of this line.

```suggestion
                using var cancellationTokenSource = new CancellationTokenSource();
```

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
```diff
@@ -0,0 +1,154 @@
...
+
+                if (!_inflightInvocations.TryAdd(request.InvocationId, cancellationTokenSource))
+                {
+                    // Dispose the CancellationTokenSource if we cannot keep track of it
+                    cancellationTokenSource?.Dispose();
```
jviau, line 64: The code as is will continue invocation of this function if we cannot track this token. This will lead to a couple of issues:

1. The `CancellationTokenSource` is now disposed, so accessing the `.Token` below will throw.
2. Even if we solve the object disposal issue, what will the expected behavior for a gRPC cancel message now? That CTS will not be correctly tracked in the dictionary. So we will potentially cancel the _wrong_ CTS and not cancel the correct one at all.

Do you see any valid reason why we should continue invocation if we fail to track the CTS? If so, how can we reconcile point 2 above? If we can't come up with a good solution to the above, it might be better to just treat this as an error case and fail the invocation. The alternative may cause some very obscure bugs.

jviau, line 64: Null operator `?.` is not necessary from what I can tell - you have new'ed it up above so it wont be null.

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
```diff
@@ -0,0 +1,154 @@
...
+            {
+                cancellationTokenSource?.Cancel();
+            }
+            catch (Exception ex)
+            {
+                _logger.LogWarning($"Unable to cancel invocation {invocationId}.", ex);
+            }
+            finally
+            {
+                cancellationTokenSource?.Dispose();
```
jviau, line 150: I wouldn't worry about disposing the token here. We can leave the disposal lifetime to be the sole responsibility of `InvokeAsync` above,

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
```diff
@@ -0,0 +1,154 @@
...
+            if (cancellationTokenSource is not CancellationTokenSource)
+            {
+                return;
+            }
+
+            try
+            {
+                cancellationTokenSource?.Cancel();
+            }
+            catch (Exception ex)
```
jviau, line 144: To reduce log noise, we can add a catch for `ObjectDisposedException` explicitly and skip logging there. Given the asynchronous nature of the `InvokeAsync` and cancellation messages, this will be an exception we encounter and should swallow as part of normal behavior.

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
```diff
@@ -0,0 +1,154 @@
...
+                    response.Result.Status = StatusResult.Types.Status.Cancelled;
+                }
+            }
+            finally
+            {
+                _inflightInvocations.TryRemove(request.InvocationId, out var cts);
+
+                if (context is IAsyncDisposable asyncContext)
+                {
+                    await asyncContext.DisposeAsync();
```
jviau, line 123: instead of these explicit disposal calls here, can we just add a `using` statement above to where the context is declared?

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
```diff
@@ -0,0 +1,154 @@
...
+                }
+
+                response.Result.Status = StatusResult.Types.Status.Success;
+            }
+            catch (Exception ex)
+            {
+                response.Result.Exception = ex.ToRpcException();
+                response.Result.Status = StatusResult.Types.Status.Failure;
+
+                if (ex.InnerException is TaskCanceledException)
```
jviau, line 112: Do we need to also check for `OperationCancelledException`?

```suggestion
                if (ex.InnerException is TaskCanceledException or OperationCancelledException)
```

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
```diff
@@ -0,0 +1,154 @@
...
+
+                    if (binding.Value is not null)
+                    {
+                        parameterBinding.Data = await binding.Value.ToRpcAsync(_serializer);
+                    }
+
+                    response.OutputData.Add(parameterBinding);
+                }
+
+                if (functionBindings.InvocationResult != null)
```
jviau, line 99: small nit for consistency with the other null checks:

```suggestion
                if (functionBindings.InvocationResult is not null)
```

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
```diff
@@ -0,0 +1,154 @@
...
+                    // Dispose the CancellationTokenSource if we cannot keep track of it
+                    cancellationTokenSource?.Dispose();
+                }
+
+                IInvocationFeatures invocationFeatures = _invocationFeaturesFactory.Create();
+                invocationFeatures.Set<FunctionInvocation>(invocation);
+                invocationFeatures.Set<IExecutionRetryFeature>(invocation);
+
+                var token = cancellationTokenSource?.Token ?? CancellationToken.None;
+                context = _application.CreateContext(invocationFeatures, token);
```
jviau, line 72: I may be missing the reason you have this variable declared above, but I _think_ we can declare it inline here (as well as dispose it):

```suggestion
                using var context = _application.CreateContext(invocationFeatures, token);
```

jviau (changes requested):
This is a great iteration! I like the simplicity of having the worker layer be responsible for providing the cancellation token, and core layer no longer tracks the CTS itself.

The only concerns I have remaining is around the logic handling a dictionary add failure for the CTS as well as ensuring the CTS is disposed. I have seen a couple memory leaks in the past due to undisposed CTS's.

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
liliankasem: Yeah this is an interesting one, the part where I'm conflicted is if a customer is not binding to a cancellation token but we fail because we can't save the CTS (that we may never use?); is it valid for us to fail an entire invocation based on that?

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
liliankasem: Just copying existing code here, happy to create an issue for cleaning up some of the old stuff after I finish the cancellation work

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
liliankasem: Same comment as below, will flag this to be addressed outside of this PR #1009

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
liliankasem: Yup, by the end of this method it's safe to assume the invocation is done

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
jviau: I think we should _always_ have a CTS for any invocation. Whether the end function binds that parameter or not, it should always be available to middleware to respond to. I believe this is what @fabiocav was mentioning.

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
```diff
@@ -0,0 +1,159 @@
...
+            }
+
+            return response;
+        }
+
+        public bool TryCancel(string invocationId)
+        {
+            _inflightInvocations.TryGetValue(invocationId, out var cancellationTokenSource);
+
+            if (cancellationTokenSource is CancellationTokenSource)
```
jviau, line 139: not blocking nit: it is nice to see defensive coding, but this can be simplified since you know you never add a `null` CTS to the dictionary:

``` CSharp
if (_inflightInvocations.TryGetValue(invocationId, out var cancellationTokenSource))
{
    try
    {
        cancellationTokenSource.Cancel();
        return true;
    }
    catch (ObjectDisposedException)
    {
    }
    catch (Exception ex)
    {
        _logger.LogWarning(ex, "Unable to cancel invocation {invocationId}.", invocationId);
    }
}

return false;
```

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
```diff
@@ -0,0 +1,159 @@
...
+                    cancellationTokenSource?.Cancel();
+                    return true;
+                }
+                catch (ObjectDisposedException)
+                {
+                    // Do nothing, normal behavior
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogWarning($"Unable to cancel invocation {invocationId}.", ex);
```
jviau, line 152: With `ILogger`, pass exception as the first argument (exceptions have special logic) and use message templates (_not_ string interpolation). The logger will actually throw here because you are supplying 1 template value, but have 0 holes to fill.
```suggestion
                    _logger.LogWarning(ex, "Unable to cancel invocation {invocationId}.", invocationId);
```

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
```diff
@@ -0,0 +1,159 @@
...
+                try
+                {
+                    cancellationTokenSource?.Cancel();
+                    return true;
+                }
+                catch (ObjectDisposedException)
+                {
+                    // Do nothing, normal behavior
+                }
+                catch (Exception ex)
```
jviau, line 150: Analyzers like to complain about broad exception catches. I like to add a `when (!ex.IsFatal())` where `IsFatal` is defined as:

``` CSharp
    /// <summary>
    /// Extensions for exceptions.
    /// </summary>
    public static class ExceptionExtensions
    {
        /// <summary>
        /// Determines if an exception is fatal, and should not be caught.
        /// </summary>
        /// <param name="exception">The exception to catch.</param>
        /// <returns>True if fatal, false otherwise (or if null).</returns>
        public static bool IsFatal(this Exception exception)
        {
            while (exception is not null)
            {
                if (exception.IsFatalCore())
                {
                    return true;
                }

                if (exception is AggregateException aggregate)
                {
                    // AggregateExceptions have a collection of inner exceptions, which may themselves be other
                    // wrapping exceptions (including nested AggregateExceptions). Recursively walk this
                    // hierarchy. The (singular) InnerException is included in the collection.
                    foreach (Exception innerException in aggregate.InnerExceptions)
                    {
                        if (innerException.IsFatal())
                        {
                            return true;
                        }
                    }

                    break;
                }
                else
                {
                    exception = exception.InnerException;
                }
            }

            return false;
        }

        private static bool IsFatalCore(this Exception exception)
        {
            return exception is OutOfMemoryException ||
                   exception is StackOverflowException || // can't actually be caught
                   exception is NullReferenceException || // this is a underlying code issue, we want to surface these.
                   exception is SEHException ||
                   exception is AccessViolationException || // can be caught, but means corrupted state.
                   exception is ThreadAbortException; // can be caught, but will always be re-thrown.
        }
    }
```

jviau (changes requested):
Only potentially impactful issue left is the logger message.

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
liliankasem: Good to know, will keep this in mind :D

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
liliankasem: We do this [in the host](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Extensions/ExceptionExtensions.cs), but looks like we don't have something similar in the worker repo.

I think probably worth doing this in a separate PR #1016

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
jviau: We may also want to consider high performance logging - but that can be done in another PR.

Essentially, logging like this requires the message template to be parsed each time. High performance lets us pre-define these log messages that are generated at compile time.

https://docs.microsoft.com/en-us/dotnet/core/extensions/logger-message-generator

jviau (approved):

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
liliankasem: Ooo very cool, that'd definitely sounds like an improvement worth making! Feel free to create an issue for this and we can get it triaged in the next backlog grooming meeting

fabiocav (approved):
:shipit:

liliankasem:
> Any idea when this will be available in the lib?

Planning to cut a release some time next week

PR #1000, opened 2022-08-23
Fix Retry Attribute Test

satvu:

Resolves e2e test that checks the generation of functions.metadata. 

### Pull request checklist

### DIFF
--- a/sdk/Sdk/Abstractions/SdkRetryOptions.cs
+++ b/sdk/Sdk/Abstractions/SdkRetryOptions.cs
@@ -5,23 +5,14 @@ namespace Microsoft.Azure.Functions.Worker.Sdk
 {
     internal class SdkRetryOptions
     {
-        public SdkRetryOptions(string strategy, int maxRetryCount, string? delayInterval, string? minimumInterval, string? maximumInterval)
-        {
-            Strategy = strategy;
-            MaxRetryCount = maxRetryCount;
-            DelayInterval = delayInterval;
-            MinimumInterval = minimumInterval;
-            MaximumInterval = maximumInterval;
-        }
+        public string? Strategy { get; set; }
 
-        public string Strategy { get; }
+        public int MaxRetryCount { get; set; }
 
-        public int MaxRetryCount { get; }
+        public string? DelayInterval { get; set; }
 
-        public string? DelayInterval { get; }
+        public string? MinimumInterval { get; set; }
 
-        public string? MinimumInterval { get; }
-
-        public string? MaximumInterval { get; }
+        public string? MaximumInterval { get; set; }
     }
 }
--- a/test/SdkE2ETests/Contents/functions.metadata
+++ b/test/SdkE2ETests/Contents/functions.metadata
@@ -64,7 +64,13 @@
         "type": "http",
         "direction": "Out"
       }
-    ]
+    ],
+    "retry": {
+      "strategy": "exponentialBackoff",
+      "maxRetryCount": 5,
+      "minimumInterval": "00:00:04",
+      "maximumInterval": "00:15:00"
+    }
   },
   {
     "name": "HttpTriggerWithDependencyInjection",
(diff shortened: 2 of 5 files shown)


test/SdkE2ETests/Contents/functions.metadata
```diff
@@ -64,7 +64,13 @@
         "type": "http",
         "direction": "Out"
       }
-    ]
+    ],
+    "retry": {
```
satvu, line 68: Pipeline was failing because of this test - had to add retry information to the JSON file.

sdk/Sdk/Abstractions/SdkRetryOptions.cs
```diff
@@ -5,23 +5,14 @@ namespace Microsoft.Azure.Functions.Worker.Sdk
...
     {
-        public SdkRetryOptions(string strategy, int maxRetryCount, string? delayInterval, string? minimumInterval, string? maximumInterval)
-        {
-            Strategy = strategy;
-            MaxRetryCount = maxRetryCount;
-            DelayInterval = delayInterval;
-            MinimumInterval = minimumInterval;
-            MaximumInterval = maximumInterval;
-        }
+        public string? Strategy { get; set; }
```
satvu, line 8: A refactor to match [SdkFunctionMetadata](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/sdk/Sdk/Abstractions/SdkFunctionMetadata.cs). In previous design there were empty retry options unexpectedly in some tests I ran (both unit and manual e2e) trying to fix the failing `functions.metadata` generation file. I believe this is the correct pattern - it matches official docs for json serialization and is passing all the tests.

kshyju:
nit: You could replace hardcoded param name with `nameof(delayInterval)`  [here](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/extensions/Worker.Extensions.Abstractions/src/FixedDelayRetryAttribute.cs#L40)?

sdk/Sdk/Abstractions/SdkRetryOptions.cs
```diff
@@ -5,23 +5,14 @@ namespace Microsoft.Azure.Functions.Worker.Sdk
...
 
-        public string Strategy { get; }
+        public int MaxRetryCount { get; set; }
```
kshyju, line 10: My OCD for keeping every possible type as immutable is kicking in.

kshyju (approved):

sdk/Sdk/Abstractions/SdkRetryOptions.cs
fabiocav: @satvu can you clarify why we're changing this here?

test/SdkE2ETests/Contents/functions.metadata
fabiocav: Will look at the failure, but is the failure caused by new code expecting this to be present? can we enhance that code to make sure it handles scenarios where it isn't?

satvu: This test runs the SDK's `FunctionMetadataGenerator` against the sample app then compares the generated `functions.metadata` file to this hard copy I've edited. 

The failure was caused by my change in the Sample App where I added RetryAttributes to two of the sample functions as an example but didn't change this file.

sdk/Sdk/Abstractions/SdkRetryOptions.cs
satvu: For JSON serialization, as far as I understand, STJ expects public properties (that you can get and set) and will only serialize those given the options I see present in `FunctionMetadataJsonWriter`. I also thought immutability was preferable, but I am now understanding why `SdkFunctionMetadata` is also written like this with get and set. Converters are another option, but we already have a pattern established by `SdkFunctionMetadata` and defined SerializationOptions in the writer that work.

Helpful documentation example [here](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-how-to?pivots=dotnet-5-0#serialization-example) to see what gets serialized and what's ignored, and also looked into properties vs fields and how STJ handles them in serialization.

fabiocav (approved):

kshyju:
@satvu You are the author of the 1st PR and 1000th PR on this repo.  🎉 🔥

PR #1006, opened 2022-08-26
Migrate ExtensionStartupRunnerGenerator to IIncrementalGenerator 

### LINKED ISSUE #1007: Migrate ExtensionStartupRunnerGenerator from ISourceGenerator to IIncrementalGenerator
IIncrementalGenerator offers better performance with value provider caching.

kshyju:
Current version of ExtensionStartupRunnerGenerator [was created](https://github.com/Azure/azure-functions-dotnet-worker/pull/830) as an implementation of ISourceGenerator. In this PR, we are migrating that to [IIncrementalGenerator](https://github.com/dotnet/roslyn/blob/main/docs/features/incremental-generators.md), which offers better performance characteristics.

resolves #1007 

### Pull request checklist


sdk/Sdk.Generators/Sdk.Generators.csproj
```diff
@@ -19,7 +19,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.11.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.2.0" />
```
kshyju, line 22: The [IIncrementalGenerator  ](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.iincrementalgenerator?view=roslyn-dotnet-4.2.0)type is available from 4.X version of CodeAnalysis package.

sdk/Sdk.Analyzers/Sdk.Analyzers.csproj
```diff
@@ -18,11 +18,11 @@
   <Import Project="..\..\build\Common.props" />
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.2">
+    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.3">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.11.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.2.0" />
```
kshyju, line 25: CodeAnalysis package was upgraded to 4.X version to use IIncrementalGenerator type. Consolidated the versions in this project as well.

test/Sdk.Generator.Tests/RoslynTestUtils.cs
```diff
@@ -0,0 +1,117 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis;
+using Xunit;
+
+namespace Microsoft.Azure.Functions.SdkGeneratorTests
+{
+    /// <summary>
+    /// This is a trimmed down version of the TestUtils from the dotnet runtime repo.
```
kshyju, line 18: This is a trimmed version of the test util class from dotnet runtime repo. We shall augment with more methods from the origin source as needed.

sdk/release_notes.md
```diff
@@ -2,7 +2,4 @@
 <!-- Please add your release notes in the following format:
 - My change description (#PR/#issue)
 -->
-- Error handling for in-proc SDK usage inside isolated app. (#824)
```
kshyju, line 5: We are yet to release the non preview version of SDK with these changes. Can get this info from source control when we do that. In my PR,  am bumping up the package version again to preview version.

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
```diff
@@ -53,21 +54,43 @@ public class ExtensionStartupRunnerGenerator : ISourceGenerator
...
         /// </summary>
         private const string StartupBaseClassName = "Microsoft.Azure.Functions.Worker.Core.WorkerExtensionStartup";
 
-        public void Execute(GeneratorExecutionContext context)
+        public void Initialize(IncrementalGeneratorInitializationContext context)
         {
-            var extensionStartupTypeNames = GetExtensionStartupTypes(context);
+            // Get the sub set of assemblies with startup attribute
+            var assembliesValueProvider = context.CompilationProvider
+                .SelectMany((compilation, token) =>
```
jviau, line 69: I see examples of incremental generators declaring their lambdas with `static`. Is that something we should be doing as well? From my understanding, it will enforce no capturing of execution context (so a perf safeguard).

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
```diff
@@ -214,8 +204,66 @@ private static void WriteStartupCodeExecutorClass(IndentedTextWriter textWriter,
...
+        }
+
+        private static AttributeData? GetStartupAttributeData(IAssemblySymbol assemblySymbol)
+        {
+            var extensionStartupAttribute = assemblySymbol.GetAttributes()
+                .FirstOrDefault(a =>
+                    (a.AttributeClass?.Name.Equals(AttributeTypeName,
+                        StringComparison.Ordinal) ?? false) &&
+                    //Call GetFullName only if class name matches.
+                    a.AttributeClass.GetFullName()
```
jviau, line 221: I think the best practice here is to use the `Compilation` object to get the `INamedTypedSymbol` we want to compare against via name, and then we can compare the symbols directly. See the logger message example [here](https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Logging.Abstractions/gen/LoggerMessageGenerator.Parser.cs#L39) And you can see how that instance is used for comparison in that class.

jviau (commented):
A couple things:

- Just want to confirm we have no intention of supporting .NET5 with this project? If we do, I _think_ we need to keep the old `ISourceGneerator` around as well. But I am fairly certain we only support NET6 and up here.
- I really like how the [logger message generator](https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Logging.Abstractions/gen/LoggerMessageGenerator.Roslyn4.0.cs) is organized into a parser and emitter sub class. If you have time or desire to refactor towards something like that, it would be nice to see.

kshyju:
> A couple things:
> 
> * Just want to confirm we have no intention of supporting .NET5 with this project? If we do, I _think_ we need to keep the old `ISourceGneerator` around as well. But I am fairly certain we only support NET6 and up here.
> * I really like how the [logger message generator](https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Logging.Abstractions/gen/LoggerMessageGenerator.Roslyn4.0.cs) is organized into a parser and emitter sub class. If you have time or desire to refactor towards something like that, it would be nice to see.

Yea, I like splitting into parser and emitter approach. Let me handle that in the next iteration.

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
```diff
@@ -41,181 +38,60 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
...
-            {
-                return;
-            }
+            IncrementalValueProvider<(Compilation Compilation, ImmutableArray<IAssemblySymbol> AssemblySymbols)>
+                compilationWithAssemblies
+                    = context.CompilationProvider.Combine(assembliesValueProvider.Collect());
 
-            SourceText sourceText;
-            using (var stringWriter = new StringWriter())
-            using (var indentedTextWriter = new IndentedTextWriter(stringWriter))
```
kshyju, line 74: Moved code from this class to the Emitter and Parser classes

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
kshyju: Done.

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
kshyju: Good suggestion. Switched to this approach.

sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.Parser.cs
```diff
@@ -0,0 +1,134 @@
...
+                }
+                
+                return new StartupTypeInfo(namedTypeSymbol.ToDisplayString(), builder.ToImmutable());
+            }
+        }
+
+        /// <summary>
+        /// A type to represent the data about the extension startup types we want to generate code for.
+        /// </summary>
+        internal struct StartupTypeInfo
```
jviau, line 114: I think you can mark this struct `readonly`. This will let the compiler avoid defensive copies (good for perf).

fabiocav:
Closing this and using #1369 to track the work once we're working on 2.x

PR #1014, opened 2022-08-31
Fixes GetHttpResponseData to return the correct entry from OutputBinding entries

kshyju:
In #987 we changed how we read the OutputBinding data. But the `GetHttpResponseData` extension method was not updated to reflect this change. Fixing that here.

### DIFF
--- a/src/DotNetWorker.Core/Context/FunctionContextHttpRequestExtensions.cs
+++ b/src/DotNetWorker.Core/Context/FunctionContextHttpRequestExtensions.cs
@@ -12,6 +12,8 @@ namespace Microsoft.Azure.Functions.Worker
     /// </summary>
     public static class FunctionContextHttpRequestExtensions
     {
+        private const string HttpBindingType = "http";
+
         /// <summary>
         /// Gets the <see cref="HttpRequestData"/> instance if the invocation is for an http trigger.
         /// </summary>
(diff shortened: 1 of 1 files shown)


satvu (approved):

liliankasem (approved):

src/DotNetWorker.Core/Context/FunctionContextHttpRequestExtensions.cs
```diff
@@ -45,7 +47,7 @@ public static class FunctionContextHttpRequestExtensions
             }
 
             // see output binding entries have a property of type HttpResponseData;
-            var httpOutputBinding = context.GetOutputBindings<HttpResponseData>().FirstOrDefault();
+            var httpOutputBinding = context.GetOutputBindings<HttpResponseData>().First(a => a.BindingType == HttpBindingType);
```
jviau, line 50: `First` will throw if no item matches the predicate.
```suggestion
            var httpOutputBinding = context.GetOutputBindings<HttpResponseData>().FirstOrDefault(a => a.BindingType == HttpBindingType);
```

kshyju: No. It should be FirstOrDefault. Copy paste mistake. This change is not absorbed in Narens other PR as these 2 are PRs are waiting for a broken test fixed in the other PR. Changed to FirstOrDefault in that PR.

PR #1022, opened 2022-09-02
Populating WorkerMetadata in init response

### LINKED ISSUE #920: Populate WorkerMetadata in WorkerInitResponse message
The protobuf has been updated to include a `WorkerMetadata` property within the `WorkerInitResponse` message. The worker should include `WorkerMetadata` when sending the `WorkerInitResponse` to the host.

[PR](https://github.com/Azure/azure-functions-language-worker-protobuf/pull/76) for reference on how to fill in the fields.

You will need to pull in the latest protobuf version first (this [doc](https://dev.azure.com/msazure/One/_git/AAPT-Antares-Docs?path=/TeamDocs/FunctionTeamDocs/EngineeringProcesses/ReleaseProcesses/ProtobufReleaseProcess.md&version=GBmaster&_a=preview) may help).

```proto
message WorkerMetadata {
  // The runtime/stack name
  string runtime_name = 1;

  // The version of the runtime/stack
  string runtime_version = 2;

  // The version of the worker or worker SDK
  string worker_version = 3;

  // The worker bitness/architecture
  string worker_bitness = 4;

  // Optional additional custom properties
  map<string, string> custom_properties = 5;
}
```

fabiocav:

resolves #920 

### Pull request checklist

### DIFF
--- a/src/DotNetWorker.Core/WorkerInformation.cs
+++ b/src/DotNetWorker.Core/WorkerInformation.cs
@@ -16,6 +16,8 @@ namespace Microsoft.Azure.Functions.Worker
         private static readonly Assembly _thisAssembly = typeof(WorkerInformation).Assembly;
         private static readonly FileVersionInfo _fileVersionInfo = FileVersionInfo.GetVersionInfo(_thisAssembly.Location);
 
+        private string? _workerVersion;
+
         public static WorkerInformation Instance = new();
 
 #if NET5_0_OR_GREATER
@@ -28,7 +30,7 @@ namespace Microsoft.Azure.Functions.Worker
         public string RuntimeIdentifier => "n/a"; // Resolve in netstandard
 #endif
 
-        public string WorkerVersion => _thisAssembly.GetName().Version?.ToString()!;
+        public string WorkerVersion => _workerVersion ??= _thisAssembly.GetName().Version?.ToString()!;
 
 
         public string? ProductVersion => _fileVersionInfo.ProductVersion;
--- a/src/DotNetWorker.Grpc/GrpcWorker.cs
+++ b/src/DotNetWorker.Grpc/GrpcWorker.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Channels;
 using System.Threading.Tasks;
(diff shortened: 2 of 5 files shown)


src/DotNetWorker.Core/WorkerInformation.cs
```diff
@@ -16,6 +16,8 @@ internal sealed class WorkerInformation
         private static readonly Assembly _thisAssembly = typeof(WorkerInformation).Assembly;
         private static readonly FileVersionInfo _fileVersionInfo = FileVersionInfo.GetVersionInfo(_thisAssembly.Location);
 
+        private string _workerVersion;
```
kshyju, line 19: Mark it as nullable?

```
private string? _workerVersion
```

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -253,9 +254,18 @@ internal static WorkerInitResponse WorkerInitRequestHandler(WorkerInitRequest re
...
+                WorkerMetadata = new WorkerMetadata
+                {
+                    RuntimeName = RuntimeInformation.FrameworkDescription,
+                    RuntimeVersion = Environment.Version.ToString(),
+                    WorkerVersion = WorkerInformation.Instance.WorkerVersion,
+                    WorkerBitness = RuntimeInformation.ProcessArchitecture.ToString()
+                }
             };
 
+            response.WorkerMetadata.CustomProperties.Add("Grpc.Worker.Version", typeof(GrpcWorker).Assembly.GetName().Version?.ToString());
```
kshyju, line 267: The package name ends with `Worker.Grpc`. Should we use that(`Worker.Grpc.Version`) here instead of `Grpc.Worker.Version`?

kshyju: Should we also include Worker.Sdk version?

Do we care about the Worker.Core package version?

kshyju (approved):
One question about including the Sdk package version in the response

src/DotNetWorker.Grpc/GrpcWorker.cs
kshyju: Oh, I see Worker.Core version is in the `WorkerVersion` property.

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -253,9 +254,18 @@ internal static WorkerInitResponse WorkerInitRequestHandler(WorkerInitRequest re
             var response = new WorkerInitResponse
             {
                 Result = new StatusResult { Status = StatusResult.Types.Status.Success },
-                WorkerVersion = WorkerInformation.Instance.WorkerVersion
+                WorkerVersion = WorkerInformation.Instance.WorkerVersion,
+                WorkerMetadata = new WorkerMetadata
+                {
+                    RuntimeName = RuntimeInformation.FrameworkDescription,
```
kshyju, line 260: Looks like WorkerInformation instance exposes some of these props. Ex: `WorkerInformation.Instance.FrameworkDescription`. It also has an `OSArchitecture` property which returns `RuntimeInformation.OSArchitecture`

surgupta-msft (approved):

src/DotNetWorker.Grpc/GrpcWorker.cs
fabiocav: > Member

That's a good question. I don't really have a strong preference for this.

fabiocav: > Should we also include Worker.Sdk version?

Could be interesting, though, we don't really have any artifacts injected today. Might be a good thing to add to the target assembly and then we could update this code to use that information.

> Do we also care about getting the Worker.Core package version?

Saw your note below. Yes, this is already tracked.

src/DotNetWorker.Grpc/GrpcWorker.cs
fabiocav: Yes. All of those end up forwarding. For this, I don't think there's a lot of value in capturing the `OSArchitecture`, as that is already known, but we're capturing the `ProcessArchitecture`

PR #1026, opened 2022-09-08
Update fabricbot.json

liliankasem:


.github/fabricbot.json
```diff
@@ -1585,7 +1585,8 @@
         "fabiocav",
         "brettsam",
         "satvu",
-        "kshyju"
+        "kshyju",
+        "jviau"
       ],
       "modifiedBy": {
         "upn": "facaval@microsoft.com",
```
kshyju, line 1592: Should we also update the modifiedBy data?

kshyju (approved):

.github/fabricbot.json
liliankasem: I don't think it's relevant xD

satvu (approved):

PR #1039, opened 2022-09-15
[AppInsights] updating dependency creation

### LINKED ISSUE #998: Dependency not tracking correctly in Application Insights
@LenaVinogradov -- can you open a separate issue with exactly what you're seeing and what you expect to see? That'd help us narrow it down. Thanks!

_Originally posted by @brettsam in https://github.com/Azure/azure-functions-dotnet-worker/issues/760#issuecomment-1220962314_

I have used new preview package and the sample code from the quoted PR: 
"The preview "packages are now available: https://www.nuget.org/packages/Microsoft.Azure.Functions.Worker.ApplicationInsights.

For usage and explanation, see the PR here (official docs will come when this is out of preview): https://github.com/Azure/azure-functions-dotnet-worker/pull/944."

Here is a code of our program.cs:
```csharp
public class Program
{
	public static async Task Main()
	{
		var host = new HostBuilder()
			{
				builder.UseNewtonsoftJson();
				builder.AddApplicationInsights().AddApplicationInsightsLogger();
			})
			.ConfigureOpenApi()
			.ConfigureAppConfiguration(builder => Startup.ConfigureAppConfiguration(builder))
			.ConfigureServices(collection => Startup.Configure(collection))
			.Build();

		await host.RunAsync();
	}
}
```
With this package we can see dependencies in appinsights (we moved from Functions v3.0 a couple of month ago, so we are very unhappy to loose the telemetry functionality that we were heavily relying on for years), but this dependencies are not properly linked to parent. The have the correct operationId set, so we can see them together with the rest of telemetry objects for the same operation, but hierarchy is still lost as the parentId of the dependencies do not point to any object in appinsights.

brettsam:
Fixes #998 

Using App Insights `StartTelemetry` to create dependencies for proper hierarchy. This method does a bunch of work to check Activity trace ids and properly set up the operation and parent ids on telemetry. 

Original example of an isolated function calling into Table storage:
image

After this change (note how each entry is a child of the other):
image

### DIFF
--- a/src/DotNetWorker.ApplicationInsights/DotNetWorker.ApplicationInsights.csproj
+++ b/src/DotNetWorker.ApplicationInsights/DotNetWorker.ApplicationInsights.csproj
@@ -8,17 +8,14 @@
     <MajorProductVersion>1</MajorProductVersion>
     <MinorProductVersion>0</MinorProductVersion>
     <PatchProductVersion>0</PatchProductVersion>
-    <VersionSuffix>-preview1</VersionSuffix>
+    <VersionSuffix>-preview2</VersionSuffix>
   </PropertyGroup>
 
   <Import Project="..\..\build\Common.props" />
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.ApplicationInsights.WorkerService" Version="2.20.0" />
-  </ItemGroup>
-
-  <ItemGroup>
-    <ProjectReference Include="..\DotNetWorker.Core\DotNetWorker.Core.csproj" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Core" Version="1.7.0-preview1" />
+    <PackageReference Include="Microsoft.ApplicationInsights.WorkerService" Version="2.21.0" />
   </ItemGroup>
 
 </Project>
--- a/src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
+++ b/src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
@@ -23,15 +23,15 @@ namespace Microsoft.Azure.Functions.Worker.ApplicationInsights
                 ShouldListenTo = source => source.Name.StartsWith("Microsoft.Azure.Functions.Worker"),
                 ActivityStarted = activity =>
                 {
-                    var dependency = new DependencyTelemetry("Azure.Functions", activity.OperationName, activity.OperationName, null);
+                    var dependency = _telemetryClient.StartOperation<DependencyTelemetry>(activity);
+                    dependency.Telemetry.Type = "Azure.Functions";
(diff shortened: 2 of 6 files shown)


src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
```diff
@@ -23,15 +23,15 @@ public void Initialize(TelemetryConfiguration configuration)
...
+                    dependency.Telemetry.Type = "Azure.Functions";
                     activity.SetCustomProperty("_depTel", dependency);
-                    dependency.Start();
                 },
                 ActivityStopped = activity =>
                 {
-                    var dependency = activity.GetCustomProperty("_depTel") as DependencyTelemetry;
-                    dependency.Stop();
-                    _telemetryClient.TrackDependency(dependency);
+                    var dependency = activity.GetCustomProperty("_depTel") as IOperationHolder<DependencyTelemetry>;
```
liliankasem, line 32: can we use constants for all the strings here?

liliankasem (approved):

src/DotNetWorker.ApplicationInsights/DotNetWorker.ApplicationInsights.csproj
```diff
@@ -8,17 +8,14 @@
...
     <MinorProductVersion>0</MinorProductVersion>
     <PatchProductVersion>0</PatchProductVersion>
-    <VersionSuffix>-preview1</VersionSuffix>
+    <VersionSuffix>-preview2</VersionSuffix>
   </PropertyGroup>
 
   <Import Project="..\..\build\Common.props" />
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Core" Version="1.7.0-preview1" />
```
kshyju, line 17: Curious to know why we switched to package reference? Is it to lock this version of AI package to specific version of Worker.Core package?

brettsam: Yeah, there's a small tweak in preview1 that we need for this package to work... but there's no need for me to keep revving that one now that I don't need additional changes. It was mostly a local dev improvement to lock this to a published package.

This will eventually go away and back to a proj reference once this is ready for GA.

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
brettsam: Yes, I'll catch that in the next rev. I'm still not completely convinced this is the write way to store the telemetry throughout the Activity. I'll need to follow up with experts.

kshyju (approved):

jviau (approved):

PR #1069, opened 2022-09-27
[Samples] Add .NET 7 sample project

liliankasem:


samples/Cancellation/HttpTriggerWithCancellation.cs
```diff
@@ -0,0 +1,52 @@
...
+using Microsoft.Azure.Functions.Worker.Http;
+using Microsoft.Extensions.Logging;
+
+namespace Cancellation
+{
+    public class HttpTriggerWithCancellation
+    {
+        private readonly ILogger _logger;
+
+        public HttpTriggerWithCancellation(ILoggerFactory loggerFactory)
```
kshyju, line 18: Should we consider adding this function to the existing FunctionApp sample instead of new project ?

liliankasem: This sample is already in the FunctionApp project but could use an update

liliankasem: Yup, this sample already exists in the FunctionApp sample, and I think it's best to keep it there. Creating this as a draft in case @cloudmelon needs the cancellation sample to be it's own project

cloudmelon: I need a project sample with functions that we can just use 'func start' to test out the capabilities

samples/Cancellation/Cancellation.csproj
```diff
@@ -0,0 +1,21 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
```
cloudmelon, line 3: @liliankasem would you change this to .NET 7 ? that's why this is why the example will have to stand out from others.  thanks :)

samples/Cancellation/HttpTriggerWithCancellation.cs
kshyju: > I need a project sample with functions that we can just use 'func start' to test out the capabilities, can we sign-off this one ?

FWIWI, We should be able to do that in the [FunctionApp](https://github.com/Azure/azure-functions-dotnet-worker/tree/main/samples/FunctionApp) sample.

samples/Cancellation/Cancellation.csproj
kshyju: Perhaps we can make this sample targeting NET7, similar to the NETFX sample we have.

kshyju (approved):

samples/Cancellation/HttpTriggerWithCancellation.cs
```diff
@@ -0,0 +1,52 @@
...
+            [HttpTrigger(AuthorizationLevel.Anonymous,"get", "post", Route = null)]
+            HttpRequestData req,
+            FunctionContext executionContext,
+            CancellationToken cancellationToken)
+        {
+            _logger.LogInformation("HttpTriggerWithCancellation function triggered");
+
+            try
+            {
+                var response = req.CreateResponse(HttpStatusCode.OK);
```
kshyju, line 34: Consider adding `cancellationToken.ThrowIfCancellationRequested();` as it is a standard usage of cancellation token.

samples/Cancellation/Cancellation.csproj
liliankasem: If we're framing is as a .net 7 sample, then I'm okay with the addition of a new sample.

PR #1071, opened 2022-09-27
[Samples] Updated the exception handling middleware sample to handle multiple output binding case

kshyju:
Updated the exception handling middleware sample to handle multiple output binding case.

### DIFF
--- a/samples/CustomMiddleware/CustomMiddleware.csproj
+++ b/samples/CustomMiddleware/CustomMiddleware.csproj
@@ -16,7 +16,7 @@
     <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.9.0" />
     <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http" Version="3.0.13" />
     <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Storage" Version="5.0.0" />
-    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.5.0-preview1" OutputItemType="Analyzer" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.7.0" />
   </ItemGroup>
   <ItemGroup>
     <None Update="host.json">
--- a/samples/CustomMiddleware/CustomMiddleware.csproj
+++ b/samples/CustomMiddleware/CustomMiddleware.csproj
@@ -13,8 +13,9 @@
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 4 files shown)


samples/CustomMiddleware/CustomMiddleware.csproj
```diff
@@ -13,9 +13,10 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.8.0-preview1" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.9.0" />
     <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http" Version="3.0.13" />
-    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.5.0-preview1" OutputItemType="Analyzer" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Storage" Version="5.0.0" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.7.0" />
```
liliankasem, line 19: Don't we want to keep `OutputItemType="Analyzer"`?

samples/CustomMiddleware/HttpTriggerWithMultipleOutputBindings.cs
```diff
@@ -0,0 +1,38 @@
...
+            response.WriteString("Success!");
+
+            return new MyOutputType()
+            {
+                Name = "Azure Functions",
+                HttpResponse = response
+            };
+        }
+    }
+    public class MyOutputType
```
liliankasem, line 31: nit: new line

liliankasem (approved):

samples/CustomMiddleware/CustomMiddleware.csproj
kshyju: Nope.

liliankasem: Just for my understanding why? :D

kshyju: It is needed only when you add as a project reference. For package reference, that is not needed.

liliankasem: Good to know, we should probably remove this from templates next time we do an update

kshyju: It is already removed from templates.

satvu (approved):
lgtm!

PR #1074, opened 2022-09-27
[Samples] Clean up FunctionApp sample

liliankasem:


samples/FunctionApp/HttpTriggerWithCancellation/HttpTriggerWithCancellation.cs
```diff
@@ -34,7 +40,7 @@ public static class HttpTriggerWithCancellation
             }
             catch (OperationCanceledException)
             {
-                logger.LogInformation("Function invocation cancelled");
+                _logger.LogInformation("Function invocation cancelled");
```
kshyju, line 43: Should we replace the `await Task.Delay(5000, cancellationToken);` with `cancellationToken.ThrowIfCancellationRequested();` ?

kshyju (approved):

samples/FunctionApp/FunctionApp.csproj
```diff
@@ -14,7 +14,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.3.0" OutputItemType="Analyzer" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.7.0" />
```
satvu, line 17: Not 100% sure, but we *might* need to keep `OutputItemType` - vaguely remember last year that we needed to make changes to this app to test the host and worker e2e ([PR here](https://github.com/Azure/azure-functions-dotnet-worker/pull/616/files)). One of the changes was adding this property. If you have time, could you test that e2e debugging works without `OutputItemType` before merging? I can also assist here if needed.

samples/FunctionApp/HttpTriggerWithCancellation/HttpTriggerWithCancellation.cs
liliankasem: Yeah I think that makes sense for the sample, I had it as a Delay for testing purposes. I wonder if I should copy the public docs here and so something like:

```csharp
if (cancellationToken.IsCancellationRequested)
{
    log.LogInformation("A cancellation token was received. Taking precautionary actions.");
    //Take precautions like noting how far along you are with processing the batch
    log.LogInformation("Precautionary activities --complete--.");
    break;
}
else
{
    //business logic as usual
    log.LogInformation($"Message: {message} was processed.");
}
```

samples/FunctionApp/FunctionApp.csproj
liliankasem: Let's sync about this offline!

liliankasem: Validated that we can still debug with the host without `OutputItemType`

Screen Shot 2022-09-28 at 2 44 39 PM

PR #1086, opened 2022-09-29
Adding support for overriding the binding property name in function metadata

### LINKED ISSUE #1059: Functions metadata for blob trigger is not matching with in-proc
For a blob trigger in a V4 in-proc app, we generate the binding JSON with `path` property. When a blob trigger is created in isolated model, the binding property name is `blobPath` (See sample below).

As per @chiangvincent, this causes scale control to behave incorrectly.

> ICM: 336192878
Vincent Chiang (vchiang) edited at 2022-09-20 13:01:05 PDT 
This appears to be a bug within the functions host SDK. Since the scale controller expects a property called "path," and not "blobPath," the scale controller will not able to create a binding to monitor the function, and will not be able to scale as a result.

#### In-proc function app

    [FunctionName("Function1")]
    public void Run([BlobTrigger("samples-workitems/{name}", Connection = "MyStorageConnStr")]Stream myBlob, 
                               string name, ILogger log)
    {
        log.LogInformation($"C# Blob trigger inproc function Processed blob\n Name:{name} \n Size: {myBlob.Length} Bytes");
    }

Relevant packages
```
  <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Storage" Version="5.0.1" />
  <PackageReference Include="Microsoft.NET.Sdk.Functions" Version="4.1.3" />
```

Generated function.json
````
{
  "generatedBy": "Microsoft.NET.Sdk.Functions.Generator-4.1.3",
  "configurationSource": "attributes",
  "bindings": [
    {
      "type": "blobTrigger",
      "connection": "MyStorageConnStr",
      "path": "samples-workitems/{name}",
      "source": "logsAndContainerScan",
      "name": "myBlob"
    }
  ],
  "disabled": false,
  "scriptFile": "../bin/FunctionApp1.dll",
  "entryPoint": "FunctionApp1.Function1.Run"
}
````
#### Isolated 
````
[Function("Function1")]
public void Run([BlobTrigger("samples-workitems/{name}", Connection = "MyStorageConnStr")] string myBlob, string name)
{
    _logger.LogInformation($"C# Blob trigger function Processed blob\n Name: {name} \n Data: {myBlob}");
}
````
Relevant packages
```
<PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Storage" Version="5.0.0" />
<PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.7.0" />
```

Functions.metadata

````
  {
    "name": "Function1",
    "scriptFile": "BlobTriggerTest.dll",
    "entryPoint": "BlobTriggerTest.Function1.Run",
    "properties": {
      "IsCodeless": false
    },
    "bindings": [
      {
        "name": "myBlob",
        "type": "blobTrigger",
        "direction": "In",
        "dataType": "String",
        "blobPath": "samples-workitems/{name}",
        "connection": "MyStorageConnStr"
      }
    ]
  }
````

### PR DESCRIPTION
**Pull request by kshyju**
Fixes #1059 

Adding support for overriding the binding property name in function metadata. 

Currently, the property name from the BindingAttribute classes (Triggers, InputBindings,OutputBindings) is used to build the binding dictionary when generating function metadata. Changes in this PR will allow us to override this behavior by specifying a different name via an attribute.

We are introducing the `MetadataBindingPropertyName` attribute which can be applied on properties of triggers and bindings. The metadata generation process will check for the presence of this attribute and use the value instead of the property name.

This PR is also updating the BlobTrigger's `BlobPath` property with this new attribute so that "path" will be used as the property name in the binding dictionary.

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/extensions/Worker.Extensions.Abstractions/src/MetadataBindingPropertyName.cs
+++ b/extensions/Worker.Extensions.Abstractions/src/MetadataBindingPropertyName.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Azure.Functions.Worker.Extensions.Abstractions
         /// <summary>
         /// Initializes a new instance of <see cref="MetadataBindingPropertyName"/> with the specified property name.
         /// </summary>
-        /// <param name="bindingPropertyName">The name of the property.</param>
+        /// <param name="bindingPropertyName">The name of the property to be used when generating function metadata.</param>
         /// <exception cref="ArgumentNullException">Throws when bindingPropertyName is null.</exception>
         public MetadataBindingPropertyName(string bindingPropertyName)
         {
--- a/sdk/Sdk/FunctionMetadataGenerator.cs
+++ b/sdk/Sdk/FunctionMetadataGenerator.cs
@@ -425,6 +425,45 @@ namespace Microsoft.Azure.Functions.Worker.Sdk
             bindingMetadata.Add(binding);
         }
 
+        /// <summary>
+        /// Creates a map of property name and it's binding property name, if specified from the attribute passed in.
+        /// </summary>
+        /// <remarks>
+        /// Example output generated from BlobTriggerAttribute.
+        ///     {
+        ///         "BlobPath" : "path"
+        ///     }
+        /// 
+        /// In BlobTriggerAttribute type, the "BlobPath" property is decorated with "MetadataBindingPropertyName" attribute
+        /// where "path" is provided as the value to be used when generating metadata binding data, as shown below.
+        ///
+        ///     [MetadataBindingPropertyName("path")]
+        ///     public string BlobPath { set;get;}
+        /// 
+        /// </remarks>
+        /// <param name="attribute">The CustomAttribute instance to use for building the map. This will be a binding attribute
+        /// used to decorate the function parameter. Ex: BlobTrigger</param>
+        /// <returns>A dictionary containing the mapping.</returns>
+        private static IDictionary<string, string> GetPropertyNameAliasMapping(CustomAttribute attribute)
+        {
+            TypeReference paramAttributeType = attribute.AttributeType;
+            var typeDefinition = paramAttributeType.Resolve();
+            IDictionary<string, string> bindingNameAliasMap = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+            foreach (var prop in typeDefinition.Properties)
+            {
+                var bindingPropAttr = prop.CustomAttributes.FirstOrDefault(a =>
+                    a.AttributeType.FullName.Equals(Constants.MetadataBindingPropertyNameType));
+
+                var firstArgument = bindingPropAttr?.ConstructorArguments.First();
+                if (firstArgument?.Value != null)
+                {
+                    bindingNameAliasMap[prop.Name] = (string)firstArgument.Value.Value;
+                }
+            }
+
+            return bindingNameAliasMap;
+        }
+
         private static ExpandoObject BuildBindingMetadataFromAttribute(CustomAttribute attribute, string bindingType, TypeReference parameterType, string? parameterName)
         {
             ExpandoObject binding = new ExpandoObject();
@@ -450,9 +489,15 @@ namespace Microsoft.Azure.Functions.Worker.Sdk
                 bindingDict["DataType"] = "Binary";
             }
 
+            var bindingNameAliasDict = GetPropertyNameAliasMapping(attribute);
+
             foreach (var property in attribute.GetAllDefinedProperties())
             {
-                bindingDict.Add(property.Key, property.Value);
+                var propertyName = bindingNameAliasDict.TryGetValue(property.Key, out var overriddenPropertyName)
+                    ? overriddenPropertyName
+                    : property.Key;
+
+                bindingDict.Add(propertyName, property.Value);
             }
 
             // Determine if we should set the "Cardinality" property based on
--- a/sdk/Sdk/FunctionMetadataGenerator.cs
+++ b/sdk/Sdk/FunctionMetadataGenerator.cs
@@ -426,19 +426,19 @@ namespace Microsoft.Azure.Functions.Worker.Sdk
         }
 
         /// <summary>
-        /// Creates a map of property name and it's binding property name, if specified from the attribute passed in.
+        /// Creates a map of property names and its binding property names, if specified from the attribute passed in.
         /// </summary>
         /// <remarks>
         /// Example output generated from BlobTriggerAttribute.
         ///     {
-        ///         "path" : "samples-workitems/{name}"
+        ///         "BlobPath" : "path"
         ///     }
         /// 
         /// In BlobTriggerAttribute type, the "BlobPath" property is decorated with "MetadataBindingPropertyName" attribute
         /// where "path" is provided as the value to be used when generating metadata binding data, as shown below.
         ///
         ///     [MetadataBindingPropertyName("path")]
-        ///     public string BlobPath { set;get;}
+        ///     public string BlobPath { get; set; }
         /// 
         /// </remarks>
         /// <param name="attribute">The CustomAttribute instance to use for building the map. This will be a binding attribute
@@ -456,7 +456,7 @@ namespace Microsoft.Azure.Functions.Worker.Sdk
 
                 // The "MetadataBindingPropertyName" attribute has only one constructor which takes the name to be used.
                 var firstArgument = bindingPropAttr?.ConstructorArguments.First();
-                if (firstArgument?.Value != null)
+                if (firstArgument?.Value is not null)
                 {
                     bindingNameAliasMap[prop.Name] = (string)firstArgument.Value.Value;
                 }
--- a/sdk/Sdk/FunctionMetadataGenerator.cs
+++ b/sdk/Sdk/FunctionMetadataGenerator.cs
@@ -448,7 +448,7 @@ namespace Microsoft.Azure.Functions.Worker.Sdk
         {
             var bindingNameAliasMap = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
             TypeDefinition typeDefinition = attribute.AttributeType.Resolve();
-            
+
             foreach (var prop in typeDefinition.Properties)
             {
                 var bindingPropAttr = prop.CustomAttributes.FirstOrDefault(a =>
@@ -472,7 +472,7 @@ namespace Microsoft.Azure.Functions.Worker.Sdk
 
             return bindingNameAliasMap;
         }
-        
+
         private static ExpandoObject BuildBindingMetadataFromAttribute(CustomAttribute attribute, string bindingType, TypeReference parameterType, string? parameterName)
         {
             ExpandoObject binding = new ExpandoObject();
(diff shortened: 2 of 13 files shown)

### CODE REVIEW

**Comment by kshyju** [comment]
Copying the [comment ](https://github.com/Azure/azure-functions-dotnet-worker/issues/1059#issuecomment-1261295874)@fabiocav left in #1059
> The issue comes from the fact that the existing serialization logic simply uses the property name to create the metadata, deriving the name from here:
> 
> https://github.com/Azure/azure-functions-dotnet-worker/blob/1f7a321d88b2f66f792f3889ba7683d83ff45b92/extensions/Worker.Extensions.Storage.Blobs/src/BlobTriggerAttribute.cs#L31
> 
> With the in-proc model, the tooling will apply transformations to some of the extensions: https://github.com/Azure/azure-functions-vs-build-sdk/blob/main/src/Microsoft.NET.Sdk.Functions.Generator/AttributeExtensions.cs#L198-L203
> 
> We need to update our metadata generation logic to look for an extension attribute that takes precedence over the property name when creating the metadata.



I checked all the replaces we are doing [here](https://github.com/Azure/azure-functions-vs-build-sdk/blob/main/src/Microsoft.NET.Sdk.Functions.Generator/AttributeExtensions.cs#L198-L203) and looks like only blobTrigger is the valid candidate to make the change.

| Trigger           | Notes                                                                                                                    |
|-------------------|--------------------------------------------------------------------------------------------------------------------------|
| TimerTrigger      | Both in-proc & oop generates `schedule` property in metadata. OOP extension does not have `ScheduleExpression` property. |
| BlobTrigger       | In-proc generates `path`. OOP generates `blobPath`. Fixing in this PR.                                                                 |
| ServiceBusTrigger | OOP extension does not have `Access` property.                                                                             |
| EventHubTrigger   | Both produces `eventHubName`. I wonder the [if expression](https://github.com/Azure/azure-functions-vs-build-sdk/blob/main/src/Microsoft.NET.Sdk.Functions.Generator/AttributeExtensions.cs#L208) is failing due to case sensitive check.                         |
| ApiHubFileTrigger | OOP does not have this extension.                                                                                      |

File: sdk/Sdk/FunctionMetadataGenerator.cs [hunk 1]
```diff
@@ -425,6 +425,46 @@ private static void AddBindingMetadata(IList<ExpandoObject> bindingMetadata, Cus
             bindingMetadata.Add(binding);
         }
 
+        /// <summary>
+        /// Creates a map of property name and it's binding property name, if specified from the attribute passed in.
```
satvu, line 429: nit: its (possessive form) not it's

satvu:
Need to change [this file](https://github.com/Azure/azure-functions-dotnet-worker/blob/7d659ced053af41d58274ea4a36d2b5db2138750/test/SdkE2ETests/Contents/functions.metadata#L52) as well for the e2e tests - it's a json file that we check json function metadata generation against.

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -425,6 +425,46 @@ private static void AddBindingMetadata(IList<ExpandoObject> bindingMetadata, Cus
...
+        /// Example output generated from BlobTriggerAttribute.
+        ///     {
+        ///         "path" : "samples-workitems/{name}"
+        ///     }
+        /// 
+        /// In BlobTriggerAttribute type, the "BlobPath" property is decorated with "MetadataBindingPropertyName" attribute
+        /// where "path" is provided as the value to be used when generating metadata binding data, as shown below.
+        ///
+        ///     [MetadataBindingPropertyName("path")]
+        ///     public string BlobPath { set;get;}
```
satvu, line 441: nit: very much not important but formatting/spacing here can be `{ set; get; }` instead

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -425,6 +425,46 @@ private static void AddBindingMetadata(IList<ExpandoObject> bindingMetadata, Cus
...
+            TypeDefinition typeDefinition = attribute.AttributeType.Resolve();
+            
+            foreach (var prop in typeDefinition.Properties)
+            {
+                var bindingPropAttr = prop.CustomAttributes.FirstOrDefault(a =>
+                                                    a.AttributeType.FullName.Equals(Constants.MetadataBindingPropertyNameType));
+
+                // The "MetadataBindingPropertyName" attribute has only one constructor which takes the name to be used.
+                var firstArgument = bindingPropAttr?.ConstructorArguments.First();
+                if (firstArgument?.Value != null)
```
liliankasem, line 459: nit
```suggestion
                if (firstArgument?.Value is not null)
```

release_notes.md
```diff
@@ -3,5 +3,4 @@
 - My change description (#PR/#issue)
 -->
 
-- Add new analyzer rules for the source-generated function metadata preview (#1048)
-- Bump gRPC dependencies to latest version (#1085)
```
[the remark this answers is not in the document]
satvu: I just released a new package b/c we had to release a bump in grpc dependencies - he's skipping the reset PR and just doing that work here.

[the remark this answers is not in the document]
kshyju: Yes. ^. , with the hope that this PR gets merged next before anything else 😛. If that is not the case, I will fix the release notes as needed. Either way I need to do that to one of the PRs.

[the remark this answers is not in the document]
liliankasem: Ahhh okay, I didn't realise we had another release - thanks!

sdk/Sdk/FunctionMetadataGenerator.cs
kshyju: Fixed

sdk/Sdk/FunctionMetadataGenerator.cs
kshyju: Fixed.

sdk/Sdk/FunctionMetadataGenerator.cs
kshyju: Fixed in latest iteration.

kshyju:
> Need to change [this file](https://github.com/Azure/azure-functions-dotnet-worker/blob/7d659ced053af41d58274ea4a36d2b5db2138750/test/SdkE2ETests/Contents/functions.metadata#L52) as well for the e2e tests - it's a json file that we check json function metadata generation against.

The E2EApp uses a package reference of SDK. Once this version of package is out, will update that project with this change.

extensions/Worker.Extensions.Abstractions/src/MetadataBindingPropertyName.cs
```diff
@@ -0,0 +1,29 @@
...
+
+using System;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Abstractions
+{
+    /// <summary>
+    /// Specifies the binding property name that is used when generating function metadata.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
+    public sealed class MetadataBindingPropertyName : Attribute
```
fabiocav, line 12: nit: [naming] This is missing the `Attribute` suffix. With that (or without) the name feels a bit long. Given how this is used, I think shortening to `BindingPropertyName` would be appropriate, so effectively `BindingPropertyNameAttribute`

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -425,6 +425,46 @@ private static void AddBindingMetadata(IList<ExpandoObject> bindingMetadata, Cus
...
+        {
+            var bindingNameAliasMap = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+            TypeDefinition typeDefinition = attribute.AttributeType.Resolve();
+            
+            foreach (var prop in typeDefinition.Properties)
+            {
+                var bindingPropAttr = prop.CustomAttributes.FirstOrDefault(a =>
+                                                    a.AttributeType.FullName.Equals(Constants.MetadataBindingPropertyNameType));
+
+                // The "MetadataBindingPropertyName" attribute has only one constructor which takes the name to be used.
```
fabiocav, line 457: This is making an assumption about that type for something that could change with no enforcement. 

Can we:

1. Have a test in place to ensure that type is not modified (new constructors added)
2. Update the generation logic to test for the assumption and log an error if it is no longer true

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -450,9 +490,15 @@ private static ExpandoObject BuildBindingMetadataFromAttribute(CustomAttribute a
                 bindingDict["DataType"] = "Binary";
             }
 
+            var bindingNameAliasDict = GetPropertyNameAliasMapping(attribute);
```
fabiocav, line 493: This will iterate through the properties twice, once to build the map and another to create the metadata. Can we delay the attribute verification (checking if the property name is overridden) for the loop below?

extensions/Worker.Extensions.Abstractions/src/MetadataBindingPropertyName.cs
kshyju: Fixed the attribute name to "BindingPropertyNameAttribute" as suggested. Thank you!

sdk/Sdk/FunctionMetadataGenerator.cs
kshyju: I added a test to ensure constructors for this type is not modified.
Added logic to check the constructor **_used_** and the parameter type.

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -425,6 +425,54 @@ private static void AddBindingMetadata(IList<ExpandoObject> bindingMetadata, Cus
...
+                // The "MetadataBindingPropertyName" attribute has only one constructor which takes the name to be used.
+                var firstArgument = bindingPropAttr?.ConstructorArguments.Single();
+                if (firstArgument?.Value is null)
+                {
+                    continue;
+                }
+
+                if (firstArgument.Value.Value is not string bindingNameArgumentValue)
+                {
+                    throw new ArgumentException(
```
kshyju, line 466: I considered using InvalidOperation ,but this exception throws when the argument is of incorrect type, so chose to use this. Feel free to suggest if we can use a better exception.

satvu (approved):

sdk/Sdk/FunctionMetadataGenerator.cs
kshyju: Hm, The collection we are looping through below is a dictionary which contains the binding entry names and it's values.  The `GetAllDefinedProperties` using the [constructor param names](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/sdk/Sdk/Extensions/CustomAttributeExtensions.cs#L50) to build the binding (_with the assumption that it matches the property names_). I did not want to hook the override name check (which we can obtain from the the new attribute decoration of the properties only) to that because that code already is working on the naming assumption I mentioned above.  Yea, we have this additional iteration through the properties collection, but that seems more cleaner.

I just noticed our source gen version is also making this assumption (using constructor arguments). Let's see investigate we can switch to reading the property values.

brettsam (approved):

PR #1091, opened 2022-09-30
Adding a new worker option property to opt-in the behavior to get empty entries in trigger payload

### LINKED ISSUE #1017: Worker options to opt-in the behavior to get empty entries in payload
We [recently updated ](https://github.com/Azure/azure-functions-host/issues/8499)host to include empty entries when sending the message payload to out of process layer when the `IncludeEmptyEntriesInMessagePayload` capability is present. We need to allow users to opt in to this behavior via worker options when they bootstrap the function app.

Something like

    var host = new HostBuilder()
        {
        }, options =>
        {
            options.IncludeEmptyEntriesInPayload = true;
        })
        .Build();

- Needs to agree on naming of property.
- Should it be on the root level or should we move it under some other property to be better organized.

kshyju:
Fixes #1017 

This change is part of a fix for [8499](https://github.com/Azure/azure-functions-host/issues/8499) which has details about what problem this is solving.

Adding a boolean property to WorkerOptions, using which customer can opt-in to get the behavior where empty entries in the message payload are not skipped. We made the host level changes in [this PR ](https://github.com/Azure/azure-functions-host/pull/8642)to support this behavior.

Sample usage.

````
var host = new HostBuilder()
    {
    }, options =>
    {
        options.IncludeEmptyEntriesInPayload = true;
    })
    .Build();
````

### Pull request checklist

* [x] My changes **do not** require documentation changes - Docs will be automatically published from type.

### Additional information

Additional PR information

### DIFF
--- a/src/DotNetWorker.Core/Hosting/WorkerOptions.cs
+++ b/src/DotNetWorker.Core/Hosting/WorkerOptions.cs
@@ -30,6 +30,10 @@ namespace Microsoft.Azure.Functions.Worker
         
         /// <summary>
         /// Gets or sets a value that determines if empty entries should be included in the function trigger message payload.
+        /// For example, if a set of entries were sent to a messaging service such as Service Bus or Event Hub and your function
+        /// app has a Service bus trigger or Event hub trigger, only the non-empty entries from the payload will be sent to the
+        /// function code as trigger data when this setting value is <see langword="false"/>. When it is <see langword="true"/>,
+        /// All entries will be sent to the function code as it is. Default value for this setting is <see langword="false"/>.
         /// </summary>
         public bool IncludeEmptyEntriesInMessagePayload { get; set; }
     }
--- a/src/DotNetWorker.Core/Hosting/WorkerOptions.cs
+++ b/src/DotNetWorker.Core/Hosting/WorkerOptions.cs
@@ -27,5 +27,10 @@ namespace Microsoft.Azure.Functions.Worker
         /// exceptions when they are surfaced to the Host. 
         /// </summary>
         public bool EnableUserCodeException { get; set; } = false;
+        
+        /// <summary>
+        /// Gets or sets a value that determines if empty entries should be included in the function trigger message payload.
+        /// </summary>
+        public bool IncludeEmptyEntriesInMessagePayload { get; set; }
     }
 }
--- a/src/DotNetWorker.Grpc/GrpcWorker.cs
+++ b/src/DotNetWorker.Grpc/GrpcWorker.cs
@@ -193,7 +193,7 @@ namespace Microsoft.Azure.Functions.Worker
             _invocationHandler.TryCancel(request.InvocationId);
         }
 
-        internal static WorkerInitResponse WorkerInitRequestHandler(WorkerInitRequest request, WorkerOptions? workerOptions = null)
+        internal static WorkerInitResponse WorkerInitRequestHandler(WorkerInitRequest request, WorkerOptions workerOptions)
         {
             var response = new WorkerInitResponse
             {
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: src/DotNetWorker.Grpc/GrpcWorker.cs [hunk 1]
```diff
@@ -214,6 +216,11 @@ internal static WorkerInitResponse WorkerInitRequestHandler(WorkerInitRequest re
             response.Capabilities.Add("HandlesWorkerTerminateMessage", bool.TrueString);
             response.Capabilities.Add("HandlesInvocationCancelMessage", bool.TrueString);
 
+            if (workerOptions.IncludeEmptyEntriesInMessagePayload)
```
kshyju, line 219: If the option is disabled, we just do not send the capability at all (instead of sending with `False` value. Host code is not checking the value of this entry. It just checks for the existence of this capability key.

https://github.com/Azure/azure-functions-host/blob/ec3c2d1049351870bbf611760520cf1e48075c8f/src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs#L328-L331

satvu (approved):

src/DotNetWorker.Core/Hosting/WorkerOptions.cs
```diff
@@ -21,5 +21,10 @@ public class WorkerOptions
         /// Gets the collection of input converters.
         /// </summary>
         public InputConverterCollection InputConverters { get; } = new InputConverterCollection();
+        
+        /// <summary>
+        /// Gets or sets a value that determines if empty entries should be included in the function trigger message payload.
```
fabiocav, line 26: It would be good to have more details here to make the behavior controlled by the option clearer, perhaps also outline the current behavior?

fabiocav (approved):
Do we have a documentation item tracking public docs updates?

liliankasem (approved):

test/DotNetWorkerTests/GrpcWorkerTests.cs
```diff
@@ -130,6 +131,44 @@ public void InitRequest_ReturnsExpectedMetadata()
             Assert.Contains(response.WorkerMetadata.CustomProperties,
                 kvp => string.Equals(kvp.Key, "Worker.Grpc.Version", StringComparison.OrdinalIgnoreCase)
                 && string.Equals(kvp.Value, grpcWorkerVersion, StringComparison.OrdinalIgnoreCase));
+            
+        }
+        
+        [Theory]
+        [InlineData("IncludeEmptyEntriesInMessagePayload", true, "IncludeEmptyEntriesInMessagePayload", true, "True")]
+        [InlineData("IncludeEmptyEntriesInMessagePayload", false, "IncludeEmptyEntriesInMessagePayload", false)]
+        public void InitRequest_ReturnsExpectedCapabilitiesBasedOnWorkerOptions(
```
madelinegordon, line 140: nit - could space out the name to make easier to read "InitRequest_ReturnsExpectedCapabilities_BasedOnWorkerOptions"

test/DotNetWorkerTests/GrpcWorkerTests.cs
```diff
@@ -130,6 +131,44 @@ public void InitRequest_ReturnsExpectedMetadata()
...
+            string booleanPropertyName, 
+            bool booleanPropertyValue, 
+            string capabilityName, 
+            bool shouldCapabilityPresent,
+            string expectedCapabilityValue = null)
+        {
+            var workerOption = new WorkerOptions();
```
madelinegordon, line 147: nit - var name "workerOptions" ?

madelinegordon (approved):

jviau (approved):

src/DotNetWorker.Core/Hosting/WorkerOptions.cs
kshyju: @fabiocav I tried to elaborate the scenario a bit more in the latest iteration. Could you take a look? Feel free to suggest a better version if we can improve this further. Thank you!

kshyju:
> Do we have a documentation item tracking public docs updates?

This [public docs page for this type](https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.functions.worker.workeroptions?view=azure-dotnet#properties) will get updated from the code comments once we merge to main. Do you think we should have a separate docs page for this property ?  Let's evaluate whether we need to add details to [this page](https://learn.microsoft.com/en-us/azure/azure-functions/dotnet-isolated-process-guide).

PR #1103, opened 2022-10-05
Use a consistent function ID via code gen

### LINKED ISSUE #1102: Function metadata codegen does not generate a consistent ID between workers
We recently dealt with a regression in the Python worker where separate worker instances would report different IDs for functions, leading to the host being unable to identifier these are the same functions between different workers. The recently introduced code gen approach for .NET function metadata has the same issue. We should make sure that for a given build, all workers will report the identical IDs for the same functions.

jviau:

This PR ensures we will report a consistent `FunctionId` for the same function between worker instances.

#### Update:
New changes to PR:

- Addressed some warnings and typos.
- DefaultFunctionMetadata now computes its `FunctionId` based on a stable hash of `Name`, `EntryPoint`, and `ScriptFile`.
    - The implementation is adapted from https://github.com/Azure/azure-functions-host/blob/71ecbb2c303214f96d7e17310681fd717180bdbb/src/WebJobs.Script/Utility.cs#L847-L863

resolves #1102

### Pull request checklist

* [x] **IN PROGRESS** I have added all required tests (Unit tests, E2E tests) 

### Additional information

The source gen is yet to be released, so documentation updates unnecessary.

/cc: @satvu

### DIFF
--- a/src/DotNetWorker.Core/FunctionMetadata/DefaultFunctionMetadata.cs
+++ b/src/DotNetWorker.Core/FunctionMetadata/DefaultFunctionMetadata.cs
@@ -49,7 +49,8 @@ namespace Microsoft.Azure.Functions.Worker.Core.FunctionMetadata
 
         private static string? HashFunctionId(DefaultFunctionMetadata function)
         {
-            // we use uint to avoid the '-' sign when we .ToString() the result.
+            // We use uint to avoid the '-' sign when we .ToString() the result.
+            // This function is adapted from https://github.com/Azure/azure-functions-host/blob/71ecbb2c303214f96d7e17310681fd717180bdbb/src/WebJobs.Script/Utility.cs#L847-L863
             static uint GetStableHash(string value)
             {
                 unchecked
--- a/src/DotNetWorker.Core/FunctionMetadata/DefaultFunctionMetadata.cs
+++ b/src/DotNetWorker.Core/FunctionMetadata/DefaultFunctionMetadata.cs
@@ -1,4 +1,8 @@
-﻿using System.Collections.Generic;
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
 
 namespace Microsoft.Azure.Functions.Worker.Core.FunctionMetadata
 {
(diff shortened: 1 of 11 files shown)


satvu:
This is another location where the worker generates the function id that I think would also need the fix: https://github.com/Azure/azure-functions-dotnet-worker/blob/9c3380d36af1ed79040707f62cdf2141410c53de/src/DotNetWorker.Grpc/FunctionMetadata/DefaultFunctionMetadataProvider.cs#L56

fabiocav (approved):

fabiocav:
> This is another location where the worker generates the function id that I think would also need the fix:
> 
> https://github.com/Azure/azure-functions-dotnet-worker/blob/9c3380d36af1ed79040707f62cdf2141410c53de/src/DotNetWorker.Grpc/FunctionMetadata/DefaultFunctionMetadataProvider.cs#L56

For the existing model, this shouldn't be an issue, but it would be good to be consistent here. I don't believe an ID is currently persisted in metadata, so we'd need to change that as well.

src/DotNetWorker.Core/FunctionMetadata/DefaultFunctionMetadata.cs
```diff
@@ -20,15 +36,69 @@ public class DefaultFunctionMetadata : IFunctionMetadata
...
+        public string? EntryPoint { get => _entryPoint; set => ClearIdAndSet(value, ref _entryPoint); }
 
         /// <inheritdoc/>
         public IList<string>? RawBindings { get; set; }
 
         /// <inheritdoc/>
-        public string? ScriptFile { get; set; }
+        public string? ScriptFile { get => _scriptFile; set => ClearIdAndSet(value, ref _scriptFile); }
+
+        private static string? HashFunctionId(DefaultFunctionMetadata function)
```
satvu, line 50: @jviau Could you add a link in the PR description to the host hash function this is adapted from? I think it would also be helpful to add a brief comment in this area mentioning that this is adapted from the host's hash function. 

If any other devs want to adapt our OOP handling of this later, we would have a clearer reverse engineering path they can follow to get the appropriate implementation context.

src/DotNetWorker.Core/FunctionMetadata/DefaultFunctionMetadata.cs
```diff
@@ -20,15 +36,69 @@ public class DefaultFunctionMetadata : IFunctionMetadata
...
+        {
+            // we use uint to avoid the '-' sign when we .ToString() the result.
```
satvu, line 52: I haven't used this before so thanks for adding the clarifying comment! Learned a new best practice today.

jviau: I wouldn't say it's a best practice, just good to know what number type to use for the problem at hand. In this case, int and uint both have the same bits and some total number _range_, but just shifted across negatives or positives only:

``` CSharp
int: –2147483648 to 2147483647 
uint: 0 to 4294967295 
```

satvu (approved):

fabiocav (approved):

Issue #1081, opened 2022-09-28
[Epic] Support SDK-type bindings for out-of-proc workers


liliankasem:
Epic to track the work to support SDK-type bindings for out-of-proc workers, focusing on dotnet isolated implementation.

Work for this feature can be found under this tag: 

- https://github.com/Azure/azure-functions-dotnet-worker/labels/feature%3A%20SDK%20bindings 

## General

- https://github.com/Azure/azure-functions-dotnet-worker/issues/1018
- https://github.com/Azure/azure-functions-dotnet-worker/issues/1080
- https://github.com/Azure/azure-functions-dotnet-worker/issues/1036
- https://github.com/Azure/azure-functions-dotnet-worker/issues/1194
- https://github.com/Azure/azure-functions-dotnet-worker/issues/1266
- https://github.com/Azure/azure-functions-host/issues/9021
- https://github.com/Azure/azure-functions-host/issues/9022
- https://github.com/Azure/azure-functions-dotnet-worker/issues/1325

#### Bypass deferred binding

- https://github.com/Azure/azure-functions-dotnet-worker/issues/1395
- https://github.com/Azure/azure-functions-dotnet-worker/issues/1418
- https://github.com/Azure/azure-functions-dotnet-worker/issues/1495
- https://github.com/Azure/azure-functions-dotnet-worker/issues/1541

#### Analyzers

- https://github.com/Azure/azure-functions-dotnet-worker/issues/1277
- https://github.com/Azure/azure-functions-dotnet-worker/issues/1144
- https://github.com/Azure/azure-functions-dotnet-worker/issues/1587
- https://github.com/Azure/azure-functions-dotnet-worker/issues/1276

#### Bugs

- https://github.com/Azure/azure-functions-dotnet-worker/issues/1100
- https://github.com/Azure/azure-functions-dotnet-worker/issues/1204
- https://github.com/Azure/azure-functions-dotnet-worker/issues/1362

#### Enhancements 

- https://github.com/Azure/azure-functions-dotnet-worker/issues/1333 
- https://github.com/Azure/azure-functions-dotnet-worker/issues/1674
- https://github.com/Azure/azure-functions-dotnet-worker/issues/1675

#### Blob Storage

- https://github.com/Azure/azure-functions-dotnet-worker/issues/1035
- https://github.com/Azure/azure-functions-dotnet-worker/issues/1156
- https://github.com/Azure/azure-functions-dotnet-worker/issues/1093
[...]

PR #1107, opened 2022-10-06
Support sdk-type binding reference type

liliankasem:

#1081

### Pull request checklist

### Additional information

Additional PR information


extensions/Worker.Extensions.Abstractions/src/ExtensionInformationAttribute.cs
```diff
@@ -14,16 +14,24 @@ public class ExtensionInformationAttribute : Attribute
 
         public bool EnableImplicitRegistration { get; }
 
+        public bool SupportsBindingReferenceType { get; }
```
mathewc, line 17: Wonder if this shouldn't be named SupportsWorkerParameterBinding?

extensions/Worker.Extensions.Abstractions/src/ReferenceTypeAttribute.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Abstractions
+{
+    [AttributeUsage(AttributeTargets.Assembly)]
+    public class ReferenceTypeAttribute : Attribute
```
mathewc, line 9: Once we agree on the naming above, this attribute would follow suit. E.g. it could be called SupportsWorkerParameterBinding. And likely it should just be a marker attribute, not bool param needed?

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -428,15 +426,20 @@ private static ExpandoObject BuildBindingMetadataFromAttribute(CustomAttribute a
             bindingDict["Type"] = bindingType;
             bindingDict["Direction"] = GetBindingDirection(attribute);
 
+            // Is sdk parameter type
```
mathewc, line 429: nit - adjust comment to "For extensions that support worker binding, set the DataType so all parameters are bound by the worker."

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -695,6 +700,12 @@ private static bool TryAddExtensionInfo(IDictionary<string, string> extensions,
                         implicitlyRegister = (bool)assemblyAttribute.ConstructorArguments[2].Value;
                     }
 
+                    if (assemblyAttribute.ConstructorArguments.Count == 4)
+                    {
+                        // SupportsBindingReferenceType
```
mathewc, line 705: nit: remove unnecessary comment?

src/DotNetWorker.Core/IBindingData.cs
```diff
@@ -0,0 +1,34 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.Functions.Worker.Core
+{
+    /// <summary>
+    /// A representation of a Microsoft.Azure.WebJobs.ParameterBindingData
+    /// </summary>
+    public interface IBindingData
```
mathewc, line 11: do we really need this interface? Doesn't seem like other types are following this pattern.

extensions/Worker.Extensions.Abstractions/src/ReferenceTypeAttribute.cs
liliankasem: Ah this should be deleted.

We had discussed either introducing a new attribute (like this one) or using the existing extension information attribute to determine if an extension supports reference types. In the end we went with using the existing attribute so I should have removed this file.

In the `ExtensionInformationAttribute` file, we introduced a `SupportsBindingReferenceType` property - do you feel we should rename this and not use bool for that?

src/DotNetWorker.Core/IBindingData.cs
liliankasem: We use this interface in the converters to determine if the converter can handle the request or not:

https://github.com/Azure/azure-functions-dotnet-worker/blob/92c5be67bcda0c3badd980ec2524f7c43eb7d859/extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs#L23-L26

src/DotNetWorker.Grpc/BindingData.cs
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Azure.Functions.Worker.Core;
+
+namespace Microsoft.Azure.Functions.Worker.Grpc.Messages
+{
+    internal sealed partial class BindingData : IBindingData
```
mathewc, line 9: For the naming of the types that are consumed by worker converters, I wonder if ParameterBindingData wouldn't be better here?

surgupta-msft:
ToDo - Adding tests

src/DotNetWorker.Grpc/BindingData.cs
liliankasem: Renaming to ModelBindingData

src/DotNetWorker.Grpc/protobuf/src/proto/FunctionRpc.proto
```diff
@@ -32,7 +32,8 @@ message StreamingMessage {
     WorkerInitRequest worker_init_request = 17;
```
liliankasem, line 32: Make sure to remove proto changes in this PR once we have a new PR to bring in the new proto version

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -425,18 +459,25 @@ private static ExpandoObject BuildBindingMetadataFromAttribute(CustomAttribute a
...
             }
 
+            var direction = GetBindingDirection(attribute);
+            bindingDict["Direction"] = direction;
             bindingDict["Type"] = bindingType;
-            bindingDict["Direction"] = GetBindingDirection(attribute);
 
+            // For extensions that support worker binding, set the DataType so parameters are bound by the worker.
+            // Only use worker binding for input and trigger bindings
+            if (supportsReferenceType && direction != Constants.OutputBindingDirection)
```
liliankasem, line 468: We probably don't need to have a check for output binding here because when we call this method (`BuildBindingMetadataFromAttribute`) from the output binding path, we always set supportsReferenceType to false

https://github.com/Azure/azure-functions-dotnet-worker/blob/08e86a23c00fd8199639c4588397b88c6bebdcae/sdk/Sdk/FunctionMetadataGenerator.cs#L444-L449

extensions/Worker.Extensions.Abstractions/src/ExtensionInformationAttribute.cs
liliankasem: With BindingData renamed to ModelBindingData, would `SupportsWorkerModelBinding` make sense?

mathewc: If we go with the defer bind naming, then this could be SupportsDeferredBinding

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -368,19 +370,57 @@ private bool TryAddOutputBindingFromMethod(IList<ExpandoObject> bindingMetadata,
...
                 foreach (CustomAttribute parameterAttribute in parameter.CustomAttributes)
                 {
                     if (IsFunctionBindingType(parameterAttribute))
                     {
-                        AddBindingMetadata(bindingMetadata, parameterAttribute, parameter.ParameterType, parameter.Name);
-                        AddExtensionInfo(_extensions, parameterAttribute);
+                        string bindingType = GetBindingType(parameterAttribute);
+                        AddExtensionInfo(_extensions, parameterAttribute, out bool supportsReferenceType);
+
+                        // if this is a trigger and any input binding contains an expression, don't use BindingData
```
mathewc, line 384: We need defer bind if any input OR output bindings have binding expressions, right? For example, otherwise, if we were to defer bind for the trigger, the host may not have the binding metadata it needs to resolve the output binding, resulting in a runtime error.

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -368,19 +370,57 @@ private bool TryAddOutputBindingFromMethod(IList<ExpandoObject> bindingMetadata,
...
                     }
                 }
             }
         }
 
+        private bool InputParameterPropertyContainsExpression(Collection<ParameterDefinition> parameters)
+        {
+            foreach (ParameterDefinition parameter in parameters)
+            {
+                foreach (CustomAttribute parameterAttribute in parameter.CustomAttributes)
```
mathewc, line 401: You're missing a check here IsFunctionBindingType which is done in the other loop above. Why do we need to loop through all the parameters twice like this anyways? You could just loop through once, build a list of binding metadatas, then at the end set the deferred property based on whether any used expressions.

sdk/Sdk/FunctionMetadataGenerator.cs
liliankasem: That's a good point. I think this class in general is confusing as this current function is called "AddInputTriggerBindingsAndExtensions" and so at this point I assume we only have input and triggers, but there's nothing upstream making that split between input/trigeer and output. I might need to get some clarity from @fabiocav and @brettsam about this, but in general agree with your point about triggers deferring binding if outputs also contain expressions

sdk/Sdk/FunctionMetadataGenerator.cs
liliankasem: This was definitly a brute force attempt to get a prototype running so definitly not expecting this to be final form :D I think your suggestion makes sense, you're saying that at the end we can just loop through `bindingMetadata` to check if any input/output bindings use expressions and use that to change the flag for the trigger? This should be a lot more efficient

src/DotNetWorker.Core/IBindingData.cs
liliankasem: Hmm, that's fair. I think the intension was around project dependencies. ModelBindingData is defined in the Grpc project, and the extension projects will be [updated to reference the Core project](https://github.com/Azure/azure-functions-dotnet-worker/blob/568045c48fcda451a62b691f34c948ad3d429dbc/extensions/Worker.Extensions.Storage.Blobs/src/Worker.Extensions.Storage.Blobs.csproj#L19), this is so they have access to converter interfaces such as `IInputConverter`. And I believe we didn't want the extensions to take a dependency on the Grpc layer. If that isn't a good enough reason, then I am happy to modify this and work without an interface.

@fabiocav want to get your input here too as you helped make the initial interface decision

sdk/Sdk/FunctionMetadataGenerator.cs
surgupta-msft: Done

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -368,19 +370,57 @@ private bool TryAddOutputBindingFromMethod(IList<ExpandoObject> bindingMetadata,
 
         private void AddInputTriggerBindingsAndExtensions(IList<ExpandoObject> bindingMetadata, MethodDefinition method)
         {
+            var inputBindingContainsExpression = InputParameterPropertyContainsExpression(method.Parameters);
```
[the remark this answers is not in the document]
surgupta-msft: Done

extensions/Worker.Extensions.Abstractions/src/ExtensionInformationAttribute.cs
```diff
@@ -14,16 +14,24 @@ public class ExtensionInformationAttribute : Attribute
 
         public bool EnableImplicitRegistration { get; }
 
+        public bool SupportsDeferredBinding { get; }
```
liliankasem, line 17: @jviau FYI: I think you commented here about this being `public bool SupportsDeferredBinding { get; init; }` but in the blob converter PR, the converter PR wasn't meant to have these base changes so that's my bad on PR setup, but should be fixed now. Just letting you know in case you wanted to do your review here too!

pack-sdk.ps1
```diff
@@ -0,0 +1,13 @@
+# FOR TESTING - WILL DELETE
```
mathewc, line 1: Remove from PR

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -763,6 +776,11 @@ private static bool TryAddExtensionInfo(IDictionary<string, string> extensions,
                         implicitlyRegister = (bool)assemblyAttribute.ConstructorArguments[2].Value;
                     }
 
+                    if (assemblyAttribute.ConstructorArguments.Count == 4)
```
mathewc, line 779: Why are we enabling things based on ctor definitions, rather than an actual instance of the assembly attribute? One problem I see with this is that if the extension specifies false for these, e.g. `[assembly: ExtensionInformation("Microsoft.Azure.WebJobs.Extensions.Storage.Blobs", "5.2.0-alpha.20221028.1", false, false)]` you'll still be enabling. Another problem is that you're doing a strict equality check for 4. If someone adds a 5th ctor overload in the future, it'll break your code.

src/DotNetWorker.Core/IModelBindingData.cs
```diff
@@ -0,0 +1,33 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.Functions.Worker.Core
+{
+    /// <summary>
+    /// A representation of a Microsoft.Azure.WebJobs.ParameterBindingData
+    /// </summary>
+    public interface IModelBindingData
```
mathewc, line 11: I still don't understand why this needs to be an interface. It seems unnecessary, and also means that we can't add things to the type in the future (e.g. new binding properties) without breaking contract changes.

liliankasem: I think we have a thread about this here: https://github.com/Azure/azure-functions-dotnet-worker/pull/1107#discussion_r1010958793

I'm not opposed to removing the interface but want to double check with Fabio around his though process - I can sync with him offline and update this

src/ci.yml
```diff
@@ -53,6 +53,7 @@ jobs:
       }
       ./setup-e2e-tests.ps1 -FunctionsRuntimeVersion $env:FUNCTIONSRUNTIMEVERSION `
                             -UseCoreToolsBuildFromIntegrationTests:$useIntegrationTestingCoreTools
+                            -SkipCosmosDBEmulator
```
surgupta-msft, line 56: This change will be removed before merging PR. It's only for testing to unblock pipeline.

pack-sdk.ps1
surgupta-msft: Removed

src/DotNetWorker.Core/IBindingData.cs
fabiocav: Core types (`Worker.Core`) do not have a dependency on any types coming from the `Worker.Grpc` project, as that is one (of possibly many) transport implementations. Any time the gRPC implementation needs to pass its types to the infrastructure/core, it does so using the abstractions defined there. `IBindingData` is just that, the abstraction that is exposed in the public APIs in Core, which is what is exposed to converters and other components. Those components do not have a dependency on any other transport specific types.

sdk/Sdk/FunctionMetadataGenerator.cs
surgupta-msft: For the last point regarding strict equality check, I have updated the check to >=4. And if Extension specifics false then below code will assign supportsDeferredBinding to False

`supportsDeferredBinding = (bool)assemblyAttribute.ConstructorArguments[3].Value;`

Regarding first question, _Why are we enabling things based on ctor definitions, rather than an actual instance of the assembly attribute?_

Need to check with @fabiocav. I remember an old discussion to extend the logic of evaluating constructor arguments.

src/DotNetWorker.Core/IModelBindingData.cs
surgupta-msft: cc @fabiocav

surgupta-msft: Resolving this one as we are discussing this under comment - https://github.com/Azure/azure-functions-dotnet-worker/pull/1107#discussion_r1002146117

src/DotNetWorker.Core/IBindingData.cs
mathewc: Guess I'm not clear on the layering here then. BindingData could itself probably just be in Worker.Core then? If you retain the interface, and we decide we need to add another bit of info later, is it easy for us to rev both the interface and the impl together?

sdk/Sdk/FunctionMetadataGenerator.cs
mathewc: Once you have the attribute instance, why can't you just check attr.SupportsDeferredBinding? Why are you checking ctor params for any of these things that the attribute exposes properties for?

src/DotNetWorker.Core/IBindingData.cs
fabiocav: `BindingData` is an auto-generated type from the `proto` file (it's what is defined in the protobuf contract), internal and also implements some gRPC specific interfaces.

We could use an abstract class instead of the interface, as that would help mitigate the concern. That's what we've done for things like the `HttpRequestData`

sdk/Sdk/FunctionMetadataGenerator.cs
surgupta-msft: I don't think we are getting attribute instance here, instead we are just getting AssemblyDefinition.

```csharp
 private static bool TryAddExtensionInfo(IDictionary<string, string> extensions, AssemblyDefinition extensionAssemblyDefinition, out bool supportsDeferredBinding, bool usedByFunction = true)
        {
            supportsDeferredBinding = false;

            foreach (var assemblyAttribute in extensionAssemblyDefinition.CustomAttributes)
            {
                if (string.Equals(assemblyAttribute.AttributeType.FullName, Constants.ExtensionsInformationType, StringComparison.Ordinal))
                {
                    string extensionName = assemblyAttribute.ConstructorArguments[0].Value.ToString();
                    string extensionVersion = assemblyAttribute.ConstructorArguments[1].Value.ToString();
                    bool implicitlyRegister = false;
                    if (assemblyAttribute.ConstructorArguments.Count >= 3)
                    {
                        // EnableImplicitRegistration
                        implicitlyRegister = (bool)assemblyAttribute.ConstructorArguments[2].Value;
                    }

                    if (assemblyAttribute.ConstructorArguments.Count >= 4)
                    {
                        supportsDeferredBinding = (bool)assemblyAttribute.ConstructorArguments[3].Value;
                    }
```

fabiocav: Yeah, it's a little awkward, but this uses `Mono.Cecil`, that isn't working with an actual instance of the attribute in this case.

extensions/Worker.Extensions.Abstractions/src/ExtensionInformationAttribute.cs
liliankasem: @surgupta-msft do you want to see if `public bool SupportsDeferredBinding { get; init; }` works, if so then you can remove the new ctor

Here's the original comment for reference: https://github.com/Azure/azure-functions-dotnet-worker/pull/1108#discussion_r1010957806

sdk/Sdk/FunctionMetadataGenerator.cs
mathewc: Don't know the backstory on that, but if you're happy, I'm happy

mathewc (approved):

src/DotNetWorker.Core/IBindingData.cs
surgupta-msft: Done

extensions/Worker.Extensions.Abstractions/src/ExtensionInformationAttribute.cs
surgupta-msft: Since the existing fields of this class are also initialized in similar way, I will not block this PR, but happy to try this refactoring out separately.

PR #1108, opened 2022-10-06
Blob converter for ParameterBindingData support

liliankasem:

Related epic: #1081

Part of the work to support SDK-type bindings for out-of-proc workers is to write a BlobConverter that can handle the new ModelBindingData type. The ModelBindingData type contains information that we can use to hydrate the client, so the job of the converter here is to parse the content provided to extract the relevant information, and use that to create the requested type (requested type being what the user is binding to). 

- ModelBindingData was added to the protobuf and brought into this repo via this [PR](https://github.com/Azure/azure-functions-dotnet-worker/pull/1148)
- We updated the SDK to use the ModelBindingData in this [PR](https://github.com/Azure/azure-functions-dotnet-worker/pull/1107). The SDK will check to see if a given extension supports deferred binding, and if it does, it will set the flag so that the host uses this ModelBindingData type for each function it is supported

The [epic](https://github.com/Azure/azure-functions-dotnet-worker/issues/1081#issuecomment-1276566098) has a list of all the PRs related to this if you'd like to follow the progress of how ModelBindingData came to be.

### Pull request checklist

  * [x] Otherwise: Documentation issue linked to PR https://github.com/MicrosoftDocs/azure-docs-pr/pull/225452
  * Working on this

### Additional information

Additional PR information


DotNetWorker.sln
```diff
@@ -115,6 +115,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "NetFxWorker", "samples\NetF
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "DotNetWorker.ApplicationInsights", "src\DotNetWorker.ApplicationInsights\DotNetWorker.ApplicationInsights.csproj", "{65DE66B6-568F-46AC-8F0D-C79A02F48214}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SdkTypeExtensions", "samples\SdkTypeExtensions\SdkTypeExtensions.csproj", "{901DA3C3-3ABA-4859-89D3-63343ED2A0AC}"
```
mathewc, line 118: nit: wondering if we shouldn't name this WorkerBindingSamples? The "SdkType" naming isn't really clear.

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,122 @@
...
+            }
+
+            if (bindingData.Source is not "Microsoft.Azure.WebJobs.Extensions.Storage.Blobs")
+            {
+                return new ValueTask<ConversionResult>(ConversionResult.Unhandled());
+            }
+
+            var content = bindingData.Content.ToObjectFromJson<Dictionary<string, string>>();
+
+            content.TryGetValue("Connection", out var connectionName);
```
mathewc, line 35: How is casing handled? E.g. if the extension ends up serializing this as lower case, will this fail?

extensions/Worker.Extensions.Storage.Blobs/src/StorageExtensionStartup.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Microsoft.Azure.Functions.Worker;
```
mathewc, line 1: Add copyright headers on all new files

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,122 @@
...
+    internal class BlobStorageConverter : IInputConverter
+    {
+        public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
+        {
+            if (context.Source is not IBindingData bindingData)
+            {
+                return new ValueTask<ConversionResult>(ConversionResult.Unhandled());
+            }
+
+            if (bindingData.Source is not "Microsoft.Azure.WebJobs.Extensions.Storage.Blobs")
```
mathewc, line 28: Should we put this in a shared constant somewhere? Particularly if we're referencing this string other places as well

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,122 @@
...
+            }
+
+            var content = bindingData.Content.ToObjectFromJson<Dictionary<string, string>>();
+
+            content.TryGetValue("Connection", out var connectionName);
+            content.TryGetValue("BlobName", out var blobName);
+            content.TryGetValue("ContainerName", out var containerName);
+
+            if (string.IsNullOrEmpty(connectionName)
+                || string.IsNullOrEmpty(blobName)
```
mathewc, line 40: You're always expecting a blob name, but blob extension also allows you to bind to a BlobContainerClient. Are you supporting that?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,122 @@
...
+            Type _ when targetType == typeof(AppendBlobClient)  => CreateBlobReference<AppendBlobClient>(connectionString, containerName, blobName),
+            _ => CreateTargetObject(targetType, connectionString, containerName, blobName)
+        };
+
+        private object CreateTargetObject(Type targetType, string connectionString, string containerName, string blobName)
+        {
+            var content = GetBlobString(connectionString, containerName, blobName);
+            MethodInfo deserializeObjectMethod = GetType()
+                                                .GetMethod(nameof(DeserializeTargetObject), BindingFlags.Instance | BindingFlags.NonPublic)
+                                                .MakeGenericMethod(new Type[] { targetType });
```
mathewc, line 74: JsonConvert.DeserializeObject has a non generic overload you can use, so you shouldn't have to do this

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,122 @@
...
+            Type _ when targetType == typeof(Stream)            => GetBlobStream(connectionString, containerName, blobName),
+            Type _ when targetType == typeof(Byte[])            => GetBlobBinaryData(connectionString, containerName, blobName),
+            Type _ when targetType == typeof(BlobClient)        => CreateBlobReference<BlobClient>(connectionString, containerName, blobName),
+            Type _ when targetType == typeof(BlockBlobClient)   => CreateBlobReference<BlockBlobClient>(connectionString, containerName, blobName),
+            Type _ when targetType == typeof(PageBlobClient)    => CreateBlobReference<PageBlobClient>(connectionString, containerName, blobName),
+            Type _ when targetType == typeof(AppendBlobClient)  => CreateBlobReference<AppendBlobClient>(connectionString, containerName, blobName),
+            _ => CreateTargetObject(targetType, connectionString, containerName, blobName)
+        };
+
+        private object CreateTargetObject(Type targetType, string connectionString, string containerName, string blobName)
```
mathewc, line 69: So CreateTargetObject is meant to be a catch all for POCOs and other types we don't recognize? E.g. in proc, we support binding a Blob to a TextReader. If they try to use that same code out of proc, you'll end up trying to deserialize into a TextReader?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        private object? DeserializeTargetObject<T>(string content)
+        {
+            return JsonConvert.DeserializeObject<T>(content);
+        }
+
+        private string GetBlobString(string connectionString, string containerName, string blobName)
+        {
+            var client = CreateBlobReference<BlobClient>(connectionString, containerName, blobName);
+            var download = client.DownloadContent();
```
mathewc, line 87: We need to be using the async APIs here and everywhere else we're downloading blobs

samples/SdkTypeExtensions/Blob/BlobInputBindingSamples.cs
```diff
@@ -0,0 +1,67 @@
...
+using System;
+using System.IO;
+using System.Net;
+using System.Text;
+using System.Threading.Tasks;
+using Azure.Storage.Blobs;
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Azure.Functions.Worker.Http;
+using Microsoft.Extensions.Logging;
+
```
mathewc, line 13: need samples binding trigger and input to POCO. I know you have the Book sample that does trigger. Maybe your trigger/input samples should also show pocos.

Are there any regression tests for this stuff, or is it only in samples that have to be manually run?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
liliankasem: Good call, we should handle casing

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
liliankasem: :o

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
liliankasem: Yeah I don't think we're planning to support TextReader, will try and get list of those together so we can deal with them

samples/SdkTypeExtensions/Blob/BlobInputBindingSamples.cs
liliankasem: I'm not sure, @kshyju @fabiocav do you know if we have regression tests in place for the dotnet worker and extensions?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,122 @@
...
+            {
+                return new ValueTask<ConversionResult>(ConversionResult.Unhandled());
+            }
+
+            if (bindingData.Source is not "Microsoft.Azure.WebJobs.Extensions.Storage.Blobs")
+            {
+                return new ValueTask<ConversionResult>(ConversionResult.Unhandled());
+            }
+
+            var content = bindingData.Content.ToObjectFromJson<Dictionary<string, string>>();
```
mathewc, line 33: Shouldn't we be using ContentType here, instead of just assuming it's json? I.e. demonstrate the proper way worker extensions should leverage the contentType/Versioning model you're introducing?

liliankasem: Yup that's right, this should defo be using version and content type, just never got round to doing it properly. Will make sure it's addressed before we publish the PR

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
liliankasem: Updated to have support for BlobContainerClient. I haven't actually tested any cases with IEnumerable or anything that might impact cardinality, will see if that works out of the box or requires more changes

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
liliankasem: Honestly, I'm not sure how we want to use version here. I added some logic to check the content type at least

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,131 @@
...
+            var containerName = content.FirstOrDefault(x => String.Equals(x.Key, Constants.ContainerName, StringComparison.OrdinalIgnoreCase)).Value;
+            var blobName = content.FirstOrDefault(x => String.Equals(x.Key, Constants.BlobName, StringComparison.OrdinalIgnoreCase)).Value;
+
+            if (string.IsNullOrEmpty(connectionName) || string.IsNullOrEmpty(containerName))
+            {
+                return new ValueTask<ConversionResult>(ConversionResult.Unhandled());
+            }
+
+            var connectionString = connectionName is null ? null : Environment.GetEnvironmentVariable(connectionName);
+            var result = Task.Run(() => ToTargetType(context.TargetType, connectionString, containerName, blobName)).Result;
```
jviau, line 46: Can this function be marked `async` and await this call directly?

Some additional points (but moving to async makes them irrelevant): (1) always use `GetAwait().GetResult()` over `.Result` as the later can deadlock. (2) `Task.Run` is not necessary - it just means post this work to the threadpool. You can remove that wrapper and do `ToTargetType(...).GetAwaiter().GetResult();` directly.

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,131 @@
...
+        private async Task<Stream> GetBlobStream(string connectionString, string containerName, string blobName)
+        {
+            var client = CreateBlobReference<BlobClient>(connectionString, containerName, blobName);
+            var download = await client.DownloadStreamingAsync();
+            return download.Value.Content;
+        }
+
+        private BlobContainerClient CreateBlobContainerClient(string connectionString, string containerName)
+        {
+            BlobContainerClient container = new(connectionString, containerName);
```
jviau, line 105: Does this need to support AAD authentication?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,131 @@
...
+        private async Task<string> GetBlobString(string connectionString, string containerName, string blobName)
+        {
+            var client = CreateBlobReference<BlobClient>(connectionString, containerName, blobName);
+            var download = await client.DownloadContentAsync();
+            return download.Value.Content.ToString();
+        }
+
+        private async Task<Byte[]> GetBlobBinaryData(string connectionString, string containerName, string blobName)
+        {
+            var stream = new MemoryStream();
```
jviau, line 90: I think this stream may need to be disposed

pack-sdk.ps1
```diff
@@ -0,0 +1,13 @@
+# FOR TESTING - WILL DELETE
+
+$packageSuffix = "dev" + [datetime]::UtcNow.Ticks.ToString()
```
jviau, line 3: If you are curious, for local builds the dotnet team likes to version everything as "42.42.42.42". I find that fun 😄

liliankasem: hahaha love that :D

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
liliankasem: We do eventually but AAD is out-of-scope for public preview, @surgupta-msft has been investigating this area

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
liliankasem: That's definitely good to know, thanks for sharing!

I couldn't make it async as then we're not following the interface https://github.com/Azure/azure-functions-dotnet-worker/blob/main/src/DotNetWorker.Core/Converters/Converter/IInputConverter.cs

and I wasn't sure the impact of me making changes to that interface 🤔

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,131 @@
...
+namespace Microsoft.Azure.Functions.Worker
+{
+    /// <summary>
+    /// Converter to bind Blob Storage type parameters.
+    /// </summary>
+    internal class BlobStorageConverter : IInputConverter
+    {
+        public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
+        {
+            if (context.Source is not IModelBindingData bindingData)
```
mathewc, line 25: As per my other comment, I don't see why we can't just bind to ModelBindingData here rather than an interface?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,131 @@
...
+            var connectionName = content.FirstOrDefault(x => String.Equals(x.Key, Constants.Connection, StringComparison.OrdinalIgnoreCase)).Value;
+            var containerName = content.FirstOrDefault(x => String.Equals(x.Key, Constants.ContainerName, StringComparison.OrdinalIgnoreCase)).Value;
+            var blobName = content.FirstOrDefault(x => String.Equals(x.Key, Constants.BlobName, StringComparison.OrdinalIgnoreCase)).Value;
+
+            if (string.IsNullOrEmpty(connectionName) || string.IsNullOrEmpty(containerName))
+            {
+                return new ValueTask<ConversionResult>(ConversionResult.Unhandled());
+            }
+
+            var connectionString = connectionName is null ? null : Environment.GetEnvironmentVariable(connectionName);
```
mathewc, line 45: I don't thing we should be defaulting the connection here. There's logic in the blob extension itself that is creating the ParameterBindingData, and we should centralize all the connection name resolution/defaulting there I think.

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,131 @@
...
+    internal class BlobStorageConverter : IInputConverter
+    {
+        public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
+        {
+            if (context.Source is not IModelBindingData bindingData)
+            {
+                return new ValueTask<ConversionResult>(ConversionResult.Unhandled());
+            }
+
+            if (bindingData.Source is not Constants.BlobExtensionName)
```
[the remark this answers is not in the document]
liliankasem: It's a preference to keep them separate

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
jviau: Any contract with a `ValueTask<T>` or `Task<T>` return can be made `async` in the implementation. This has no impact on the contract, it is mostly compile-time directives, callers are not aware of `async` or not, they only care about the return type..

kshyju: @liliankasem This existing converter uses async await. https://github.com/Azure/azure-functions-dotnet-worker/blob/main/src/DotNetWorker.Core/Converters/JsonPocoConverter.cs

liliankasem: Yup its working now, why I was getting interface errors before? I'll never know 😫

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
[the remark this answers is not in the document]
liliankasem: I have no idea xD @kshyju might be more familiar here on the worker side?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
surgupta-msft: Will be done as part of this work - https://github.com/Azure/azure-functions-dotnet-worker/issues/1156

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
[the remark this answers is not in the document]
surgupta-msft: Got similar feedback from Fabio to extract `ConfigurationSection`

```csharp
 IConfigurationSection connectionSection = _configuration.GetWebJobsConnectionStringSection(name);
```

Reference code - [Link](https://github.com/Azure/azure-sdk-for-net/blob/126ed7e63f99e305a962ac43613d55a25ab5fbfe/sdk/extensions/Microsoft.Azure.WebJobs.Extensions.Clients/src/Shared/StorageClientProvider.cs)

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,235 @@
...
+            {
+                // Try using Managed Identity
+                context.FunctionContext.BindingContext.BindingData.TryGetValue("Uri", out var blobEndpoint);
+                string endpoint = blobEndpoint.ToString().Trim('\\', '"');
+
+                if (Environment.GetEnvironmentVariable(Constants.ConnectionAccountName) is not null
+                    || Environment.GetEnvironmentVariable(Constants.ConnectionBlobUri) is not null)
+                {
+                    // for AzureWebJobsStorage
+                    blob = new BlobClient(new Uri(endpoint), new DefaultAzureCredential());
```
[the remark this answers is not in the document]
liliankasem: Trying some thing like this with the [cosmos converter](https://github.com/Azure/azure-functions-dotnet-worker/pull/1109/files#diff-646bb57f7c89f3e846afcc0bffd40b5c6a20871f16879322091004498708bf26) - will try an E2E test in the new year

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,235 @@
...
+
+        private async Task<string> GetBlobContentString(BlobClient client)
+        {
+            var download = await client.DownloadContentAsync();
+            return download.Value.Content.ToString();
+        }
+
+        private async Task<Byte[]> GetBlobBinaryData(string connectionString, string containerName, string blobName)
+        {
+            using (MemoryStream stream = new MemoryStream())
```
jviau, line 153: nit: can use simplified `using` block:

``` CSharp
using MemoryStream stream = new();
var client = CreateBlobReference<BlobClient>(connectionString, containerName, blobName);
await client.DownloadToAsync(stream);
return stream.ToArray();
```

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,235 @@
...
+
+        private bool TryGetBindingDataContent(ModelBindingData bindingData, out IDictionary<string, string> bindingDataContent)
+        {
+            bindingDataContent = bindingData.ContentType switch
+            {
+                Constants.JsonContentType => new Dictionary<string, string>(bindingData.Content.ToObjectFromJson<Dictionary<string, string>>(), StringComparer.OrdinalIgnoreCase),
+                _ => null
+            };
+
+            if (bindingDataContent is null)
```
jviau, line 105: Can simplify to:

``` CSharp
return bindingDataContent is not null;
```

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,235 @@
...
+            content.TryGetValue(Constants.ContainerName, out var containerName);
+            content.TryGetValue(Constants.BlobName, out var blobName);
+
+            if (string.IsNullOrEmpty(connectionName) || string.IsNullOrEmpty(containerName))
+            {
+                return new ValueTask<ConversionResult>(ConversionResult.Unhandled());
+            }
+
+            var connectionString = connectionName is null ? null : Environment.GetEnvironmentVariable(connectionName);
+            var result = ToTargetType(targetType, connectionString, containerName, blobName, context).GetAwaiter().GetResult();
```
jviau, line 73: I believe this function and its caller / call chain can be made properly `async`.

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,235 @@
...
+            Type _ when targetType == typeof(AppendBlobClient) => CreateBlobReference<AppendBlobClient>(connectionString, containerName, blobName),
+            Type _ when targetType == typeof(BlobContainerClient) => CreateBlobContainerClient(connectionString, containerName),
+            _ => await DeserializeToTargetObject(targetType, connectionString, containerName, blobName)
+        };
+
+        private IEnumerable<object> ToTargetCollectionType(Type targetType, IEnumerable<object> array) => targetType switch
+        {
+            Type _ when targetType == typeof(BlobClient) => array.Select(p => (BlobClient)p),
+            Type _ when targetType == typeof(String) => array.Select(p => GetBlobContentString((BlobClient)p).GetAwaiter().GetResult()),
+            _ => throw new Exception()
```
jviau, line 130: nit: use a more applicable exception type.

extensions/Worker.Extensions.Storage.Blobs/src/Constants.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    internal static class Constants
+    {
+        internal const string BlobExtensionName = "AzureStorageBlobs";
+        internal const string Connection = "Connection";
+        internal const string ContainerName = "ContainerName";
+        internal const string BlobName = "BlobName";
+
+        // Media content types
+        internal const string JsonContentType = "application/json";
```
jviau, line 14: I think this constant is available: https://learn.microsoft.com/en-us/dotnet/api/system.net.mime.mediatypenames.application.json?view=net-7.0

extensions/Worker.Extensions.Storage.Blobs/src/Worker.Extensions.Storage.Blobs.csproj
```diff
@@ -4,9 +4,10 @@
     <AssemblyName>Microsoft.Azure.Functions.Worker.Extensions.Storage.Blobs</AssemblyName>
     <RootNamespace>Microsoft.Azure.Functions.Worker.Extensions.Storage.Blobs</RootNamespace>
     <Description>Azure Blob Storage extensions for .NET isolated functions</Description>
-    
+
     <!--Version information-->
-    <VersionPrefix>5.0.1</VersionPrefix>
+    <VersionPrefix>5.0.2</VersionPrefix>
```
jviau, line 9: Should this be a minor version rev as you are introducing new features? `5.1.0`

jviau (commented):
I think the entry point of `ConvertAsync` can be made `async`, then it would be good to do a pass through the call chain in `BlodStorageConverter` to make sure it is all `async` as appropriate, removing the `GetAwait().GetResult()` calls.

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,235 @@
...
+        };
+
+        private async Task<object?> DeserializeToTargetObject(Type targetType, string connectionString, string containerName, string blobName)
+        {
+            var content = await GetBlobString(connectionString, containerName, blobName);
+            return JsonConvert.DeserializeObject(content, targetType);
```
kshyju, line 136: I see we are using Newtonsoft.JSON.NET for deserialization here. Is that intentional? Wonder we should use the [serializer available on the worker option](https://github.com/Azure/azure-functions-dotnet-worker/blob/853fc0a0a8ff48464dda7b49aa7004410022fb12/src/DotNetWorker.Core/Hosting/WorkerOptions.cs#L18)?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,235 @@
...
+            return download.Value.Content;
+        }
+
+        private BlobContainerClient CreateBlobContainerClient(string connectionString, string containerName)
+        {
+            BlobContainerClient container = new(connectionString, containerName);
+            return container;
+        }
+
+        private T CreateBlobReference<T>(string connectionString, string containerName, string blobName, ConverterContext context = null) where T : BlobBaseClient
```
kshyju, line 174: should connectionString be nullable?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,235 @@
...
+        {
+            if (string.IsNullOrEmpty(blobName))
+            {
+                throw new ArgumentNullException(nameof(blobName));
+            }
+
+            BlobBaseClient blob = null;
```
kshyju, line 181: `BlobBaseClient? blob = null;` ?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,235 @@
...
+
+                if (result is not null)
+                {
+                    return new ValueTask<ConversionResult>(ConversionResult.Success(result));
+                }
+            }
+
+            return new ValueTask<ConversionResult>(ConversionResult.Unhandled());
+        }
+
```
kshyju, line 49: nit: extra line break.

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,235 @@
...
+
+        private object? ConvertModelBindingDataAsync(ConverterContext context, Type targetType, ModelBindingData bindingData)
+        {
+            if (bindingData.Source is not Constants.BlobExtensionName)
+            {
+                return new ValueTask<ConversionResult>(ConversionResult.Unhandled());
```
kshyju, line 55: line 55 and 60 causes boxing since the method return type is object. Is there anyway we can avoid that?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,235 @@
...
+            content.TryGetValue(Constants.Connection, out var connectionName);
+            content.TryGetValue(Constants.ContainerName, out var containerName);
+            content.TryGetValue(Constants.BlobName, out var blobName);
+
+            if (string.IsNullOrEmpty(connectionName) || string.IsNullOrEmpty(containerName))
+            {
+                return new ValueTask<ConversionResult>(ConversionResult.Unhandled());
+            }
+
+            var connectionString = connectionName is null ? null : Environment.GetEnvironmentVariable(connectionName);
```
kshyju, line 72: Is `connectionName is null` necessary?  Is line 67 already not handing that case?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,235 @@
...
+        {
+            BlobContainerClient container = new(connectionString, containerName);
+            return container;
+        }
+
+        private T CreateBlobReference<T>(string connectionString, string containerName, string blobName, ConverterContext context = null) where T : BlobBaseClient
+        {
+            if (string.IsNullOrEmpty(blobName))
+            {
+                throw new ArgumentNullException(nameof(blobName));
```
kshyju, line 178: Is any of the public methods in the caller chain handling this exception being thrown (catching and returning a ConversionStatus.Failed()?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,235 @@
...
+                {
+                    // For user assigned identity
+                    blob = new BlobClient(new Uri(endpoint),
+                                new ClientSecretCredential(
+                                    Environment.GetEnvironmentVariable(Constants.ConnectionTenantId),
+                                    Environment.GetEnvironmentVariable(Constants.ConnectionClientId),
+                                    Environment.GetEnvironmentVariable(Constants.ConnectionClientSecret)
+                                ));
+                }
+                else if (Environment.GetEnvironmentVariable(Constants.ConnectionCredential) == "ManagedIdentity")
```
kshyju, line 208: Should this be case insensitive string comparison?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
liliankasem: Good question, will need to test this out and confirm

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
liliankasem: Moved up to ConvertAsync

extensions/Worker.Extensions.Storage.Blobs/src/Constants.cs
liliankasem: Yeah I tried using that before but we're on .net standard 2.0 and `Json` was added in 2.1

When I use `MediaTypeNames.Application.Json` I get this error: `MediaTypeNames.Application' does not contain a definition for 'Json' `

We can maybe change the target to <TargetFramework>netstandard2.1</TargetFramework> but I'm not sure of the impact of that

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -151,7 +144,9 @@ private bool TryGetBindingDataContent(ModelBindingData bindingData, out IDiction
         private IEnumerable<object> ToTargetTypeCollection(Type targetType, IEnumerable<object> blobCollection) => targetType switch
         {
             Type _ when targetType == typeof(BlobBaseClient) => blobCollection.Select(b => (BlobBaseClient)b),
-            Type _ when targetType == typeof(String) => blobCollection.Select(async (b) => await GetBlobContentString((BlobClient)b)),
+            Type _ when targetType == typeof(BlobClient) => blobCollection.Select(b => (BlobClient)b),
+            Type _ when targetType == typeof(String) => blobCollection.Select((b) =>  ((String)b)),
```
jviau, line 148: small nit:
```suggestion
            Type _ when targetType == typeof(String) => blobCollection.Select(b =>  (string)b),
```

jviau (approved):

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,241 @@
...
+                        modelBindingData);
+
+                    if (element is not null)
+                    {
+                        collectionBlob.Add(element);
+                    }
+                }
+
+                var collectionResult = context.TargetType.IsArray ?
+                    ToTargetTypeArray(context.TargetType, collectionBlob) :
```
liliankasem, line 86: todo: look into using `.ToArray<T>`

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,241 @@
...
+            }
+
+            return true;
+        }
+
+        private bool TryGetBindingDataContent(ModelBindingData bindingData, out IDictionary<string, string> bindingDataContent)
+        {
+            bindingDataContent = bindingData.ContentType switch
+            {
+                Constants.JsonContentType => new Dictionary<string, string>(bindingData.Content.ToObjectFromJson<Dictionary<string, string>>(), StringComparer.OrdinalIgnoreCase),
```
liliankasem, line 114: Do we need to support more content-types for preview

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,241 @@
...
+            Type _ when targetType == typeof(BlobBaseClient) => CreateBlobClient<BlobBaseClient>(connectionName, containerName, blobName),
+            Type _ when targetType == typeof(BlobClient) => CreateBlobClient<BlobClient>(connectionName, containerName, blobName),
+            Type _ when targetType == typeof(BlockBlobClient) => CreateBlobClient<BlockBlobClient>(connectionName, containerName, blobName),
+            Type _ when targetType == typeof(PageBlobClient) => CreateBlobClient<PageBlobClient>(connectionName, containerName, blobName),
+            Type _ when targetType == typeof(AppendBlobClient) => CreateBlobClient<AppendBlobClient>(connectionName, containerName, blobName),
+            Type _ when targetType == typeof(BlobContainerClient) => CreateBlobContainerClient(connectionName, containerName),
+            _ => await DeserializeToTargetObject(targetType, connectionName, containerName, blobName)
+        };
+
+        private IEnumerable<object> ToTargetTypeCollection(Type targetType, IEnumerable<object> blobCollection) => targetType switch
```
liliankasem, line 150: Wondering if this can be simplified the same way we're using generics for `CreateBlobClient<T>`

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
liliankasem: Not for preview, only if required by a specific extension

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,272 @@
...
+            Type _ when targetType == typeof(BlobClient[]) => blobCollection.Select((b) => ((BlobClient)b)).ToArray(),
+            Type _ when targetType == typeof(String[]) => blobCollection.Select((b) => ((String)b)).ToArray(),
+            Type _ when targetType == typeof(Stream[]) => blobCollection.Select((b) => ((Stream)b)).ToArray(),
+            Type _ when targetType == typeof(Byte[][]) => blobCollection.Select((b) => ((Byte[])b)).ToArray(),
+            Type _ when targetType == typeof(BlockBlobClient[]) => blobCollection.Select((b) => ((BlockBlobClient)b)).ToArray(),
+            Type _ when targetType == typeof(PageBlobClient[]) => blobCollection.Select((b) => ((PageBlobClient)b)).ToArray(),
+            Type _ when targetType == typeof(AppendBlobClient[]) => blobCollection.Select((b) => ((AppendBlobClient)b)).ToArray(),
+            Type _ when targetType == typeof(BlobContainerClient) => blobCollection.Select((b) => ((BlobClient)b)).ToArray(),
+            _ => null
+        };
```
liliankasem, line 184: nit on the collection & array switch expressions: if would be neat if they were in the same order as `ToTargetType` or we change `ToTargetType` to match the order here

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,272 @@
...
+        public static T[] CloneToArray<T>(IList<object> source)
+        {
+            var result = source.Cast<T>().ToList();
+            return result.ToArray();
+        }
+
+        public static IEnumerable<T> CloneToList<T>(IList<object> source)
+        {
+            return source.Cast<T>().ToList();
+        }
```
liliankasem, line 213: Wondering if we should create a utilities sort of file for things like this, I think this class is getting rather large and breaking things down might help us better maintain this. Esp if we might be adding output binding support

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,272 @@
...
+
+        private async Task<object?> DeserializeToTargetObject(Type targetType, string connectionName, string containerName, string blobName)
+        {
+            var content = await GetBlobStream(connectionName, containerName, blobName);
+            return _workerOptions.Value.Serializer.Deserialize(content, targetType, CancellationToken.None);
+        }
+
+        private object? DeserializeToTargetObjectCollection(IEnumerable<object> blobCollection, Type targetType)
+        {
+            (string methodName, Type type) = targetType.IsArray ? (nameof(CloneToArray), targetType.GetElementType())
```
liliankasem, line 194: We're checking `targetType.IsArray` a couple times (maybe 3?) - I wonder if there is a way to refactor some of this so that we only need to check IsArray once?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,272 @@
...
+                                                            : (nameof(CloneToList), targetType.GenericTypeArguments[0]);
+
+            blobCollection = blobCollection.Select(b => Convert.ChangeType(b, type));
+            MethodInfo method = typeof(BlobStorageConverter).GetMethod(methodName);
+            MethodInfo genericMethod = method.MakeGenericMethod(type);
+
+            return genericMethod.Invoke(null, new[] {blobCollection.ToList()});
```
liliankasem, line 201:
```suggestion
            return genericMethod.Invoke(null, new[] { blobCollection.ToList() } );
```

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,272 @@
...
+        }
+        
+        public static T[] CloneToArray<T>(IList<object> source)
+        {
+            var result = source.Cast<T>().ToList();
+            return result.ToArray();
```
liliankasem, line 207: I'm a little nervous about the number of conversions here, to list -> to array feels wrong to me. Happy to help look into this but don't have any answers at the top of my head

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,272 @@
...
+            Type _ when targetType == typeof(BlockBlobClient) => CreateBlobClient<BlockBlobClient>(connectionName, containerName, blobName),
+            Type _ when targetType == typeof(PageBlobClient) => CreateBlobClient<PageBlobClient>(connectionName, containerName, blobName),
+            Type _ when targetType == typeof(AppendBlobClient) => CreateBlobClient<AppendBlobClient>(connectionName, containerName, blobName),
+            Type _ when targetType == typeof(BlobContainerClient) => CreateBlobContainerClient(connectionName, containerName),
+            _ => await DeserializeToTargetObject(targetType, connectionName, containerName, blobName)
+        };
+
+        private IEnumerable<object>? ToTargetTypeCollection(Type targetType, IEnumerable<object> blobCollection) => targetType switch
+        {
+            Type _ when targetType == typeof(IEnumerable<BlobBaseClient>) => blobCollection.Select(b => (BlobBaseClient)b),
```
jviau, line 160: Just to be clear - we expect `targetType` to only ever be `IEnumerable<T>` here right? Not something like `ICollection<T>`?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
jviau: Can you drop the `ToList()` entirely and do `source.Cast<T>().ToArray()`?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
jviau: You could if you want, but the way our extension files are structured would require a shared internals project. Might be more work than it is worth. Also, just inlining `*.Cast<T>().ToList()` instead of an extension method isn't the worst either.

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,272 @@
...
+            var result = source.Cast<T>().ToList();
+            return result.ToArray();
+        }
+
+        public static IEnumerable<T> CloneToList<T>(IList<object> source)
+        {
+            return source.Cast<T>().ToList();
+        }
+
+        private async Task<string> GetBlobString(string connectionName, string containerName, string blobName)
```
jviau, line 215: nit: add `Async` suffix to all `Task`-returning methods.

extensions/Worker.Extensions.Storage.Blobs/src/Constants.cs
fabiocav: We can't move to `netstandard2.1` here as it would impact our .NET Framework support. For this simple change/API usage, I would also not consider any of the other options, as the benefit here wouldn't justify the added complexity.

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
liliankasem: I think the extension method currently exists so it can be called by reflection, any ideas on how we can improve that?

```csharp
        private object? DeserializeToTargetObjectCollection(IEnumerable<object> blobCollection, Type targetType)
        {
            (string methodName, Type type) = targetType.IsArray
                                            ? (nameof(CloneToArray), targetType.GetElementType())
                                            : (nameof(CloneToList), targetType.GenericTypeArguments[0]);

            blobCollection = blobCollection.Select(b => Convert.ChangeType(b, type));
            MethodInfo method = typeof(BlobStorageConverter).GetMethod(methodName);
            MethodInfo genericMethod = method.MakeGenericMethod(type);

            return genericMethod.Invoke(null, new[] { blobCollection.ToList() });
        }
```

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
liliankasem: Yup, as far as I'm aware, that's all we advertise as what we support - mostly going off the sdk docs: https://github.com/surgupta-msft/azure-sdk-for-net/tree/main/sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Blobs

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,273 @@
...
+
+        private static T[] CloneToArray<T>(IList<object> source)
+        {
+            var result = source.Cast<T>().ToList();
+            return result.ToArray();
+        }
+
+        private static IEnumerable<T> CloneToList<T>(IList<object> source)
+        {
+            return source.Cast<T>().ToList();
```
liliankasem, line 213: Is the purpose of this to just change IList<object> to IList<T> - does it matter too much? What happens if we return IList<object> to users?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
liliankasem: Looking up the stack, I don't see anything handling the exception. Do you suggest we return a failure status here? We could turn this into `TryCreateBlobClient` and return a bool and have the bubble up to ConvertAsync to return Failure result?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
jviau: That depends how this is used. If this is for a function parameter binding then whatever we return must be assignable to the declared parameter type. If they declare `IEnumerable<SomeType> input` then `IList<object>` is **not** assignable and would throw. The `T` must be `SomeType` or a more derived class of `SomeType`.

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,273 @@
...
+            Type _ when targetType == typeof(BlockBlobClient) => CreateBlobClient<BlockBlobClient>(connectionName, containerName, blobName),
+            Type _ when targetType == typeof(PageBlobClient) => CreateBlobClient<PageBlobClient>(connectionName, containerName, blobName),
+            Type _ when targetType == typeof(AppendBlobClient) => CreateBlobClient<AppendBlobClient>(connectionName, containerName, blobName),
+            Type _ when targetType == typeof(BlobContainerClient) => CreateBlobContainerClient(connectionName, containerName),
+            _ => await DeserializeToTargetObjectAsync(targetType, connectionName, containerName, blobName)
+        };
+
+        private object[]? ToTargetTypeArray(Type targetType, IEnumerable<object> blobCollection) => targetType switch
+        {
+            Type _ when targetType == typeof(String[]) => blobCollection.Select(b => (string)b).ToArray(),
```
jviau, line 160: Can this implementation be replaced with `ToTargetTypeCollection(targetType, blobCollection).ToArray()`? Or can even the usage of `ToTargetTypeArray` be replaced entirely with something around calling `ToArray()` on the result of `ToTargetTypeCollection`?

jviau (changes requested):
The primary concerns I have are `ConfigurationExtensions` being public and add the package to `Worker.Extensions.Abstractions`.

extensions/Worker.Extensions.Abstractions/src/Configuration/ConfigurationExtensions.cs
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Abstractions
+{
+    public static class ConfigurationExtensions
```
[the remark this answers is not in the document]
liliankasem: Yeah I think we will need to use this for all of the extensions as we start adding sdk binding support, would love to discuss improvements here

extensions/Worker.Extensions.Abstractions/src/Worker.Extensions.Abstractions.csproj
```diff
@@ -11,4 +11,8 @@
 
   <Import Project="..\..\..\build\Common.props" />
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Azure" Version="1.1.0" />
```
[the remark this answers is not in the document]
liliankasem: Yeah we do need it for the configuration extension

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
surgupta-msft: Dropping ToList() works. Updated code here.

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,272 @@
...
+                                            : (nameof(CloneToList), targetType.GenericTypeArguments[0]);
+
+            blobCollection = blobCollection.Select(b => Convert.ChangeType(b, type));
+            MethodInfo method = typeof(BlobStorageConverter).GetMethod(methodName);
+            MethodInfo genericMethod = method.MakeGenericMethod(type);
+
+            return genericMethod.Invoke(null, new[] { blobCollection.ToList() });
+        }
+
+        public static T[] CloneToArray<T>(IList<object> source)
```
surgupta-msft, line 205: Updated both methods (`CloneToArray`, `CloneToList`) to public from private for keeping E2E working. We can discuss better approach here.

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
surgupta-msft: Yes, seeing exception similar to what Jacob has mentioned.

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
liliankasem: @jviau - just in case you missed this!

extensions/Worker.Extensions.Abstractions/src/Worker.Extensions.Abstractions.csproj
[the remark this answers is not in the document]
jviau: Is it this package that is needed, or `Microsoft.Extensions.Configuration`?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
jviau: Ah I see how these are used. I can't really think of anything better off the top of my head. It would be nice to move these to a shared extension. I'll start an offline chat with the team to see how we want to do that.

extensions/Worker.Extensions.Abstractions/src/Worker.Extensions.Abstractions.csproj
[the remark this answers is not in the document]
liliankasem: Ah yup you're right, I must have had something else in there that used Azure but was removed. Thanks!

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
jviau: Why did you need to update them? Was the `GetMethod` reflection call not finding them? If so, you need to pass in some flags to `GetMethod` to tell it to search for `private static`.

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
liliankasem: Added a catch!

extensions/Worker.Extensions.Abstractions/src/Configuration/ConfigurationExtensions.cs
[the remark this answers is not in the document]
liliankasem: Moved to shared!

extensions/Worker.Extensions.Storage.Blobs/src/Properties/AssemblyInfo.cs
```diff
@@ -1,6 +1,6 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
-﻿using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
+using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
 
-[assembly: ExtensionInformation("Microsoft.Azure.WebJobs.Extensions.Storage.Blobs", "5.0.0")]
+[assembly: ExtensionInformation("Microsoft.Azure.WebJobs.Extensions.Storage.Blobs", "5.2.0-alpha.20221103.1")]
```
liliankasem, line 6: Use public package when ready (eta 7th feb)

extensions/Worker.Extensions.Storage.Blobs/src/BlobOutputAttribute.cs
```diff
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
```
fabiocav, line 3: nit: Consider removing from this PR

extensions/Worker.Extensions.Shared/Worker.Extensions.Shared.csproj
```diff
@@ -17,7 +17,7 @@
 
   <ItemGroup>
     <!-- Include packages here as appropriate so that intellisense still works. -->
-    <!-- <PackageReference Include="PackageId" Version="PackageVersion" /> -->
+    <PackageReference Include="Microsoft.Extensions.Configuration" Version="7.0.0" />
```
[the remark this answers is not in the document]
liliankasem: It does cause intellisense issues in VS Code which just makes it less fun to deal with, but I can remove this reference if it causes issues!

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,250 @@
...
+            }
+
+            return true;
+        }
+
+        private bool TryGetBindingDataContent(ModelBindingData bindingData, out IDictionary<string, string> bindingDataContent)
+        {
+            bindingDataContent = bindingData?.ContentType switch
+            {
+                Constants.JsonContentType => new Dictionary<string, string>(bindingData?.Content?.ToObjectFromJson<Dictionary<string, string>>(), StringComparer.OrdinalIgnoreCase),
```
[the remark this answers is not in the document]
liliankasem: That's a good point, if an extension ever changes or provides a different content type we would have to handle that after manually updating the version of the extension we're using. So right now we know for sure we only get Json.

Right now, we fail if we don't get a match - I can change this to throw? Or are you suggesting we don't even need the switch statement here?

RE: serialization failure - that's a great catch we should definitely be catching that. Might be better to have the `trycatch` in the calling method as here we're not expected to return a ConversionResult. Or we would need to have the `try-catch` in here and return the exception details as well. Right now, if bindingDataContent is null, we fail

extensions/Worker.Extensions.Shared/Configuration/ConfigurationExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    internal static class ConfigurationExtensions
+    {
+        private const string WebJobsConfigurationSectionName = "AzureWebJobs";
+        private const string ConnectionStringsConfigurationSectionName = "ConnectionStrings";
+
+        internal static IConfigurationSection GetWebJobsConnectionStringSection(this IConfiguration configuration, string connectionName)
```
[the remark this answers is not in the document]
liliankasem: Documented and made private!

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,250 @@
...
+                _logger.LogTrace("Source '{source}' is not supported by {converter}", bindingData?.Source, nameof(BlobStorageConverter));
+                return false;
+            }
+
+            return true;
+        }
+
+        private bool TryGetBindingDataContent(ModelBindingData bindingData, out IDictionary<string, string> bindingDataContent)
+        {
+            bindingDataContent = bindingData?.ContentType switch
```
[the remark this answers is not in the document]
liliankasem: I think this was fixed my me throwing instead?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,250 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
```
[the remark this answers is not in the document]
liliankasem: Do we have a test project for extensions? Wondering if we need to setup something new or if we can put tests in `test/DotNetWorkerTests/Converters`?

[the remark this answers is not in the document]
liliankasem: @fabiocav I'll work on tests in a new PR to avoid blocking this release - is that okay? Otherwise will priortize tomorrow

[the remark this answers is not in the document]
fabiocav: That's OK for this change. We'll definitely want this in place before GA/RCs, but as long as we have an issue tracking, I'm good with flagging this as resolved.

extensions/Worker.Extensions.Shared/Configuration/ConfigurationExtensions.cs
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions
+{
+    internal static class ConfigurationExtensions
```
jviau, line 8: To address your concern we discussed offline: `PackageReference` is **not** transitive when using `SharedReference`. When `Worker.Extensions.Shared.csproj` includes a `PackageReference`, any projects point to this via a `SharedReference` will **not** implicitly get that `PackageReference` as well. The referencing projects must include the appropriate dependencies themselves. This means keeping the packages consistent between the shared project and the referencing projects is a manual effort by us devs. But mistakes will be readily apparent via build errors.

Including the `PackageReference` in `Worker.Extensions.Shared.csproj` is optional and is only for design time assembly resolution in VS / VS Code so that intellisense works properly. However, I don't think that is even functioning due to the use of `Microsoft.Build.NoTargets` SDK here. Definitely not a perfect experience, so it is fine to leave it out for now. I might find some time and see if I can give a more optimal development experience for these shared projects.

For example, I added `Microsoft.Extensioms.Configuration/7.0.0` to the shared project, then built and packed the Blob project (from your PR). Here are the nuget dependencies of that package:

image

Notice there is no `Microsoft.Extensions.Configuration (>= 7.0.0)`? That is because `Worker.Extensions.Storage.Blob` does **not** depend on it - even though `Worker.Extensions.Shared` does depend on it.

jviau (approved):

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,236 @@
...
+
+            return true;
+        }
+
+        private Dictionary<string, string> GetBindingDataContent(ModelBindingData bindingData)
+        {
+            return bindingData?.ContentType switch
+            {
+                Constants.JsonContentType => new Dictionary<string, string>(bindingData?.Content?.ToObjectFromJson<Dictionary<string, string>>(), StringComparer.OrdinalIgnoreCase),
+                _ => throw new InvalidOperationException($"Unexpected content-type. Currently only {Constants.JsonContentType} is supported.")
```
liliankasem, line 122: Not Supported

Issue #1056, opened 2022-09-20
[AOT] Make Isolated worker AOT friendly


kshyju:
This issue is to track the progress of making isolated worker AOT friendly.

Task list

- [ ] #1053 
- [ ] #1054 
- [ ] #1055

kshyju:
See Team's update in https://github.com/Azure/azure-functions-dotnet-worker/issues/2428#issuecomment-2620277971

PR #1109, opened 2022-10-06
Cosmos DB converter for SDK-type support and samples

### LINKED ISSUE #1034: Implement Cosmos DB rules and converter
Expand the existing sdk-type binding prototype to also include hydrating a Cosmos DB client.

- [x] Create input and trigger binding rules in the [webjobs extension for cosmos ](https://github.com/surgupta-msft/azure-sdk-for-net/tree/main/sdk/cosmosdb)
- [x] Write a converter in the [worker extension](https://github.com/Azure/azure-functions-dotnet-worker/tree/main/extensions/Worker.Extensions.CosmosDB/src) to hydrate the cosmos DB client
- [x] Support POCO bindings for the trigger

liliankasem:

resolves #1034

### Pull request checklist

### Additional information

Additional PR information


extensions/Worker.Extensions.CosmosDB/src/Worker.Extensions.CosmosDB.csproj
```diff
@@ -17,6 +17,8 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\Worker.Extensions.Abstractions\src\Worker.Extensions.Abstractions.csproj" />
+    <ProjectReference Include="..\..\..\src\DotNetWorker.Core\DotNetWorker.Core.csproj" />
+    <PackageReference Include="Azure.Cosmos" Version="4.0.0-preview3" />
```
jviau, line 21: CosmosDB v4 has not been actively developed for a while. While [v3](https://www.nuget.org/packages/Microsoft.Azure.Cosmos) is still being worked on. Should we use v3 instead? Or does this extension only work with v4?

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
```diff
@@ -0,0 +1,85 @@
...
+            return content.ToObjectFromJson<T>();
+        }
+
+        private object CreateCosmosReference<T>(BinaryData content)
+        {
+            var cosmosAttribute = content.ToObjectFromJson<CosmosDBInputAttribute>();
+            var connectionString = Environment.GetEnvironmentVariable(cosmosAttribute?.Connection);
+
+            Type targetType = typeof(T);
+            CosmosClient cosmosClient = new (connectionString);
```
jviau, line 74: style nit:

```suggestion
            CosmosClient cosmosClient = new(connectionString);
```

I think @brettsam has updated (or has a PR to do so, or did it in the host... can't remember exactly) to update our StyleCop package to a newer version to fix the warning on `new()` without the space.

jviau, line 74: Do we have plans to support identity-based cosmos clients?

jviau (commented):
I think the main concern is the use of preview `Azure.Cosmos` over `Microsoft.Azure.Cosmos`.

If we have to choose one, I would say v3. If we want to support both, we either need separate packages or get crazy with some reflection. And don't reference either package! discover and cache the types/ctors we want via reflection.

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
```diff
@@ -0,0 +1,85 @@
...
+            Type _ when targetType == typeof(CosmosContainer)           => CreateCosmosReference<CosmosContainer>(content),
+            _ => CreateTargetObject(targetType, content)
+        };
+
+        private object CreateTargetObject(Type targetType, BinaryData content)
+        {
+            MethodInfo deserializeObjectMethod = GetType()
+                                                .GetMethod(nameof(DeserializeTargetObject), BindingFlags.Instance | BindingFlags.NonPublic)
+                                                .MakeGenericMethod(new Type[] { targetType });
+            return deserializeObjectMethod.Invoke(this, new object[] { content });
```
jviau, line 60: Interesting `BinaryData` does not expose this itself, but we can roll our own extension method to avoid the reflection here.

Here is the `ToObjectFromJson<T>` implementation of BinaryData. It already is just using `typeof(T)` so it could so easily have a non-generic overload which takes in `Type type`.

``` CSharp
    public T ToObjectFromJson<T>(JsonSerializerOptions? options = null)
    {
        return (T)JsonSerializer.Deserialize(_bytes.Span, typeof(T), options);
    }
```

We can add:

``` CSharp
    public static object? ToObjectFromJson(this BinaryData data, Type type, JsonSerializerOptions? options = null)
    {
        return JsonSerializer.Deserialize(_bytes.Span, type, options);
    }
```

liliankasem: Was trying to figure out how to avoid reflection so definitely appreciate the suggestion here - I'll try it out!

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
liliankasem: Yup! Working on it :)

extensions/Worker.Extensions.CosmosDB/src/Worker.Extensions.CosmosDB.csproj
liliankasem: Let me get back to you on this one. When we started this work we always had the intension of targeting v4 (was supposed to be released in November). I can sync with the v-team on this topic

liliankasem: I had a chat with the Cosmos team and we should indeed be using v3!

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
```diff
@@ -0,0 +1,221 @@
...
+            while (iterator.HasMoreResults)
+            {
+                FeedResponse<T> response = await iterator.ReadNextAsync();
+                IEnumerable<T> documents = response.Resource;
+                documentList.AddRange(documents);
+            }
+            return documentList;
+        }
+
+        private object CreateCosmosClient<T>(CosmosDBInputAttribute cosmosAttribute)
```
jviau, line 175: Can the return type be changed to `T`?

extensions/Worker.Extensions.CosmosDB/src/CosmosDBTriggerAttribute.cs
```diff
@@ -1,7 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
-﻿using System;
+using System;
```
jviau, line 4: Is this a newline CRLF -> LF or something like that?

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
```diff
@@ -0,0 +1,221 @@
...
+        {
+            if (cosmosAttribute is null)
+            {
+                // What do?
+                throw new InvalidOperationException("Cosmos attribute cannot be null");
+            }
+
+            CosmosClient cosmosClient;
+            var connectionString = Environment.GetEnvironmentVariable(cosmosAttribute.Connection);
```
jviau, line 184: Just to make sure - are you still planning on switching to `IConfiguration` for this? Looking at your `Constants.cs` file, all your properties are `AzureWebJobsStorage__*`, so `IOptions<T>` would actually be a perfect fit!

``` CSharp
internal class CosmosBindingOptions
{
    public string AccountName { get; set; }
    
    // other properties also
}

// in extension startup:
applicationBuilder.Services.AddOptions<CosmosBindingOptions>().BindConfiguration("AzureWebJobsStorage");

// inject IOptions into converter:

public CosmosDbConverter(IOptions<CosmosBindingOptions> options)
{
}
```

liliankasem: Yup that's the plan! Will work on that next after making sure managed ID is working as expected

extensions/Worker.Extensions.CosmosDB/src/CosmosDBTriggerAttribute.cs
liliankasem: I think so! It was showing up as a character that looked like this`<?>`

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
liliankasem: @jviau  I went with a slightly different approach here, still using IConfiguration, thoughts?

jviau: I would recommend using `IOptions` here, as it is more flexible that `IConfiguration`. Maybe even make the options class public. This will be beneficial for two reasons:

1. Unit testing! You now only need to create an options object, no need to set up `IConfiguration`.
2. Customers can configure the options object in a variety of ways, not just via `IConfiguration`.

My opinion is that you should only very sparingly consume `IConfiguration` directly in code. In most cases you want to use options pattern - this separates the concern between how you consume config/settings and where/how those settings are built/populated.

jviau: An additional benefit, is you can even move the `TokenCredential` resolution from `AzureComponentFactory` into the options configuration, which simplifies your cosmos DB factory, as it will have a built `TokenCredential` provided to it.

extensions/Worker.Extensions.CosmosDB/src/Constants.cs
```diff
@@ -0,0 +1,14 @@
...
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    internal static class Constants
+    {
+        internal const string ExtensionName = "CosmosDB";
+        internal const string ConfigurationSectionName = "AzureWebJobs";
+        internal const string ConnectionStringsSectionName = "ConnectionStrings";
```
jviau, line 11: I think `IConfiguration` has an extension method for this: [IConfiguration.GetConnectionString(string)](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.configurationextensions.getconnectionstring?view=dotnet-plat-ext-7.0#microsoft-extensions-configuration-configurationextensions-getconnectionstring(microsoft-extensions-configuration-iconfiguration-system-string))

extensions/Worker.Extensions.CosmosDB/src/Config/ICosmosDBServiceFactory.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.Cosmos;
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    internal interface ICosmosDBServiceFactory
+    {
+        CosmosClient CreateService(string connectionString, CosmosClientOptions cosmosClientOptions);
```
jviau, line 10: nit: the implementation has this parameter as `connectionName`. Can you make them consistent? Whichever is a more suitable name (I think `connectionName` is more appropriate).

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
```diff
@@ -0,0 +1,201 @@
...
+        {
+            // if target type is a collection and NOT of type IList, should we handle this early
+            // and let users know we only support IList types?
+
+            if (targetType.GenericTypeArguments.Any())
+            {
+                targetType = targetType.GenericTypeArguments.FirstOrDefault();
+            }
+
+            MethodInfo createPOCOFromReferenceMethod = GetType()
```
jviau, line 117: Needing to use reflection here is less than ideal. Is it possible to do it without reflection? For example, items can be read without generics via `Container.ReadItemStreamAsync`, which will give the item as a `Stream`, which leaves deserializing as the next challenge. Do we use the serializer from Cosmos? Do we have access to that (I think it is just default JSON)? Or do we use the serializer from `WorkerOptions`? And is that worth it or do we live with the reflection?

I'll leave this up to you. I would like to see the reflection removed as it is always nice to avoid it. But the potential serialization inconsistency may be even worse.

liliankasem: Yeah that's a fair point - I think either way handling this whole POCO conversion is going to be difficult :( I can look into another approach but it might be worth a design discussion with the team

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
liliankasem: I'll be honest, I'm not exactly sure how this works or what it is supposed to look like but I will make an attempt! I'd love to setup time for us to chat about this so I can learn more!

extensions/Worker.Extensions.CosmosDB/src/CosmosExtensionStartup.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        public override void Configure(IFunctionsWorkerApplicationBuilder applicationBuilder)
+        {
+            if (applicationBuilder == null)
+            {
+                throw new ArgumentNullException(nameof(applicationBuilder));
+            }
+
+            applicationBuilder.Services.AddAzureClientsCore(); // Adds AzureComponentFactory
+            applicationBuilder.Services.AddOptions<CosmosDBBindingOptions>();
```
jviau, line 22: You could also add a `.Validate()` delegate to verify that _at least_ one of the two auth settings are configured correctly (conn string or AAD).

jviau (approved):

extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBBindingOptionsSetup.cs
```diff
@@ -0,0 +1,56 @@
...
+            {
+                options.ConnectionString = connectionSection.Value;
+            }
+            else
+            {
+                options.AccountEndpoint  = connectionSection[Constants.AccountEndpoint];
+                if (string.IsNullOrWhiteSpace(options.AccountEndpoint))
+                {
+                    // Not found
+                    throw new InvalidOperationException($"Connection should have an '{Constants.AccountEndpoint}' property or be a " +
```
jviau, line 48: Oh, I see you have the validation equivalent here. I recommend moving it to [Options Validation](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-7.0#options-validation).

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
```diff
@@ -0,0 +1,210 @@
...
+                    object result = await ToTargetType(context.TargetType, cosmosAttribute);
+
+                    if (result is not null)
+                    {
+                        return ConversionResult.Success(result);
+                    }
+                }
+                catch (Exception ex)
+                {
+                    // What do we want to do for error handling?
```
kshyju, line 49: log using ILogger ? You can DI `ILogger<CosmosDBConverter>` into the constructor of CosmosDBConverter

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
```diff
@@ -0,0 +1,210 @@
...
+                    Console.WriteLine(ex);
+
+                    if (ex is CosmosException docEx)
```
kshyju, line 52: Why throwing only for CosmosException?

kshyju: Should we return conversion failed response here?

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
kshyju: We are trying to avoid/rewrite reflection code as much as we can to make isolated AOT friendly. #1056 tracks the rewrite needed for existing code.

fabiocav:
@liliankasem are you planning on rebasing this PR and making it ready for review?

liliankasem:
> @liliankasem are you planning on rebasing this PR and making it ready for review?

Yes; currently deprioritized with focus on testing but will restart work on this PR soon

liliankasem:
Created an issue to restart work on this PR

- https://github.com/Azure/azure-functions-dotnet-worker/issues/1371

PR #1115, opened 2022-10-11
Update cancellation sample to show valid use case

liliankasem:
Cancellation tokens are signaled by the platform when there is a sudden restart or shutdown of the host, or when a function invocation exceeds its timeout limit.

Also using this PR to bump all samples to the latest worker version

### DIFF
--- a/samples/Net7Worker/EventHubCancellationToken.cs
+++ b/samples/Net7Worker/EventHubCancellationToken.cs
@@ -17,7 +17,7 @@ namespace Net7Worker
 
         [Function(nameof(EventHubCancellationToken))]
         public void Run(
-            [EventHubTrigger("src", Connection = "EventHubConnection")] string[] messages,
+            [EventHubTrigger("sample-workitems", Connection = "EventHubConnection")] string[] messages,
             FunctionContext context,
             CancellationToken cancellationToken)
         {
@@ -27,23 +27,13 @@ namespace Net7Worker
             {
                 foreach (var message in messages)
                 {
-                    if (cancellationToken.IsCancellationRequested)
-                    {
-                        _logger.LogInformation("A cancellation token was received. Taking precautionary actions.");
-                        // Take precautions like noting how far along you are with processing the batch
-                        _logger.LogInformation("Precautionary activities --complete--.");
-                        break;
-                    }
-                    else
-                    {
-                        // Business logic as usual
-                        _logger.LogInformation($"Message: {message} was processed.");
-                    }
+                    cancellationToken.ThrowIfCancellationRequested();
+                    _logger.LogInformation($"Message: {message} was processed.");
                 }
             }
-            catch (Exception ex)
+            catch (OperationCanceledException)
             {
-                _logger.LogInformation($"Something unexpected happened: {ex.Message}");
+                _logger.LogInformation("A cancellation token was received - invocation cancelled.");
             }
         }
     }
--- a/samples/Net7Worker/EventHubCancellationToken.cs
+++ b/samples/Net7Worker/EventHubCancellationToken.cs
@@ -1,0 +1,50 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Extensions.Logging;
+
+namespace Net7Worker
+{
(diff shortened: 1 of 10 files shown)


samples/Net7Worker/local.settings.json
```diff
@@ -2,6 +2,7 @@
   "IsEncrypted": false,
   "Values": {
     "FUNCTIONS_WORKER_RUNTIME": "dotnet-isolated",
-    "AzureWebJobsStorage": "UseDevelopmentStorage=true"
+    "AzureWebJobsStorage": "UseDevelopmentStorage=true",
+    "EventHubConnection": ""
```
cloudmelon, line 6: < your Eventhub connection string >

samples/Net7Worker/EventHubCancellationToken.cs
```diff
@@ -0,0 +1,50 @@
...
+                    else
+                    {
+                        // Business logic as usual
+                        _logger.LogInformation($"Message: {message} was processed.");
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                _logger.LogInformation($"Something unexpected happened: {ex.Message}");
```
fabiocav, line 46: Let's avoid a sample catching all exceptions. Would lead to function failures reported as success.

samples/Net7Worker/EventHubCancellationToken.cs
```diff
@@ -0,0 +1,50 @@
...
+        private readonly ILogger _logger;
+
+        public EventHubCancellationToken(ILoggerFactory loggerFactory)
+        {
+            _logger = loggerFactory.CreateLogger<EventHubCancellationToken>();
+        }
+
+        [Function(nameof(EventHubCancellationToken))]
+        public void Run(
+            [EventHubTrigger("src", Connection = "EventHubConnection")] string[] messages,
```
cloudmelon, line 20: "src" => would you use a more easy-understable name for the trigger ? "sample-workitems" is the default

samples/Net7Worker/EventHubCancellationToken.cs
```diff
@@ -0,0 +1,50 @@
...
+            FunctionContext context,
+            CancellationToken cancellationToken)
+        {
+            _logger.LogInformation("C# EventHub trigger function processing a request.");
+
+            try
+            {
+                foreach (var message in messages)
+                {
+                    if (cancellationToken.IsCancellationRequested)
```
fabiocav, line 30: The sample could simply throw at the beginning of the look (using `ThrowIfCancellationRequested`). You can document the pattern of checking `IsCancellationRequested` for scenarios where additional work/cleanup is required. But if we just to short circuit the loop, throwing would be sufficient.

samples/Net7Worker/EventHubCancellationToken.cs
```diff
@@ -0,0 +1,40 @@
...
+            FunctionContext context,
+            CancellationToken cancellationToken)
+        {
+            _logger.LogInformation("C# EventHub trigger function processing a request.");
+
+            try
+            {
+                foreach (var message in messages)
+                {
+                    cancellationToken.ThrowIfCancellationRequested();
```
SeanFeldman, line 30: Is the intent here to demonstrate a cancellation token invoked by the underlying SDK? If so, I'd have a hard time guessing that, given this is an isolated worker and the process doesn't work directly with the SDK. Perhaps the sample should have something in the readme.md to describe the possible scenarios where the cancellation token would be activated. Thank you.

liliankasem: I'll make a note in the readme, but the public docs should be updated with this information (and new sample) @cloudmelon

satvu (approved):

samples/Net7Worker/EventHubCancellationToken.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        private readonly ILogger _logger;
+
+        public EventHubCancellationToken(ILoggerFactory loggerFactory)
+        {
+            _logger = loggerFactory.CreateLogger<EventHubCancellationToken>();
+        }
+
+        // Sample showing how to handle a cancellation token being received
+        // In this example, the function invocation status will be "Cancelled"
```
surgupta-msft, line 19: I may be overlooking something, but are we asserting on _Cancelled_ function Invocation status anywhere?

surgupta-msft (approved):

PR #1130, opened 2022-10-18
[SourceGen] Add support for overriding binding property name

### LINKED ISSUE #1087: Adding support for overriding the binding property name in function metadata - SourceGen provider
Update FunctionMetadataProviderGenerator to support overriding binding property name. 

#1086 added support for this in the legacy metadata generation code path.

satvu:

resolves #1087

### Pull request checklist

### Additional information

[Example usage](https://github.com/Azure/azure-functions-dotnet-worker/blob/2b7986b81eb72da91468291b80a9cd51c470eb97/extensions/Worker.Extensions.Storage.Blobs/src/BlobTriggerAttribute.cs#L30) of the attribute that overrides binding property names. 

On the issue thread you can find a comment on the [design assumptions](https://github.com/Azure/azure-functions-dotnet-worker/issues/1087#issuecomment-1282658117). Given the limitations of Roslyn analyzers, this design relies on the convention that attribute constructor parameter names will match the name of the property they are assigned to (matching is NOT case-sensitive).

### DIFF
--- a/sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
+++ b/sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
@@ -535,7 +535,7 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
                 // that the constructor names would match the property names
                 for (int i = 0; i < attributeData.ConstructorArguments.Length; i++)
                 {
-                    string argumentName = attribMethodSymbol.Parameters[i].Name;// going to have to do something here to get the attribute on the parameter?
+                    string argumentName = attribMethodSymbol.Parameters[i].Name;
 
                     if (TryOverrideBindingName(attributeData.AttributeClass, argumentName, out string? newName))
                     {
--- a/sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
+++ b/sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
@@ -574,9 +574,8 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
                     {
                         var bindingNameAttrList = prop.GetAttributes().Where(attr => SymbolEqualityComparer.Default.Equals(attr.AttributeClass, bindingPropertyNameSymbol));
 
-                        if (bindingNameAttrList.SingleOrDefault() is { } bindingNameAttr)
+                        if (bindingNameAttrList.SingleOrDefault() is { } bindingNameAttr) // there will only be one BindingAttributeName attribute b/c there can't be duplicate attributes on a piece of syntax
                         {
-                            bindingNameAttr = bindingNameAttrList.FirstOrDefault(); // there will only be one BindingAttributeName attribute b/c there can't be duplicate attributes on a piece of syntax
                             argumentName = bindingNameAttr.ConstructorArguments.First().Value!.ToString(); // there is only one constructor argument for this binding attribute (the binding name override)
                         }
                     }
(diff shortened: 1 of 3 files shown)


test/Sdk.Generator.Tests/FunctionMetadataProviderGeneratorTests/StorageBindingTests.cs
```diff
@@ -236,7 +236,7 @@ public Task<ImmutableArray<IFunctionMetadata>> GetFunctionMetadataAsync(string d
                 name = 'blob',
                 type = 'BlobTrigger',
                 direction = 'In',
-                blobPath = 'container2/%file%',
+                path = 'container2/%file%',
```
satvu, line 239: This test has been changed to reflect that blob triggers should have an arg name of "path", not "blobPath". In line 201 above for the Blob output binding, we still keep the arg name "blobPath". 

[Blob bindings classess](https://github.com/Azure/azure-functions-dotnet-worker/tree/main/extensions/Worker.Extensions.Storage.Blobs/src) for reference.

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -562,6 +568,35 @@ private bool TryLoadConstructorArguments(AttributeData attributeData, IDictionar
                 return true;
             }
 
+            private bool TryOverrideBindingName(INamedTypeSymbol? attributeClass, string argumentName, out string? newName)
```
jviau, line 571: You can change this signature to a ref to remove the if statement usage of it:
```suggestion
            private void OverrideBindingName(INamedTypeSymbol? attributeClass, ref string argumentName)
```

with `ref` you can choose to update that value or not. So the usage will be:

``` CSharp
string argumentName = attribMethodSymbol.Parameters[i].Name;
OverrideBindingName(attributeData.AttributeClass, argumentName); // either argumentName will remain unchanged OR be updated to the overridden name at the end of this.
```

jviau (approved):
Looks good - just some minor suggestions.

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -562,6 +564,25 @@ private bool TryLoadConstructorArguments(AttributeData attributeData, IDictionar
...
+
+                foreach (var prop in attributeClass.GetMembers().Where(a => a is IPropertySymbol))
+                {
+                    if (String.Equals(prop.Name, argumentName, StringComparison.OrdinalIgnoreCase)) // relies on convention where constructor parameter names match the property their value will be assigned to (JSON serialization is a precedence for this convention)
+                    {
+                        var bindingNameAttrList = prop.GetAttributes().Where(attr => SymbolEqualityComparer.Default.Equals(attr.AttributeClass, bindingPropertyNameSymbol));
+
+                        if (bindingNameAttrList.SingleOrDefault() is { } bindingNameAttr)
+                        {
+                            bindingNameAttr = bindingNameAttrList.FirstOrDefault(); // there will only be one BindingAttributeName attribute b/c there can't be duplicate attributes on a piece of syntax
```
kshyju, line 579: Do we need this line? `bindingNameAttr` is already the single item from line 577. So 580 should just work.

kshyju (approved):

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
satvu: oh good catch let me update

PR #1131, opened 2022-10-18
[SourceGen] Handle default values in binding attributes

### LINKED ISSUE #1101: Handle DefaultValueAttribute when building function metadata in GeneratedFunctionMetadataProvider
We expose the [DefaultValueAttribute ](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/extensions/Worker.Extensions.Abstractions/src/DefaultValueAttribute.cs)which allows extension authors to set a default property value for triggers/bindings and this value will be used when generating binding part of the function metadata. 

Currently EventHubTrigger uses this.

https://github.com/Azure/azure-functions-dotnet-worker/blob/9c3380d36af1ed79040707f62cdf2141410c53de/extensions/Worker.Extensions.EventHubs/src/EventHubTriggerAttribute.cs#L43-L48

Looks like we missed porting this in the source gen metadata provider.

For reference, below is how legacy metadata provider getting the default attribute values when building bindings dictionary.

https://github.com/Azure/azure-functions-dotnet-worker/blob/9c3380d36af1ed79040707f62cdf2141410c53de/sdk/Sdk/Extensions/CustomAttributeExtensions.cs#L25-L32

satvu:

resolves #1101 

### Pull request checklist

Currently, only EventHubsTrigger has a property with "DefaultValue" on it. This PR additionally standardizes casing for bindings on the source-generated file by choosing PascalCase when creating fields on the anonymous types we represent bindings with (before serializing them). 

Values taken from properties with DefaultValue attributes keep their casing, and values whose names come from constructor parameters will uppercase the first letter.

### DIFF
--- a/sdk/Sdk.Generators/Constants.cs
+++ b/sdk/Sdk.Generators/Constants.cs
@@ -11,8 +11,12 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
         internal const string FunctionNameType = "Microsoft.Azure.Functions.Worker.FunctionAttribute";
         internal const string HttpResponseType = "Microsoft.Azure.Functions.Worker.Http.HttpResponseData";
         internal const string EventHubsTriggerType = "Microsoft.Azure.Functions.Worker.EventHubTriggerAttribute";
+<<<<<<< HEAD
         internal const string BindingPropertyNameAttributeType = "Microsoft.Azure.Functions.Worker.Extensions.Abstractions.BindingPropertyNameAttribute";
         internal const string DefaultValueType = "Microsoft.Azure.Functions.Worker.Extensions.Abstractions";
+=======
+        internal const string DefaultValueType = "Microsoft.Azure.Functions.Worker.Extensions.Abstractions.DefaultValueAttribute";
+>>>>>>> 8d65547 (resolve comment, add event hubs check)
 
         // System types
         internal const string IEnumerableType = "System.Collections.IEnumerable";
--- a/sdk/Sdk.Generators/Constants.cs
+++ b/sdk/Sdk.Generators/Constants.cs
@@ -12,6 +12,7 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
         internal const string HttpResponseType = "Microsoft.Azure.Functions.Worker.Http.HttpResponseData";
         internal const string EventHubsTriggerType = "Microsoft.Azure.Functions.Worker.EventHubTriggerAttribute";
         internal const string BindingPropertyNameAttributeType = "Microsoft.Azure.Functions.Worker.Extensions.Abstractions.BindingPropertyNameAttribute";
+        internal const string DefaultValueType = "Microsoft.Azure.Functions.Worker.Extensions.Abstractions";
 
         // System types
         internal const string IEnumerableType = "System.Collections.IEnumerable";
--- a/sdk/Sdk.Generators/Constants.cs
+++ b/sdk/Sdk.Generators/Constants.cs
@@ -11,12 +11,8 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
         internal const string FunctionNameType = "Microsoft.Azure.Functions.Worker.FunctionAttribute";
         internal const string HttpResponseType = "Microsoft.Azure.Functions.Worker.Http.HttpResponseData";
         internal const string EventHubsTriggerType = "Microsoft.Azure.Functions.Worker.EventHubTriggerAttribute";
-<<<<<<< HEAD
         internal const string BindingPropertyNameAttributeType = "Microsoft.Azure.Functions.Worker.Extensions.Abstractions.BindingPropertyNameAttribute";
-        internal const string DefaultValueType = "Microsoft.Azure.Functions.Worker.Extensions.Abstractions";
-=======
         internal const string DefaultValueType = "Microsoft.Azure.Functions.Worker.Extensions.Abstractions.DefaultValueAttribute";
->>>>>>> 8d65547 (resolve comment, add event hubs check)
 
         // System types
         internal const string IEnumerableType = "System.Collections.IEnumerable";
--- a/sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
+++ b/sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
@@ -523,10 +523,9 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
 
                     if (defaultValAttrList.SingleOrDefault() is { } defaultValAttr) // list will only be of size one b/c there cannot be duplicates of an attribute on one piece of syntax
                     {
-                        // only one constructor argument in DefaultValue attribute
                         if (!attrProperties.Keys.Any(a => string.Equals(a, default, StringComparison.OrdinalIgnoreCase))) // check if this property has been assigned a value already in constructor or named args
                         {
-                            attrProperties[member.Name.ToString()] = defaultValAttr.ConstructorArguments.SingleOrDefault().Value!.ToString(); // add property with default value to func metadata, using binding property as the arg name
+                            attrProperties[member.Name.ToString()] = defaultValAttr.ConstructorArguments.SingleOrDefault().Value!.ToString(); // only one constructor arg in DefaultValue attribute (the default value)
                         }
                     }
                 }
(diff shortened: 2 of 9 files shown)


sdk/Sdk.Generators/Constants.cs
```diff
@@ -11,6 +11,7 @@ internal static class Constants
         internal const string FunctionNameType = "Microsoft.Azure.Functions.Worker.FunctionAttribute";
         internal const string HttpResponseType = "Microsoft.Azure.Functions.Worker.Http.HttpResponseData";
         internal const string EventHubsTriggerType = "Microsoft.Azure.Functions.Worker.EventHubTriggerAttribute";
+        internal const string DefaultValueType = "Microsoft.Azure.Functions.Worker.Extensions.Abstractions";
```
jviau, line 14: Is this the correct attribute name? This looks like a namespace.

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -517,6 +517,33 @@ private bool TryGetAttributeProperties(AttributeData attributeData, Location att
...
 
+                // some properties have default values, so if these properties were not already defined in constructor or named arguments, we will auto-add them here
+                foreach (var member in attributeData.AttributeClass!.GetMembers().Where(a => a is IPropertySymbol))
+                {
+                    var defaultValAttrList = member.GetAttributes().Where(attr => SymbolEqualityComparer.Default.Equals(attr.AttributeClass, Compilation.GetTypeByMetadataName(Constants.DefaultValueType)));
+
+                    if (defaultValAttrList.SingleOrDefault() is { } defaultValAttr) // list will only be of size one b/c there cannot be duplicates of an attribute on one piece of syntax
+                    {
+                        var argName = member.Name;
+                        var argDefaultVal = defaultValAttr.ConstructorArguments.SingleOrDefault().Value!.ToString(); // only one constructor arg in DefaultValue attribute (the default value)
```
jviau, line 528: The value can be a string or a bool here. Would pattern matching fit better for branching?

``` CSharp
object arg = defaultValAttr.ConstructorArguments.SingleOrDefault().Value!;
switch arg
{
    case bool b:
        // used bool constructor
    case string s:
        // use string constructor
    defaullt:
        throw new NotSupportedException(); // used unknown ctor.
}
```

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -608,6 +635,18 @@ private bool IsEventHubsTriggerValid(IParameterSymbol parameterSymbol, TypeSynta
                     }
                 }
 
+                // Check the default value of IsBatched
+                var eventHubsAttr = attribute.AttributeClass;
+                var isBatchedProp = eventHubsAttr!.GetMembers().Where(m => string.Equals(m.Name, Constants.IsBatchedKey, StringComparison.OrdinalIgnoreCase)).SingleOrDefault();
+                AttributeData defaultValAttr = isBatchedProp.GetAttributes().Where(attr => SymbolEqualityComparer.Default.Equals(attr.AttributeClass, Compilation.GetTypeByMetadataName(Constants.DefaultValueType))).SingleOrDefault();
+                var defaultVal = defaultValAttr.ConstructorArguments.SingleOrDefault().Value!.ToString(); // there is only one constructor arg, the default value
```
jviau, line 642: If we are expecting a bool here, can we do something strongly typed instead of string comparisons?

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
satvu: The dictionary created here is <string, object> so I don't think we need to differentiate much - I revised this to just take the value as a generic object and set the dictionary value to that. 

The only special handling we need in this spot is to change the IsBatched value to "Cardinality" for the function metadata info (which I've also revised slightly in this iteration to use arg as a bool).

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -504,7 +504,7 @@ private bool TryGetAttributeProperties(AttributeData attributeData, Location att
                 {
                     if (namedArgument.Value.Value != null)
                     {
-                        if (string.Equals(namedArgument.Key, Constants.IsBatchedKey) && !attrProperties.Keys.Any(k => string.Equals(k, "Cardinality", StringComparison.OrdinalIgnoreCase)))
+                        if (String.Equals(namedArgument.Key, Constants.IsBatchedKey) && !attrProperties.Keys.Contains("Cardinality"))
```
jviau, line 507: nit: lower case `string` was a nice to have change:
```suggestion
                        if (string.Equals(namedArgument.Key, Constants.IsBatchedKey) && !attrProperties.Keys.Contains("Cardinality"))
```

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -525,21 +525,17 @@ private bool TryGetAttributeProperties(AttributeData attributeData, Location att
                     if (defaultValAttrList.SingleOrDefault() is { } defaultValAttr) // list will only be of size one b/c there cannot be duplicates of an attribute on one piece of syntax
                     {
                         var argName = member.Name;
-                        var argDefaultVal = defaultValAttr.ConstructorArguments.SingleOrDefault().Value!.ToString(); // only one constructor arg in DefaultValue attribute (the default value)
-                        
-                        if (string.Equals(argName, Constants.IsBatchedKey))
+                        object arg = defaultValAttr.ConstructorArguments.SingleOrDefault().Value!; // only one constructor arg in DefaultValue attribute (the default value)
+                        if ( arg is bool b && string.Equals(argName, Constants.IsBatchedKey))
```
jviau, line 529: nit: whitespace
```suggestion
                        if (arg is bool b && string.Equals(argName, Constants.IsBatchedKey))
```

jviau (approved):

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -517,6 +517,29 @@ private bool TryGetAttributeProperties(AttributeData attributeData, Location att
...
+                {
+                    var defaultValAttrList = member.GetAttributes().Where(attr => SymbolEqualityComparer.Default.Equals(attr.AttributeClass, Compilation.GetTypeByMetadataName(Constants.DefaultValueType)));
+
+                    if (defaultValAttrList.SingleOrDefault() is { } defaultValAttr) // list will only be of size one b/c there cannot be duplicates of an attribute on one piece of syntax
+                    {
+                        var argName = member.Name;
+                        object arg = defaultValAttr.ConstructorArguments.SingleOrDefault().Value!; // only one constructor arg in DefaultValue attribute (the default value)
+                        if ( arg is bool b && string.Equals(argName, Constants.IsBatchedKey))
+                        {
+                            if (!attrProperties.Keys.Contains("Cardinality"))
```
jviau, line 531:
```suggestion
                            if (!attrProperties.ContainsKey("Cardinality"))
```

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -517,6 +517,29 @@ private bool TryGetAttributeProperties(AttributeData attributeData, Location att
...
+                            {
+                                attrProperties["Cardinality"] = b ? "Many" : "One";
+                            }
+                        }
+                        else if (!attrProperties.Keys.Any(a => string.Equals(a, argName, StringComparison.OrdinalIgnoreCase))) // check if this property has been assigned a value already in constructor or named args
```
jviau, line 536: I noticed we actually own this dictionary. I am curious about why we sometimes want case sensitive and other times not case sensitive? I know I commented before about being case sensitive for a property check. Either way, if we can settle on a single correct way for key comparison, you can provide that comparer to the dictionary ctor:

``` CSharp
 Dictionary<string, object> attrProperties = new(StringComparer.OrdinalIgnoreCase);
```

Then you are free to use `attrProperties.ContainsKey({key})` and it will use the comparer from the ctor. Maybe ordinal ignore case is correct? But it might be more complex than that: when comparing a _property_ here, you want `Ordinal`, when comparing a _constructor parameter_, you want a comparator that will upper-case the first character only, then do an Ordinal comparison.

This is all getting into the weeds and may not be that important - so I'll leave it up to you to decide how accurate you want this to be. But give some thought to the scenario where a type here has both properties `MyProperty` and `Myproperty` (for reasons unknown to us) - will we consider them separately? How will we match a constructor parameter of `myProperty` vs `myproperty`?

satvu: Yeah the non-case-sensitive check here is because we may have added the constructor parameter (which would be _isBatched_ while the property is _IsBatched_). It would be a stricter comparison to upper-case the first character of the constructor parameter and based on the convention we're following the end result should be the same. I guess it would also give attribute authors the flexibility to have `MyProperty` and `Myproperty` though I can't imagine why yet 😆. Leaning towards this change - will walk through it and test it out a bit first but if I do implement it I'll try to summarize the context we discussed here in the comments for future reference.

jviau (approved):

satvu:
\check-enforcer evaluate

Issue #1121, opened 2022-10-13
Add support for bindings to provide information about input converters


kshyju:
Add support for bindings to provide input converter information to be used. This allows binding authors(extension authors) to specify one or more converters to be used when doing input conversion for that binding entry.

Initial plans for the implementation:

As part of this change, we plan to expose a new interface `IConverterProvider`

```
/// <summary>
/// Provides information about what converters to be used.
/// </summary>
public interface IConverterProvider
{
    /// <summary>
    /// Gets an <see cref="IEnumerable{T}"/> of <see cref="Type"/> instances representing the converters to be used.
    /// </summary>
    public IList<Type> ConverterTypes { get; }
}
```
Here is an example binding which implements this interface and specifying `MyFooConverter` as the converter type to be used for this binding.

```
public class FooTriggerBinding : TriggerBindingAttribute, IConverterProvider
{
    public IList<Type> ConverterTypes => new[] { typeof(MyFooConverter) };

    // Other properties relevant to the binding.
}
```

The `ConverterTypes` property is an ordered collection. So the binding authors can define multiple converters in the order they want.

```
public IList<Type> ConverterTypes =>
    new[]
    {
        typeof(MyPrimaryConverter),
        typeof(MyBackupConverter)
    };
```

If the function parameter has an Input converter specifically defined(see sample below), that will be used and converters defined on the binding level will not be used.

```
[Function("products")]
public HttpResponseData Run(
                 [HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = null)] HttpRequestData req,
                 [InputConverter(typeof(MyCustomConverter))] ProductViewModel productvm)
{
   ....    
}

```

SeanFeldman:
I was just wondering, since this is something that will be released in the future, would it be possible to support C# 11 generic interfaces so that

```C#
[InputConverter(typeof(MyCustomConverter))] ProductViewModel productvm
```

would become

```C#
[InputConverter<MyCustomConverter>] ProductViewModel productvm
```

jviau:
For completeness of the design discussion, have we considered this alternative: instead of introducing the new contract, let the binding attribute implement `IInputConverter` itself and we will test for that contract and use it if present.

PR #1136, opened 2022-10-21
Introducing IInputConverterProvider interface 

kshyju:
Introducing `IInputConverterProvider` interface  to provide information about what converters to be used for bindings. This will allow binding authors to provide information about what specific converter to be used during the input conversion of that binding entry.

This is part 1 of the fix for #1121. I will have a follow up PR which consumes this version of the `Microsoft.Azure.Functions.Worker.Extensions.Abstractions` nuget package in the Grpc layer and use it.

### DIFF
--- a/extensions/Worker.Extensions.Abstractions/release_notes.md
+++ b/extensions/Worker.Extensions.Abstractions/release_notes.md
@@ -1,0 +1,5 @@
+## Release notes
+<!-- Please add your release notes in the following format:
+- My change description (#PR/#issue)
+-->
+- Introducing `IInputConverterProvider` interface (#1121)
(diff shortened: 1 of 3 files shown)


satvu (approved):
lgtm!

jviau:
Do we want to leverage the existing [TypeConverter](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.typeconverter?view=net-7.0) in dotnet?

fabiocav (approved):

liliankasem (approved):

jviau:
I also brought up a question in the linked issue:

> have we considered this alternative: instead of introducing the new contract, let the binding attribute implement `IInputConverter` itself and we will test for that contract and use it if present.

kshyju:
Abandoning this PR as the approach used here won't work to get the needed information we need when writing an analyzer. Going to switch to the attribute approach in a new PR.

kshyju:
> Do we want to leverage the existing [TypeConverter](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.typeconverter?view=net-7.0) in dotnet?

Let's sync offline on this.

PR #1152, opened 2022-11-04
[Samples] WorkerBindingSamples project with blob function examples

liliankasem:
Sample project showcasing how to use SDK-type bindings with blob storage with the new sdk-binding feature.

### DIFF
--- a/extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
+++ b/extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
@@ -179,59 +179,56 @@ namespace Microsoft.Azure.Functions.Worker
                 throw new ArgumentNullException(nameof(blobName));
             }
 
-            /*
             BlobBaseClient blob = null;
 
             if (connectionString == null)
             {
                 // Try using Managed Identity
                 context.FunctionContext.BindingContext.BindingData.TryGetValue("Uri", out var blobEndpoint);
-                
-                if(Environment.GetEnvironmentVariable("AzureWebJobsStorage__accountName") != null)
+                string endpoint = blobEndpoint.ToString().Trim('\\', '"');
+
+                if(Environment.GetEnvironmentVariable(Constants.ConnectionAccountName) != null
+                || Environment.GetEnvironmentVariable(Constants.ConnectionBlobUri) != null)
                 {
-                    // for logged in user and AzureWebJobsStorage
-                    blob = new BlobClient(new Uri(blobEndpoint.ToString().Trim('\\', '"')), new DefaultAzureCredential());
+                    // for AzureWebJobsStorage
+                    blob = new BlobClient(new Uri(endpoint), new DefaultAzureCredential());
                 }
-                if(Environment.GetEnvironmentVariable("AzureWebJobsStorage__blobServiceUri") != null)
+                else if (
+                    Environment.GetEnvironmentVariable(Constants.ConnectionTenantId) != null
+                    && Environment.GetEnvironmentVariable(Constants.ConnectionClientId) != null
+                    && Environment.GetEnvironmentVariable(Constants.ConnectionClientSecret) != null)
                 {
-                    // for logged in user and AzureWebJobsStorage
-                    blob = new BlobClient(new Uri(blobEndpoint.ToString().Trim('\\', '"')), new DefaultAzureCredential());
-                }
-                else if (Environment.GetEnvironmentVariable("AzureWebJobsStorage__tenantId") != null)
-                {
-                    // if specific clientid to be userd - userassigned
-                    blob = new BlobClient(new Uri(blobEndpoint.ToString().Trim('\\', '"')),
+                    // For user assigned identity
+                    blob = new BlobClient(new Uri(endpoint),
                     new ClientSecretCredential(
-                        Environment.GetEnvironmentVariable("AzureWebJobsStorage__tenantId"),
-                        Environment.GetEnvironmentVariable("AzureWebJobsStorage__clientId"),
-                        Environment.GetEnvironmentVariable("AzureWebJobsStorage__clientSecret")
+                        Environment.GetEnvironmentVariable(Constants.ConnectionTenantId),
+                        Environment.GetEnvironmentVariable(Constants.ConnectionClientId),
+                        Environment.GetEnvironmentVariable(Constants.ConnectionClientSecret)
                     ));
                 }
-                else if (Environment.GetEnvironmentVariable("AzureWebJobsStorage__credential") == "ManagedIdentity")
+                else if (Environment.GetEnvironmentVariable(Constants.ConnectionCredential) == "ManagedIdentity")
                 {
-                    // Credential and ClientId. For portal apps.
-                    blob = new BlobClient(new Uri(blobEndpoint.ToString().Trim('\\', '"')),
+                    // For portal apps only - using credential and clientId
+                    blob = new BlobClient(new Uri(endpoint),
                                 new ManagedIdentityCredential(
-                                    Environment.GetEnvironmentVariable("AzureWebJobsStorage__clientId")
+                                    Environment.GetEnvironmentVariable(Constants.ConnectionClientId)
                                 ));
                 }
             }
             else
             {
-                */
+                Type targetType = typeof(T);
+                var container = CreateBlobContainerClient(connectionString, containerName);
 
-            Type targetType = typeof(T);
-            var container = CreateBlobContainerClient(connectionString, containerName);
-
-            BlobBaseClient blob = targetType switch
-            {
-                Type _ when targetType == typeof(BlobClient) => container.GetBlobClient(blobName),
-                Type _ when targetType == typeof(BlockBlobClient) => container.GetBlockBlobClient(blobName),
-                Type _ when targetType == typeof(PageBlobClient) => container.GetPageBlobClient(blobName),
-                Type _ when targetType == typeof(AppendBlobClient) => container.GetAppendBlobClient(blobName),
-                _ => new(connectionString, containerName, blobName)
-            };
-        //}
+                blob = targetType switch
+                {
+                    Type _ when targetType == typeof(BlobClient) => container.GetBlobClient(blobName),
+                    Type _ when targetType == typeof(BlockBlobClient) => container.GetBlockBlobClient(blobName),
+                    Type _ when targetType == typeof(PageBlobClient) => container.GetPageBlobClient(blobName),
+                    Type _ when targetType == typeof(AppendBlobClient) => container.GetAppendBlobClient(blobName),
+                    _ => new(connectionString, containerName, blobName)
+                };
+        }
 
             return (T)blob;
         }
--- a/extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
+++ b/extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
@@ -57,13 +57,6 @@ namespace Microsoft.Azure.Functions.Worker
 
             if (context.Source is CollectionModelBindingData collectionBindingData)
             {
-                if (!IsSupportedEnumerable(context.TargetType))
-                {
-                    return ConversionResult.Failed(new InvalidOperationException($"Target type '{context.TargetType}' is not supported"));
-                }
-
-                Type individualTargetType = context.TargetType.GenericTypeArguments.FirstOrDefault();
-
                 var collectionBlob = new List<object>(collectionBindingData.ModelBindingDataArray.Length);
 
                 foreach (ModelBindingData modelBindingData in collectionBindingData.ModelBindingDataArray)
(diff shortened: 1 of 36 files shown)


samples/WorkerBindingSamples/WorkerBindingSamples.csproj
```diff
@@ -0,0 +1,38 @@
...
+    <PackageReference Include="System.Net.NameResolution" Version="4.3.0" />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\..\extensions\Worker.Extensions.Http\src\Worker.Extensions.Http.csproj" />
+    <ProjectReference Include="..\..\extensions\Worker.Extensions.CosmosDB\src\Worker.Extensions.CosmosDB.csproj" />
+    <ProjectReference Include="..\..\extensions\Worker.Extensions.Storage.Blobs\src\Worker.Extensions.Storage.Blobs.csproj" />
+    <ProjectReference Include="..\..\extensions\Worker.Extensions.Storage.Queues\src\Worker.Extensions.Storage.Queues.csproj" />
+    <ProjectReference Include="..\..\src\DotNetWorker.ApplicationInsights\DotNetWorker.ApplicationInsights.csproj" />
+    <ProjectReference Include="..\..\src\DotNetWorker\DotNetWorker.csproj" />
+  </ItemGroup>
```
liliankasem, line 28: Use real published packages instead of project ref?

samples/WorkerBindingSamples/Program.cs
```diff
@@ -0,0 +1,22 @@
...
+using System;
+using Microsoft.Extensions.Hosting;
+
+namespace SampleApp
+{
+    public class Program
+    {
+        public static void Main()
+        {
+            Console.WriteLine($"Azure Functions .NET Worker (PID: { Environment.ProcessId }) initialized in debug mode. Waiting for debugger to attach...");
```
kshyju, line 13: Do we need this console.writeline message? The IDE is not going to wait here. rite ?

samples/WorkerBindingSamples/Blob/BlobInputBindingSamples.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        [Function(nameof(BlobInputCollectionFunction))]
+        public HttpResponseData BlobInputCollectionFunction(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "get", "post", Route = null)] HttpRequestData req,
+            [BlobInput("input-container")] IEnumerable<BlobClient> blobs)
+        {
+            _logger.LogInformation("Blobs within container:");
+            foreach (BlobClient blob in blobs)
+            {
+                _logger.LogInformation(blob.Name);
```
kshyju, line 96: name of the blob is something the current model also get. So may be read a property which current model does not provide. Ex: `BlobContainerName`. ?

kshyju (approved):

samples/WorkerBindingSamples/Program.cs
liliankasem: I use it for debugging, I can attach to the PID without having the IDE wait here but we don't need to include it in the sample

samples/WorkerBindingSamples/Blob/BlobInputBindingSamples.cs
```diff
@@ -0,0 +1,104 @@
...
+        private readonly ILogger<BlobInputBindingSamples> _logger;
+
+        public BlobInputBindingSamples(ILogger<BlobInputBindingSamples> logger)
+        {
+            _logger = logger;
+        }
+
+        [Function(nameof(BlobInputClientFunction))]
+        public async Task<HttpResponseData> BlobInputClientFunction(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "get", "post", Route = null)] HttpRequestData req,
```
mathewc, line 28: Recommend removing `Route = null` and `Connection = "AzureWebJobsStorage"` from these samples to keep the code clean and focused. The storage account will be defaulted to the value you're providing, right?

mathewc, line 28: Recommend not using Anonymous for the samples, since we don't want to guide users to do that. Samples will be run normally via core tools, and auth will be disabled there anyways, so Function level auth annotations will work.

samples/WorkerBindingSamples/Blob/BlobInputBindingSamples.cs
```diff
@@ -0,0 +1,104 @@
...
+            return req.CreateResponse(HttpStatusCode.OK);
+        }
+
+        [Function(nameof(BlobInputStreamFunction))]
+        public HttpResponseData BlobInputStreamFunction(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "get", "post", Route = null)] HttpRequestData req,
+            [BlobInput("input-container/sample1.txt", Connection = "AzureWebJobsStorage")] Stream stream)
+        {
+            using var blobStreamReader = new StreamReader(stream);
+            _logger.LogInformation("Blob content: {stream}", blobStreamReader.ReadToEnd());
```
mathewc, line 44: Use async APIs in all these samples, e.g. `ReadToEndAsync`

samples/WorkerBindingSamples/Blob/BlobInputBindingSamples.cs
```diff
@@ -0,0 +1,104 @@
...
+
+            return req.CreateResponse(HttpStatusCode.OK);
+        }
+
+        [Function(nameof(BlobInputStreamFunction))]
+        public HttpResponseData BlobInputStreamFunction(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "get", "post", Route = null)] HttpRequestData req,
+            [BlobInput("input-container/sample1.txt", Connection = "AzureWebJobsStorage")] Stream stream)
+        {
+            using var blobStreamReader = new StreamReader(stream);
```
mathewc, line 43: This using statement needs braces around the usage, so it's disposed at the right scope, right?

samples/WorkerBindingSamples/Blob/BlobInputBindingSamples.cs
```diff
@@ -0,0 +1,104 @@
...
+            _logger.LogInformation("Blob content: {stream}", blobStreamReader.ReadToEnd());
+
+            return req.CreateResponse(HttpStatusCode.OK);
+        }
+
+        [Function(nameof(BlobInputByteArrayFunction))]
+        public HttpResponseData BlobInputByteArrayFunction(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "get", "post", Route = null)] HttpRequestData req,
+            [BlobInput("input-container/sample1.txt", Connection = "AzureWebJobsStorage")] Byte[] data)
```
mathewc, line 52: Should we show at least one example demonstrating how the blob name can come from request query string params rather than hard coded?

samples/WorkerBindingSamples/Blob/BlobTriggerBindingSamples.cs
```diff
@@ -0,0 +1,64 @@
...
+namespace SampleApp
+{
+    public static class BlobTriggerBindingSamples
+    {
+        [Function(nameof(BlobClientFunction))]
+        public static async Task BlobClientFunction(
+            [BlobTrigger("client-trigger/{name}", Connection = "AzureWebJobsStorage")] BlobClient client,
+            FunctionContext context)
+        {
+            var logger = context.GetLogger(nameof(BlobClientFunction));
```
mathewc, line 21: Any reason why you're using FunctionContext in these samples, rather than taking the logger as a ctor param as you did in the input samples?

samples/WorkerBindingSamples/Blob/BlobInputBindingSamples.cs
liliankasem: Yup it should be!

samples/WorkerBindingSamples/Blob/BlobTriggerBindingSamples.cs
liliankasem: Wanted to show that both options are available, but I think we want folks to use the ctor pattern right? Happy to switch these to match input bindings

samples/WorkerBindingSamples/Blob/BlobInputBindingSamples.cs
liliankasem: I believe it will still get disposed within the right scope, we're using a `using declaration` here and this is what the docs say about it:

>The using declaration calls the [Dispose](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable.dispose) method on the object in the correct way when it goes out of scope. The using statement causes the object itself to go out of scope as soon as [Dispose](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable.dispose) is called.

>Within the using block, the object is read-only and can't be modified or reassigned. A variable declared with a using declaration is read-only. If the object implements IAsyncDisposable instead of IDisposable, either using form calls the [DisposeAsync](https://learn.microsoft.com/en-us/dotnet/api/system.iasyncdisposable.disposeasync) and awaits the returned [ValueTask](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask). For more information on [IAsyncDisposable](https://learn.microsoft.com/en-us/dotnet/api/system.iasyncdisposable), see [Implement a DisposeAsync method](https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync).

If a `using statement` is a better fit here I'd like to learn more and happy to make the changes suggested

mathewc: I think you're good - I haven't actually used this new C# 8 using declaration syntax - I've always defined using blocks :)

samples/WorkerBindingSamples/Blob/BlobInputBindingSamples.cs
surgupta-msft: Added in sample function `BlobInputStringFunction`

samples/WorkerBindingSamples/Blob/BlobInputBindingSamples.cs
```diff
@@ -0,0 +1,118 @@
...
+        private readonly ILogger<BlobInputBindingSamples> _logger;
+
+        public BlobInputBindingSamples(ILogger<BlobInputBindingSamples> logger)
+        {
+            _logger = logger;
+        }
+
+        [Function(nameof(BlobInputClientFunction))]
+        public async Task<HttpResponseData> BlobInputClientFunction(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "get", "post")] HttpRequestData req,
```
mathewc, line 28: Any reason why the functions in this class are explicit about get/post methods, while your trigger samples don't list declare verbs?

liliankasem: This is a HTTP trigger so we're being explicit, but the trigger samples are all BlobTriggers which don't need/use get/post

samples/WorkerBindingSamples/Blob/BlobTriggerBindingSamples.cs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        [Function(nameof(BlobStreamFunction))]
+        public static void BlobStreamFunction(
+            [BlobTrigger("stream-trigger/{name}", Connection = "AzureWebJobsStorage")] Stream stream,
+            FunctionContext context)
+        {
+            var logger = context.GetLogger(nameof(BlobStreamFunction));
+            using var blobStreamReader = new StreamReader(stream);
+            logger.LogInformation("Blob content: {stream}", blobStreamReader.ReadToEnd());
```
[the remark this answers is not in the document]
liliankasem: @mathewc Would that be available in the FunctionContext?

mathewc (approved):

samples/WorkerBindingSamples/Blob/BlobTriggerBindingSamples.cs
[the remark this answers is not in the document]
mathewc: When using expressions like "book-trigger/{name}" the name value can then be defined as a string "name" parameter and the platform will bind that so you can reference it in your code. If you're not using it in your code, you don't need to use the {name} expression. See examples in [doc](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-storage-blob-trigger?tabs=in-process%2Cextensionv5&pivots=programming-language-csharp#example).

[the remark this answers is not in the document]
liliankasem: Thanks! I'll update the samples to either use it or remove it :)

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -0,0 +1,250 @@
...
+        public BlobStorageConverter(IOptions<WorkerOptions> workerOptions, IOptionsSnapshot<BlobStorageBindingOptions> blobOptions, ILogger<BlobStorageConverter> logger)
+        {
+            _workerOptions = workerOptions ?? throw new ArgumentNullException(nameof(workerOptions));
+            _blobOptions = blobOptions ?? throw new ArgumentNullException(nameof(blobOptions));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
+        public async ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
+        {
+            if (context is null)
```
kshyju, line 39: I think Fabio also mentioned this. But throw an exception if context is null. context is not expected to be null and if it is, it is a fundamental problem we need to fix.

liliankasem: this is behind the blob converter pr, should be fixed when i rebase

PR #1180, opened 2022-11-16
Worker-indexing duplicate binding info fix

satvu:

Resolves Function metadata loading error where duplicate `BindingInfo` dictionaries are created for `RpcFunctionMetadata` in `GrpcWorker.cs`. 

This occurs when you use an SDK that has the worker-indexing version that reads `functions.metadata` (not the source-gen version) combined with the latest worker package. The latest package has some components related to source-generated Function metadata that is not compatible with the first worker-indexing iteration.

Repro steps:

1. Use worker SDK package 1.7.0 (does not have source gen).
2. Use worker package 1.11.0-preview2.
3. Set MSBuild property `FunctionsEnableWorkerIndexing` to true in `.csproj` file.
4. Run app. 

### Pull request checklist

### Additional information

May need to discuss work item for E2E testing for different function metadata generation options and different preview options? Could be in this PR or we can create a new work item.

### DIFF
--- a/src/DotNetWorker.Grpc/FunctionMetadata/DefaultFunctionMetadataProvider.cs
+++ b/src/DotNetWorker.Grpc/FunctionMetadata/DefaultFunctionMetadataProvider.cs
@@ -60,12 +60,6 @@ namespace Microsoft.Azure.Functions.Worker
                     foreach (var binding in rawBindings.EnumerateArray())
                     {
                         functionMetadata.RawBindings.Add(binding.GetRawText());
-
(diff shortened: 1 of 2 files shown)


satvu:
> Whoever found this issue is a genius.

😆 thank you @kshyju for the find

jviau (approved):

jviau:
> May need to discuss work item for E2E testing for different function metadata generation options and different preview options? Could be in this PR or we can create a new work item.

I would suggest a different PR / work item for this. The testing work won't be trivial for this, but definitely beneficial. We are going to need to devise some way to test the MSBuild task as well as `IIncrementalGenerator` (when we switch to that - the existing source gen test helpers don't work for that type).

PR #1183, opened 2022-11-16
adding ability to manipulate capabilities

brettsam:
Adding an API for an (advanced) scenario where a customer needs to disable or change a capability.

### DIFF
--- a/samples/FunctionApp/Program.cs
+++ b/samples/FunctionApp/Program.cs
@@ -1,9 +1,6 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
-using System;
-using System.Diagnostics;
-using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.Functions.Worker;
 using Microsoft.Extensions.DependencyInjection;
@@ -21,11 +18,10 @@ namespace FunctionApp
             //<docsnippet_startup>
             var host = new HostBuilder()
                 //<docsnippet_configure_defaults>
-                .ConfigureFunctionsWorkerDefaults(builder =>
+                .ConfigureFunctionsWorkerDefaults((WorkerOptions options) =>
                 {
-                    builder
-                        .AddApplicationInsights()
-                        .AddApplicationInsightsLogger();
+                    // example (will remove before merge)
+                    options.Capabilities.Remove("HandlesWorkerTerminateMessage");
                 })
                 //</docsnippet_configure_defaults>
                 //<docsnippet_dependency_injection>
--- a/samples/FunctionApp/Program.cs
+++ b/samples/FunctionApp/Program.cs
@@ -18,10 +18,11 @@ namespace FunctionApp
             //<docsnippet_startup>
             var host = new HostBuilder()
                 //<docsnippet_configure_defaults>
-                .ConfigureFunctionsWorkerDefaults((WorkerOptions options) =>
+                .ConfigureFunctionsWorkerDefaults(builder =>
                 {
-                    // example (will remove before merge)
-                    options.Capabilities.Remove("HandlesWorkerTerminateMessage");
+                    builder
+                        .AddApplicationInsights()
+                        .AddApplicationInsightsLogger();
                 })
                 //</docsnippet_configure_defaults>
                 //<docsnippet_dependency_injection>
--- a/src/DotNetWorker.Core/Context/Features/IDictionaryExtensions.cs
+++ b/src/DotNetWorker.Core/Context/Features/IDictionaryExtensions.cs
@@ -8,49 +8,59 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.ComponentModel;
 
-namespace Microsoft.Azure.Functions.Worker
+internal static class IDictionaryExtensions
 {
-    internal static class IDictionaryExtensions
+    internal static bool TryAdd<TKey, TValue>(this IDictionary<TKey, TValue> dictionary, TKey key, TValue value)
     {
-        internal static bool TryAdd<TKey, TValue>(this IDictionary<TKey, TValue> dictionary, TKey key, TValue value)
+        if (key == null)
         {
-            if (key == null)
-            {
-                throw new ArgumentNullException(nameof(key));
-            }
-
-            if (dictionary.ContainsKey(key))
-            {
-                return false;
-            }
-
-            dictionary.Add(key, value);
-            return true;
+            throw new ArgumentNullException(nameof(key));
         }
-    }
 
-    internal static class ConcurrentDictionaryExtensions
-    {
-        public static TValue GetOrAdd<TKey, TValue, TArg>(this ConcurrentDictionary<TKey, TValue> dictionary, TKey key, Func<TKey, TArg, TValue> valueFactory, TArg factoryArgument)
+        if (dictionary.ContainsKey(key))
         {
-            if (dictionary == null)
-            {
-                throw new ArgumentNullException(nameof(dictionary));
-            }
-
-            if (key == null)
-            {
-                throw new ArgumentNullException(nameof(key));
-            }
-
-            if (valueFactory == null)
-            {
-                throw new ArgumentNullException(nameof(valueFactory));
-            }
-
-            return dictionary.GetOrAdd(key, k => valueFactory(k, factoryArgument));
+            return false;
         }
+
+        dictionary.Add(key, value);
+        return true;
     }
 }
+
+internal static class ConcurrentDictionaryExtensions
+{
+    public static TValue GetOrAdd<TKey, TValue, TArg>(this ConcurrentDictionary<TKey, TValue> dictionary, TKey key, Func<TKey, TArg, TValue> valueFactory, TArg factoryArgument)
+    {
+        if (dictionary == null)
+        {
+            throw new ArgumentNullException(nameof(dictionary));
+        }
+
+        if (key == null)
+        {
+            throw new ArgumentNullException(nameof(key));
+        }
+
+        if (valueFactory == null)
+        {
+            throw new ArgumentNullException(nameof(valueFactory));
+        }
+
+        return dictionary.GetOrAdd(key, k => valueFactory(k, factoryArgument));
+    }
+}
+
+internal static class KeyValuePairExtensions
+{
+    // Based on https://source.dot.net/#System.Private.CoreLib/KeyValuePair.cs,aa57b8e336bf7f59
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public static void Deconstruct<TKey, TValue>(this KeyValuePair<TKey, TValue> pair, out TKey key, out TValue value)
+    {
+        key = pair.Key;
+        value = pair.Value;
+    }
+}
+
 #endif
--- a/src/DotNetWorker.Core/Context/Features/IDictionaryExtensions.cs
+++ b/src/DotNetWorker.Core/Context/Features/IDictionaryExtensions.cs
@@ -54,7 +54,7 @@ internal static class ConcurrentDictionaryExtensions
 
 internal static class KeyValuePairExtensions
 {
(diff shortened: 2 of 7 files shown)


samples/FunctionApp/Program.cs
```diff
@@ -21,11 +18,10 @@ static async Task Main(string[] args)
             //<docsnippet_startup>
             var host = new HostBuilder()
                 //<docsnippet_configure_defaults>
-                .ConfigureFunctionsWorkerDefaults(builder =>
+                .ConfigureFunctionsWorkerDefaults((WorkerOptions options) =>
                 {
-                    builder
```
jviau, line 26: AppInsights logic is no longer needed in this sample?

src/DotNetWorker.Core/Hosting/WorkerOptions.cs
```diff
@@ -22,12 +23,17 @@ public class WorkerOptions
         /// </summary>
         public InputConverterCollection InputConverters { get; } = new InputConverterCollection();
 
+        /// <summary>
+        /// Gets the worker capabilities.
+        /// </summary>
+        public IDictionary<string, string> Capabilities { get; } = new Dictionary<string, string>();
```
jviau, line 29: What do you think of having this dictionary initialize to the default set of capabilities? Will remove the need for our configure call.

src/DotNetWorker.Core/Hosting/WorkerOptions.cs
```diff
@@ -22,12 +23,17 @@ public class WorkerOptions
...
+
         /// <summary>
         /// Gets and sets the flag for opting in to unwrapping user-code-thrown
         /// exceptions when they are surfaced to the Host. 
         /// </summary>
         public bool EnableUserCodeException { get; set; } = false;
```
jviau, line 35: Since this directly maps to a capability, I suggest updating the get & set to interact with the `Capabilities` dictionary:

``` CSharp
public bool EnableUserCodeException
{
    get => GetBoolCapability(nameof(EnableUserCodeException));
    set => Capabilities[nameof(EnableUserCodeException)] = value.ToString();
}

public bool IncludeEmptyEntriesInMessagePayload
{
    get => GetBoolCapability(nameof(IncludeEmptyEntriesInMessagePayload));
    set => Capabilities[nameof(IncludeEmptyEntriesInMessagePayload)] = value.ToString();
}

private bool GetBoolCapability(string name)
{
    return Capabilities.TryGetValue(name, out string value) && bool.TryParse(value, out bool b) && b;
}
```

samples/FunctionApp/Program.cs
```diff
@@ -21,11 +18,10 @@ static async Task Main(string[] args)
...
-                        .AddApplicationInsights()
-                        .AddApplicationInsightsLogger();
+                    // example (will remove before merge)
+                    options.Capabilities.Remove("HandlesWorkerTerminateMessage");
```
surgupta-msft, line 24: We can add a unit test for this case if not done already.

surgupta-msft (approved):

samples/FunctionApp/Program.cs
brettsam: This change won't go into the final commit -- was just for testing locally.

samples/FunctionApp/Program.cs
brettsam: yes -- need to add some tests

src/DotNetWorker.Core/Hosting/ServiceCollectionExtensions.cs
```diff
@@ -117,6 +117,37 @@ internal static IServiceCollection AddDefaultInputConvertersToWorkerOptions(this
...
+        {
+            return services.Configure<WorkerOptions>((workerOptions) =>
+            {
+                workerOptions.AddDefaultCapabilities();
+            });
+        }
+
+        internal static void AddDefaultCapabilities(this WorkerOptions workerOptions)
+        {
+            workerOptions.Capabilities["RpcHttpBodyOnly"] = bool.TrueString;
```
fabiocav, line 130: Do we need some validation in place to ensure some capabilities have the expected value?  Some of these (e.g., `UseNullableValueDictionaryForHttp`) could potentially cause problems if overridden, right?

src/DotNetWorker.Core/Hosting/WorkerOptions.cs
brettsam: Nice.

src/DotNetWorker.Core/Hosting/WorkerOptions.cs
brettsam: Yeah, I'm transitioning this that way. Good idea.

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -210,22 +211,18 @@ internal static WorkerInitResponse WorkerInitRequestHandler(WorkerInitRequest re
...
             response.Capabilities.Add("RpcHttpTriggerMetadataRemoved", bool.TrueString);
             response.Capabilities.Add("UseNullableValueDictionaryForHttp", bool.TrueString);
             response.Capabilities.Add("TypedDataCollection", bool.TrueString);
             response.Capabilities.Add("WorkerStatus", bool.TrueString);
-            response.Capabilities.Add("HandlesWorkerTerminateMessage", bool.TrueString);
-            response.Capabilities.Add("HandlesInvocationCancelMessage", bool.TrueString);
 
-            if (workerOptions.EnableUserCodeException)
+            // Add additional capabilites defined by WorkerOptions
+            foreach (KeyValuePair<string, string> entry in workerOptions.Capabilities)
```
jviau, line 223: I think you can deconstruct KVP:

```suggestion
            foreach ((string key, string value) in workerOptions.Capabilities)
```

It might not be available in `netstandard2.0` though. But if you are interested, you can include it yourself. I tend to have a lang-compat file for netstandard2.0 projects: https://github.com/microsoft/durabletask-dotnet/blob/main/src/Shared/Core/CSharpLangCompat.cs

``` CSharp
// These extension methods are defined in the global namespace so that they're available everywhere.
internal static class KeyValuePairExtensions
{
    // Based on https://source.dot.net/#System.Private.CoreLib/KeyValuePair.cs,aa57b8e336bf7f59
    [EditorBrowsable(EditorBrowsableState.Never)]
    public static void Deconstruct<TKey, TValue>(this KeyValuePair<TKey, TValue> pair, out TKey key, out TValue value)
    {
        key = pair.Key;
        value = pair.Value;
    }
}
```

This is all if you are interested in that change. No worries if not interested.

jviau (approved):

src/DotNetWorker.Core/Hosting/WorkerOptions.cs
```diff
@@ -22,19 +23,56 @@ public class WorkerOptions
...
+            return Capabilities.TryGetValue(name, out string value) && bool.TryParse(value, out bool b) && b;
+        }
+
+        // For false values, the host does not expect the capability to exist; there are some cases where this
+        // will be interpreted as "true" just because the key is there.
+        private void SetBoolCapability(string name, bool value)
+        {
+            if (value)
+            {
+                Capabilities[name] = value.ToString();
```
kshyju, line 70: Should we use the static `bool.TrueString`  instead of calling `ToString` on the value ?

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -210,22 +211,18 @@ internal static WorkerInitResponse WorkerInitRequestHandler(WorkerInitRequest re
...
             response.Capabilities.Add("RawHttpBodyBytes", bool.TrueString);
             response.Capabilities.Add("RpcHttpTriggerMetadataRemoved", bool.TrueString);
             response.Capabilities.Add("UseNullableValueDictionaryForHttp", bool.TrueString);
             response.Capabilities.Add("TypedDataCollection", bool.TrueString);
             response.Capabilities.Add("WorkerStatus", bool.TrueString);
-            response.Capabilities.Add("HandlesWorkerTerminateMessage", bool.TrueString);
-            response.Capabilities.Add("HandlesInvocationCancelMessage", bool.TrueString);
 
-            if (workerOptions.EnableUserCodeException)
+            // Add additional capabilites defined by WorkerOptions
```
kshyju, line 222: nit: typo

kshyju (approved):

src/DotNetWorker.Core/Hosting/WorkerOptions.cs
```diff
@@ -22,19 +23,56 @@ public class WorkerOptions
         /// </summary>
         public InputConverterCollection InputConverters { get; } = new InputConverterCollection();
 
+        /// <summary>
+        /// Gets the optional worker capabilities.
+        /// </summary>
+        public IDictionary<string, string> Capabilities { get; } = new Dictionary<string, string>()
```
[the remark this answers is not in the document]
brettsam: Can't do that here since it's an interface...

src/DotNetWorker.Core/Hosting/ServiceCollectionExtensions.cs
brettsam: I changed this so there are some immutable values.

surgupta-msft (approved):

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -210,24 +210,20 @@ internal static WorkerInitResponse WorkerInitRequestHandler(WorkerInitRequest re
...
             {
-                response.Capabilities.Add("EnableUserCodeException", bool.TrueString);
-            }
-            if (workerOptions.IncludeEmptyEntriesInMessagePayload)
-            {
-                response.Capabilities.Add("IncludeEmptyEntriesInMessagePayload", bool.TrueString);
+                response.Capabilities[key] = value;
             }
 
+            // Add required capabilities; these cannot be modified and will override anything from WorkerOptions
```
liliankasem, line 219: Doesn't really make a difference, but would expect the required items to be added before the additional capabilities

src/DotNetWorker.Core/Hosting/WorkerOptions.cs
```diff
@@ -22,19 +23,56 @@ public class WorkerOptions
...
+        {
+            // Enable these by default, although they are not strictly required and can be removed
+            { "HandlesWorkerTerminateMessage", bool.TrueString },
+            { "HandlesInvocationCancelMessage", bool.TrueString }
```
liliankasem, line 33: Does it make sense to also add all the required items here by default instead of adding them in GrpcWorker?

liliankasem (approved):

src/DotNetWorker.Grpc/GrpcWorker.cs
brettsam: I wanted to make sure that the required ones "win" -- so even if you set one of these to false, it ends up being true.

src/DotNetWorker.Core/Hosting/WorkerOptions.cs
brettsam: I only wanted to expose the editable Capabilities -- but calling them that is too confusing. And I didn't want to have some ReadOnlyDictionary here alongside an editable Dictionary -- so I ended up with what I've got today where we "force" our mandatory settings after adding the editable ones.

src/DotNetWorker.Core/Context/Features/IDictionaryExtensions.cs
```diff
@@ -8,49 +8,59 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.ComponentModel;
 
-namespace Microsoft.Azure.Functions.Worker
```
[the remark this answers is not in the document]
brettsam: See @jviau's comment above -- https://github.com/Azure/azure-functions-dotnet-worker/pull/1183#discussion_r1048970457.

In short, these are for compiled netstandard2.0 only and bridge some gaps that newer .NET versions have added. By removing the namespace, they're all available everywhere.

fabiocav (approved):
Small nit

PR #1191, opened 2022-11-21
[Source Gen] Create JSON directly instead of using anonymous type

### LINKED ISSUE #1188: [Source Gen] Refactor to create json strings directly instead of using anonymous type
Instead of emitting anonymous types then serializing them into JSON (see [here](https://github.com/Azure/azure-functions-dotnet-worker/blob/c936ce7ed411612dfb88a892579ec98b977f6a6a/sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Emitter.cs#L135)), create JSON strings from the start to improve performance.

satvu:

resolves #1188 

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/sdk/Sdk.Generators/Abstractions/GeneratorFunctionMetadata.cs
+++ b/sdk/Sdk.Generators/Abstractions/GeneratorFunctionMetadata.cs
@@ -21,6 +21,6 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
 
         public IDictionary<string, object> Properties { get; set; } = new Dictionary<string, object>();
 
-        public IList<IDictionary<string, string>> RawBindings { get; set; } = new List<IDictionary<string, string>>(); // List of <propertyName, propertyValue> bindings.
+        public IList<IDictionary<string, object>> RawBindings { get; set; } = new List<IDictionary<string, object>>(); // List of <propertyName, propertyValue> bindings.
     }
 }
--- a/sdk/Sdk.Generators/Extensions/StringExtensions.cs
+++ b/sdk/Sdk.Generators/Extensions/StringExtensions.cs
@@ -27,26 +27,7 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
                 result = result.TrimStringFromEnd(s);
             }
 
-            return result; 
-        }
-
-        public static string UppercaseFirst(this string str)
-        {
-            // Check for empty string.
-            if (string.IsNullOrEmpty(str))
-            {
-                return string.Empty;
-            }
-
-            if (!char.IsUpper(str[0]))
-            {
-                // Return char and concat substring.
-                return char.ToUpper(str[0]) + str.Substring(1);
-            }
-            else
-            {
-                return str;
-            }
+            return result;
         }
 
         public static string LowercaseFirst(this string str)
--- a/sdk/Sdk.Generators/Extensions/StringExtensions.cs
+++ b/sdk/Sdk.Generators/Extensions/StringExtensions.cs
@@ -47,7 +47,25 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
             {
                 return str;
             }
+        }
 
+        public static string LowercaseFirst(this string str)
+        {
+            // Check for empty string.
+            if (string.IsNullOrEmpty(str))
+            {
+                return string.Empty;
+            }
+
+            if (!char.IsLower(str[0]))
+            {
+                // Return char and concat substring.
+                return char.ToLower(str[0]) + str.Substring(1);
+            }
+            else
+            {
+                return str;
+            }
         }
     }
 }
--- a/sdk/Sdk.Generators/Extensions/StringExtensions.cs
+++ b/sdk/Sdk.Generators/Extensions/StringExtensions.cs
@@ -41,7 +41,7 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
             if (!char.IsLower(str[0]))
             {
                 // Return char and concat substring.
-                return char.ToLower(str[0]) + str.Substring(1);
+                return char.ToLowerInvariant(str[0]) + str.Substring(1);
             }
             else
             {
--- a/sdk/Sdk.Generators/Extensions/StringExtensions.cs
+++ b/sdk/Sdk.Generators/Extensions/StringExtensions.cs
@@ -29,24 +29,5 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
 
             return result;
         }
-
-        public static string LowercaseFirst(this string str)
-        {
-            // Check for empty string.
-            if (string.IsNullOrEmpty(str))
-            {
-                return string.Empty;
-            }
-
-            if (!char.IsLower(str[0]))
-            {
-                // Return char and concat substring.
-                return char.ToLowerInvariant(str[0]) + str.Substring(1);
-            }
-            else
-            {
-                return str;
-            }
-        }
     }
 }
--- a/sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Emitter.cs
+++ b/sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Emitter.cs
@@ -5,6 +5,7 @@ using System.CodeDom.Compiler;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text;
 using System.Threading;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Text;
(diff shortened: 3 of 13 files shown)


satvu:
Pushing out initial prototype quickly for cold-start testing, but to do:

- Fix generation test cases
- Clean up casing logic. I think we can drop uppercasing the first letter since we are no longer using an anonymous type and just going straight to a JSON string.
- Only E2E tested one HTTP Trigger so need to make sure others are working as expected.

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Emitter.cs
```diff
@@ -122,18 +123,31 @@ private void AddBindingInfo(IndentedTextWriter indentedTextWriter, string functi
                 foreach (var binding in bindings)
                 {
                     var bindingVarName = functionVarName + "binding" + bindingCount.ToString();
-                    indentedTextWriter.WriteLine($"var {bindingVarName} = new {{");
-                    indentedTextWriter.Indent++;
-                    
+                    StringBuilder jsonStr = new StringBuilder("$\"{{");
```
fabiocav, line 126: @satvu is there a reason we're building the JSON string by hand here? We should be able to use JSON serialization in the context of the source generator, but persist the JSON string in the generated code.

satvu: Let me check if I can make that work with the AuthorizationLevel - we can't parse it to the correct value until runtime b/c it depends on the HTTP Extensions package which we won't have access to.

sdk/Sdk.Generators/Extensions/StringExtensions.cs
```diff
@@ -27,27 +27,26 @@ public static string TrimStringsFromEnd(this string str, IReadOnlyList<string> s
...
             {
                 return string.Empty;
             }
 
-            if (!char.IsUpper(str[0]))
+            if (!char.IsLower(str[0]))
             {
                 // Return char and concat substring.
-                return char.ToUpper(str[0]) + str.Substring(1);
+                return char.ToLower(str[0]) + str.Substring(1);
```
jviau, line 44: I _think_ you want invariant culture conversion:

```suggestion
                return char.ToLowerInvariant(str[0]) + str.Substring(1);
```

sdk/Sdk.Generators/Sdk.Generators.csproj
```diff
@@ -20,6 +21,7 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.11.0" />
+    <PackageReference Include="System.Text.Json" Version="5.0.2" />
```
jviau, line 24: Analyzers have to package their dependencies manually. See [here](https://github.com/dotnet/roslyn/blob/main/docs/features/source-generators.cookbook.md#use-functionality-from-nuget-packages). Specifically, the section which starts with:

> However, any _generation-time_ dependencies, that is, used by the generator while it is is running and generating code, must be packaged directly alongside the generator assembly inside the generator NuGet package.

jviau, line 24: This might need `PrivateAssets="all"`. Same with the `Microsoft.CodeAnalaysis.CSharp` package. Analyzer packages should not cause transitive references.

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -570,7 +525,17 @@ private bool TryLoadConstructorArguments(AttributeData attributeData, IDictionar
                             dict[argumentName] = arg.Value;
                             break;
                         case TypedConstantKind.Enum:
-                            dict[argumentName] = $"(AuthorizationLevel){arg.Value}"; // the only enum type we have in function metadata is authlevel
+                            // Currently investigating why different scenarios fail here and a Type can't get loaded
+                            var typeName = arg.Type!.GetAssemblyQualifiedName();
+                            var type = Type.GetType(typeName);
+                            if (type != null)
```
jviau, line 531: Consider a ternary for conciseness:

``` CSharp
dict[argumentName] = type is null ? "Anonymous" : Enum.GetName(type, arg.Value);
```

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Emitter.cs
```diff
@@ -115,25 +121,17 @@ private void AddFunctionMetadataInfo(IndentedTextWriter indentedTextWriter, IRea
...
-                        indentedTextWriter.WriteLine($"{key} = {binding[key]},");
-                    }
 
-                    indentedTextWriter.Indent--;
-                    indentedTextWriter.WriteLine("};");
-                    indentedTextWriter.WriteLine($"var {bindingVarName}JSON = JsonSerializer.Serialize({bindingVarName});");
-                    indentedTextWriter.WriteLine($"{functionBindingsListVarName}.Add({bindingVarName}JSON);");
+                    var jsonBinding = JsonSerializer.Serialize(binding, _jsonOptions);
+
+                    indentedTextWriter.WriteLine($"{functionBindingsListVarName}.Add($\"{{{jsonBinding.Replace("\"", "\\\"")}}}\");");
```
jviau, line 134: Would be helpful to see tests with varied data to see how this works. Is there any input that would escape this string and break compilation?

sdk/Sdk.Generators/Abstractions/GeneratorFunctionMetadata.cs
```diff
@@ -21,6 +21,6 @@ internal class GeneratorFunctionMetadata
 
         public IDictionary<string, object> Properties { get; set; } = new Dictionary<string, object>();
 
-        public IList<IDictionary<string, string>> RawBindings { get; set; } = new List<IDictionary<string, string>>(); // List of <propertyName, propertyValue> bindings.
+        public IList<IDictionary<string, object>> RawBindings { get; set; } = new List<IDictionary<string, object>>(); // List of <propertyName, propertyValue> bindings.
```
satvu, line 24: `string` -> `object` change is due to switching to JSON serialization. Previously, arrays were stored as a string that would be written right to the generated file, but now we need the actual array object.

sdk/Sdk.Generators/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "profiles": {
+    "SourceGenDebug": {
+      "commandName": "DebugRoslynComponent",
```
satvu, line 4: Used to debug against a target project at compilation time - I think this is worth keeping.

satvu: Debating on how to add the other component for this ([example setup](https://github.com/JoanComasFdz/dotnet-how-to-debug-source-generator-vs2022)) in the `FunctionApp.csproj` file since it can make other debugging things work poorly with another project reference. Other options:

1. Add the necessary component but commented out w/ additional warnings/instructions.
2. Add SourceGen debugging instructions to README.md or Wiki, leaving just this launch profile here and the other component in the documentation for devs to copy/paste over.

sdk/Sdk.Generators/Extensions/StringExtensions.cs
```diff
@@ -27,27 +27,26 @@ public static string TrimStringsFromEnd(this string str, IReadOnlyList<string> s
...
             }
 
-            return result; 
+            return result;
         }
 
-        public static string UppercaseFirst(this string str)
+        public static string LowercaseFirst(this string str)
         {
             // Check for empty string.
```
surgupta-msft, line 35: Nit: We can delete this comment

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -276,23 +232,23 @@ private bool TryGetParameterInputAndTriggerBindings(MethodDeclarationSyntax meth
...
                             if (dataType is not DataType.Undefined)
                             {
-                                bindingDict!.Add("DataType", FormatObject(Enum.GetName(typeof(DataType), dataType)));
+                                bindingDict!.Add("dataType", Enum.GetName(typeof(DataType), dataType));
                             }
 
                             // special handling for EventHubsTrigger - we need to define a property called "Cardinality"
                             if (validEventHubs)
                             {
                                 if (!bindingDict!.ContainsKey("Cardinality"))
```
surgupta-msft, line 249: Wanted to double check if this code is replacing the key from "Cardinality" to "cardinality" and not adding a new one.

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -570,7 +525,10 @@ private bool TryLoadConstructorArguments(AttributeData attributeData, IDictionar
                             dict[argumentName] = arg.Value;
                             break;
                         case TypedConstantKind.Enum:
-                            dict[argumentName] = $"(AuthorizationLevel){arg.Value}"; // the only enum type we have in function metadata is authlevel
+                            var enumValue = arg.Type!.GetMembers()
+                              .FirstOrDefault(m => m is IFieldSymbol field
+                                                            && field.ConstantValue is object value
```
surgupta-msft, line 530: Nit: Indentation update

test/Sdk.Generator.Tests/FunctionMetadataProviderGeneratorTests/ExtensionsTests.cs
```diff
@@ -35,12 +35,12 @@ public void TrimStringsFromEndWorks(string input, string expectedOutput)
             }
 
             [Theory]
-            [InlineData("isBatched", "IsBatched")]
-            [InlineData("MyProperty", "MyProperty")]
-            [InlineData("myproperty", "Myproperty")]
+            [InlineData("isBatched", "isBatched")]
+            [InlineData("MyProperty", "myProperty")]
+            [InlineData("myproperty", "myproperty")]
             public void UpperCaseFirstLetter(string input, string expectedOutput)
```
surgupta-msft, line 41: Do we want to update the test name?

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Emitter.cs
```diff
@@ -5,16 +5,20 @@
...
 
 namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
 {
     public partial class FunctionMetadataProviderGenerator
     {
         internal sealed class Emitter
         {
+            JsonSerializerOptions _jsonOptions = new JsonSerializerOptions
+            {
+                PropertyNamingPolicy = JsonNamingPolicy.CamelCase
```
fabiocav, line 20: To make the world a happier place, also use: `DictionaryKeyPolicy = JsonNamingPolicy.CamelCase,`

satvu: This is what I needed!! Thank you - I think it removes the need for the lowercasing logic discussed offline but will validate first.

sdk/Sdk.Generators/Sdk.Generators.csproj
```diff
@@ -7,24 +7,32 @@
...
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.11.0" PrivateAssets="all" />
+    <PackageReference Include="System.Text.Json" Version="5.0.2" PrivateAssets="all" GeneratePathProperty="true" />
   </ItemGroup>
-
+    
   <ItemGroup>
     <!-- Package the generator in the analyzer directory of the nuget package -->
     <None Include="$(OutputPath)\$(AssemblyName).dll" Pack="true" PackagePath="analyzers/dotnet/cs" Visible="false" />
+  
+    <None Include="$(PKGSystem_Text_Json)\lib\netstandard2.0\*.dll" Pack="true" PackagePath="analyzers/dotnet/cs" Visible="false" />
```
jviau, line 33: I can see you are using this package, but where/why are the other two needed? Are they dependencies of STJ?

jviau (approved):
This is great to see! Much simpler in the end I think.

sdk/Sdk.Generators/Sdk.Generators.csproj
satvu: Yes they're dependencies of STJ and they won't get loaded with STJ ... had to add each package associated with a "couldn't find file or assembly" until the generator built successfully.

fabiocav (approved):

PR #1200, opened 2022-11-28
[SourceGen] Add MSBuild prop to disable source-generated function metadata

### LINKED ISSUE #1155: Add MSBuild Property to disable/enable function metadata source generator
Add additional feature to enable/disable source generator in the case an app is caught by edge case (or other) failure in the generated file. (see #1154)

satvu:

resolves #1155 

### Pull request checklist

### Additional information

Using guidance from [this section](https://github.com/dotnet/roslyn/blob/main/docs/features/source-generators.cookbook.md#consume-msbuild-properties-and-metadata) of the generators cookbook.

For manual testing, I recommend creating a local nuget package of the SDK and referencing that in a simple app (like http trigger template app).

### DIFF
--- a/sdk/Sdk.Generators/Constants.cs
+++ b/sdk/Sdk.Generators/Constants.cs
@@ -5,6 +5,12 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
 {
     internal static class Constants
     {
+        // MSBuild Props
+        internal const string EnableSourceGenProp = "build_property.FunctionsMetadataSourceGen_Enabled";
+
+        // Generated File Names
+        internal const string GeneratedFunctionMetadataFileName = "GeneratedFunctionMetadataProvider.g.cs";
+
         // Our types
         internal const string BindingAttributeType = "Microsoft.Azure.Functions.Worker.Extensions.Abstractions.BindingAttribute";
         internal const string OutputBindingAttributeType = "Microsoft.Azure.Functions.Worker.Extensions.Abstractions.OutputBindingAttribute";
--- a/sdk/Sdk.Generators/Constants.cs
+++ b/sdk/Sdk.Generators/Constants.cs
@@ -12,37 +12,37 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
 
         public static class FileNames
         {
-            internal const string GeneratedFunctionMetadataFileName = "GeneratedFunctionMetadataProvider.g.cs";
+            internal const string GeneratedFunctionMetadata = "GeneratedFunctionMetadataProvider.g.cs";
         }
 
         public static class Types
         {
             // Our types
-            internal const string BindingAttributeType = "Microsoft.Azure.Functions.Worker.Extensions.Abstractions.BindingAttribute";
-            internal const string OutputBindingAttributeType = "Microsoft.Azure.Functions.Worker.Extensions.Abstractions.OutputBindingAttribute";
-            internal const string FunctionNameType = "Microsoft.Azure.Functions.Worker.FunctionAttribute";
-            internal const string HttpResponseType = "Microsoft.Azure.Functions.Worker.Http.HttpResponseData";
-            internal const string HttpTriggerBindingType = "Microsoft.Azure.Functions.Worker.HttpTriggerAttribute";
-            internal const string EventHubsTriggerType = "Microsoft.Azure.Functions.Worker.EventHubTriggerAttribute";
-            internal const string BindingPropertyNameAttributeType = "Microsoft.Azure.Functions.Worker.Extensions.Abstractions.BindingPropertyNameAttribute";
-            internal const string DefaultValueType = "Microsoft.Azure.Functions.Worker.Extensions.Abstractions.DefaultValueAttribute";
+            internal const string BindingAttribute = "Microsoft.Azure.Functions.Worker.Extensions.Abstractions.BindingAttribute";
+            internal const string OutputBindingAttribute = "Microsoft.Azure.Functions.Worker.Extensions.Abstractions.OutputBindingAttribute";
+            internal const string FunctionName = "Microsoft.Azure.Functions.Worker.FunctionAttribute";
+            internal const string HttpResponse = "Microsoft.Azure.Functions.Worker.Http.HttpResponseData";
+            internal const string HttpTriggerBinding = "Microsoft.Azure.Functions.Worker.HttpTriggerAttribute";
+            internal const string EventHubsTrigger = "Microsoft.Azure.Functions.Worker.EventHubTriggerAttribute";
+            internal const string BindingPropertyNameAttribute = "Microsoft.Azure.Functions.Worker.Extensions.Abstractions.BindingPropertyNameAttribute";
+            internal const string DefaultValue = "Microsoft.Azure.Functions.Worker.Extensions.Abstractions.DefaultValueAttribute";
 
             // System types
-            internal const string IEnumerableType = "System.Collections.IEnumerable";
-            internal const string IEnumerableGenericType = "System.Collections.Generic.IEnumerable`1";
-            internal const string IEnumerableOfStringType = "System.Collections.Generic.IEnumerable`1<System.String>";
-            internal const string IEnumerableOfBinaryType = "System.Collections.Generic.IEnumerable`1<System.Byte[]>";
+            internal const string IEnumerable = "System.Collections.IEnumerable";
+            internal const string IEnumerableGeneric = "System.Collections.Generic.IEnumerable`1";
+            internal const string IEnumerableOfString = "System.Collections.Generic.IEnumerable`1<System.String>";
+            internal const string IEnumerableOfBinary = "System.Collections.Generic.IEnumerable`1<System.Byte[]>";
             internal const string IEnumerableOfT = "System.Collections.Generic.IEnumerable`1<T>";
             internal const string IEnumerableOfKeyValuePair = "System.Collections.Generic.IEnumerable`1<System.Collections.Generic.KeyValuePair`2<TKey,TValue>>";
-            internal const string StringType = "System.String";
-            internal const string ByteArrayType = "System.Byte[]";
-            internal const string ByteStructType = "System.Byte";
-            internal const string TaskGenericType = "System.Threading.Tasks.Task`1";
-            internal const string TaskType = "System.Threading.Tasks.Task";
-            internal const string VoidType = "System.Void";
+            internal const string String = "System.String";
+            internal const string ByteArray = "System.Byte[]";
+            internal const string ByteStruct = "System.Byte";
+            internal const string TaskGeneric = "System.Threading.Tasks.Task`1";
+            internal const string Task = "System.Threading.Tasks.Task";
+            internal const string Void = "System.Void";
             internal const string ReadOnlyMemoryOfBytes = "System.ReadOnlyMemory`1<System.Byte>";
-            internal const string LookupGenericType = "System.Linq.Lookup`2";
-            internal const string DictionaryGenericType = "System.Collections.Generic.Dictionary`2";
+            internal const string LookupGeneric = "System.Linq.Lookup`2";
+            internal const string DictionaryGeneric = "System.Collections.Generic.Dictionary`2";
         }
         
         public static class FunctionMetadataBindingProps {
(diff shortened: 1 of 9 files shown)


sdk/Sdk.Generators/Sdk.Generators.csproj
```diff
@@ -10,6 +10,7 @@
     <EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>
     <IncludeBuildOutput>false</IncludeBuildOutput>
     <VersionSuffix>-preview2</VersionSuffix>
+    <IsRoslynComponent>true</IsRoslynComponent>
```
satvu, line 13: Trying to get #1191 merged before this and will rebase after - leaving these changes checked in for now in case anyone wants to try this out locally.

sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.props
```diff
@@ -20,6 +20,9 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
         <ProjectCapability Include="AzureFunctions"/>
   </ItemGroup>
 
+  <ItemGroup>
+    <CompilerVisibleProperty Include="FunctionsDisableFuncMetadataSourceGen" />
```
satvu, line 24: Very open to name suggestions - this is just a first thought that I used for prototyping

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.cs
```diff
@@ -26,15 +26,23 @@ public void Execute(GeneratorExecutionContext context)
...
             var p = new Parser(context);
-            IReadOnlyList<GeneratorFunctionMetadata> functionMetadataInfo = p.GetFunctionMetadataInfo(receiver.CandidateMethods);
 
-            // Proceed to generate the file if function metadata info was successfully returned
-            if (functionMetadataInfo.Count > 0)
+            context.AnalyzerConfigOptions.GlobalOptions.TryGetValue("build_property.FunctionsDisableFuncMetadataSourceGen", out var sourceGenSwitch);
+
+            bool.TryParse(sourceGenSwitch, out bool disableSourceGen);
+
+            if (!disableSourceGen)
```
liliankasem, line 34: Nit: do we want to check for disableSourceGen == true and return early, so that we can move logic out of the `{ }` similar to the patten on line 22

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.cs
```diff
@@ -26,15 +26,23 @@ public void Execute(GeneratorExecutionContext context)
 
             // attempt to parse user compilation
             var p = new Parser(context);
-            IReadOnlyList<GeneratorFunctionMetadata> functionMetadataInfo = p.GetFunctionMetadataInfo(receiver.CandidateMethods);
 
-            // Proceed to generate the file if function metadata info was successfully returned
-            if (functionMetadataInfo.Count > 0)
+            context.AnalyzerConfigOptions.GlobalOptions.TryGetValue("build_property.FunctionsDisableFuncMetadataSourceGen", out var sourceGenSwitch);
```
liliankasem, line 30: Can we make this a const somewhere?

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.cs
```diff
@@ -26,15 +26,23 @@ public void Execute(GeneratorExecutionContext context)
...
+                IReadOnlyList<GeneratorFunctionMetadata> functionMetadataInfo = p.GetFunctionMetadataInfo(receiver.CandidateMethods);
 
-                context.AddSource($"GeneratedFunctionMetadataProvider.g.cs", SourceText.From(result, Encoding.UTF8));
+                // Proceed to generate the file if function metadata info was successfully returned
+                if (functionMetadataInfo.Count > 0)
+                {
+                    var e = new Emitter();
+                    string result = e.Emit(functionMetadataInfo, context.CancellationToken);
+
+                    context.AddSource($"GeneratedFunctionMetadataProvider.g.cs", SourceText.From(result, Encoding.UTF8));
```
liliankasem, line 44: Same here about using const

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.cs
```diff
@@ -26,15 +26,23 @@ public void Execute(GeneratorExecutionContext context)
...
             {
-                var e = new Emitter();
-                string result = e.Emit(functionMetadataInfo, context.CancellationToken);
+                IReadOnlyList<GeneratorFunctionMetadata> functionMetadataInfo = p.GetFunctionMetadataInfo(receiver.CandidateMethods);
 
-                context.AddSource($"GeneratedFunctionMetadataProvider.g.cs", SourceText.From(result, Encoding.UTF8));
+                // Proceed to generate the file if function metadata info was successfully returned
+                if (functionMetadataInfo.Count > 0)
+                {
+                    var e = new Emitter();
```
liliankasem, line 41: fancy nit: `Emitter e = new();`

sdk/Sdk.Generators/Properties/launchSettings.json
```diff
@@ -1,6 +1,6 @@
 {
   "profiles": {
-    "SourceGenDebug": {
+    "Source Generator": {
```
liliankasem, line 3: Could having a space in the name lead to issues? `SourceGenerator` feels safer to me for some reason

sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.props
liliankasem: Only note I would have is removing fun in the middle but I'm fine as is: `FunctionsDisableMetadataSourceGen`

liliankasem (approved):

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.cs
jviau, line 30: Small suggestion: can we start a source gen property naming format? So that if we introduce other build props to control source gen later on, they will all follow the same naming. I'd recommend something like: `FunctionsMetadataSourceGen_{Property}`, so `FunctionsMetadataSourceGen_Enabled` (and have it default to `true` if not present).

The choice for "Enabled" is also a different preference of mine: I like to name bools in the _positive_ term (I do make occasional exceptions though).

sdk/Sdk.Generators/Properties/launchSettings.json
jviau: There shouldn't be an issue. This is mainly a display string for Visual Studio launch/debug dropdown. But I am also not against removing the space if you feel strongly for that.

sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.props
jviau: I left another comment with a naming suggestion: `FunctionsMetadataSourceGen_Enabled`. That comment has an explanation for why.

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.cs
satvu: @jviau I was thinking that if we plan to make source-gen the default, then the action should be to disable it. Positive term is definitely more intuitive though - changed it to that.

sdk/Sdk.Generators/Constants.cs
```diff
@@ -5,6 +5,12 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
 {
     internal static class Constants
     {
+        // MSBuild Props
+        internal const string EnableSourceGenProp = "build_property.FunctionsMetadataSourceGen_Enabled";
```
jviau, line 9: Small suggestion to consider: I like using nested classes to organize my constants as the file gets bigger. Something like:

``` CSharp
internal static class Constants
{
    public static class BuildProperties
    {
        public const string SourceGenEnabled = "build_property.FunctionsMetadataSourceGen_Enabled";
    }

    public static class Types
    {
        public const string BindingAttribute = "Microsoft.Azure.Functions.Worker.Extensions.Abstractions.BindingAttribute";
        public const string OutputBindingAttribute = "Microsoft.Azure.Functions.Worker.Extensions.Abstractions.OutputBindingAttribute";
    }
}
```

But of course, this is just how I like to do it.

sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
```diff
@@ -34,6 +34,7 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
         <MSBuildFunctionsTargetsPath>$(MSBuildExtensionsPath)\Microsoft\VisualStudio\Managed.Functions\</MSBuildFunctionsTargetsPath>
       
         <FunctionsEnableWorkerIndexing Condition="$(FunctionsEnableWorkerIndexing) == ''">false</FunctionsEnableWorkerIndexing>
+        <FunctionsDisableFuncMetadataSourceGen Condition="$(FunctionsMetadataSourceGen_Enabled) == ''">true</FunctionsDisableFuncMetadataSourceGen>
```
jviau, line 37: Need to update this properter name:

```suggestion
        <FunctionsMetadataSourceGen_Enabled Condition="$(FunctionsMetadataSourceGen_Enabled) == ''">true</FunctionsMetadataSourceGen_Enabled>
```

jviau (changes requested):

jviau:
bump to avoid closure.

sdk/Sdk.Generators/Constants.cs
```diff
@@ -5,41 +5,50 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
...
+        public static class BuildProperties
+        {
+            internal const string EnableSourceGenProp = "build_property.FunctionsMetadataSourceGen_Enabled";
+        }
 
-        // Generated File Names
-        internal const string GeneratedFunctionMetadataFileName = "GeneratedFunctionMetadataProvider.g.cs";
+        public static class FileNames
+        {
+            internal const string GeneratedFunctionMetadataFileName = "GeneratedFunctionMetadataProvider.g.cs";
```
jviau, line 15: With the subclass, you can drop some of the suffixes now as the subclass covers that:

```suggestion
            internal const string GeneratedFunctionMetadata = "GeneratedFunctionMetadataProvider.g.cs";
```

This will still be clear in usage as we would write:

``` CSharp
Constants.FileNames.GeneratedFunctionMetadata;
```

jviau (approved):

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.cs
```diff
@@ -26,15 +26,25 @@ public void Execute(GeneratorExecutionContext context)
...
+
+            context.AnalyzerConfigOptions.GlobalOptions.TryGetValue(Constants.BuildProperties.EnableSourceGenProp, out var sourceGenSwitch);
+
+            bool.TryParse(sourceGenSwitch, out bool enableSourceGen);
+
+            if (!enableSourceGen)
+            {
+                return;
+            }
+
```
kshyju, line 38: Move the `Parser` instance creation(line 28) here. No need to run that constructor if we are going to return.

kshyju (approved):
💰

sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
```diff
@@ -34,6 +34,7 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
         <MSBuildFunctionsTargetsPath>$(MSBuildExtensionsPath)\Microsoft\VisualStudio\Managed.Functions\</MSBuildFunctionsTargetsPath>
       
         <FunctionsEnableWorkerIndexing Condition="$(FunctionsEnableWorkerIndexing) == ''">false</FunctionsEnableWorkerIndexing>
+        <FunctionsMetadataSourceGen_Enabled Condition="$(FunctionsMetadataSourceGen_Enabled) == ''">true</FunctionsMetadataSourceGen_Enabled>
```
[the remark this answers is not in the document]
satvu: Yes - the current behavior is that the generation of the file happens automatically so I followed that and added this prop to disable it. Pros and cons for either default - mostly depends on how we want this rolled out and how we want customers to opt-in (more steps, less steps?).

PR #1213, opened 2022-12-06
Native host integration

fabiocav:
Managed side of the custom host work introduced in #1212 

The code here detects when running in the Functions .NET host and replaces the client to communicate through it.

### Pull request checklist

### DIFF
--- a/src/DotNetWorker.Grpc/Definition/GrpcFunctionDefinition.cs
+++ b/src/DotNetWorker.Grpc/Definition/GrpcFunctionDefinition.cs
@@ -22,6 +22,15 @@ namespace Microsoft.Azure.Functions.Worker.Definition
             Id = loadRequest.FunctionId;
 
             string? scriptRoot = Environment.GetEnvironmentVariable("AzureWebJobsScriptRoot")!;
+            if (string.IsNullOrWhiteSpace(scriptRoot))
+            {
+                throw new ArgumentException("AzureWebJobsScriptRoot env variable value is null or empty");
+            }
+            if (string.IsNullOrWhiteSpace(loadRequest.Metadata.ScriptFile))
+            {
+                throw new ArgumentException("loadRequest.Metadata.ScriptFile value is null or empty");
+            }
+
             string scriptFile = Path.Combine(scriptRoot, loadRequest.Metadata.ScriptFile);
             PathToAssembly = Path.GetFullPath(scriptFile);
 
--- a/src/DotNetWorker.Grpc/Definition/GrpcFunctionDefinition.cs
+++ b/src/DotNetWorker.Grpc/Definition/GrpcFunctionDefinition.cs
@@ -19,8 +19,11 @@ namespace Microsoft.Azure.Functions.Worker.Definition
         {
             EntryPoint = loadRequest.Metadata.EntryPoint;
             Name = loadRequest.Metadata.Name;
-            PathToAssembly = Path.GetFullPath(loadRequest.Metadata.ScriptFile);
             Id = loadRequest.FunctionId;
+
+            string? scriptRoot = Environment.GetEnvironmentVariable("AzureWebJobsScriptRoot")!;
+            string scriptFile = Path.Combine(scriptRoot, loadRequest.Metadata.ScriptFile);
+            PathToAssembly = Path.GetFullPath(scriptFile);
 
             var grpcBindingsGroup = loadRequest.Metadata.Bindings.GroupBy(kv => kv.Value.Direction);
             var grpcInputBindings = grpcBindingsGroup.Where(kv => kv.Key == BindingInfo.Types.Direction.In).FirstOrDefault();
--- a/src/DotNetWorker.Grpc/Definition/GrpcFunctionDefinition.cs
+++ b/src/DotNetWorker.Grpc/Definition/GrpcFunctionDefinition.cs
@@ -24,11 +24,12 @@ namespace Microsoft.Azure.Functions.Worker.Definition
             string? scriptRoot = Environment.GetEnvironmentVariable("AzureWebJobsScriptRoot");
             if (string.IsNullOrWhiteSpace(scriptRoot))
             {
-                throw new ArgumentException("AzureWebJobsScriptRoot env variable value is null or empty");
+                throw new InvalidOperationException("The 'AzureWebJobsScriptRoot' environment variable value is not defined. This is a required environment variable that is automatically set by the Azure Functions runtime.");
             }
+
             if (string.IsNullOrWhiteSpace(loadRequest.Metadata.ScriptFile))
             {
-                throw new ArgumentException("loadRequest.Metadata.ScriptFile value is null or empty");
+                throw new InvalidOperationException($"Metadata for function '{loadRequest.Metadata.Name} ({loadRequest.Metadata.FunctionId})' does not specify a 'ScriptFile'.");
             }
 
             string scriptFile = Path.Combine(scriptRoot, loadRequest.Metadata.ScriptFile);
--- a/src/DotNetWorker.Grpc/Definition/GrpcFunctionDefinition.cs
+++ b/src/DotNetWorker.Grpc/Definition/GrpcFunctionDefinition.cs
@@ -21,7 +21,7 @@ namespace Microsoft.Azure.Functions.Worker.Definition
             Name = loadRequest.Metadata.Name;
             Id = loadRequest.FunctionId;
 
-            string? scriptRoot = Environment.GetEnvironmentVariable("AzureWebJobsScriptRoot")!;
+            string? scriptRoot = Environment.GetEnvironmentVariable("AzureWebJobsScriptRoot");
             if (string.IsNullOrWhiteSpace(scriptRoot))
             {
                 throw new ArgumentException("AzureWebJobsScriptRoot env variable value is null or empty");
--- a/src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
+++ b/src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
@@ -10,6 +10,7 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <MinorProductVersion>7</MinorProductVersion>
     <VersionSuffix>-preview2</VersionSuffix>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
 
   <Import Project="..\..\build\Common.props" />
@@ -39,6 +40,7 @@
     Grpc.Net.Client implementation-->
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="Grpc.Core" Version="2.45.0" />
+    <Compile Remove="NativeHostIntegration/./*.*" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
+++ b/src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
@@ -9,7 +9,7 @@
     <RootNamespace>Microsoft.Azure.Functions.Worker.Grpc</RootNamespace>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <MinorProductVersion>7</MinorProductVersion>
-    <VersionSuffix>-preview2</VersionSuffix>
+    <VersionSuffix>-preview7</VersionSuffix>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
 
--- a/src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
+++ b/src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
@@ -40,7 +40,7 @@
     Grpc.Net.Client implementation-->
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="Grpc.Core" Version="2.45.0" />
-    <Compile Remove="NativeHostIntegration/./*.*" />
+    <Compile Remove="NativeHostIntegration/**" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
+++ b/src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
@@ -8,8 +8,8 @@
     <AssemblyName>Microsoft.Azure.Functions.Worker.Grpc</AssemblyName>
     <RootNamespace>Microsoft.Azure.Functions.Worker.Grpc</RootNamespace>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <MinorProductVersion>7</MinorProductVersion>
-    <VersionSuffix>-preview7</VersionSuffix>
+    <MinorProductVersion>8</MinorProductVersion>
+    <VersionSuffix>-preview1</VersionSuffix>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
 
--- a/src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs
+++ b/src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs
@@ -3,22 +3,16 @@
 
 using System;
 using System.Threading.Channels;
-using Grpc.Core;
+using Microsoft.Azure.Functions.Core;
 using Microsoft.Azure.Functions.Worker;
 using Microsoft.Azure.Functions.Worker.Core.FunctionMetadata;
 using Microsoft.Azure.Functions.Worker.Grpc.Messages;
-using static Microsoft.Azure.Functions.Worker.Grpc.Messages.FunctionRpc;
 using Microsoft.Azure.Functions.Worker.Logging;
 using Microsoft.Azure.Functions.Worker.Grpc;
 using Microsoft.Azure.Functions.Worker.Diagnostics;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Options;
-
-#if NET5_0_OR_GREATER
-using Grpc.Net.Client;
-#endif
 
 namespace Microsoft.Extensions.DependencyInjection
 {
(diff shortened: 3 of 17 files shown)


fabiocav:
Pending release notes update and documentation work.

src/DotNetWorker.Core/Invocation/IFunctionExecutor.cs
```diff
@@ -1,12 +1,12 @@
...
+﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System.Threading.Tasks;
 using Microsoft.Azure.Functions.Worker.Pipeline;
 
 namespace Microsoft.Azure.Functions.Worker.Invocation
 {
-    internal interface IFunctionExecutor
+    public interface IFunctionExecutor
```
fabiocav, line 9: This was a temporary set of changes required to add a custom executor. Will remove, but we'll eventually need this.

src/DotNetWorker.Core/WorkerHostedService.cs
```diff
@@ -14,17 +14,17 @@ internal class WorkerHostedService : IHostedService
         private readonly IWorker _worker;
         private readonly IWorkerDiagnostics _diagnostics;
 
-        public WorkerHostedService(IWorker worker, IWorkerDiagnostics diagnostics)
+        public WorkerHostedService(IWorker worker)
         {
             _worker = worker ?? throw new ArgumentNullException(nameof(worker));
-            _diagnostics = diagnostics ?? throw new ArgumentNullException(nameof(diagnostics));
+            //_diagnostics = diagnostics ?? throw new ArgumentNullException(nameof(diagnostics));
```
fabiocav, line 20: @brettsam will remove this and replace with an issue link to address the logging at initialization

src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
```diff
@@ -25,10 +26,18 @@
...
-    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="5.0.0" />
   </ItemGroup>
 
+    <ItemGroup Condition="'$(TargetFramework)' == 'net5.0'">
+        <PackageReference Include="Microsoft.Extensions.Hosting" Version="5.0.0" />
+        <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="5.0.0" />
+    </ItemGroup>
+
+    <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
+        <PackageReference Include="Microsoft.Extensions.Hosting" Version="6.0.0" />
```
fabiocav, line 37: These changes need to be discussed

src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
```diff
@@ -39,6 +48,7 @@
     Grpc.Net.Client implementation-->
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="Grpc.Core" Version="2.45.0" />
+    <Compile Remove="NativeHostIntegration/./*.*" />
```
jviau, line 51: `**` can be used for "everything recursively".

```suggestion
    <Compile Remove="NativeHostIntegration/**" />
```

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -78,54 +70,16 @@ internal class GrpcWorker : IWorker
...
-                await ProcessRequestAsync(responseStream.Current);
-            }
-        }
-
-        private Task ProcessRequestAsync(StreamingMessage request)
+        Task IMessageProcessor.ProcessMessageAsync(StreamingMessage message)
         {
             // Dispatch and return.
-            Task.Run(() => ProcessRequestCoreAsync(request));
+            Task.Run(() => ProcessRequestCoreAsync(message));
```
jviau, line 81: This seems slightly odd to me: the contract is `Task` returning, but this will fire and forget regardless. In these cases, shouldn't the caller decide if they want to fire & forget vs wait on the completion?

Of course, I may be missing some context here. But maybe there is a way we can shape this contract to line up with the implementation a bit more? Might be as simple as `Task ProcessMessageAsync` -> `ValueTask PostMessageAsync`.

Additionally - is `Task.Run()` necessary here? Or would a:

``` CSharp
_ = ProcessRequestCoreAsync(message);
```

be good enough?

src/DotNetWorker.Grpc/IGrpcRequestHandler.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.Functions.Worker.Grpc.Messages;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Functions.Worker.Grpc
+{
+    internal interface IMessageProcessor
+    {
+        Task ProcessMessageAsync(StreamingMessage request);
```
jviau, line 11: Do we want `CancellationToken` on these APIs? Or is the cancellation inside `StreamingMessage`?

Same question for `IWorkerClient.SendMessageAsync`.

src/DotNetWorker.Grpc/NativeHostIntegration/NativeWorkerClient.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        public ValueTask SendMessageAsync(StreamingMessage message)
+        {
+            return _outputChannelWriter.WriteAsync(message);
+        }
+
+        [UnmanagedCallersOnly]
+        private static unsafe IntPtr HandleRequest(byte** nativeMessage, int nativeMessageSize, IntPtr grpcHandler)
+        {
+            var span = new ReadOnlySpan<byte>(*nativeMessage, nativeMessageSize);
```
jviau, line 66: style nit:
```suggestion
            ReadOnlySpan<byte> span = new(*nativeMessage, nativeMessageSize);
```

src/DotNetWorker.Grpc/NativeHostIntegration/NativeWorkerClientFactory.cs
```diff
@@ -0,0 +1,29 @@
...
+        public NativeWorkerClientFactory(GrpcHostChannel hostChannel)
+        {
+            _hostChannel = hostChannel;
+        }
+
+        public Task<IWorkerClient> StartClientAsync(IMessageProcessor messageProcessor, CancellationToken token)
+        {
+            var nativeHostData = NativeMethods.GetNativeHostData();
+
+            var client = new NativeWorkerClient(messageProcessor, _hostChannel, nativeHostData);
```
jviau, line 23: style nit:
```suggestion
            NativeWorkerClient client = new(messageProcessor, _hostChannel, nativeHostData);
```

src/DotNetWorker.Grpc/IWorkerClientFactory.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Functions.Worker.Grpc
+{
+    internal interface IWorkerClientFactory
+    {
+        Task<IWorkerClient> StartClientAsync(IMessageProcessor messageProcessor, CancellationToken token);
```
jviau, line 11: What do you think of `ValueTask<IWorkerClient>` here? Considering one of our implementations is synchronous.

But also not a big deal at all as this is called once per worker I think?

src/DotNetWorker/DotNetWorker.csproj
```diff
@@ -2,24 +2,33 @@
...
-    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="5.0.0" />
   </ItemGroup>
 
+    <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
+        <PackageReference Include="Microsoft.Extensions.Hosting" Version="6.0.0" />
+        <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="6.0.0" />
+    </ItemGroup>
+
+    <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0' OR '$(TargetFramework)' == 'net5.0'">
+        <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="5.0.0" />
```
jviau, line 27: As with the other csproj, I think the packages can be simplified to just `Microsoft.Extensions.Hosting`

src/DotNetWorker.Core/DotNetWorker.Core.csproj
```diff
@@ -19,16 +19,22 @@
...
-  </ItemGroup>
-
-  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <ItemGroup>
+        <PackageReference Include="Azure.Core" Version="1.10.0" />
+    </ItemGroup>
+    
+    <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
+      <PackageReference Include="Microsoft.Extensions.Hosting" Version="6.0.0" />
+      <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="6.0.0" />
```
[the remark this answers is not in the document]
bobbyangers: @jviau since this is the core module, there is nothing to transitive from?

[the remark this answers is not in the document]
jviau: https://www.nuget.org/packages/Microsoft.Extensions.Hosting#dependencies-body-tab includes `Microsoft.Extensions.Hosting.Abstractions` as a dependency. Including `Microsoft.Extensions.Hosting` as a package will bring in all of its dependencies, and their dependencies, and so on. This is called "transitive" references.

And now, anything that references this package (as either a p2p or package reference) will also get that entire dependency chain.

https://learn.microsoft.com/en-us/nuget/concepts/dependency-resolution

src/DotNetWorker.Core/Invocation/IFunctionExecutor.cs
kshyju: I noticed FunctionApp26 has an implementation of this ([DirectFunctionExecutor](https://github.com/kshyju/FunctionApp26/blob/fabiocav/tweaks/FunctionApp26/DirectFunctionExecutor.cs#L24)). Is our plan to add some source generator to generate that for the customer function?

src/DotNetWorker.Grpc/NativeHostIntegration/NativeMethods.cs
```diff
@@ -0,0 +1,45 @@
...
+using System;
+using System.Runtime.InteropServices;
+using Google.Protobuf;
+using Microsoft.Azure.Functions.Worker.Grpc.Messages;
+
+namespace Microsoft.Azure.Functions.Worker.Grpc.NativeHostIntegration
+{
+    internal static unsafe partial class NativeMethods
+    {
+        internal const string NativeWorkerDll = "FunctionsNetHost.exe";
```
kshyju, line 13: Should we change this to private?

src/DotNetWorker.Core/DotNetWorker.Core.csproj
```diff
@@ -19,16 +19,22 @@
...
+      <PackageReference Include="System.Collections.Immutable" Version="6.0.0" />
+   </ItemGroup>
+    
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0' OR '$(TargetFramework)' == 'net5.0'">
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="5.0.0" />
     <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
     <PackageReference Include="Microsoft.Extensions.Options" Version="5.0.0" />
+      <PackageReference Include="Microsoft.Extensions.Hosting" Version="5.0.0" />
+      <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="5.0.0" />
```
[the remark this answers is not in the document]
bobbyangers: but why not use version `7.0.0`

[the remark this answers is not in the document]
jviau: As an SDK we have several restrictions around what packages we can and cannot use, and also their versions. We have many competing customer requirements we need to balance, back compat / breaking change requirements, and general guidelines to follow.

1. This is multi-targeting, so we are matching the .NET version with the packages released around that time. When you target net6.0, you should get the 6.0.0 dependencies. We will probably only bring in 7.0.0 packages if we have a net7.0 target. This is more of a pattern core .NET packages and frameworks use.
2. Reving major versions of dependencies can be considered a breaking change and would require us to also rev our matching version to follow SemVer semantics.
3. As a library, it is actually beneficial for us to target the lowest dependency versions possible (provided it has the features we want). Customers can pin to a higher version themselves via their own package reference to the desired higher version. This lets us cast as wide of a compatibility net as possible. We do not want to _force_ an upgrade, but we also don't want to _block_ an upgrade. We want to give as much control over package versions to the customer as possible.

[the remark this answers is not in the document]
bobbyangers: @jviau 
Thank you for your explanation; I certainly grasp the intent and your concerns.

This said, providing a base packages, namely `DotNetWorker.Core`, `DotnetWorker`, `DotNetWorker.Grpc` can safely target `netstandard2.0`, `net6.0`, `net7.0` and be adjusted with some conditionals in the project to reference the proper libraries.

This can be also implemented in the tests projects that perform the same tests and verify the specifics versions equally, which (in my opinion) be much better than relying only on `netstandard2.0' and 'net5.0' that it will work with the officially supported TFM.

In this example however, `Microsoft.Extensions.Hosting` version **7.0.0** is fully backward compatible to `netstandard2.0`, `net6.0`, `net7.0` ; since itself support those frameworks, please refer to [Microsoft.Extensions.Hosting](https://www.nuget.org/packages/Microsoft.Extensions.Hosting).    So if the `.csproj` refer to this version and has multi-targeting activated and tests are successful, it is reasonable to think it will work with your users.

Please note that this is also and **SDK** that are concerns with the same as stated earlier.

This also applies to the SDK portion, generator.

samples/FunctionApp/NuGet.Config
```diff
@@ -3,5 +3,6 @@
   <packageSources>
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
     <add key="azure_app_service" value="https://www.myget.org/F/azure-appservice/api/v2" />
+    <add key="local" value="D:\src\azure-functions-dotnet-worker/local" />
```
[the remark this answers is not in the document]
fabiocav: Yes. Will clean this all up before marking as ready.

src/DotNetWorker.Grpc/GrpcWorker.cs
fabiocav: This was pre-existing code, but yes, we should be able to use a discard here, so I'll update.

I think the suggested interface change makes sense, but let me follow up with a different PR so I'm not increasing the scope of the change here (as that is also pre-existing).

src/DotNetWorker.Grpc/NativeHostIntegration/NativeWorkerClient.cs
fabiocav: Tempted to stick with the original here as that matches the style in this codebase :)

src/DotNetWorker.Grpc/NativeHostIntegration/NativeMethods.cs
fabiocav: 👍🏼

src/DotNetWorker.Grpc/IGrpcRequestHandler.cs
fabiocav: I think it makes sense for these interfaces to take in a `CancellationToken`. Will update

src/DotNetWorker.Core/DotNetWorker.Core.csproj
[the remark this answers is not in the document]
fabiocav: @bobbyangers those dependencies will be revised and updated in a new major version, but can you share more information about whether that is blocking any of your scenarios? 

Multi-targeting with conditional dependencies is already the approach followed by SDK, but this also has to be done with care to ensure there will be no experiences where we force a major dependency bump within a given major version of our packages.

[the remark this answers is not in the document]
bobbyangers: @fabiocav, of course, I understand.

I have proposed a separate PR (#1242) to handle multiple - simultaneous versions that would avoid address this concern.

src/DotNetWorker.Grpc/Definition/GrpcFunctionDefinition.cs
```diff
@@ -19,9 +19,21 @@ public GrpcFunctionDefinition(FunctionLoadRequest loadRequest, IMethodInfoLocato
         {
             EntryPoint = loadRequest.Metadata.EntryPoint;
             Name = loadRequest.Metadata.Name;
-            PathToAssembly = Path.GetFullPath(loadRequest.Metadata.ScriptFile);
             Id = loadRequest.FunctionId;
 
+            string? scriptRoot = Environment.GetEnvironmentVariable("AzureWebJobsScriptRoot");
+            if (string.IsNullOrWhiteSpace(scriptRoot))
```
kshyju, line 25: These checks and throw were added to figure out why the process was crashing, while we were building the first working version. Should we keep this?

src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
```diff
@@ -9,7 +9,8 @@
     <RootNamespace>Microsoft.Azure.Functions.Worker.Grpc</RootNamespace>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <MinorProductVersion>7</MinorProductVersion>
```
kshyju, line 11: Should we bring these changes in a minor version update?

src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs
```diff
@@ -56,40 +49,21 @@ public static IServiceCollection AddGrpc(this IServiceCollection services)
...
-
-                var options = new ChannelOption[]
-                {
-                    new ChannelOption(Grpc.Core.ChannelOptions.MaxReceiveMessageLength, arguments.GrpcMaxMessageLength),
-                    new ChannelOption(Grpc.Core.ChannelOptions.MaxSendMessageLength, arguments.GrpcMaxMessageLength)
-                };
-
-                Grpc.Core.Channel grpcChannel = new Grpc.Core.Channel(arguments.Host, arguments.Port, ChannelCredentials.Insecure, options);
-
+            services.AddSingleton<IWorkerClientFactory, GrpcWorkerClientFactory>();
```
kshyju, line 65: Will NETFX case falls into this else case? We want the nativehost in NetFX as well. correct?

src/DotNetWorker.Grpc/NativeHostIntegration/NativeWorkerClientFactory.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        private readonly GrpcHostChannel _hostChannel;
+
+        public NativeWorkerClientFactory(GrpcHostChannel hostChannel)
+        {
+            _hostChannel = hostChannel;
+        }
+
+        public Task<IWorkerClient> StartClientAsync(IMessageProcessor messageProcessor, CancellationToken token)
+        {
```
kshyju, line 20: Should we `token.ThrowIfCancellationRequested();`?

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -78,54 +67,16 @@ internal class GrpcWorker : IWorker
...
-                await ProcessRequestAsync(responseStream.Current);
-            }
-        }
-
-        private Task ProcessRequestAsync(StreamingMessage request)
+        Task IMessageProcessor.ProcessMessageAsync(StreamingMessage message)
         {
             // Dispatch and return.
-            Task.Run(() => ProcessRequestCoreAsync(request));
+            _ = ProcessRequestCoreAsync(message);
```
kshyju, line 78: We should also update env reload case (line 116) to return capabilities and worker metadata. Can be a follow up PR I can take care of.

kshyju (approved):

src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs
fabiocav: No native host for NetFX

src/DotNetWorker.Grpc/GrpcWorker.cs
fabiocav: Has the host been updated to support that? yes, we should, but either way, I'm inclined to say this should be a separate change

src/DotNetWorker.Grpc/Definition/GrpcFunctionDefinition.cs
fabiocav: Discussed offline. Keeping checks and updating verbiage.

PR #1221, opened 2022-12-08
Moving proto files to repo root

kshyju:
Moving proto files to the repo root so that both DotNetWorker.Grpc project and FunctionsNetHost can use this single source of truth. Currently we have 2 copies.

1 . https://github.com/Azure/azure-functions-dotnet-worker/tree/main/src/DotNetWorker.Grpc/protobuf/src/proto
2. https://github.com/Azure/azure-functions-dotnet-worker/tree/main/host/src/Protos


brettsam (approved):

host/src/Protos/CMakeLists.txt
```diff
@@ -2,6 +2,17 @@ find_package(protobuf CONFIG REQUIRED)
 find_package(gRPC CONFIG REQUIRED)
 find_package(Threads)
 
+# We cannot directly refer the proto files in the repo root as that dir is not a CMAKE Source dir.
```
satvu, line 5: great clarifying comments in this file

satvu (approved):

liliankasem (changes requested):
We have a process around how we import the protobuf updates. The protobuf in the Grpc project is a clone of the protobuf repo. If we're moving it to protos folder at the root, we need to make sure we're following the same pattern

kshyju:
> We have a process around how we import the protobuf updates from the proto repo into the host and workers. The protobuf files in the Grpc folder are an exact copy of the protobuf repo. If we're moving it to protos folder at the root, we need to make sure we're following the same pattern and folder structure
> 
> Here's the protobuf repo: https://github.com/Azure/azure-functions-language-worker-protobuf
> 
> Protobuf release process: https://dev.azure.com/msazure/One/_git/AAPT-Antares-Docs?path=/TeamDocs/FunctionTeamDocs/EngineeringProcesses/ReleaseProcesses/ProtobufReleaseProcess.md&version=GBmaster&line=1

Kept the same folder structure as the protobuf repo.

liliankasem (approved):

PR #1232, opened 2022-12-14
Kafka Extension upgrade from 3.6.0 to 3.7.0 & Fixing the null attribute issue

shrohilla:

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

Additional PR information


raorugan:
@kshyju  - can you please approve this PR as it is pending for out-proc worker extension nuget package release?

extensions/Worker.Extensions.Kafka/src/KafkaOutputAttribute.cs
```diff
@@ -32,38 +32,38 @@ public KafkaOutputAttribute(string brokerList, string topic)
         /// Gets or sets the Avro schema.
         /// Should be used only if a generic record should be generated
         /// </summary>
-        public string? AvroSchema { get; set; }
+        public string AvroSchema { get; set; }
```
kshyju, line 35: Why changing from nullable to non nullable? Does tooling/static analysis shows a warning about this (non nullable)prop not being initialized? I think it is best to embrace nullable type feature the language is offering.

Will customers start seeing warnings about props not being initialized after they upgrade from 3.6.0 to 3.70?

extensions/Worker.Extensions.Kafka/src/KafkaOutputAttribute.cs
```diff
@@ -32,38 +32,38 @@ public KafkaOutputAttribute(string brokerList, string topic)
...
 
         /// <summary>
         /// Gets or sets the Maximum transmit message size. Default: 1MB
```
kshyju, line 38: nit: 
Use lowercase maximum. 
Consider using consistent casing across the file (ex: default or Default?).
Consider rephrasing to include the unit as well (like "in bytes").

extensions/Worker.Extensions.Kafka/src/KafkaOutputAttribute.cs
```diff
@@ -32,38 +32,38 @@ public KafkaOutputAttribute(string brokerList, string topic)
...
         /// </summary>
-        public int? RequestTimeoutMs { get; set; }
+        public int RequestTimeoutMs { get; set; } = 5_000;
 
         /// <summary>
         /// How many times to retry sending a failing Message. **Note:** default: 2 
         /// </summary>
         /// <remarks>Retrying may cause reordering unless <c>EnableIdempotence</c> is set to <c>true</c>.</remarks>
-        public int? MaxRetries { get; set; }
+        public int MaxRetries { get; set; } = int.MaxValue;
```
kshyju, line 66: 2 billion max retries? 😵

extensions/Worker.Extensions.Kafka/src/KafkaTriggerAttribute.cs
```diff
@@ -77,25 +77,25 @@ public KafkaTriggerAttribute(string brokerList, string topic)
...
         /// <summary>
         /// Path to client's certificate.
         /// ssl.certificate.location in librdkafka
         /// </summary>
-        public string? SslCertificateLocation { get; set; }
+        public string SslCertificateLocation { get; set; }
 
         /// <summary>
         /// Password for client's certificate.
         /// ssl.key.password in librdkafka
```
kshyju, line 96: Q: Is this a valid comment to present here?

kshyju (commented):
Noticed some public methods/types are missing comments/summary. Could we add that as well? ex: https://github.com/Azure/azure-functions-dotnet-worker/blob/main/extensions/Worker.Extensions.Kafka/src/KafkaTriggerAttribute.cs#L15

extensions/Worker.Extensions.Kafka/src/KafkaOutputAttribute.cs
```diff
@@ -32,38 +32,38 @@ public KafkaOutputAttribute(string brokerList, string topic)
...
         /// When set to `true`, the producer will ensure that messages are successfully produced exactly once and in the original produce order. default: false
         /// </summary>
-        public bool? EnableIdempotence { get; set; }
+        public bool EnableIdempotence { get; set; } = false;
 
         /// <summary>
         /// Local message timeout. This value is only enforced locally and limits the time a produced message waits for successful delivery. A time of 0 is infinite. This is the maximum time used to deliver a message (including retries). Delivery error occurs when either the retry count or the message timeout are exceeded. default: 300000
         /// </summary>
-        public int? MessageTimeoutMs { get; set; }
+        public int MessageTimeoutMs { get; set; } = 300_000;
```
kshyju, line 55: Question: Would this change be considered as a breaking change? Is the code which consumes this property (may be in host/kafka webjobs extn) updated to handle the default value change? Earlier this was sent as null and now it will send 300000 when sending the serialized version of this class instance.

kshyju:
@raorugan @shrohilla Ping on this PR.

fabiocav:
@shrohilla / @raorugan can you take a look at the comments above? Thanks!

extensions/Worker.Extensions.Kafka/src/KafkaOutputAttribute.cs
jainharsh98: The reason for the change is the compilation error with use of nullable attributes as discussed [here](https://github.com/Azure/azure-functions-kafka-extension/pull/362). 
Agreed that we should embrace the nullable types - something we will definitely look into going forward. But for now - In order to maintain parity with the kafka extension we are keeping the attributes non-nullable.

extensions/Worker.Extensions.Kafka/src/KafkaOutputAttribute.cs
jainharsh98: This is in line with the extension and the underlying Confluent Kafka sdk. Even when set to null this is the default value. Explicitly setting as default will help customers.

extensions/Worker.Extensions.Kafka/src/KafkaTriggerAttribute.cs
jainharsh98: Comment in line with the extension.  https://github.com/Azure/azure-functions-kafka-extension/blob/82de33298da77b7e5066db8d6b902370641ae765/src/Microsoft.Azure.WebJobs.Extensions.Kafka/Output/KafkaAttribute.cs#L132

extensions/Worker.Extensions.Kafka/src/KafkaOutputAttribute.cs
jainharsh98: No. This would not be a breaking change. When set to null the value used was 300000 in the underlying librdkafka library.

jainharsh98:
@kshyju The Kafka Extension pipeline seems to be waiting for some permission. Can you please help with that?

extensions/Worker.Extensions.Kafka/src/KafkaOutputAttribute.cs
jainharsh98: Done.

satvu (approved):

kshyju:
> @kshyju The Kafka Extension pipeline seems to be waiting for some permission. Can you please help with that?

All good now. https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=113680&view=results

jainharsh98:
Closing in favor of #1336

PR #1242, opened 2022-12-19
proposal multi-target

bobbyangers:
Based on your recent changes regarding test projects to `net7.0`;
As stated in recent exchanges, I propose to enable multi-target in the test projects, but also in the main projects.

Please review the proposed changes and comment.

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/build/Common.props
+++ b/build/Common.props
@@ -20,7 +20,7 @@
     <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)/../key.snk</AssemblyOriginatorKeyFile>
     <DebugType>embedded</DebugType>
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
-    <ContinuousIntegrationBuild Condition="'$(TF_BUILD)' == 'true'">true"</ContinuousIntegrationBuild>
+    <ContinuousIntegrationBuild Condition="'$(TF_BUILD)' == 'true'">true</ContinuousIntegrationBuild>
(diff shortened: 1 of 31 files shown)


jviau:
@bobbyangers can you explain what the end customer benefit of this PR is? What scenarios does this unblock? What issues does it address?

bobbyangers:
> @bobbyangers can you explain what the end customer benefit of this PR is? What scenarios does this unblock? What issues does it address?

In a previous exchanged you mentionned the concern about supporting a wide range of customer, and reducing the risk to excisting consumer of the package.  In this proposal, I am summitting a proposal to expand the sdk to natively support `netstandard2.0`, `net6` and `net7.0`.    Using the method already in place (ie multi-targeting), and expanding this to the test projects, and validating the versions against the same tests.     This in effect, removes the guess work (and assumptions) and validates the core modules and SDK actually works with all supported versions (not simply assume that everything is backward compatible from 'net7.0`).

The problematic I am trying to solve is that when writing a `net6.0` function, I want the `core module` and `sdk` to use `net6.0`,  when using `net7.0`, then I want to use `net7.0` at all levels as well.

For example, when the sdk and core module are compile against `net5.0` and that I get warning from the compiler and from my static code analyzer that my modules rely on 'unsupported` or even 'unsafe` modules,  policy required that I use more recent modules and framework.

PR #1248, opened 2022-12-20
WorkerMetadata: Update runtime name to exclude the version

### LINKED ISSUE #1098: Update worker metadata to remove dotnet version from "RuntimeName"

liliankasem:

resolves #1098

### Pull request checklist

### Additional information

Regex extracts framework name as expected:

Screenshot 2022-12-21 at 11 44 29 AM

Tested E2E by deploying to the portal

Screenshot 2022-12-21 at 1 18 24 PM

Screenshot 2022-12-21 at 2 11 40 PM
Screenshot 2022-12-21 at 2 12 21 PM


jviau:
Can you explain why we want to stop the version from this string?

liliankasem:
> Can you explain why we want to stop the version from this string?

Two reasons:

a) we want to have a standard around what the worker metadata fields are like and runtime name should just be the name; and
b) since we already have a version field, we don't want to also have that same information in the name 

Currently we will see this in our telemetry:

```
"RuntimeName": .NET 6.0.9, // should be just .NET 
"RuntimeVersion": 6.0.9,
"WorkerVersion": 1.8.0.0,
"WorkerBitness": X64,
"GrpcVersion": 1.6.0.0
```

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -44,6 +45,8 @@ internal class GrpcWorker : IWorker
         private readonly IInvocationHandler _invocationHandler;
         private readonly IFunctionMetadataProvider _functionMetadataProvider;
 
+        internal static readonly Regex _frameworkDescriptionRegex = new Regex(@"^(\D*)");
```
kshyju, line 48: Should this be static? This Regex is only used for worker init response message handling use case. No reason to keep this in the memory. If we remove static and move closer to usage/method scope, GC can reclaim that memory.

liliankasem: Doesn't have to be; was just trying to follow the practice of having this resolved early as suggested by Fabio every time I use regex xD I would make it not static and pass the variable to the `WorkerInitRequestHandler` method as a param instead? Or are you saying we should create this regex inside of `WorkerInitRequestHandler` - any perf concerns here?

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -200,7 +203,7 @@ internal static WorkerInitResponse WorkerInitRequestHandler(WorkerInitRequest re
                 WorkerVersion = WorkerInformation.Instance.WorkerVersion,
                 WorkerMetadata = new WorkerMetadata
                 {
-                    RuntimeName = RuntimeInformation.FrameworkDescription,
+                    RuntimeName = _frameworkDescriptionRegex.Match(RuntimeInformation.FrameworkDescription).Value.Trim() ?? RuntimeInformation.FrameworkDescription,
```
kshyju, line 206: Wonder the regex can be updated to handle the space split so you can remove the Trim call here.

src/DotNetWorker.Grpc/GrpcWorker.cs
kshyju: I think it might be best to keep it close to usage, within the method context. since this class is singleton, the private variable may stay alive I think.

kshyju: Do you know compiled regex may provide better perf? This gets used only once. But might be good to measure both compiled regex and non compiled.

src/DotNetWorker.Grpc/GrpcWorker.cs
liliankasem: Ah good point, can definitely do that!

src/DotNetWorker.Grpc/GrpcWorker.cs
liliankasem: I'm sure that complied regex does provide better perf (Fabio and Nick have called this out in a couple of my PRs)

liliankasem: Since `WorkerInitRequestHandler ` in theory only gets called once in the lifetime of a worker(?), I can move it inside the method scope as suggested. I think ideally it we do the precompiled regex but it might not be worth it xD

jviau (approved):

kshyju (approved):

PR #1275, opened 2023-01-17
Implement SupportsDeferredBindingAttribute

### LINKED ISSUE #1266: Refactor how we determine which extensions support deferred binding
Today we have have a single `SupportsDeferredBinding` attribute that extensions can set to indicate that `ParameterBindingData` can be used (i.e. the extension supports sdk-type bindings). This is used by the FunctionMetadataGenerator to set the "supportsDeferringBinding" property for every function.

The goal for this issue is to refactor the SupportsDeferredBinding attribute so that instead of being a blanket "all or nothing" approach for extensions, we can set the supportsDeferredBinding flag per binding instead. For example, the CosmosDBTrigger does not support deferred binding but we can use deferred binding for the input binding.

- [ ] Remove the SupportsDeferredBinding property from the [ExtensionInformationAttribute](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/extensions/Worker.Extensions.Abstractions/src/ExtensionInformationAttribute.cs)
- [ ] Implement an attribute that binding attributes can use to set `SupportsDeferredBinding` to true or false
- [ ] Update the [FunctionMetadataGenerator](https://github.com/Azure/azure-functions-dotnet-worker/blob/ae0cbe0f08b2ee7418fc6ded8d020b5b4b9bdfb2/sdk/Sdk/FunctionMetadataGenerator.cs#L492) to remove the code that gets the supportsDeferredBinding flag from `ExtensionInformationAttribute`, and to get that information from the CustomAttribute instead

liliankasem:

resolves #1266

Today we have have a single SupportsDeferredBinding attribute that extensions can set to indicate that ParameterBindingData can be used (i.e. the extension supports sdk-type bindings). This is used by the FunctionMetadataGenerator to set the "supportsDeferringBinding" property for every function.

The goal for this issue is to refactor the SupportsDeferredBinding attribute so that instead of being a blanket "all or nothing" approach for extensions, we can set the supportsDeferredBinding flag per binding instead. For example, the CosmosDBTrigger does not support deferred binding but we can use deferred binding for the input binding.

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/extensions/Worker.Extensions.Abstractions/src/ExtensionInformationAttribute.cs
+++ b/extensions/Worker.Extensions.Abstractions/src/ExtensionInformationAttribute.cs
@@ -14,24 +14,16 @@ namespace Microsoft.Azure.Functions.Worker.Extensions.Abstractions
 
         public bool EnableImplicitRegistration { get; }
 
-        public bool SupportsDeferredBinding { get; }
-
         public ExtensionInformationAttribute(string extensionPackage, string extensionVersion)
-            : this(extensionPackage, extensionVersion, false, false)
+            : this(extensionPackage, extensionVersion, false)
         {
         }
 
         public ExtensionInformationAttribute(string extensionPackage, string extensionVersion, bool enableImplicitRegistration)
-            : this(extensionPackage, extensionVersion, enableImplicitRegistration, false)
-        {
-        }
-
-        public ExtensionInformationAttribute(string extensionPackage, string extensionVersion, bool enableImplicitRegistration, bool supportsDeferredBinding)
         {
             ExtensionPackage = extensionPackage;
             ExtensionVersion = extensionVersion;
             EnableImplicitRegistration = enableImplicitRegistration;
-            SupportsDeferredBinding = supportsDeferredBinding;
         }
     }
 }
--- a/sdk/Sdk/FunctionMetadataGenerator.cs
+++ b/sdk/Sdk/FunctionMetadataGenerator.cs
@@ -574,12 +574,6 @@ namespace Microsoft.Azure.Functions.Worker.Sdk
             return binding;
         }
 
-        private static bool SupportsDeferredBinding(CustomAttribute attribute)
-        {
-            var typeDefinition = attribute.AttributeType.Resolve();
-            return typeDefinition.CustomAttributes.Any(a => a.AttributeType.FullName == Constants.SupportsDeferredBindingAttributeType);
-        }
-
         private static bool IsIterableCollection(TypeReference type, out DataType dataType)
         {
             // Array and not byte array
@@ -799,6 +793,19 @@ namespace Microsoft.Azure.Functions.Worker.Sdk
             }
 
             return Constants.InputBindingDirection;
+        }
+
+        private static bool SupportsDeferredBinding(CustomAttribute attribute)
+        {
+            var typeDefinition = attribute?.AttributeType?.Resolve();
+
+            if (typeDefinition is null)
+            {
+                return false;
+            }
+
+            return typeDefinition.CustomAttributes
+                                 .Any(a => string.Equals(a.AttributeType.FullName, Constants.SupportsDeferredBindingAttributeType, StringComparison.Ordinal));
         }
 
         private static bool IsOutputBindingType(CustomAttribute attribute)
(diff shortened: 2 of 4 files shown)


extensions/Worker.Extensions.Abstractions/src/ExtensionInformationAttribute.cs
```diff
@@ -14,24 +14,16 @@ public class ExtensionInformationAttribute : Attribute
...
+            : this(extensionPackage, extensionVersion, false)
         {
         }
 
         public ExtensionInformationAttribute(string extensionPackage, string extensionVersion, bool enableImplicitRegistration)
-            : this(extensionPackage, extensionVersion, enableImplicitRegistration, false)
-        {
-        }
-
-        public ExtensionInformationAttribute(string extensionPackage, string extensionVersion, bool enableImplicitRegistration, bool supportsDeferredBinding)
```
jviau, line 29: I think removing this constructor would be a breaking change.

jviau (commented):
@liliankasem I am concerned this is both a public API breaking change as well as behavior breaking change. Do you have more context on if this is okay to perform now or will it need to wait until next version?

Or can we support _both_ ways for now, and remove the legacy way later?

extensions/Worker.Extensions.Abstractions/src/ExtensionInformationAttribute.cs
liliankasem: We should be okay; we added this constructor and nothing is using it yet other than the draft SDK-Binding PRs. Plan is to switch to using the new attribute instead of this constructor before we release anything and cause breaking changes later

jviau: Oh, so this has not been released yet?

liliankasem: Here's the last [PR](#1107) for reference, I don't believe we ever published this and if we ever did it was preview

jviau: If it has not been released (or only released as preview), then I have no worries about the change.

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -489,7 +487,7 @@ private static ExpandoObject BuildBindingMetadataFromAttribute(CustomAttribute a
 
             // For extensions that support deferred binding, set the supportsDeferredBinding property so parameters are bound by the worker
             // Only use deferred binding for input and trigger bindings, output is out not currently supported
-            if (supportsDeferredBinding && direction != Constants.OutputBindingDirection)
+            if (SupportsDeferredBinding(attribute) && direction != Constants.OutputBindingDirection)
```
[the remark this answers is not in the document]
liliankasem: I haven't written any analyzers before so I'm not sure, but maybe? @kashimiz might know better :D 

I don't know yet that not having support for output bindings will be permanent, it's not something we're targeting for preview but we might need them in place for GA

surgupta-msft:
Can we add some tests for these changes? One place could be - [FunctionMetadataGeneratorTests](https://github.com/Azure/azure-functions-dotnet-worker/blob/a8f64c27234b0e0c4048d5946ff6a8b7cc65d339/test/FunctionMetadataGeneratorTests/FunctionMetadataGeneratorTests.cs)

surgupta-msft (approved):

liliankasem:
> Can we add some tests for these changes? One place could be - [FunctionMetadataGeneratorTests](https://github.com/Azure/azure-functions-dotnet-worker/blob/a8f64c27234b0e0c4048d5946ff6a8b7cc65d339/test/FunctionMetadataGeneratorTests/FunctionMetadataGeneratorTests.cs) and a sample app/function with the attribute `SupportsDeferredBinding`

I'll see if I can add a test for it! But I don't believe we need a sample app for it, this attribute goes onto the worker extension input/output/trigger attribute classes not on customer function classes e.g.:

```csharp
    [SupportsDeferredBinding]
    public sealed class CosmosDBInputAttribute : InputBindingAttribute
    {
```

sdk/Sdk/FunctionMetadataGenerator.cs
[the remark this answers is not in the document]
liliankasem: Done! https://github.com/Azure/azure-functions-dotnet-worker/issues/1277

fabiocav (approved):

PR #1280, opened 2023-01-19
Bumping up DotNetIsolatedNativeHost version to preview5

kshyju:
Bumping up DotNetIsolatedNativeHost version to preview5


satvu (approved):
why the big jump?

kshyju:
> why the big jump?

We manually created some artifacts (up to preview 4) before holidays and used them.

PR #1286, opened 2023-01-23
Hosted service for logging a function's resolved binding types

### LINKED ISSUE #1257: Adding logging to indicate what the ParameterBindingData resolved type is
We need a way to determine what we resolving the ParameterBindingData type to on the worker side. For example, if a user binds to a `BlobClient` are we correcting hydrating and returning a `BlobClient` to the user when we use the ParameterBindingData reference type. This aim of this is to help us with potential CRIs and telemetry to 

a) help troubleshoot if things go wrong, i.e. if a customer is binding to a BlobClient and we provide them with a stream
b) have telemetry to better understand adoption and what customers are most commonly binding to

- [ ] Add a GrpcLog statement to indicate the type resolved by the converter

liliankasem:
Would love some early feedback on:

1. Is this the right approach for logging something once?
2. What is a good timeframe for sending this log out to avoid cold start impact?
3. Should I be using WorkerDiagnostics or something else?
4. Any ideas on how I can get the resolved binding information here so I can log it out properly?

resolves #1257

We need a way to determine what we are resolving the ParameterBindingData type to on the worker side. For example, if a user binds to a BlobClient are we correctly hydrating and returning a BlobClient to the user when we use the ParameterBindingData reference type. This aim of this is to give us telemetry to better understand the SDK-binding feature adoption and what customers are most commonly binding to.

- Data: what customers are binding to in their Functions i.e. BlobClient, string, etc.
- Interval: once, we only need this information once per function app
- Usage: telemetry on feature adoption and most commonly used bindings
    - This will help us figure out what to prioritize in terms of building out more extension support using SDK-type bindings
    - Currently we can't know if the work is actually being leveraged for SDK types, people can still be using strings even with the deferred binding feature which doesn't indicate value of the Azure SDKs themselves.
    - Potentially helpful for debugging if customers bind to BlobClient and we provide them a string or other type
    -  (I can get more information from our PMs about what else they want to use this for)

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/DotNetWorker.Core/Logging/BindingLoggerHostedService.cs
+++ b/src/DotNetWorker.Core/Logging/BindingLoggerHostedService.cs
@@ -1,0 +1,92 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+
+namespace  Microsoft.Azure.Functions.Worker.Logging
+{
+    /// <summary>
+    /// Service responsible for logging the binding type that has been resolved for each function.
+    /// This log only occurs once 10 seconds the worker has been initialized to avoid cold start
+    /// impact.
+    /// </summary>
+    public class BindingLoggerHostedService : IHostedService, IDisposable
+    {
+        private readonly ILogger _logger;
+        private readonly Timer _timer;
+        private readonly int _due = 10000; // 10 seconds
+        private bool _disposed;
+
+        /// <summary>
+        /// TODO: should we be using IWorkerDiagnostics?
+        /// </summary>
+        public BindingLoggerHostedService(ILogger<BindingLoggerHostedService> logger)
+        {
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _timer = new Timer(PublishLogs);
+        }
+
+        /// <inheritdoc/>
+        public Task StartAsync(CancellationToken cancellationToken)
+        {
+            if (_disposed || _timer is null)
+            {
+                return Task.FromException(new InvalidOperationException("Service has already been disposed"));
+            }
+
+            try
+            {
+                // start the timer by setting the due time
+                _timer.Change(_due, Timeout.Infinite);
+            }
+            catch (Exception exc)
+            {
+                _logger.LogWarning(exc, "Unable to set timer interval");
+            }
+
+            return Task.CompletedTask;
+        }
+
+        /// <inheritdoc/>
+        public Task StopAsync(CancellationToken cancellationToken)
+        {
+            // stop the timer if it has been started
+            _timer?.Change(Timeout.Infinite, Timeout.Infinite);
+            return Task.CompletedTask;
+        }
+
+        /// <summary>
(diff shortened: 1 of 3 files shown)


src/DotNetWorker.Core/Logging/BindingLoggerHostedService.cs
```diff
@@ -0,0 +1,92 @@
...
+        private readonly Timer _timer;
+        private readonly int _due = 10000; // 10 seconds
+        private bool _disposed;
+
+        /// <summary>
+        /// TODO: should we be using IWorkerDiagnostics?
+        /// </summary>
+        public BindingLoggerHostedService(ILogger<BindingLoggerHostedService> logger)
+        {
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
```
liliankasem, line 29: @fabiocav wondering if I should be using `WorkerDiagnostics` here instead of logger or if there is something else I should be using

src/DotNetWorker.Core/Logging/BindingLoggerHostedService.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        /// <summary>
+        /// Log the binding type that has been resolved for each function
+        /// </summary>
+        private void PublishLogs(object? state)
+        {
+            try
+            {
+                // TODO: figure out how to get the resolved binding type
+                _logger.LogInformation("Function 'blah' resolved binding to 'blah'");
```
liliankasem, line 70: Will likely have to log this for every binding, for every function registered

jviau:
Not sure if this is based on my logging comment in the one PR. If so, my comment was more around putting some debug or trace messages in the converters themselves to help customers self-diagnose any issues and observe behavior. This PR description seems to be around collecting usage data for this feature for our own consumption - do we do that often? It would be helpful if you break down exactly what you want to collect - the data itself, the interval, how it is supposed to be used, etc. I want to make sure we only collect data like this if we have a good use for it.

liliankasem:
> Not sure if this is based on my logging comment in the one PR. If so, my comment was more around putting some debug or trace messages in the converters themselves to help customers self-diagnose any issues and observe behavior. This PR description seems to be around collecting usage data for this feature for our own consumption - do we do that often? It would be helpful if you break down exactly what you want to collect - the data itself, the interval, how it is supposed to be used, etc. I want to make sure we only collect data like this if we have a good use for it.

This isn't related to your comments in another PR, this is a different user story all together about collecting telemetry on what customers are binding to; I'll update the PR description with the data you asked for

surgupta-msft:
Curious, did we test this approach with a function app, if it is actually emitting the binding logs we are interested in?

liliankasem:
> Curious, did we test this approach with a function app, if it is emitting the binding logs we are interested in?
> 
> Also, will be helpful if you can share more details/design about the need to use Hosted Service here?

I haven't tested anything out yet, just laying the foundation to see if this is the right approach to follow - hoping for @fabiocav's feedback before I keep going.

As for hosted services, this is the pattern we use in the host and that Fabio suggested we use here too - I'm not aware of any existing documentation for this pattern from our team. Here are the public docs on using hosted services for background tasks: https://learn.microsoft.com/en-us/aspnet/core/fundamentals/host/hosted-services?view=aspnetcore-7.0&tabs=visual-studio

liliankasem:
Working on design for a more generic approach to this problem #1311

PR #1307, opened 2023-02-01
adding FunctionActivitySource to Worker; refactoring AppInsights support

brettsam:
This change:
- moves the `FunctionActivitySource` into the core worker project and provides OTel versioning capabillities for the future. 
- updates the handling of this Activity in the App Insights package
- makes the App Insights handling richer (now sends Exceptions, sets Dependency status, for example)
- removes a bunch of other custom App Insights code we had to simplify some of our behavior
-  sets `"WorkerApplicationInsightsLoggingEnabled"` capability for when functions host can handle this (soon)


src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
```diff
@@ -20,18 +24,36 @@ public void Initialize(TelemetryConfiguration configuration)
...
+                    foreach (ActivityEvent activityEvent in activity.Events)
+                    {
+                        TrackExceptionTelemetryFromActivityEvent(activityEvent, _telemetryClient);
+                    }
+
+                    if (activity.GetCustomProperty(DependencyTelemetryKey) is IOperationHolder<DependencyTelemetry> dependencyHolder)
+                    {
+                        var dependency = dependencyHolder.Telemetry;
+
+                        foreach (var item in activity.Tags)
```
lmolkova, line 45: nit: I think it's not needed and done within `StartOperation` https://github.com/microsoft/ApplicationInsights-dotnet/blob/main/BASE/src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs#L393

src/DotNetWorker.Core/Diagnostics/FunctionActivitySource.cs
```diff
@@ -0,0 +1,59 @@
...
+            _attributeMap = new Lazy<IReadOnlyDictionary<string, string>>(() => GetMapping(options.Value.OpenTelemetrySchemaVersion));
+            _schemaVersionUrl = TraceConstants.OpenTelemetrySchemaMap[options.Value.OpenTelemetrySchemaVersion];
+        }
+
+        public Activity? StartInvoke(FunctionContext context)
+        {
+            var activity = _activitySource.StartActivity(TraceConstants.FunctionsInvokeActivityName, ActivityKind.Internal, context.TraceContext.TraceParent,
+                tags: GetTags(context));
+
+            return activity;
```
lmolkova, line 29: would be great to add `activity.Tracestate = context.TraceContext.Tracestate`

src/DotNetWorker.Core/Diagnostics/TraceConstants.cs
```diff
@@ -0,0 +1,30 @@
...
+        public const string FunctionsActivitySource = "Microsoft.Azure.Functions.Worker";
+        public const string FunctionsActivitySourceVersion = "1.0.0.0";
+        public const string FunctionsInvokeActivityName = "Invoke";
+
+        public const string AttributeExceptionEventName = "exception";
+        public const string AttributeExceptionType = "exception.type";
+        public const string AttributeExceptionMessage = "exception.message";
+        public const string AttributeExceptionStacktrace = "exception.stacktrace";
+
+        public const string AttributeSchemaUrl = "schema_url";
```
lmolkova, line 19: nit: we're using `az.schema_url` in Azure SDKs and agreed on it  with azmon folks https://github.com/microsoft/common-schema/blob/main/Mappings/AzureSDKs.md#versioning

Can we align on the name here? I'm open to renaming and generalizing it in MS common-schema beyond azure SDKs

src/DotNetWorker.Core/FunctionsApplication.cs
```diff
@@ -61,12 +65,24 @@ public void LoadFunction(FunctionDefinition definition)
...
-            using (_logger.BeginScope(scope))
+
+            using var logScope = _logger.BeginScope(scope);
+            using Activity? invokeActivity = _functionActivitySourceFactory.StartInvoke(context);
+
+            try
             {
-                return _functionExecutionDelegate(context);
+                await _functionExecutionDelegate(context);
+                invokeActivity?.SetStatus(ActivityStatusCode.Ok);
```
lmolkova, line 78: faas spec does mention status, but it's common in other conventions to leave status unset. This way if user sets error/ok status on current activity that finished without exception, we won't override it here

src/DotNetWorker.Core/FunctionsApplication.cs
```diff
@@ -61,12 +65,24 @@ public void LoadFunction(FunctionDefinition definition)
...
+            }
+            catch (Exception ex)
+            {
+                invokeActivity?.SetStatus(ActivityStatusCode.Error, ex.Message);
+                invokeActivity?.RecordException(ex);
```
RohitRanjanMS, line 83: Do you think we should also set exception.escapd? https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/exceptions/#:~:text=Requirement%20Level-,exception.escaped,-boolean

lmolkova (approved):
Left a few minor(-ish) comments, looks great overall. Super-excited to see it coming! 🚀

src/DotNetWorker.Core/Diagnostics/FunctionActivitySource.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Diagnostics;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.Functions.Worker.Diagnostics
+{
+    internal class FunctionActivitySourceFactory
+    {
+        private static readonly ActivitySource _activitySource = new(TraceConstants.FunctionsActivitySource);
```
RohitRanjanMS, line 14: we can also set the version here.

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
```diff
@@ -20,18 +24,36 @@ public void Initialize(TelemetryConfiguration configuration)
 
             _listener = new ActivityListener
             {
-                ShouldListenTo = source => source.Name.StartsWith("Microsoft.Azure.Functions.Worker"),
+                ShouldListenTo = source => source.Name.StartsWith(TraceConstants.FunctionsActivitySource),
                 ActivityStarted = activity =>
                 {
                     var dependency = _telemetryClient.StartOperation<DependencyTelemetry>(activity);
```
RohitRanjanMS, line 30: will this create a `RequestTelemetry`?

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
```diff
@@ -20,18 +24,36 @@ public void Initialize(TelemetryConfiguration configuration)
...
-                    dependency.Telemetry.Type = "Azure.Functions";
-                    activity.SetCustomProperty("_depTel", dependency);
+                    activity.SetCustomProperty(DependencyTelemetryKey, dependency);
```
lmolkova, line 31: oh, btw, we need to set dependency.Type to  "InProc" to make AI UX happy

https://github.com/microsoft/ApplicationInsights-dotnet/blob/18be714998a846aba7e9b673d47eece67be2d3dc/WEB/Src/DependencyCollector/DependencyCollector/Implementation/RemoteDependencyConstants.cs#L17

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
brettsam: This doesn't capture any tags added by someone later, though -- like inside the function itself with `Activity.Current.SetTag()`.

I had added a test for that -- is that not expected to work for auto-tracking stuff like this?

If I recall -- I copied the behavior from here: 
- https://github.com/microsoft/ApplicationInsights-dotnet/blob/main/WEB/Src/DependencyCollector/DependencyCollector/Implementation/EventHandlers/DiagnosticsEventHandlerBase.cs#L83

Which gets called from places like this:
- https://github.com/microsoft/ApplicationInsights-dotnet/blob/main/WEB/Src/DependencyCollector/DependencyCollector/Implementation/EventHandlers/ServiceBusDiagnosticsEventHandler.cs#L100

src/DotNetWorker.Core/Diagnostics/TraceConstants.cs
brettsam: Oh yep -- easy to change.

src/DotNetWorker.Core/Diagnostics/FunctionActivitySource.cs
brettsam: Done.

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
lmolkova: yep, you're right, thanks for checking!

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
brettsam: No -- `DependencyTelemetry`.

Although I guess this is a good question for App Insights experts... this is technically a distributed call across processes. Should we treat this as a separate service and create a `RequestTelemetry` with a parent of the host's `RequestTelemetry`?

Any opinions @lmolkova?

src/DotNetWorker.Core/FunctionsApplication.cs
brettsam: so -- only set it to Error if we catch the Exception here? Otherwise leave it alone?

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
RohitRanjanMS: what's the appropriate Activity.Kind in this case? If we create an activity with Server/Consumer, Az Mon exporter translates that to a RequestTelemetry. We definitely have clinet and server relationship between host and worker. Another thing that we have to check is how it looks in the Application Map.

src/DotNetWorker.Core/FunctionsApplication.cs
brettsam: Yup!

src/DotNetWorker.Core/Diagnostics/FunctionActivitySource.cs
brettsam: Done.

src/DotNetWorker.Core/FunctionsApplication.cs
lmolkova: yep!

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
```diff
@@ -40,6 +62,42 @@ public void Initialize(TelemetryConfiguration configuration)
...
+
+                ExceptionTelemetry et = new(new[] { edi }, SeverityLevel.Error, null, new Dictionary<string, string>() { }, new Dictionary<string, double>() { });
+
+                telemetryClient.TrackException(et);
+            }
+        }
+
         public void Dispose()
         {
             _telemetryClient?.Flush();
```
RohitRanjanMS, line 103: Let's use FlushAsync, https://github.com/Azure/azure-webjobs-sdk/pull/2939/files#:~:text=var%20task%20%3D%20_client.FlushAsync(_cancellationTokenSource.Token)%3B

@michaelpeng36 is making this change in host as well.

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
lmolkova: Interesting question. Apparently, Java workers don't report any telemetry about worker invocation fully relying on host trigger `RequestTelemetry` and faking parent on the nested calls. It might be possible in .NET, but would be hacky. 

Since this package is in preview, I would not block on that, especially if you want users to have a similar experience on OTel.

Then the next question is how to make it look best in AI:
- host will record RequestTelemetry
- worker would need to report DependencyTelemetry (as a child of request). Adding another request would mess up AI UX. To make it work properly, the type of this telemetry should be "InProc". Otherwise AppMap will be broken among other things. 
- there will be child auto-traced activities nested under worker invocation dependency.

Internal activity maps to InProc dependency telemetry - we use in Azure SDKs for the same reasons to tell UX where is the real network call and where is a logical operation.

Regarding Activity.Kind - It seems faas otel spec only talks about outer (host) request, but afaik AWS lambda reports several spans too - it'd be nice to clarify with otel faas WG and write down in the spec.

brettsam: This is awesome information, thanks @lmolkova! It does seem like we want this tracked as an "InProc" Dependency for the best App Insights experience.

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
brettsam: Already done :-)

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
```diff
@@ -20,18 +27,37 @@ public void Initialize(TelemetryConfiguration configuration)
...
                 },
                 ActivityStopped = activity =>
                 {
-                    var dependency = activity.GetCustomProperty("_depTel") as IOperationHolder<DependencyTelemetry>;
-                    _telemetryClient.StopOperation(dependency);
-                    dependency?.Dispose();
+                    // Check for Exceptions events
+                    foreach (ActivityEvent activityEvent in activity.Events)
+                    {
+                        TrackExceptionTelemetryFromActivityEvent(activityEvent, _telemetryClient);
```
jviau, line 42: I am not sure we want to track exceptions separately like this. The reason you see both exception & request telemetry when using app insights is because they are tracked separately. It just happens that something is either calling `ILogger` with the exception and also the `Activity` is being tracked separately.

I also think AppInsights SDK has some code to hook into unhandled exceptions and export them.

In short: I think we should trust any exception here would be logged separately. We can also ensure we log unhandled exceptions from a function invocation - to make sure that scenario is covered.

src/DotNetWorker.Core/Diagnostics/FunctionActivitySource.cs
```diff
@@ -0,0 +1,64 @@
...
+        public FunctionActivitySourceFactory(IOptions<WorkerOptions> options)
+        {
+            _attributeMap = new Lazy<IReadOnlyDictionary<string, string>>(() => GetMapping(options.Value.OpenTelemetrySchemaVersion));
+            _schemaVersionUrl = TraceConstants.OpenTelemetrySchemaMap[options.Value.OpenTelemetrySchemaVersion];
+        }
+
+        public Activity? StartInvoke(FunctionContext context)
+        {
+            var activity = _activitySource.StartActivity(TraceConstants.FunctionsInvokeActivityName, ActivityKind.Internal, context.TraceContext.TraceParent,
+                tags: GetTags(context));
```
jviau, line 27: To optimize the non-listener scenario, we can build and add tags only if `activity != null`:

``` CSharp
var activity = _activitySource.StartActivity(TraceConstants.FunctionsInvokeActivityName, ActivityKind.Internal, context.TraceContext.TraceParent);

if (activity is null)
{
    return null;
}

activity.TraceStateString = context.TraceContext.TraceState;

// add tags here.
return activity;
```

I also like to invert the null check and short circuit, allowing the rest of my logic to be outside of an `if` statement. But that is all personal preference.

src/DotNetWorker.Core/Diagnostics/FunctionActivitySource.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        public FunctionActivitySourceFactory(IOptions<WorkerOptions> options)
+        {
+            _attributeMap = new Lazy<IReadOnlyDictionary<string, string>>(() => GetMapping(options.Value.OpenTelemetrySchemaVersion));
+            _schemaVersionUrl = TraceConstants.OpenTelemetrySchemaMap[options.Value.OpenTelemetrySchemaVersion];
+        }
+
+        public Activity? StartInvoke(FunctionContext context)
+        {
+            var activity = _activitySource.StartActivity(TraceConstants.FunctionsInvokeActivityName, ActivityKind.Internal, context.TraceContext.TraceParent,
```
jviau, line 26: Should this be `ActivityKind.Server` as we are ultimately serving a function invocation request?

jviau, line 26: There is a `StartActivity` overload where you can pass in `ActivityContext parentContext`, which is a struct to hold traceParent and traceState.

``` CSharp
ActivityContext.TryParse(context.TraceContext.TraceParent, context.TraceContext.TraceState, out ActivityContext parentContext);
```

jviau (commented):
This is great to see!

A question that I have with this and with DurableTask tracing: I would like to ship the activity source as "preview" to start. This is so we can iterate on schema. But how? How do we ship the activity source _within_ this already GA package? Do we need to split this out into its own package `Microsoft.Azure.Functions.Worker.Instrumentation` and have some internal hook point for that separate package to start/stop its activity at the right point?

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
```diff
@@ -20,18 +27,37 @@ public void Initialize(TelemetryConfiguration configuration)
 
             _listener = new ActivityListener
             {
-                ShouldListenTo = source => source.Name.StartsWith("Microsoft.Azure.Functions.Worker"),
+                ShouldListenTo = source => source.Name.StartsWith(TraceConstants.FunctionsActivitySource),
                 ActivityStarted = activity =>
                 {
                     var dependency = _telemetryClient.StartOperation<DependencyTelemetry>(activity);
```
jviau, line 33: I think it would be helpful to inspect the `ActivityKind` and instantiate an appropriate telemetry item from there.

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
```diff
@@ -98,10 +102,16 @@ internal static void TrackExceptionTelemetryFromActivityEvent(ActivityEvent acti
...
 
-        public void Dispose()
+        public async ValueTask DisposeAsync()
         {
-            _telemetryClient?.Flush();
             _listener?.Dispose();
+
+            if (_telemetryClient is not null)
+            {
+                await _telemetryClient.FlushAsync(CancellationToken.None);
```
RohitRanjanMS, line 111: FlushAsync may take longer, especially if the telemetry client is in backoff retry mode. It would be safe to cancel flush after few seconds if it's not complete.

brettsam: Agree. We're working through getting this right in the host as well -- at some point we just give up -- but that's not even enough time in some cases and we end up with CRIs with missing telemetry during a shutdown (I think we give up after 2 seconds there).

I could start with 5 second cancelation and let people configure it? I was noticing that even AppInsights doesn't have great handling for this: https://github.com/microsoft/ApplicationInsights-dotnet/blob/main/LOGGING/src/ILogger/ApplicationInsightsLoggerProvider.cs#L111-L113

src/DotNetWorker.Core/Diagnostics/FunctionActivitySource.cs
brettsam: `GetTags()` is using `yield return` for this situation. No tags are instantiated unless the `Activity` is created. The trick is that I wanted as much info as possible sent to the ActivityStarted handler in the listener -- in order to have any tags at that point, they need to be passed into `StartActivity()`.

brettsam: omg -- 2 seconds after sending this, I see there's a `CreateActivity()` that lets you handle this in a more sane matter -- you can call that, then separately call `Start()`. This was added in 6.0.0 -- but I originally wrote all this with 5.0.0 which is why I never saw it.

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
brettsam: I tested without this -- and it had no `ExceptionTelemetry` associated. You'd get the `DependencyTelemetry` marked as "Error" but no context for what the error was.

The behavior here is very similar to what they're doing in the OTel exporter by pulling out the exception and logging it separately: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/monitor/Azure.Monitor.OpenTelemetry.Exporter/src/Internals/TraceHelper.cs#L37

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
RohitRanjanMS: Michael has PR open to improve this on host. https://github.com/Azure/azure-webjobs-sdk/pull/2939
He is logging if the flush fail to complete in stipulated time.

src/DotNetWorker.Core/Diagnostics/FunctionActivitySource.cs
brettsam: See the comment here -- https://github.com/Azure/azure-functions-dotnet-worker/pull/1307#discussion_r1095296512. It seems that this is currently not really defined by the otel spec, but that App Insights and others treat this as an internal call with the host log being the "Server" piece. If they were both "Server", rendering of the Span would not be correct.

I think this is something we can handle with the OTel schema versioning we're doing -- if things change in the future, we can use that to change the Kind as well.

brettsam: Wouldn't we have to do this first, though, even if there's nothing Listening? I wanted to only do stuff like this if we actually create an Activity. Or maybe there's some other benefit to using the ActivityContext?

brettsam: Oh we could wrap all this in `_activitySource.HasListeners()`...

src/DotNetWorker.Core/Diagnostics/FunctionActivitySource.cs
brettsam: I've actually wrapped this in `_activitySource.HasListeners()` before I do anything -- this should cover it I believe. It also makes it more clear for future maintainers of this.

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
brettsam: I'd vote that we do that when (if) we add new activities. I'm really hoping that this assembly is a short-lived "bridge" solution until the Azure Monitor otel support GAs -- then we can recommend folks go there instead.

brettsam:
@jviau re: versioning

I was planning to get this in as GA inside the Worker assembly without any preview. (The ActivitySource; not the App Insights stuff). I was intentionally keeping this initial Activity small -- with really only a single tag added plus the schema. I figured the only real changes I see coming are:
- add new tags to make things richer -- which I wouldn't see as "breaking"
- some tag needs to change due to the otel schema changing. This is factored in (i.e. "faas.execution" -> "faas.invocation" mapping would already happen and customer would choose the schema they want to run with in WorkerOptions).
- we'd need to change the Activity name or kind? We can hide this behind the schema selection as well.

We're getting lots of requests for this to be "official" -- so I didn't want to sit in Preview for a long time. I know changes will come which is why I chatted with @lmolkova about how everyone else is handling schema changes. 

We can certainly move to an external diagnostics assembly -- it'd be more flexible. But I'm sure even after we GA that we'd get more changes coming in the future. Maybe it'd be easier to bump that one's major we have to do a major overhaul?

Curious how others feel -- @fabiocav / @lmolkova / @RohitRanjanMS

src/DotNetWorker.Core/Diagnostics/FunctionActivitySource.cs
jviau: I think we need to have a discussion about the spans we will all have. I don't think we should be trying to hide the gRPC hop from Host to Worker. Our Host/Worker design is a distributed system and it should be represented as such in spans. Customers should see the following at minimum:

Incoming request (Host Server Span) --> Host Calling Worker (Host Client Span) --> Worker handling invocation (Worker Server Span)

And then we also have to see how gRPC spans factor into all of this. Given that N gRPC spans may relate to a single invocation, I think they should be links on the Functions spans and not parent/child

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
jviau: Yes, the exception telemetry would need to be logged separately. But traces and logs are separate telemetry items - we should keep their implementations separate. Is it feasible for us to add an exception log via `ILogger` somewhere for a failed function invocation? A separate middleware?

src/DotNetWorker.Core/Diagnostics/FunctionActivitySource.cs
brettsam: So -- the spec does say it should be Server: https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/faas/#incoming-invocations. But I also see really no notion of a "host". It seems to only care about the instance handling the invocation itself.

I think in most other clouds, the "host" -- the thing polling/sending you events -- is part of infrastructure and not even present in the distributed trace? anyone have any idea?

@lmolkova -- if we do indeed make this a Server (which is what it seems like it should be) -- that would turn this span into a Request in App Insights, right?

src/DotNetWorker.ApplicationInsights/FunctionsApplicationInsightsExtensions.cs
```diff
@@ -44,11 +43,13 @@ public static IFunctionsWorkerApplicationBuilder AddApplicationInsightsLogger(th
         {
             builder.AddCommonServices();
 
+            // Lets the host know that the worker is sending logs to App Insights. The host will now ignore these.
+            builder.Services.Configure<WorkerOptions>(workerOptions => workerOptions.Capabilities["WorkerApplicationInsightsLoggingEnabled"] = bool.TrueString);
```
kshyju, line 47: Would this work fine with the changes exposed in #1183 where we allow users to override/disable default values of capability ?

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
```diff
@@ -20,18 +27,37 @@ public void Initialize(TelemetryConfiguration configuration)
...
+
+                        foreach (var item in activity.Tags)
+                        {
+                            if (!dependency.Properties.ContainsKey(item.Key))
+                            {
+                                dependency.Properties[item.Key] = item.Value;
+                            }
+                        }
+
+                        dependency.Success = activity.Status != ActivityStatusCode.Error;
```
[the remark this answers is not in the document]
jviau: There are three values - `Unset`, `Ok`, and `Error`. Both `Unset` and `Ok` translate to success - so this is correct.

src/DotNetWorker.Core/FunctionsApplication.cs
```diff
@@ -61,12 +65,23 @@ public void LoadFunction(FunctionDefinition definition)
...
-            using (_logger.BeginScope(scope))
+
+            using var logScope = _logger.BeginScope(scope);
+            using Activity? invokeActivity = _functionActivitySourceFactory.StartInvoke(context);
+
+            try
             {
-                return _functionExecutionDelegate(context);
+                await _functionExecutionDelegate(context);
+            }
```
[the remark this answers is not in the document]
jviau: Setting `OK` is optional. It is fine to leave it as `Unset` for a success case.

src/DotNetWorker.Core/Diagnostics/FunctionActivitySource.cs
jviau: Yes, `Server` becomes `request` in application insights. This is also beneficial because app insights has built in aggregation for request telemetry - failure rate, performance, etc. I see this as a great benefit for customers.

lmolkova: If host reports queue trigger, it will report it as consumer, but I think the question here is how should we treat worker, i.e.
Option 1:
- Function-host (Server/Consumer span)
  - Call to worker (client span)
    - Incoming on worker (Server span)

Option 2:
- Function-host (Server/Consumer span)
  - Incoming on worker (Internal span)


Option 3 (possible with hacks only):
- Function-host (Server/Consumer span)
  - no new span for worker invocation, but host context is picked up in `Activity.Current`

I.e. if we consider host and worker as the same instance and want to hide that they are different processed, we should pick option 2. Otherwise we should proceed with option 1. 

I agree that FaaS otel spec does not specify it at all, but should at least recommend a default. 

My opinion here is that Option2 seems like a good start:
- presumably, communication between worker and host is super-fast and reliable and therefore is not quite interesting. Users might not like the idea to pay for extra telemetry that does not explain anything new. I would even prefer option 3 if only it was possible without hacks
- it keeps the door open to implement Option1 or 3 in the future with minimal behavior changes to users

lmolkova: Created https://github.com/open-telemetry/opentelemetry-specification/issues/3213

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
brettsam: I'm not sure I understand the comment. I'm doing the same thing that the App Insights OTel exporter is doing: if there's an Exception in the Activity, create an `ExceptionTelemetry`. The App Insights SDK doesn't do this automatically so we need to do it here.

If we add an `ILogger.LogError()` upon an Exception, that will log a `TraceTelemetry` in App Insights. It will also log an `ExceptionTelemetry`, but this can be disabled via options:

```
.AddApplicationInsightsLogger((ApplicationInsightsLoggerOptions options) => options.TrackExceptionsAsExceptionTelemetry = false)
```

I'm trying to keep the experience close to what OTel will have so that switching will be easier...

test/DotNetWorkerTests/TestFunctionInvocation.cs
```diff
@@ -24,6 +24,6 @@ public TestFunctionInvocation(string id = null, string functionId = null)
 
         public override string FunctionId { get; } = Guid.NewGuid().ToString();
 
-        public override TraceContext TraceContext { get; } = new DefaultTraceContext(Guid.NewGuid().ToString(), Guid.NewGuid().ToString());
+        public override TraceContext TraceContext { get; } = new DefaultTraceContext($"00-{Guid.NewGuid()}-00", Guid.NewGuid().ToString());
```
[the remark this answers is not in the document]
brettsam: Good idea. Yes, I needed this App Insights logic to kick in: https://github.com/microsoft/ApplicationInsights-dotnet/blob/main/BASE/src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs#L312

src/DotNetWorker.ApplicationInsights/FunctionsApplicationInsightsExtensions.cs
brettsam: Yes, that's why I introduced that in the first place. B/c these are `Options<>`, you can register your own `ConfigureOptions<>` later to remove this if you need to for some reason. They'll chain together.

src/DotNetWorker.Core/FunctionsApplication.cs
```diff
@@ -61,12 +65,23 @@ public void LoadFunction(FunctionDefinition definition)
...
+            catch (Exception ex)
+            {
+                invokeActivity?.SetStatus(ActivityStatusCode.Error, ex.Message);
+                invokeActivity?.RecordException(ex, escaped: true);
```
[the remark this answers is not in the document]
brettsam: Yes, I'll follow up with that change.

src/DotNetWorker.Core/Diagnostics/FunctionActivitySource.cs
brettsam: Ah, yes. Any other ActivitySource will be differentiated. That's why this one is named with ".Worker" -- it's only meant to emit Activities directly related to the Worker.

Any Activity related to the host will be appropriately named and, won't exist in the worker process -- those will be emitted via the Host process.

Does that address the concern?

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
jviau: Did not realize that implementation had this behavior. I don't exactly like the idea of mixing _log_ telemetry into _trace_ telemetry (with `ExceptionTelemetry` being a form of _log_). Can we include an option to enable/disable emitting `ExceptionTelemetry` from an `Activity`?

jviau: I talk about separation of _trace_ vs _log_ from a general telemetry concept point of view (3 pillars of telemetry: traces, logs, metrics), not app insights specifically. App Insights predates this general concept (I think), so yeah you will see a bunch of conflicting naming/decisions there.

Ultimately, I am fine with whatever you want to do here. So long as we also add the explicit `.LogError(exception, ...)` in the `catch` statement I commented on separately. I do still think it is odd for App Insights exporter to be mixing logs and traces - it doesn't feel right. Right now, I am going to see it produce redundant exception telemetry. When we add the LogError that is, but that has to happen as errors should be logged regardless of if tracing is enabled or not.

src/DotNetWorker.Core/Diagnostics/FunctionActivitySource.cs
lmolkova: > My justification is I believe that we should give Server spans for the Worker (and specifically not Internal)

Having two nested server spans will break the application map in application insights and could be confusing to other backends. It also goes against otel spec that says server spans should have client parents - this creates expectations on backends that would be broken. So if the decision is to use server span, it'd be best to add client span that tracks call to worker right away

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
jviau: Going to re-open this for reconsideration given our discussion of using `ActivityKind=Server` instead of `Internal`. In which case, should we use `RequestTelemetry` for `Server`?

Regarding the app map UX being messed up if we use request telemetry, as the host will already be emitting that. Should host and worker be seen as two separate apps? This way the UX will draw lines between the Host and Worker.

jviau, line 33: I was reading through `StartOperation` and I noticed it will call [Activity.Start](https://github.com/microsoft/ApplicationInsights-dotnet/blob/main/BASE/src/Microsoft.ApplicationInsights/TelemetryClientExtensions.cs#L330) - but this is an already started activity. This will cause an [exception to be raised](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs#L683) and immediately caught as this activity will always be already started.

This might not be a big deal, just something to think about.

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
```diff
@@ -20,18 +27,37 @@ public void Initialize(TelemetryConfiguration configuration)
...
+                        TrackExceptionTelemetryFromActivityEvent(activityEvent, _telemetryClient);
+                    }
+
+                    if (activity.GetCustomProperty(DependencyTelemetryKey) is IOperationHolder<DependencyTelemetry> dependencyHolder)
+                    {
+                        var dependency = dependencyHolder.Telemetry;
+
+                        foreach (var item in activity.Tags)
+                        {
+                            if (!dependency.Properties.ContainsKey(item.Key))
```
jviau, line 51: Are there any special tags we should be looking out for and be set on an explicit telemetry property? Such as `ResultCode`, `Data`, or `Target` (for `DependencyTelemetry`).

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
```diff
@@ -20,18 +27,37 @@ public void Initialize(TelemetryConfiguration configuration)
...
+                            {
+                                dependency.Properties[item.Key] = item.Value;
+                            }
+                        }
+
+                        dependency.Success = activity.Status != ActivityStatusCode.Error;
+                        _telemetryClient.StopOperation(dependencyHolder);
```
jviau, line 58: I think we will need to do some additional manual work on stop.

1. Setting of `.Success` - I don't think `StopOperation` will do that.
2. Update `Operation.Name` if the user has changed `Activity.DisplayName`.

Any others?

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
```diff
@@ -40,10 +66,64 @@ public void Initialize(TelemetryConfiguration configuration)
...
+        }
+
+        public async ValueTask DisposeAsync()
         {
-            _telemetryClient?.Flush();
             _listener?.Dispose();
+
+            if (_telemetryClient is not null)
+            {
+                CancellationTokenSource cts = new(millisecondsDelay: 5000);
```
jviau, line 111: small nit:

```suggestion
                using CancellationTokenSource cts = new(millisecondsDelay: 5000);
```

and remove the `finally`?

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryInitializer.cs
```diff
@@ -1,90 +1,42 @@
...
-                instanceName = Environment.GetEnvironmentVariable(ComputerNameKey);
-                if (string.IsNullOrEmpty(instanceName))
-                {
-                    instanceName = Environment.GetEnvironmentVariable(ContainerNameKey);
-                }
-            }
+            string? version = typeof(FunctionsTelemetryInitializer).Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion;
 
-            return instanceName ?? Environment.MachineName;
+            return version == null ? null : $"azurefunctions-netiso: {version}";
```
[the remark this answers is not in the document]
brettsam: I made it up -- App Insights wants us to include our own version here (of our choosing) so they can tell how many folks are using our SDKs. We do similar for:
- WebJobs: https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ISdkVersionProvider.cs#L15
- Functions: https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Diagnostics/FunctionsSdkVersionProvider.cs#L12

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
brettsam: Yes, I noticed as well. The problem is there's so much subtle code inside of the operation tracking w/ Activity that I didn't see a great way around it (I didn't want to copy all that logic). Can follow up with App Insights separately before we GA that to see if there's a better way.

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
brettsam: `Success` is set directly above the call to `StopOperation`. Or was there another `Success` you were referring to?

DisplayName needs some work from the host to really get right, which is why I haven't touched it yet. I see it as additive in the future. I've filed https://github.com/Azure/azure-functions-host/issues/9150 to help us with this. Will try to get this in soon so we can take advantage of it.

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
brettsam: The LogError is something I'll do in a follow-up -- it's not really related to ActivitySource and it's not related to App Insights so I didn't want to bundle that together.

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
brettsam: So that a customer can add a tag with `Target` and it'd overwrite whatever it was set to? I think we can see if that's something people want and come back to it -- or they can likely add their own Initializer for that as well.

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
brettsam: So this part (the App Insights package) will still be in Preview -- I'd like to get this in and we can come back to it after getting some feedback.

src/DotNetWorker.Core/Diagnostics/FunctionActivitySource.cs
jviau: Yes, option 1 does include the client span from Host -> Worker.

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
jviau: Ah yeah they can do their own initializer true. That is good enough for me.

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
jviau: Ah woops, missed the `Success` setting. Thanks!

jviau (approved):

PR #1331, opened 2023-02-10
Cleanup blob sdk-binding samples

liliankasem:


surgupta-msft (approved):

samples/WorkerBindingSamples/Blob/BlobInputBindingSamples.cs
```diff
@@ -42,7 +42,7 @@ public BlobInputBindingSamples(ILogger<BlobInputBindingSamples> logger)
             [BlobInput("input-container/sample1.txt")] Stream stream)
```
kshyju, line 42: nit: Remove unused using directives on top of this file.

kshyju (approved):

PR #1368, opened 2023-02-22
more ci improvements

brettsam:
- The emulator retry wasn't working because it was exiting upon timeout. I think I've fixed this.
- We have always been generating the incorrect `AssemblyInformationalVersion` string because we've been packing the assemblies from the *test build* rather than from the initial build. These were always built with Release config, but used a different InformationalVersion so we've always had "-local_____" in that string. I'm now specifiying `--no-build` during testing so we don't re-build those assemblies ever. Confirmed that things look right.
- Switched another place to use `PackageOutputPath` in case we get hit with the `-o` error in the near future.

### DIFF
--- a/tools/start-emulators.ps1
+++ b/tools/start-emulators.ps1
@@ -118,7 +118,7 @@ if (!$SkipCosmosDBEmulator -and $startedCosmos -eq $true)
     $cosmosStatus = Get-CosmosDbEmulatorStatus
     Write-Host "CosmosDB emulator status: $cosmosStatus"
 
-    $waitSuccess = Wait-CosmosDbEmulator -Status Running -Timeout 60
+    $waitSuccess = Wait-CosmosDbEmulator -Status Running -Timeout 60 -ErrorAction Continue
 
     if ($waitSuccess -ne $true)
     {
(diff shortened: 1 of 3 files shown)


tools/start-emulators.ps1
```diff
@@ -118,7 +118,7 @@ if (!$SkipCosmosDBEmulator -and $startedCosmos -eq $true)
     $cosmosStatus = Get-CosmosDbEmulatorStatus
     Write-Host "CosmosDB emulator status: $cosmosStatus"
 
-    $waitSuccess = Wait-CosmosDbEmulator -Status Running -Timeout 60
+    $waitSuccess = Wait-CosmosDbEmulator -Status Running -Timeout 60 -ErrorAction Continue
```
fabiocav, line 121: Could this lead to difficult to diagnose test failures?

fabiocav (approved):

tools/start-emulators.ps1
brettsam: no -- it'll fail again later if the emulator doesn't start successfully or it times out (this is what happens today) -- but this was short-circuiting the retry logic and failing too early

PR #1385, opened 2023-02-24
[Source Gen] Fixing incorrect entrypoint property value in metadata when namespace is different than assembly name

### LINKED ISSUE #1299: Incorrect EntryPoint value in function metadata when namespace is different than assembly name
In our new function metadata provider, The `entryPoint` property value in function metadata has the assembly name instead of namespace name. This causes function load to fail when the namespace of function class is different than the assembly name of the function app.

Relevant part of the code generated by legacy metadata generator, where the assembly name is "FunctionApp18" and namespace is "MyCompany". The entrypoint value is correct here.
```
"name": "Function1",
"scriptFile": "FunctionApp18.dll",
"entryPoint": "MyCompany.MyHttpFunctions.Run",
````
In the new metadata provider, the entryPoint value has the assembly name.

```
EntryPoint = "FunctionApp18.MyHttpFunctions.Run",
```

https://github.com/Azure/azure-functions-dotnet-worker/blob/14cc88859a1b7bbf68ee43409a1ad45a6e8362cb/sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs#L42

Logs which shows exception during function load.

> [2023-01-27T20:04:51.201Z] Worker process started and initialized.
[2023-01-27T20:04:51.336Z] Worker failed to load function: 'Function1' with function id: '2268186449'.
[2023-01-27T20:04:51.337Z] Result: Failure
Exception: System.InvalidOperationException: Method 'Run' specified in EntryPoint was not found. This function cannot be created.
[2023-01-27T20:04:51.338Z]    at Microsoft.Azure.Functions.Worker.Invocation.DefaultMethodInfoLocator.GetMethod(String pathToAssembly, String entryPoint)

### PR DESCRIPTION
**Pull request by kshyju**
Fixes #1299 

- Fixed the source gen so that it will use the correct fully qualified name of the parent class when building the EntryPoint property value in metadata. Previously we were using the assembly name to build the namespace part of this value.
- Fixed tests to reflect this.
- Added a new test for nested types scenario.

### DIFF
--- a/test/Sdk.Generator.Tests/FunctionMetadataProviderGeneratorTests/EventHubsBindingsTests.cs
+++ b/test/Sdk.Generator.Tests/FunctionMetadataProviderGeneratorTests/EventHubsBindingsTests.cs
@@ -1,9 +1,6 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
-using System;
-using System.ComponentModel.DataAnnotations;
-using System.Linq;
 using System.Reflection;
 using System.Text;
 using Microsoft.Azure.Functions.Worker.Sdk.Generators;
@@ -15,7 +12,7 @@ namespace Microsoft.Azure.Functions.SdkGeneratorTests
     {
         public class EventHubsBindingsTests
         {
-            private Assembly[] referencedExtensionAssemblies;
+            private readonly Assembly[] _referencedExtensionAssemblies;
             private readonly string _usingStringsForInput;
 
             public EventHubsBindingsTests()
@@ -28,7 +25,7 @@ namespace Microsoft.Azure.Functions.SdkGeneratorTests
                 var hostingAbExtension = Assembly.LoadFrom("Microsoft.Extensions.Hosting.Abstractions.dll");
                 var diAbExtension = Assembly.LoadFrom("Microsoft.Extensions.DependencyInjection.Abstractions.dll");
 
-                referencedExtensionAssemblies = new[]
+                _referencedExtensionAssemblies = new[]
                 {
                     abstractionsExtension,
                     httpExtension,
@@ -110,7 +107,7 @@ namespace Microsoft.Azure.Functions.SdkGeneratorTests
                                 {
                                     Language = "dotnet-isolated",
                                     Name = "{{functionName}}",
-                                    EntryPoint = "TestProject.EventHubsInput.{{functionName}}",
+                                    EntryPoint = "FunctionApp.EventHubsInput.{{functionName}}",
                                     RawBindings = Function0RawBindings,
                                     ScriptFile = "TestProject.dll"
                                 };
@@ -139,7 +136,7 @@ namespace Microsoft.Azure.Functions.SdkGeneratorTests
                     """);
 
                 await TestHelpers.RunTestAsync<FunctionMetadataProviderGenerator>(
-                    referencedExtensionAssemblies,
+                    _referencedExtensionAssemblies,
                     inputCodeBuilder.ToString(),
                     expectedGeneratedFileName,
                     expectedOutputBuilder.ToString());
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: test/Sdk.Generator.Tests/FunctionMetadataProviderGeneratorTests/NestedTypesTest.cs [hunk 1]
```diff
@@ -0,0 +1,215 @@
...
+            {
+                string inputCode = """
+                using System;
+                using System.Collections.Generic;
+                using Microsoft.Azure.Functions.Worker;
+                using Microsoft.Azure.Functions.Worker.Http;
+
+                namespace MyCompany
+                {
+                    public class  WebApis
```
kshyju, line 138: The legacy metadata generator fails to generate for this case (nested classes).

test/Sdk.Generator.Tests/FunctionMetadataProviderGeneratorTests/EventHubsBindingsTests.cs
```diff
@@ -110,7 +107,7 @@ public Task<ImmutableArray<IFunctionMetadata>> GetFunctionMetadataAsync(string d
                                 {
                                     Language = "dotnet-isolated",
                                     Name = "{{functionName}}",
-                                    EntryPoint = "TestProject.EventHubsInput.{{functionName}}",
+                                    EntryPoint = "FunctionApp.EventHubsInput.{{functionName}}",
```
kshyju, line 110: FunctionApp is the namespace of the input code.

test/Sdk.Generator.Tests/FunctionMetadataProviderGeneratorTests/NestedTypesTest.cs
```diff
@@ -0,0 +1,215 @@
...
+{
+    public partial class FunctionMetadataProviderGeneratorTests
+    {
+        public class NestedTypesTests
+        {
+            private readonly Assembly[] _referencedExtensionAssemblies;
+
+            public NestedTypesTests()
+            {
+                // load all extensions used in tests (match extensions tested on E2E app? Or include ALL extensions?)
```
satvu, line 19: whatever extensions are used in the inputCode snippets in the tests in this file need to be loaded and others can be dropped.

test/Sdk.Generator.Tests/FunctionMetadataProviderGeneratorTests/NestedTypesTest.cs
satvu: Do we want to fail to generate in this case? Not sure what the pros/cons are, but I see the test is allowing this behavior.

liliankasem: It's probably a good thing to supported nested classes

liliankasem (approved):

test/Sdk.Generator.Tests/FunctionMetadataProviderGeneratorTests/NestedTypesTest.cs
kshyju: > Do we want to fail to generate in this case? Not sure what the pros/cons are, but I see the test is allowing this behavior.

No. The legacy generator behavior is wrong here. It should still generate.

satvu (approved):

test/Sdk.Generator.Tests/FunctionMetadataProviderGeneratorTests/NestedTypesTest.cs
kshyju: I think we need the 6 we are loading. Do you notice anything which is not needed here?

satvu: I was responding to the comment - is that copy/pasted from somewhere? It reads like my writing ... so if it is copy/pasted will note that comment for a cleanup PR

PR #1397, opened 2023-03-02
Reading AzureWebJobsScriptRoot env variable instead of FUNCTIONS_WORKER_DIRECTORY to get script root

kshyju:
Reverted PR #1330. There was merge conflict on the csproj file of dotnet worker. Kept the version of DotnetWorker package as what is in latest main(1.13.0-preview1).

In #1330 we made the change to read from `FUNCTIONS_WORKER_DIRECTORY` instead of `AzureWebJobsScriptRoot` to get the script root of the isolated process binary. This causes placeholder specialization use case to fail. In placeholder use case, the env variable `FUNCTIONS_WORKER_DIRECTORY` has the path to the worker location for the language worker (where we will have the `workerconfig.json` and our native placeholder process. Ex: "WebJobs.Script.WebHost\bin\Debug\net6.0\workers\dotnet-isolated")

For a normal use case (non placeholder), I observed both env variables has the same value , the path to the customer payload binaries. 

image

### DIFF
--- a/src/DotNetWorker.Grpc/Definition/GrpcFunctionDefinition.cs
+++ b/src/DotNetWorker.Grpc/Definition/GrpcFunctionDefinition.cs
@@ -21,10 +21,10 @@ namespace Microsoft.Azure.Functions.Worker.Definition
             Name = loadRequest.Metadata.Name;
             Id = loadRequest.FunctionId;
 
-            string? scriptRoot = Environment.GetEnvironmentVariable("FUNCTIONS_WORKER_DIRECTORY");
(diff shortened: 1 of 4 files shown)


fabiocav:
This is reverting a tactical change without clarity on the root cause of the original issue reported by @liliankasem . Can we follow up on the original issue to understand what was impacting her scenarios before reverting the change?

fabiocav (changes requested):
Let's wait on changing this in `main` until we have clarity on the issue impacting the original code

PR #1271, opened 2023-01-13
Assign cardinality correctly for Blob collection scenarios

surgupta-msft:
### Issue describing the changes in this PR

resolves #1204

Adding isBatched attribute as the [FunctionMetadataGenerator](https://github.com/Azure/azure-functions-dotnet-worker/blob/7b0860ade7ec3ac362e3e80cfe16fb36d67afac1/sdk/Sdk/FunctionMetadataGenerator.cs#L541) already has the logic to correctly assign the `cardinality` to `Many` if `isBatched = true`

### Pull request checklist

* [ ] My changes **do not** require documentation changes
  * [ ] Otherwise: Documentation issue linked to PR
* [ ] My changes **should not** be added to the release notes for the next release
  * [ ] Otherwise: I've added my notes to `release_notes.md`
* [ ] My changes **do not** need to be backported to a previous version
  * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [x] I have added all required tests (Unit tests, E2E tests)

### Additional information

Additional PR information

PR #1402, opened 2023-03-07
[Source Gen] Fix cardinality validation to include all scenarios

### LINKED ISSUE #1372: [Source Gen] Validating Cardinality Scenarios
From https://github.com/Azure/azure-functions-dotnet-worker/issues/1254, investigation showed that cardinality was assigned correctly for types that have cardinality (and therefore set `IsBatched` by default), but validation is only on one binding type (instead of all bindings that support cardinality). Items to fix:

- Refactor [FunctionMetadataGeneratorProvider.Parser](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs) to validate Cardinality based on whether or not IsBatched is set (by default or by user)
- Update tests and diagnostic descriptors

satvu:

resolves #1372 

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/docs/analyzer-rules/AZFW0008.md
+++ b/docs/analyzer-rules/AZFW0008.md
@@ -8,19 +8,19 @@
 
 ## Cause
 
-This rule is triggered when an Input Binding or Trigger of an Azure Function has an invalid Cardinality.
+This rule is triggered when an Input or Trigger Binding of an Azure Function has an invalid Cardinality.
 
 ## Rule description
 
-"Cardinality" dictates whether or not an input is batched together or processed individually. It is defined by using the argument "IsBatched" in an Azure Functions Input Binding or Trigger attribute. When IsBatched is true, Cardinality is set to `Many`. When IsBatched is false, Cardinality is set to `One`.
+"Cardinality" dictates whether or not an input is batched together or processed individually. It is defined by using the argument "IsBatched" in an Azure Functions Input or Trigger Binding attribute. When IsBatched is true, Cardinality is set to `Many`. When IsBatched is false, Cardinality is set to `One`.
 
-All Input Bindings and Triggers must correctly declare a compatible "IsBatched" value and parameter input type. For example, for bindings and triggers where `IsBatched = true`, the input parameter type must be an iterable collection like `string[]` or `List<string>`. Combining `IsBatched = true` with a parameter of `string[]` is valid, but combining `IsBatched = true` with a parameter of `string` is invalid.
+All Input and Trigger Bindings must correctly declare a compatible "IsBatched" value and parameter input type. For example, for bindings where `IsBatched = true`, the input parameter type must be an iterable collection like `string[]` or `List<string>`. Combining `IsBatched = true` with a parameter of `string[]` is valid, but combining `IsBatched = true` with a parameter of `string` is invalid.
 
-_**Note:**_ The default value of `IsBatched` changes depending on the Input Binding or Trigger type. The default value of `IsBatched` for each type can be found in the dotnet-isolated [extensions libraries](https://github.com/Azure/azure-functions-dotnet-worker/tree/main/extensions). If a value is not explicitly provided in the named arguments of the attribute, the default value will be used.
+_**Note:**_ The default value of `IsBatched` changes depending on the Input or Trigger Binding type. The default value of `IsBatched` for each type can be found in the dotnet-isolated [extensions libraries](https://github.com/Azure/azure-functions-dotnet-worker/tree/main/extensions). If a value is not explicitly provided in the named arguments of the attribute, the default value will be used.
 
 ## How to fix violations
 
-If an Input Binding or Trigger has `IsBatched = true` (explicitly or by default), the input parameter type must be changed to an iterable collection. Otherwise, `IsBatched` needs to be set to false.
+If an Input or Trigger Bindings has `IsBatched = true` (explicitly or by default), the input parameter type must be changed to an iterable collection. Otherwise, `IsBatched` needs to be set to false.
 
 ## When to suppress warnings
 
--- a/docs/analyzer-rules/AZFW0008.md
+++ b/docs/analyzer-rules/AZFW0008.md
@@ -14,13 +14,13 @@ This rule is triggered when an Input Binding or Trigger of an Azure Function has
 
 "Cardinality" dictates whether or not an input is batched together or processed individually. It is defined by using the argument "IsBatched" in an Azure Functions Input Binding or Trigger attribute. When IsBatched is true, Cardinality is set to `Many`. When IsBatched is false, Cardinality is set to `One`.
 
-All Input Bindings and Triggers must correctly declare a compatible "IsBatched" value and parameter input type. For example, for EventHubs triggers where `IsBatched = true`, the input parameter type must be an iterable collection. Combining `IsBatched = true` with a parameter of `string[]` is valid, but combining `IsBatched = true` with a parameter of `string` is invalid.
+All Input Bindings and Triggers must correctly declare a compatible "IsBatched" value and parameter input type. For example, for bindings and triggers where `IsBatched = true`, the input parameter type must be an iterable collection like `string[]` or `List<string>`. Combining `IsBatched = true` with a parameter of `string[]` is valid, but combining `IsBatched = true` with a parameter of `string` is invalid.
 
-_**Note:**_ The default value of IsBatched changes depending on the Input Binding or Trigger type. The default value of `IsBatched` for each type can be found in the dotnet-isolated [extensions libraries](https://github.com/Azure/azure-functions-dotnet-worker/tree/main/extensions). If a value is not explicitly provided in the named arguments of the attribute, the default value will be used.
+_**Note:**_ The default value of `IsBatched` changes depending on the Input Binding or Trigger type. The default value of `IsBatched` for each type can be found in the dotnet-isolated [extensions libraries](https://github.com/Azure/azure-functions-dotnet-worker/tree/main/extensions). If a value is not explicitly provided in the named arguments of the attribute, the default value will be used.
 
 ## How to fix violations
 
-If you have an EventHubs trigger where `IsBatched = true` (explicitly or by default), make sure your input parameter type is an iterable collection.
+If an Input Binding or Trigger has `IsBatched = true` (explicitly or by default), the input parameter type must be changed to an iterable collection. Otherwise, `IsBatched` needs to be set to false.
 
 ## When to suppress warnings
 
--- a/docs/analyzer-rules/AZFW0008.md
+++ b/docs/analyzer-rules/AZFW0008.md
@@ -1,4 +1,4 @@
-# AZFW0008: Invalid EventHubs Trigger
+# AZFW0008: Invalid Cardinality
 
 | | Value |
 |-|-|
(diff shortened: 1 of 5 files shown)


sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -576,57 +563,71 @@ private void OverrideBindingName(INamedTypeSymbol attributeClass, ref string arg
...
-            /// This method verifies that an EventHubsTrigger matches our expectations on cardinality (isBatched property). If isBatched is set to true, the parameter with the
+            /// This method verifies that a binding that has Cardinality (isBatched property) is valid. If isBatched is set to true, the parameter with the
             /// attribute must be an enumerable type.
             /// </summary>
-            private bool IsEventHubsTriggerValid(IParameterSymbol parameterSymbol, TypeSyntax? parameterTypeSyntax, SemanticModel model, AttributeData attribute, out DataType dataType, out Cardinality cardinality)
+            private bool IsCardinalityValid(IParameterSymbol parameterSymbol, TypeSyntax? parameterTypeSyntax, SemanticModel model, AttributeData attribute, out DataType dataType)
             {
                 dataType = DataType.Undefined;
-                cardinality = Cardinality.Undefined;
+                var cardinalityIsNamedArg = false;
```
satvu, line 587: Note: No longer handling cardinality as a special case. All _Named Arguments_ and _Arguments with default values_ are handled in [this method](https://github.com/Azure/azure-functions-dotnet-worker/blob/0fd7bf6aef005e4b8a14874506bf7a8ad7ad73ef/sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs#L445).

satvu:
Still working on adding more test cases but wanted to push the main components for review.

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -576,57 +563,71 @@ private void OverrideBindingName(INamedTypeSymbol attributeClass, ref string arg
                 }
             }
 
+            private bool HasCardinality(AttributeData attribute)
```
surgupta-msft, line 566: very nit - may be `SupportsCardinality()` or `IsCardinalitySupported()`

satvu:
Tests cases for assigning cardinality on a blob were already included in [previous PR](https://github.com/Azure/azure-functions-dotnet-worker/pull/1271/files#diff-1001ece56b095f3bfb83ee0ecdf227ad496cae0e330b02fe58a9abb383d07dbc) from @surgupta-msft. This PR adds failure cases + appropriate diagnostics.

test/Sdk.Generator.Tests/FunctionMetadataProviderGeneratorTests/StorageBindingTests.cs
```diff
@@ -257,6 +259,51 @@ public static IHostBuilder ConfigureGeneratedFunctionMetadataProvider(this IHost
                     expectedGeneratedFileName,
                     expectedOutput);
             }
+
+            [Fact]
+            public async void TestInvalidBlobCardinalityMany()
```
liliankasem, line 264: Can we add another test for IsBatched false but with List<string>? I would expect that to fail as well.

And a valid scenario test would also be good to have e.g.

IsBatched = false or not set / string -> no error
IsBatched = true / IList<string> - no error

docs/analyzer-rules/AZFW0008.md
```diff
@@ -8,17 +8,19 @@
...
+All Input Bindings and Triggers must correctly declare a compatible "IsBatched" value and parameter input type. For example, for EventHubs triggers where `IsBatched = true`, the input parameter type must be an iterable collection. Combining `IsBatched = true` with a parameter of `string[]` is valid, but combining `IsBatched = true` with a parameter of `string` is invalid.
+
+_**Note:**_ The default value of IsBatched changes depending on the Input Binding or Trigger type. The default value of `IsBatched` for each type can be found in the dotnet-isolated [extensions libraries](https://github.com/Azure/azure-functions-dotnet-worker/tree/main/extensions). If a value is not explicitly provided in the named arguments of the attribute, the default value will be used.
 
 ## How to fix violations
 
-If you have an EventHubs trigger where `IsBatched = true` (it is true by default), make sure your input parameter type is an iterable collection.
+If you have an EventHubs trigger where `IsBatched = true` (explicitly or by default), make sure your input parameter type is an iterable collection.
```
liliankasem, line 23: It could also be a blob trigger or blob input

docs/analyzer-rules/AZFW0008.md
```diff
@@ -1,4 +1,4 @@
-# AZFW0008: Invalid EventHubs Trigger
```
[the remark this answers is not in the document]
liliankasem: As I see, you're making it more generic and not specific to eventhub - got it

docs/analyzer-rules/AZFW0008.md
satvu: Ah forgot to generalize this section thx for catching

docs/analyzer-rules/AZFW0008.md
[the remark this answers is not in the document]
satvu: The invalid EH trigger is captured in the cardinality - this was created b/c I mistook what extensions had cardinality or not and didn't generalize this error/validation in the initial iteration.

test/Sdk.Generator.Tests/FunctionMetadataProviderGeneratorTests/StorageBindingTests.cs
satvu: Both of those valid scenarios for Blob were captured in Surbhi's PR (test cases above in this same file). I wrote some but erased them after I realized they were already there the next day xD. Similar scenarios are also completely captured in the EH tests.

I haven't written tests for the scenario of `IsBatched = false` with an iterable collection because I don't believe the legacy version checks for that (or has tests for it). [Here](https://github.com/Azure/azure-functions-dotnet-worker/blob/0fd7bf6aef005e4b8a14874506bf7a8ad7ad73ef/sdk/Sdk/FunctionMetadataGenerator.cs#L522) is the section in the legacy generator that deals with cardinality. 

Let me check if this is a valid scenario or something we should fail - if it's the latter, I can create issues for both the legacy generator and this one and follow up in another PR.

satvu: Offline discussion conclusion:

If a queue message has a string like "['a','b']" -- we can still send that to a user as an IEnumerable<string>. Even though there's a single message, a function can take an IEnumerable.

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -576,57 +563,71 @@ private void OverrideBindingName(INamedTypeSymbol attributeClass, ref string arg
...
+                return false;
+            }
+
             /// <summary>
-            /// This method verifies that an EventHubsTrigger matches our expectations on cardinality (isBatched property). If isBatched is set to true, the parameter with the
+            /// This method verifies that a binding that has Cardinality (isBatched property) is valid. If isBatched is set to true, the parameter with the
             /// attribute must be an enumerable type.
             /// </summary>
-            private bool IsEventHubsTriggerValid(IParameterSymbol parameterSymbol, TypeSyntax? parameterTypeSyntax, SemanticModel model, AttributeData attribute, out DataType dataType, out Cardinality cardinality)
+            private bool IsCardinalityValid(IParameterSymbol parameterSymbol, TypeSyntax? parameterTypeSyntax, SemanticModel model, AttributeData attribute, out DataType dataType)
```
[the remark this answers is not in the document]
satvu: Similar to the legacy generator [here](https://github.com/Azure/azure-functions-dotnet-worker/blob/584cab006b82e0cbc2b95a8b66e2546d3a8ddfe4/sdk/Sdk/FunctionMetadataGenerator.cs#LL546C36-L546C36), for iterable collections, we also have to find the core data type while verifying that the type is an iterable collection.

docs/analyzer-rules/AZFW0008.md
```diff
@@ -8,17 +8,19 @@
 
 ## Cause
 
-This rule is triggered when the EventHubs trigger of an Azure Function is invalid.
+This rule is triggered when an Input Binding or Trigger of an Azure Function has an invalid Cardinality.
```
liliankasem, line 11: nit: Noticed this throughout the code and docs, trigger is also a binding technically so this makes a little more sense to me

```suggestion
This rule is triggered when an Input or Trigger Binding of an Azure Function has an invalid Cardinality.
```

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -210,22 +209,19 @@ private bool TryGetParameterInputAndTriggerBindings(MethodDeclarationSyntax meth
...
-                            // There are two special cases we need to handle: HttpTrigger and EventHubsTrigger.
                             if (SymbolEqualityComparer.Default.Equals(attribute.AttributeClass, Compilation.GetTypeByMetadataName(Constants.Types.HttpTriggerBinding)))
                             {
                                 hasHttpTrigger = true;
                             }
-                            else if (SymbolEqualityComparer.Default.Equals(attribute.AttributeClass, Compilation.GetTypeByMetadataName(Constants.Types.EventHubsTrigger)))
+
+                            DataType dataType = GetDataType(parameterSymbol.Type);
+
+                            if(IsCardinalitySupported(attribute))
```
liliankasem, line 220:
```suggestion
                            if (IsCardinalitySupported(attribute))
```

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -576,67 +563,95 @@ private void OverrideBindingName(INamedTypeSymbol attributeClass, ref string arg
...
             }
 
+            private bool IsCardinalitySupported(AttributeData attribute)
+            {
+                var attrClass = attribute.AttributeClass;
+                var isBatchedProp = attrClass!
+                    .GetMembers()
+                    .SingleOrDefault(m => string.Equals(m.Name, Constants.FunctionMetadataBindingProps.IsBatchedKey, StringComparison.OrdinalIgnoreCase));
+
+                if (isBatchedProp != null)
```
liliankasem, line 573:
```suggestion
               return isBatchedProp != null ? true : false;
```

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -576,67 +563,95 @@ private void OverrideBindingName(INamedTypeSymbol attributeClass, ref string arg
...
+                // When "IsBatched" is not a named arg, we have to check the default value
+                if (!cardinalityIsNamedArg)
                 {
-                    dataType = GetDataType(parameterSymbol.Type);
-                    cardinality = Cardinality.One;
-                    return true;
+                    var attrClass = attribute.AttributeClass;
+                    var isBatchedProp = attrClass!
+                        .GetMembers()
+                        .SingleOrDefault(m => string.Equals(m.Name, Constants.FunctionMetadataBindingProps.IsBatchedKey, StringComparison.OrdinalIgnoreCase));
```
liliankasem, line 613: This code block exists a couple times I think? Might be worth it's own private method

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -576,67 +563,95 @@ private void OverrideBindingName(INamedTypeSymbol attributeClass, ref string arg
...
+                    
+                    if(isBatchedProp is null)
```
liliankasem, line 615:
```suggestion
                    if (isBatchedProp is null)
```

liliankasem (approved):

docs/analyzer-rules/AZFW0008.md
liliankasem: + the debug log!

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -576,67 +563,90 @@ private void OverrideBindingName(INamedTypeSymbol attributeClass, ref string arg
...
+            }
+
+            private bool TryGetIsBatchedProp(AttributeData attribute, out ISymbol? isBatchedProp)
+            {
+                var attrClass = attribute.AttributeClass;
+                isBatchedProp = attrClass!
+                    .GetMembers()
+                    .SingleOrDefault(m => string.Equals(m.Name, Constants.FunctionMetadataBindingProps.IsBatchedKey, StringComparison.OrdinalIgnoreCase));
+
+                return isBatchedProp != null ? true : false;
```
kshyju, line 578: nit; Can remove `? true : false;`

sdk/Sdk.Generators/FunctionMetadataProviderGenerator.Parser.cs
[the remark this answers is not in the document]
kshyju: so the value set in 218 will be overwritten? Ex: If the type was List<MyPoco> it will be first List<T> and then this out param set it to MyPoco? Should we use 2 separate variables for those?

PR #1408, opened 2023-03-09
Add query as property to HttpRequestData

### LINKED ISSUE #502: How  to query HttpRequestData .net 5  (Query) gone!
Hi
This is not an issue at all, but rather a question on how to do something 

In the "old" .net core we could do  ]

    ` var myfield =req.Query["MyField"]`

How do you do it using HttpRequestData now?

Hopefully it will be a quick straight answer 

thanks

madelinegordon:

resolves #502 

### Pull request checklist

### DIFF
--- a/src/DotNetWorker.Core/Http/HttpRequestData.cs
+++ b/src/DotNetWorker.Core/Http/HttpRequestData.cs
@@ -3,8 +3,10 @@
 
 using System;
 using System.Collections.Generic;
+using System.Data;
 using System.IO;
 using System.Security.Claims;
+using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.Azure.Functions.Worker.Http
 {
@@ -62,5 +64,10 @@ namespace Microsoft.Azure.Functions.Worker.Http
         /// </summary>
         /// <returns>The response instance.</returns>
         public abstract HttpResponseData CreateResponse();
+
+        /// <summary>
+        /// Gets the <see cref="Query"/> containing the request query. 
+        /// </summary>
+        public abstract IDictionary<string, StringValues> Query { get; }
     }
 }
--- a/src/DotNetWorker.Core/Http/HttpRequestData.cs
+++ b/src/DotNetWorker.Core/Http/HttpRequestData.cs
@@ -4,10 +4,8 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Specialized;
-using System.Data;
 using System.IO;
 using System.Security.Claims;
-using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.Azure.Functions.Worker.Http
 {
--- a/src/DotNetWorker.Core/Http/HttpRequestData.cs
+++ b/src/DotNetWorker.Core/Http/HttpRequestData.cs
@@ -68,7 +68,7 @@ namespace Microsoft.Azure.Functions.Worker.Http
         public abstract HttpResponseData CreateResponse();
 
         /// <summary>
-        /// Gets the <see cref="Query"/> containing the request query. 
+        /// Gets the <see cref="NameValueCollection"/> containing the request query. 
         /// </summary>
         public virtual NameValueCollection Query => _query ??= HttpUtility.ParseQueryString(Url.Query);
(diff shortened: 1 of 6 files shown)


madelinegordon:
Open question - in the case where there are multiple queries with the same key and different values, do we want to expose the values as type `StringValues`, or as a comma-separated string? The protobuf sends that data on the request to the worker as a comma-separated string, and ASP net core exposes multiple values to the same key as a collection of separate strings.

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
```diff
@@ -100,6 +102,8 @@ public override IReadOnlyCollection<IHttpCookie> Cookies
 
         public override Uri Url => _url ??= new Uri(_httpData.Url);
 
+        public override IDictionary<string, StringValues> Query => _query ??= _httpData.NullableQuery.ToDictionary(q => q.Key, q => new StringValues(q.Value?.Value.ToString().Split(',')));
```
liliankasem, line 105: Do we have a test that tests out this logic?

madelinegordon: That is a good point - not one that tests multiple values for one key. I will add one.

madelinegordon:
> Open question - in the case where there are multiple queries with the same key and different values, do we want to expose the values as type `StringValues`, or as a comma-separated string? The protobuf sends that data on the request to the worker as a comma-separated string, and ASP net core exposes multiple values to the same key as a collection of separate strings.

synced with @fabiocav - we will use the System.Web.HttpUtility ParseQueryString method rather than parsing ourselves, for parity with how customer are currently parsing the query string off the Uri.

src/DotNetWorker.Core/Http/HttpRequestData.cs
```diff
@@ -62,5 +63,10 @@ public HttpRequestData(FunctionContext functionContext)
         /// </summary>
         /// <returns>The response instance.</returns>
         public abstract HttpResponseData CreateResponse();
+
+        /// <summary>
+        /// Gets the <see cref="Query"/> containing the request query. 
+        /// </summary>
+        public abstract NameValueCollection Query { get; }
```
fabiocav, line 70: Would be good to have this defined as a `virtual` method in this type with the same implementation that was added to `GrpcHttpRequestData`. That would make this simpler and remove the requirement to have derived types implement this property and figure out the proper logic.

src/DotNetWorker.Core/Http/HttpRequestData.cs
```diff
@@ -62,5 +64,16 @@ public HttpRequestData(FunctionContext functionContext)
...
         public abstract HttpResponseData CreateResponse();
+
+        /// <summary>
+        /// Gets the <see cref="Query"/> containing the request query. 
+        /// </summary>
+        public virtual NameValueCollection Query
+        {
+            get
+            {
+                return HttpUtility.ParseQueryString(Url.Query);
```
kshyju, line 75: Every time this prop is accessed, the parse method gets executed. Should we store the results of parse in local variable and return that? (basically moving the logic you have in the GrpcHttpRequestData)?

madelinegordon: I think that would make sense.

src/DotNetWorker.Core/Http/HttpRequestData.cs
fabiocav: Discussed the reasons (including breaking/impact) offline. Flagging this as resolved.

fabiocav (approved):

test/DotNetWorkerTests/OutputBindings/OutputBindingsMiddlewareTests.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Specialized;
```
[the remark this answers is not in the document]
madelinegordon: removed yesterday :)

PR #1411, opened 2023-03-09
Changed TableInputAttribute.Take type from int? to int. (#1403)

### LINKED ISSUE #1403: [Tables extension] Take attribute cannot be defined as nullable
The Take property is defined as a nullable int in the TableInputAttribute, but this is not [allowed for attributes](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/attributes?WT.mc_id=DOP-MVP-4039670#2124-attribute-parameter-types).

This means that the property cannot be used. e.g. you can't do this:
```c#
[TableInput(nameof(TablesPoco), "foo", Take = 1)]
```

Instead, it should be defined as an int, like it is for the [in-process extension](https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.webjobs.tableattribute.take?view=azure-dotnet#microsoft-azure-webjobs-tableattribute-take).

kashimiz:

resolves #1403 

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/extensions/Worker.Extensions.Tables/src/TableInputAttribute.cs
+++ b/extensions/Worker.Extensions.Tables/src/TableInputAttribute.cs
@@ -60,7 +60,7 @@ namespace Microsoft.Azure.Functions.Worker
         /// <summary>
         /// Used with filter. RowKey should be null. 
         /// </summary>
-        public int? Take
+        public int Take
         {
             get; set;
         }
--- a/extensions/Worker.Extensions.Tables/release_notes.md
+++ b/extensions/Worker.Extensions.Tables/release_notes.md
@@ -3,4 +3,4 @@
 - My change description (#PR/#issue)
 -->
 
-- Publishing Worker.Extension.Tables package v 1.0.0 (#1173)
+- Changed TableInputAttribute.Take from int? to int (#1403)


liliankasem (approved):

extensions/Worker.Extensions.Tables/src/TableInputAttribute.cs
```diff
@@ -43,31 +43,31 @@ public TableInputAttribute(string tableName, string partitionKey, string rowKey)
 
         /// <summary>When binding to a table entity, gets the partition key of the entity.</summary>
         /// <remarks>When binding to an entire table, returns <see langword="null"/>.</remarks>
-        public string? PartitionKey { get; }
```
satvu, line 46: Is the ask to also change this prop to be non-nullable? Or is it just "Take"? If so, could you also update the comments that say "returns null".

satvu:
is a version change/bump needed? will it be in this PR?

extensions/Worker.Extensions.Tables/src/TableInputAttribute.cs
kashimiz: The original issue only asks to change `int?` to `int` for `Take`, but I noticed that we had `string?` for some properties. Since `string` is a reference type (and therefore intrinsically nullable,) the `?` is unnecessary (it's intended for value types), so I removed them as cleanup.

kashimiz: Okay, reading a little further and looking at other worker extensions, it looks like we do want nullable strings in this case: [Nullable reference types and static analysis | Microsoft Learn](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-reference-types#nullable-references-and-static-analysis)

Reverting the string? changes.

satvu (approved):

brettsam (approved):

jviau:
I believe this is a breaking change (`Nullable<int>` to `int`) and will need to go into a major version rev.

PR #1413, opened 2023-03-10
switch to use FUNCTIONS_APPLICATION_DIRECTORY

brettsam:


src/DotNetWorker.Grpc/Definition/GrpcFunctionDefinition.cs
```diff
@@ -7,24 +7,26 @@
...
+        private const string FunctionsApplicationDirectoryKey = "FUNCTIONS_APPLICATION_DIRECTORY";
+
         public GrpcFunctionDefinition(FunctionLoadRequest loadRequest, IMethodInfoLocator methodInfoLocator)
         {
             EntryPoint = loadRequest.Metadata.EntryPoint;
             Name = loadRequest.Metadata.Name;
             Id = loadRequest.FunctionId;
 
-            string? scriptRoot = Environment.GetEnvironmentVariable("FUNCTIONS_WORKER_DIRECTORY");
+            string? scriptRoot = Environment.GetEnvironmentVariable(FunctionsApplicationDirectoryKey);
```
kshyju, line 26: Should we check both for timebeing (until host changes to set the new env variable is fully rolled out). That allows us to merge and release this version without waiting for host changes?

`Environment.GetEnvironmentVariable("FUNCTIONS_APPLICATION_DIRECTORY")
                ?? Environment.GetEnvironmentVariable("FUNCTIONS_WORKER_DIRECTORY");`

brettsam: Yeah, good point -- that would be a good approach. Then this will just start working once the host supports it.

kshyju (approved):

kshyju:
Do we need the patch version bump in this PR?

brettsam:
> Do we need the patch version bump in this PR?

Not here -- we'll make sure that's all done before we release though.

PR #1421, opened 2023-03-16
logging error upon invocation exception

brettsam:
Emitting a new log message if there is an exception while executing an invocation. Previously, this exception would only be sent to the host to be logged, but now it can be seen/filtered/redirected/etc straight from the worker.

This exception message will still be logged on the host, so it may be desirable to disable it on the worker. To do this, you can use any of [the logging configuration mechanisms available to you in .NET](https://learn.microsoft.com/en-us/dotnet/core/extensions/logging?tabs=command-line#configure-logging) and set the category `Microsoft.Azure.Functions.Worker.FunctionsApplication` to `LogLevel.None`.

For example:
```csharp
var host = new HostBuilder() 
    .ConfigureLogging(logging =>
    {
        logging.AddFilter("Microsoft.Azure.Functions.Worker.FunctionsApplication", LogLevel.None);
    })
    .Build();
```


brettsam:
@paulbatum, FYI

paulbatum (commented):
I am curious - is there a chance that this will translate into new exceptions in customer logs in a way that they might get the incorrect impression that the new host has broken them, when in reality the app behavior hasn't changed, just the way the errors are logged?

src/DotNetWorker.Core/FunctionsApplication.cs
```diff
@@ -81,6 +81,8 @@ public async Task InvokeFunctionAsync(FunctionContext context)
                 invokeActivity?.SetStatus(ActivityStatusCode.Error, ex.Message);
                 invokeActivity?.RecordException(ex, escaped: true);
```
jviau, line 82: I think part of the offline conversation with @lmolkova recommended removing this?

> so there is no valid reason to use RecordException if you have ILogger

If you don't want to remove it just yet / have further discussion that is fine - can leave that work separate to this PR.

jviau (approved):

brettsam:
> I am curious - is there a chance that this will translate into new exceptions in customer logs in a way that they might get the incorrect impression that the new host has broken them, when in reality the app behavior hasn't changed, just the way the errors are logged?

@paulbatum -- yes -- it's possible. There will now be an error logged here from the worker along with the "Function failed" log written by the host. Although these should be identical and come from the same invocation id. I still think this is the correct way to go -- any other ideas?

Have updated release notes to point back to this PR, which now explains how to disable this log in the worker.

src/DotNetWorker.Core/FunctionsApplication.cs
brettsam: Yep -- removed it in an updated commit and added some Activity-only tests as well (previous tests were all App Insights-based)

paulbatum:
> > I am curious - is there a chance that this will translate into new exceptions in customer logs in a way that they might get the incorrect impression that the new host has broken them, when in reality the app behavior hasn't changed, just the way the errors are logged?
> 
> @paulbatum -- yes -- it's possible. There will now be an error logged here from the worker along with the "Function failed" log written by the host. Although these should be identical and come from the same invocation id. I still think this is the correct way to go -- any other ideas?
> 
> Have updated release notes to point back to this PR, which now explains how to disable this log in the worker.

Yeah no other ideas at this time, I agree with you this is the correct way to go. Was just trying to think through any support implications of the change.

PR #1425, opened 2023-03-21
Including worker metadata & capabilities in env reload response

kshyju:
Worker side changes to address https://github.com/Azure/azure-functions-host/issues/8983.

Including worker metadata and capabilities in env reload response.

Host side changes were made in https://github.com/Azure/azure-functions-host/pull/9020

There is some changes in the native worker as well to make this work. But will keep that changes in separate PR.


src/DotNetWorker.Grpc/AppContextUtils.cs
```diff
@@ -0,0 +1,15 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.Functions.Worker.Grpc
+{
+    internal class AppContextUtils
+    {
+        /// <summary>
+        /// Returns a boolean value indicating whether the worker is running with native host(placeholder support).
```
satvu, line 11:
```suggestion
        /// Returns a boolean value indicating whether the worker is running with native host (placeholder support).
```

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -146,36 +143,13 @@ internal void InvocationCancelRequestHandler(InvocationCancel request)
...
-            }
-
-            // Add required capabilities; these cannot be modified and will override anything from WorkerOptions
-            response.Capabilities["RpcHttpBodyOnly"] = bool.TrueString;
-            response.Capabilities["RawHttpBodyBytes"] = bool.TrueString;
-            response.Capabilities["RpcHttpTriggerMetadataRemoved"] = bool.TrueString;
-            response.Capabilities["UseNullableValueDictionaryForHttp"] = bool.TrueString;
-            response.Capabilities["TypedDataCollection"] = bool.TrueString;
-            response.Capabilities["WorkerStatus"] = bool.TrueString;
+            response.Capabilities.Add(GetWorkerCapabilities(workerOptions));
```
satvu, line 152: nit: line spacing

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -284,5 +258,57 @@ internal static FunctionLoadResponse FunctionLoadRequestHandler(FunctionLoadRequ
 
             return response;
         }
+        internal static FunctionEnvironmentReloadResponse EnvReloadRequestHandler(FunctionEnvironmentReloadRequest request, WorkerOptions workerOptions)
```
satvu, line 261: nit: line spacing

src/DotNetWorker.Grpc/AppContextUtils.cs
```diff
@@ -0,0 +1,15 @@
...
+        /// </summary>
+        internal static bool IsRunningWithNativeHost() => AppContext.GetData("AZURE_FUNCTIONS_NATIVE_HOST") is not null;
```
satvu, line 13: Might be out of scope for this PR but between capabilities that already exist (and more props to come with this work + http proxying) we might want to make a constants file similar to [the host one](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Workers/Rpc/RpcWorkerConstants.cs).

satvu (approved):

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -284,5 +258,57 @@ internal static FunctionLoadResponse FunctionLoadRequestHandler(FunctionLoadRequ
...
+            };
+            workerMetadata.CustomProperties.Add("Worker.Grpc.Version", typeof(GrpcWorker).Assembly.GetName().Version?.ToString());
+
+            return workerMetadata;
+        }
+
+        private static IDictionary<string, string> GetWorkerCapabilities(WorkerOptions workerOptions)
+        {
+            var capabilities = new Dictionary<string, string>();
+            // Add additional capabilities defined by WorkerOptions
```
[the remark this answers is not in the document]
kshyju: added linebreak.

src/DotNetWorker.Grpc/GrpcWorker.cs
kshyju: added linebreak.

src/DotNetWorker.Grpc/GrpcWorker.cs
kshyju: added line break.

src/DotNetWorker.Grpc/AppContextUtils.cs
kshyju: fixed

src/DotNetWorker.Grpc/AppContextUtils.cs
liliankasem: +1 to using a const

liliankasem (approved):
Does the host change need to be released for this feature to go out? And if so, is the host version you need fully rolled out

src/DotNetWorker.Grpc/AppContextUtils.cs
fabiocav, line 11: Running native and placeholder mode are different and although our setup will run in native in placeholder mode only, that is not a requirement. I think you can drop the placeholder comment.

Suggest something like this:

```
 /// <summary>
 /// Indicates whether the worker is running in the native host.
 /// </summary>
 /// <returns><see langword="true"/> if the worker is running in the native host; otherwise,<see langword="false"/>.</returns>
```

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -284,5 +258,57 @@ internal static FunctionLoadResponse FunctionLoadRequestHandler(FunctionLoadRequ
...
         }
+        internal static FunctionEnvironmentReloadResponse EnvReloadRequestHandler(FunctionEnvironmentReloadRequest request, WorkerOptions workerOptions)
+        {
+            var envReloadResponse = new FunctionEnvironmentReloadResponse
+            {
+                Result = new StatusResult { Status = StatusResult.Types.Status.Success }
+            };
+
+            if (AppContextUtils.IsRunningWithNativeHost())
+            {
```
fabiocav, line 269: I'm concerned about this logic here. This layer shouldn't behave differently based on whether we're running in the native worker. The behavior here should be consistent and all of the differences scoped to the native client instead. We want to make sure we don't have discrepancies between the two models.

fabiocav (changes requested):
Can we discuss the need for this changes? The request/response handling behavior should be the same and the gRPC worker ideally shouldn't be aware of the native host.

src/DotNetWorker.Grpc/GrpcWorker.cs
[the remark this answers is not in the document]
kshyju: Renamed as suggested.

src/DotNetWorker.Grpc/AppContextUtils.cs
kshyju: Removed this helper in the new iteration based on the other feedback about conditionally setting those props.

src/DotNetWorker.Grpc/AppContextUtils.cs
kshyju: Removed this helper.

kshyju:
> Does the host change need to be released for this feature to go out? And if so, is the host version you need fully rolled out

No. The host gracefully handles this situation.

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -284,5 +259,54 @@ internal static FunctionLoadResponse FunctionLoadRequestHandler(FunctionLoadRequ
...
+                Result = new StatusResult { Status = StatusResult.Types.Status.Success },
+                WorkerMetadata = GetWorkerMetadata()
+            };
+            envReloadResponse.Capabilities.Add(GetWorkerCapabilities(workerOptions));
+
+            return envReloadResponse;
+        }
+
+        private static WorkerMetadata GetWorkerMetadata()
```
kshyju, line 275: content of this method is existing code. Moved to this method so that it can be called from worker init handler and env reload response handler.

src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -284,5 +259,54 @@ internal static FunctionLoadResponse FunctionLoadRequestHandler(FunctionLoadRequ
...
+                RuntimeVersion = Environment.Version.ToString(),
+                WorkerVersion = WorkerInformation.Instance.WorkerVersion,
+                WorkerBitness = RuntimeInformation.ProcessArchitecture.ToString()
+            };
+            workerMetadata.CustomProperties.Add("Worker.Grpc.Version", typeof(GrpcWorker).Assembly.GetName().Version?.ToString());
+
+            return workerMetadata;
+        }
+
+        private static IDictionary<string, string> GetWorkerCapabilities(WorkerOptions workerOptions)
```
kshyju, line 291: content of this method is existing code. Moved to this method so that it can be called from worker init handler and env reload response handler.

liliankasem (approved):

PR #1426, opened 2023-03-22
Add E2E tests for Cosmos extension

liliankasem:
Adding E2E tests to test out function bindings that use the deferring binding feature.

This PR is blocked on fixing the Cosmos emulator for testing.


test/E2ETests/E2ETests/Helpers/HttpHelpers.cs
```diff
@@ -19,7 +19,7 @@ public static async Task<HttpResponseMessage> InvokeHttpTrigger(string functionN
             return await GetResponseMessage(request);
         }
 
-        public static async Task<HttpResponseMessage> InvokeHttpTriggerWithBody(string functionName, string body, HttpStatusCode expectedStatusCode, string mediaType, int expectedCode = 0)
```
aishwaryabh, line 22: Is there a reason we don't have to have expectedStatusCode checked anymore?

aishwaryabh (approved):

test/E2ETests/E2ETests/Helpers/HttpHelpers.cs
liliankasem: `expectedStatusCode` was never used in the method so I just removed it

PR #1436, opened 2023-03-23
Forwarding env reload request to worker(customer app payload) after it is loaded

kshyju:
Forwarding env reload request to worker(customer app payload) after it is loaded.

We call [NativeMethods.RegisterCallbacks ](https://github.com/Azure/azure-functions-dotnet-worker/blob/2b78e7b03503f499ae71507b2a842d1848b86fb2/src/DotNetWorker.Grpc/NativeHostIntegration/NativeWorkerClient.cs#L36)when we start the Native worker client. 

On the native side, this invokes [SetCallbackHandles](https://github.com/Azure/azure-functions-dotnet-worker/blob/2b78e7b03503f499ae71507b2a842d1848b86fb2/host/src/funcgrpc/nativehostapplication.cpp#L70). This gets called only once. I am using the call to this method as a signal that worker app (customer payload) was loaded successfully and is ready to take any other streaming messages.  

Once #1425 is also merged, env reload request will be properly handled by the worker even in placeholder/specialiazation use case. Currently we are sending a dummy response (with status = "success") from native host.

### DIFF
--- a/host/src/funcgrpc/handlers/funcgrpc_native_handler.cpp
+++ b/host/src/funcgrpc/handlers/funcgrpc_native_handler.cpp
@@ -93,7 +93,7 @@ void AzureFunctionsRpc::NativeHostMessageHandler::HandleMessage(ByteBuffer *rece
                     application_->ExecuteApplication(exePath);
                 }
 
-                FUNC_LOG_INFO("Going to wait for worker payload to load.");
+                FUNC_LOG_INFO("Waiting for worker initialization.");
                 std::unique_lock lk(application_->mtx_workerLoaded);
                 application_->cv_workerLoaded.wait(lk, [this] { return application_->hasWorkerLoaded; });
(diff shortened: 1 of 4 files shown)


host/src/funcgrpc/handlers/funcgrpc_native_handler.cpp
```diff
@@ -93,15 +93,12 @@ void AzureFunctionsRpc::NativeHostMessageHandler::HandleMessage(ByteBuffer *rece
                     application_->ExecuteApplication(exePath);
                 }
 
-                StreamingMessage streamingMsg;
-                streamingMsg.mutable_function_environment_reload_response()->mutable_result()->set_status(
-                    AzureFunctionsRpcMessages::StatusResult::Success);
+                FUNC_LOG_INFO("Going to wait for worker payload to load.");
```
fabiocav, line 96: nit: perhaps something like `"Waiting for worker initialization."

PR #1309, opened 2023-02-03
[Source Gen] Generate function executor code

kshyju:
Fixes #1284 

This source generator creates an implementation of `IFunctionExecutor`. Our [current implementation of `IFunctionExecutor`](https://github.com/Azure/azure-functions-dotnet-worker/blob/storage-extension-5.0.1/src/DotNetWorker.Core/Invocation/DefaultFunctionExecutor.cs#L25) relies on reflection at runtime (so it has an impact on coldstart). We have tried a [hard coded version](https://github.com/kshyju/FunctionApp26/blob/fabiocav/tweaks/FunctionApp26/DirectFunctionExecutor.cs) in our coldstart test environment and it proved to improve cold start numbers since no reflection code is being executed. In this PR, the source gen creates an implementation which is similar to the one linked above. We rely on IFunctionActivator in this version to create an instance of the function class.

Generator needs to be enabled explicitly by using ` <FunctionsEnableExecutorSourceGen>True</FunctionsEnableExecutorSourceGen>` in project file (of the function app)

PR #1453, opened 2023-04-05
[Source Gen] Cleanup 

satvu:

General cleanup following refactoring examples set in #1309 

### Pull request checklist

### Additional information

Additional PR information


sdk/Sdk.Generators/KnownTypes.cs
```diff
@@ -14,12 +18,34 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
...
+        internal readonly INamedTypeSymbol DictionaryGeneric;
+
         internal KnownTypes(Compilation compilation)
         {
             TaskType = compilation.GetTypeByMetadataName(typeof(Task).FullName)!;
             TaskOfTType = compilation.GetTypeByMetadataName(typeof(Task<>).FullName)!;
             ValueTaskType = compilation.GetTypeByMetadataName(typeof(ValueTask).FullName)!;
             ValueTaskOfTTypeOpt = compilation.GetTypeByMetadataName(typeof(ValueTask<>).FullName)!;
+
+            IEnumerable = compilation.GetTypeByMetadataName(typeof(IEnumerable).FullName)!;
```
satvu, line 39: Not sure if valid to assume these all won't be null.

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/KnownFunctionMetadataTypes.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    internal readonly struct KnownFunctionMetadataTypes
+    {
+        internal readonly INamedTypeSymbol? BindingAttribute;
+        internal readonly INamedTypeSymbol? OutputBindingAttribute;
+        internal readonly INamedTypeSymbol? FunctionName;
+        internal readonly INamedTypeSymbol? BindingPropertyNameAttribute;
+        internal readonly INamedTypeSymbol? DefaultValue;
+        internal readonly INamedTypeSymbol? HttpResponse;
+        internal readonly INamedTypeSymbol? HttpTriggerBinding;
```
satvu, line 16: Nullable since customer may not be using http extension package. The 5 non-Http types above should be present in any functions app (as part of abstractions package) and we could throw an error when initializing if it's `null` and make these props non nullable. Would have to add error handling, but as is, these props are used in Symbol comparisons in `FunctionMetadataProviderGenerator.Parser` and returning `null` is still fine.

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -721,18 +726,17 @@ private DataType GetDataType(ITypeSymbol symbol)
...
             {
-                return SymbolEqualityComparer.Default.Equals(symbol, Compilation.GetTypeByMetadataName(Constants.Types.String))
-                    || (symbol is IArrayTypeSymbol arraySymbol && SymbolEqualityComparer.Default.Equals(arraySymbol.ElementType, Compilation.GetTypeByMetadataName(Constants.Types.String)));
+                return SymbolEqualityComparer.Default.Equals(symbol, _knownTypes.StringType)
+                    || (symbol is IArrayTypeSymbol arraySymbol && SymbolEqualityComparer.Default.Equals(arraySymbol.ElementType, _knownTypes.StringType));
             }
 
             private bool IsBinaryType(ITypeSymbol symbol)
             {
-                var isByteArray = SymbolEqualityComparer.Default.Equals(symbol, Compilation.GetTypeByMetadataName(Constants.Types.ByteArray))
```
surgupta-msft, line 730: To understand this PR better, any reason why we were not doing something like below directly here?

`compilation.GetTypeByMetadataName(typeof(byte[]).FullName)!`

satvu: That sample snippet you shared is what is being done inside of the `KnownTypes` class

sdk/Sdk.Generators/KnownTypes.cs
jviau, line 39: I would slightly prefer all the fields/properties here to be lazily initialized. This way we only read from `compilation` the types we actually use as the generator executes.

jviau: They are all part of the BCL, so I think it is safe.

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/KnownFunctionMetadataTypes.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.CodeAnalysis;
+
+namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
+{
+    internal readonly struct KnownFunctionMetadataTypes
+    {
+        internal readonly INamedTypeSymbol? BindingAttribute;
```
jviau, line 10: same suggestion here: would be slightly preferable to have these all be lazily initialized properties. Also, I recommend exposing properties and not fields - even when they are internal.

jviau (approved):

liliankasem (approved):

PR #1454, opened 2023-04-07
Fix HttpTriggers with POCO deserialization not working

### LINKED ISSUE #164: HttpTrigger with complex type does not work
Using the below HttpTrigger signature with preview 4.

`[FunctionName(nameof(CreateTodosList))]`
`public async Task<HttpResponseData> CreateTodosList([HttpTrigger(AuthorizationLevel.Anonymous, "post", Route = "todolists")] `
            `TodoList todoList,`
            `HttpRequestData req,`
            `FunctionExecutionContext executionContext)`
`{`
``
`}`

TodoList instance is always null. Even HttpRequestData instance is null.

This signature works fine with .NET core 3.1 solution with HttpRequest.

njqdev:

resolves #164

### Pull request checklist

### Additional information
* Fixes POCO deserialization
* Populates HttpRequestData if it’s a secondary parameter
* Resets the request body stream position to allow `ReadAsStringAsync` and similar methods
* Moved `TestHttpRequestData.cs`

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -5,3 +5,4 @@
 
 - Logging an error on a function invocation exception. This error will also be logged by the host, but this allows for richer telemetry coming directly from the worker. See #1421 for details on how to disable this log if desired.
 - Add query as property to HttpRequestData (#1408)
+- Enable HttpTriggers with object deserialization and a secondary HttpRequestData parameter (#1454)
--- a/src/DotNetWorker.Core/Context/Features/DefaultFunctionInputBindingFeature.cs
+++ b/src/DotNetWorker.Core/Context/Features/DefaultFunctionInputBindingFeature.cs
@@ -56,9 +56,11 @@ namespace Microsoft.Azure.Functions.Worker.Context.Features
                     FunctionParameter param = context.FunctionDefinition.Parameters[i];
 
                     // Check InputData first, then TriggerMetadata
-                    if (!functionBindings.InputData.TryGetValue(param.Name, out object? source))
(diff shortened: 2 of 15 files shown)


fabiocav:
@njqdev thanks for sending this!

Could you scope the PR to the POCO deserialization change only? The approach for the loose parameter binding in general is fragile and in general, the implicit model (even if with a more specific target type like `HttpRequestData`) leads to issues we're trying to avoid in the Worker model.

I actually have work in progress for POCO support [here](https://github.com/Azure/azure-functions-dotnet-worker/tree/fabiocav/httpbindingfix) with the explicit model that will also work with the ASP.NET Core abstractions once the integration is in place. Please take a look as that would also address what the issue this change is tracking. That uses an explicit model as the binding target enabling you to specify the binding source (body, query string, form, etc.)

njqdev:
@fabiocav thanks for the info. Looks like you have a better solution in that case. And binding both parameters is also possible with your changes ie.

```cs
        public static HttpResponseData POCOAndHttpRequest(
            [HttpTrigger(AuthorizationLevel.Anonymous, "post", Route = null)] HttpRequestData req,
            [FromBody] Book book)
```

If you want I could adjust and gift you some of the E2E tests. But either way I assume this should be closed right?

fabiocav:
> If you want I could adjust and gift you some of the E2E tests. But either way I assume this should be closed right?

It is, other items have taken priority, otherwise I'd have that out by now. I would seriously love the help with the E2E tests :) . I'll try to have the PR open this week and that would make it easier to iterate. Thank you!

PR #1462, opened 2023-04-11
Implement bypass deferred binding

### LINKED ISSUE #1495: [SDK Binding] Implement the design to bypass deferred binding
Implement the design to bypass deferred binding for those scenarios where we want the extension to continue to handle it on the host side.

Design doc - [Link](https://dev.azure.com/msazure/One/_git/AAPT-Antares-Docs?path=/TeamDocs/FunctionTeamDocs/Design/OOPWorkers/SdkTypeBindings/BypassDeferredBinding.md&version=GBmaster&_a=preview)

surgupta-msft:

Resolves #1495

#### Implementation Summary

1. Binding Attributes are updated as per design doc - ([Refer section: class design](https://dev.azure.com/msazure/One/_git/AAPT-Antares-Docs?path=/TeamDocs/FunctionTeamDocs/Design/OOPWorkers/SdkTypeBindings/BypassDeferredBinding.md&version=GBmaster&_a=preview&anchor=3.2-class-design))

      - Binding Attributes updated - `BlobInputAttribute`, `BlobTriggerAttribute`, `CosmosDBInputAttribute`
      - Converters updated - `BlobStorageConverter`, `CosmosDBConverter`
      - New Attributes added - `AllowConverterFallback`, `SupportedConverterTypes`
      ```csharp
        // Binding Attributes to advertise all the converters they support and enable/disable fallback to registered converters
        [AllowConverterFallback]        // Attribute indicates that fallback to default converters is enabled by BindingAttribute
        [InputConverter(typeof(MyConverter1))] // Attribute to indicate the converter supported by BindingAttribute 
        [InputConverter(typeof(MyConverter2))]
        class MyTriggerAttribute // or MyInputAttribute
        {   
        
        }

        // Converters to advertise the support for deferred binding and types that are supported
        [SupportsDeferredBinding]        // Attribute indicates that converter supports deferred binding
        [SupportedConverterTypes(typeof(MyType1))]  // Attribute to indicate the type supported by converter
        [SupportedConverterTypes(typeof(MyType2))]
        [SupportedConverterTypes(typeof(IEnumerable<MyType2>))]
        class MyConverter
        {
        
        }
        ```
2. Changes in Function Metadata Generator include - 
    - Logic to correctly assign `SupportsDeferredBinding` by analyzing attributes advertised by `BindingAttributes` and `Converters`
    - Handling for cases such as POCOs, Collections - ([Refer section: Metadata Generation flow](https://dev.azure.com/msazure/One/_git/AAPT-Antares-Docs?path=/TeamDocs/FunctionTeamDocs/Design/OOPWorkers/SdkTypeBindings/BypassDeferredBinding.md&version=GBmaster&_a=preview&anchor=3.3.1--metadata-generation-flow))

3. Changes in `GrpcFunctionDefinition.cs` -
     - This class will cache information about Binding Attributes and Converters in `PropertyBags` during Function indexing time to avoid type evaluation during invocation time.
     - Information cached here -
          - Whether BindingAttribute supports fallback to registered converters
          - A Dictionary with Key as `converter` and value as Properties for that converter such as all the types supported by it and support for Json Deserializable objects.

4. Invocation flow - ([Refer section: Invocation flow](https://dev.azure.com/msazure/One/_git/AAPT-Antares-Docs?path=/TeamDocs/FunctionTeamDocs/Design/OOPWorkers/SdkTypeBindings/BypassDeferredBinding.md&version=GBmaster&_a=preview&anchor=3.3.2-invocation-flow))
     - Changes in `DefaultInputConversionFeature.cs`
          - Get list of all the converters advertised by BindingAttribute and hence all the types supported by each converter.
          - Check if `TargetType` is supported by the converter before invoking it.
          - Check if Fallback to registered converters is enabled by Binding attribute. If yes. continue with default probing.

### Pull request checklist

### Additional information

Additional PR information


extensions/Worker.Extensions.Storage.Blobs/src/BlobInputAttribute.cs
```diff
@@ -1,11 +1,12 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using Microsoft.Azure.Functions.Worker.Converters;
 using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
 
 namespace Microsoft.Azure.Functions.Worker
 {
-    [SupportsDeferredBinding]
+    [InputConverter(disableConverterFallback: false, typeof(BlobStorageConverter))]
```
liliankasem, line 9: I think I'd like to to find a better name for this (disableConverterFallback ) parameter. The purpose of it is to enable or disable using the builtin converters right? So maybe `UseBuiltinConverters`? or something along those lines?

I think in terms of ease of usability, can we default it to false? It would be nice to also not require a bool value; so for example:

`[InputConverter(typeof(BlobStorageConverter))]`

^ does not use builtin converters

`[InputConverter(UseBuiltinConverters, typeof(BlobStorageConverter))]`

^ would use builtin converters

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -14,12 +14,25 @@
...
+using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
 
 namespace Microsoft.Azure.Functions.Worker
 {
     /// <summary>
     /// Converter to bind Blob Storage type parameters.
     /// </summary>
+    [SupportsDeferredBinding]
+    [SupportedConverterTypes(
+        typeof(string), typeof(IEnumerable<string>), typeof(string[]),
```
JoshLove-msft, line 26: What happens if the user binds some parameters to supported converter types, and others to unsupported types? I'm thinking of a scenario where a user binds to `ServiceBusReceivedMessage` but also binds to a various string properties. Curious if this is a supported scenario.

surgupta-msft: Can you share a basic function with the scenario you are mentioning to make sure I understand it correctly? I think it should bind per parameter, but I will double check.

surgupta-msft: For the sample blob function below, Stream supports deferred binding while BinaryData does not support it. This will work.

```csharp
        [Function(nameof(ExpressionFunction))]
        public void Run(
            [BlobTrigger("expression-trigger")] Stream book,
            [BlobInput("input-container/sample1.txt")] BinaryData myBlob,
            FunctionContext context)
        {
            _logger.LogInformation("Trigger content: {content}", book);
            _logger.LogInformation("Blob content: {content}", myBlob);
        }
```

```json
   "bindings": [
      {
        "name": "book",
        "direction": "In",
        "type": "blobTrigger",
        "path": "expression-trigger",
        "properties": {
          "supportsDeferredBinding": "True"
        }
      },
      {
        "name": "myBlob",
        "direction": "In",
        "type": "blob",
        "blobPath": "input-container/sample1.txt",
        "cardinality": "One",
        "properties": {}
      }
    ]
 
```

JoshLove-msft: Sounds good - in the example I was thinking of there would only be one attribute, but multiple parameters, e.g.:
```c#
public void Run(
            [ServiceBusTrigger("expression-trigger")] ServiceBusReceivedMessage message,
            string MessageId,
            string LockToken,
            FunctionContext context)
```

surgupta-msft: It's a good scenario. For Blob, I have a similar function handy. In case of below function, `Stream` is the binding type and `name` is part of the expression. Irrespective of the type of `name`, if Stream supports deferred binding, it will get a `true` in metadata, otherwise `false`.

```csharp
        [Function(nameof(BlobStreamFunction))]
        public async Task BlobStreamFunction([BlobTrigger("stream-trigger/{name}")] Stream stream, string name)
```

I am interested to know if we can infer similarly for ServiceBus or not. Meaning `message` will be considered as Binding type and `messageId`, `lockToken` are part of expression? Let me know if this not as you expected. We can definitely discuss this offline as well. Thanks for bringing this up.

JoshLove-msft: The other parameters are not part of an expression. Instead they are intended to bind to properties of the message. I think this scenario is probably unusual because users can get the same properties from the ServiceBusReceivedMessage type itself. I mostly was just curious as to what would happen in this case - I think it is reasonable if we say it is not supported.

surgupta-msft: Had a quick chat with Fabio on this. There will be an upcoming change for custom binding sources where such scenarios will be supported. Binding source will help identifying where is the parameter coming from. For example, is it coming from body/query/header and so on. Will share if I find any docs related to it.

samples/WorkerBindingSamples/WorkerBindingSamples.csproj
```diff
@@ -7,12 +7,16 @@
...
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.9.0-preview1-local" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
     <PackageReference Include="System.Net.NameResolution" Version="4.3.0" />
+  </ItemGroup>
+    <ItemGroup>
+    <ProjectReference Include="..\..\extensions\Worker.Extensions.Abstractions\src\Worker.Extensions.Abstractions.csproj" />
+    <ProjectReference Include="..\..\extensions\Worker.Extensions.Http\src\Worker.Extensions.Http.csproj" />
+    <ProjectReference Include="..\..\extensions\Worker.Extensions.Storage\src\Worker.Extensions.Storage.csproj" />
+    <ProjectReference Include="..\..\src\DotNetWorker.ApplicationInsights\DotNetWorker.ApplicationInsights.csproj" />
+    <ProjectReference Include="..\..\src\DotNetWorker\DotNetWorker.csproj" />
```
liliankasem, line 19: Are these changes for testing or permenant?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -14,12 +14,24 @@
...
+    [SupportsJsonDeserialization]
+    [SupportedConverterTypes(typeof(string), supportsCollection: true)]
+    [SupportedConverterTypes(typeof(Stream), supportsCollection: true)]
+    [SupportedConverterTypes(typeof(byte[]), supportsCollection: true)]
+    [SupportedConverterTypes(typeof(BlobBaseClient), supportsCollection: true)]
+    [SupportedConverterTypes(typeof(BlobClient), supportsCollection: true)]
+    [SupportedConverterTypes(typeof(BlockBlobClient), supportsCollection: true)]
+    [SupportedConverterTypes(typeof(PageBlobClient), supportsCollection: true)]
+    [SupportedConverterTypes(typeof(AppendBlobClient), supportsCollection: true)]
+    [SupportedConverterTypes(typeof(BlobContainerClient), supportsCollection: true)]
```
liliankasem, line 34: I know we technically support this, but its not a valid scenario

samples/WorkerBindingSamples/WorkerBindingSamples.csproj
surgupta-msft: only for testing

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
surgupta-msft: I did not follow. Which scenario are we talking about?

surgupta-msft: ohh got it, you mean `BlobContainerClient`. I will remove it.

extensions/Worker.Extensions.Storage.Blobs/src/BlobInputAttribute.cs
surgupta-msft: I am with you on the naming issue. I think the intent is that by default all converters will support fallback to default converters. But we will give option to the converter to disable fallback using this flag. We can use name `DisableFallbackToBuiltinConverters` but I am not sure if we want to refer them as `default` or `builtIn` converters? We certainly need to brainstorm on naming at various places. Will wait for the design to get finalized and come back to it in case there is any significant change in flow.

extensions/Worker.Extensions.CosmosDB/src/CosmosDBTriggerAttribute.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 ﻿using System;
+using Microsoft.Azure.Functions.Worker.Converters;
```
liliankasem, line 5: nit: is this needed?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -15,12 +15,40 @@
...
+    [SupportedConverterType(typeof(BlockBlobClient))]
+    [SupportedConverterType(typeof(BlockBlobClient[]))]
+    [SupportedConverterType(typeof(IEnumerable<BlockBlobClient>))]
+    [SupportedConverterType(typeof(PageBlobClient))]
+    [SupportedConverterType(typeof(PageBlobClient[]))]
+    [SupportedConverterType(typeof(IEnumerable<PageBlobClient>))]
+    [SupportedConverterType(typeof(AppendBlobClient))]
+    [SupportedConverterType(typeof(AppendBlobClient[]))]
+    [SupportedConverterType(typeof(IEnumerable<AppendBlobClient>))]
+    [SupportedConverterType(typeof(BlobContainerClient))]
```
liliankasem, line 51: We should make a new GH issue to break this converter out into one converter per type

samples/WorkerBindingSamples/WorkerBindingSamples.csproj
```diff
@@ -1,29 +1,31 @@
...
+    <PropertyGroup>
+        <TargetFramework>net7.0</TargetFramework>
+        <AzureFunctionsVersion>v4</AzureFunctionsVersion>
+        <OutputType>Exe</OutputType>
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
+    <ItemGroup>
+        <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.9.0-preview1" />
+        <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
```
liliankasem, line 11: Do we need the newtonsoft import?

samples/WorkerBindingSamples/WorkerBindingSamples.csproj
```diff
@@ -1,29 +1,31 @@
...
+        <PackageReference Include="System.Net.NameResolution" Version="4.3.0" />
+    </ItemGroup>
+    <ItemGroup>
+        <ProjectReference Include="..\..\extensions\Worker.Extensions.Abstractions\src\Worker.Extensions.Abstractions.csproj" />
+        <ProjectReference Include="..\..\extensions\Worker.Extensions.Http\src\Worker.Extensions.Http.csproj" />
+        <ProjectReference Include="..\..\extensions\Worker.Extensions.Storage\src\Worker.Extensions.Storage.csproj" />
+        <ProjectReference Include="..\..\src\DotNetWorker.ApplicationInsights\DotNetWorker.ApplicationInsights.csproj" />
+        <ProjectReference Include="..\..\src\DotNetWorker\DotNetWorker.csproj" />
+        <ProjectReference Include="..\..\extensions\Worker.Extensions.CosmosDB\src\Worker.Extensions.CosmosDB.csproj" />
```
liliankasem, line 20: nit: order imports so that all extensions are together

samples/WorkerBindingSamples/WorkerBindingSamples.csproj
```diff
@@ -1,29 +1,31 @@
...
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
+    <ItemGroup>
+        <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.9.0-preview1" />
+        <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+        <PackageReference Include="System.Net.NameResolution" Version="4.3.0" />
+    </ItemGroup>
+    <ItemGroup>
+        <ProjectReference Include="..\..\extensions\Worker.Extensions.Abstractions\src\Worker.Extensions.Abstractions.csproj" />
+        <ProjectReference Include="..\..\extensions\Worker.Extensions.Http\src\Worker.Extensions.Http.csproj" />
```
liliankasem, line 16: We didn't make changes to Http so we should keep the package reference 

`<PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http" Version="3.0.13" />`

src/DotNetWorker.Core/Context/Features/DefaultFunctionInputBindingFeature.cs
```diff
@@ -69,18 +70,27 @@ public async ValueTask<FunctionInputBindingResult> BindFunctionInputAsync(Functi
...
+                        // Pass info about input converter types defined for this parameter, if present.
+                        if (param.Properties.TryGetValue(PropertyBagKeys.BindingAttributeSupportedConverters, out var converters))
                         {
-                            { PropertyBagKeys.ConverterType, converterTypeAssemblyFullName }
-                        };
+                            properties.Add(PropertyBagKeys.BindingAttributeSupportedConverters, converters);
                         }
 
-                        var converterContext = _converterContextFactory.Create(param.Type, source, context, properties);
+                        var converterContext = _converterContextFactory.Create(param.Type, source, context, properties.Count() != 0 ? properties.ToImmutableDictionary() : ImmutableDictionary<string, object>.Empty);
```
liliankasem, line 93: nit: use multiple lines for easier readability i.e.

```csharp
var converterContext = _converterContextFactory.Create(param.Type, source, context, properties.Count() != 0 
                     ? properties.ToImmutableDictionary()
                     : ImmutableDictionary<string, object>.Empty);
```

liliankasem (commented):
part 1, will do a deeper review on some of the core sdk changes

src/DotNetWorker.Core/Converters/Converter/EnableConvertersFallbackAttribute.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using System;
+
+namespace Microsoft.Azure.Functions.Worker.Converters
+{
+    /// <summary>
+    /// An attribute that specifies if Converters fallback is enabled
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class)]
+    public sealed class EnableConvertersFallbackAttribute : Attribute
```
jviau, line 12: This needs a bool parameter so that it can be explicitly enabled _or_ disabled.

src/DotNetWorker.Core/Converters/Converter/EnableConvertersFallbackAttribute.cs
```diff
@@ -0,0 +1,15 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.Functions.Worker.Converters
+{
+    /// <summary>
+    /// An attribute that specifies if Converters fallback is enabled
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class)]
```
jviau, line 11: I believe we want this to be usable anywhere `InputConverterAttribute` is usable (so parameters, types, etc). Although keep `AllowMultiple = false`

src/DotNetWorker.Core/Properties/AssemblyInfo.cs
```diff
@@ -8,3 +8,4 @@
...
 [assembly: InternalsVisibleTo("DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7")]
+[assembly: InternalsVisibleTo("Microsoft.Azure.Functions.Worker.Extensions.Storage.Blobs, PublicKey=00240000048000009400000006020000002400005253413100040000010001005148be37ac1d9f58bd40a2e472c9d380d635b6048278f7d47480b08c928858f0f7fe17a6e4ce98da0e7a7f0b8c308aecd9e9b02d7e9680a5b5b75ac7773cec096fbbc64aebd429e77cb5f89a569a79b28e9c76426783f624b6b70327eb37341eb498a2c3918af97c4860db6cdca4732787150841e395a29cfacb959c1fd971c1")]
```
jviau, line 11: I don't think we should be adding `InternalsVisibleTo` to any extension package. This feature must work without that, and also internals visible to makes all our `internal` types now need to follow public breaking change rules, as extension packages tend to be versioned separately.

src/DotNetWorker.Grpc/Definition/GrpcFunctionDefinition.cs
```diff
@@ -71,16 +76,116 @@ public GrpcFunctionDefinition(FunctionLoadRequest loadRequest, IMethodInfoLocato
...
+            }
+
+            var output = new Dictionary<string, object>();
+            bool isInputConverterAttributeAdvertised = false;
+
+            // ConverterTypesDictionary will be "object" part of the return value of this method - ImmutableDictionary<string, object>
+            // The dictionary has key of type IInputConverter and value as Properties of that converter (specifies supported types and support for Json Deserialization)
+            var converterTypesDictionary = new Dictionary<Type, ConverterProperties>();
+
+            IEnumerable<Attribute> customAttributes = bindingAttribute.GetType().GetCustomAttributes();
```
jviau, line 111: instead of enumerating this twice, can we refactor to something like:

```suggestion
            Type type = bindingAttribute.GetType();
            if (type.GetCustomerAttribute<InputConverterAttribute>() is { } attribute)
            {
                // do something
                isInputConverterAttributeAdvertised = true;
            }
            
            if (isInputConverterAttributeAdvertised)
            {
                bool fallback = type.GetCustomAttribute<EnableConvertersFallbackAttribute>()?.Enabled ?? true;
            }
```

src/DotNetWorker.Grpc/Definition/GrpcFunctionDefinition.cs
```diff
@@ -71,16 +76,116 @@ public GrpcFunctionDefinition(FunctionLoadRequest loadRequest, IMethodInfoLocato
...
+                    }
+                }
+            }
 
-            return ImmutableDictionary<string, object>.Empty;
+            output.Add(PropertyBagKeys.BindingAttributeSupportedConverters, converterTypesDictionary);
+
+            if (isInputConverterAttributeAdvertised)
+            {
+                output.Add(PropertyBagKeys.EnableFallbackConverters, false);
```
jviau, line 133: The default should be `true` to keep logic identical with the way it is today.

src/DotNetWorker.Core/Properties/AssemblyInfo.cs
liliankasem: IIRC: It's only added because of the unit tests, the extension itself does not actually need this. I don't actually remember why this was the case - we should be able to write unit tests without this

src/DotNetWorker.Core/Converters/Converter/SupportedConverterType.cs
```diff
@@ -0,0 +1,29 @@
...
+
+using System;
+
+namespace Microsoft.Azure.Functions.Worker.Converters
+{
+    /// <summary>
+    /// An attribute that can specify a type of <see cref="Type"/> supported by function input conversion.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
+    public sealed class SupportedConverterTypeAttribute : Attribute
```
kshyju, line 12: Is this attribute used to specify the target types supported by the converter? Should the name be something like that? ex: InputConversionTargetType ?

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
```diff
@@ -12,12 +12,19 @@
...
+using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
+using System.IO;
 
 namespace Microsoft.Azure.Functions.Worker
 {
     /// <summary>
     /// Converter to bind Cosmos DB type parameters.
     /// </summary>
+    [SupportsDeferredBinding]
+    [SupportsJsonDeserialization]
```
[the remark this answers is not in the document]
surgupta-msft: `Poco` and `Object` are different. We will be using this definition of Poco for our check -https://learn.microsoft.com/en-us/aspnet/core/mvc/models/model-binding?view=aspnetcore-7.0#pocos-without-parameterless-constructors. And yes, this attribute will cover Poco and Poco collections.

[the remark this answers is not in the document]
liliankasem: Will there be any distinction between arrays and IEnumerable types? I.e. IReadOnlyList<POCO> vs POCO[] Or does this not matter to the converter?

What about for things like the blob trigger where we support POCO but not POCO collections?

```
        [Function(nameof(BlobBookFunction))]
        public void BlobBookFunction(
            [BlobTrigger("book-trigger")] Book data)
        {
            _logger.LogInformation("Id: {id} - Name: {name}", data.Id, data.Name);
        }
```

Trying to think about the impact this will have on the analyzer - like how we can determine that a given collection is supported if not directly listed by `SupportedConverterType` 

E.g. some one tries to bind to `Book[]` on a BlobTrigger - it's not supported. 

Maybe for things like this we just have to build a different analyzer that checks trigger attributes for collection use? Independent of the Supported Types analyzer that will only check `SupportedConverterType` attributes?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -15,12 +15,40 @@
...
 using Microsoft.Extensions.Logging;
+using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
 
 namespace Microsoft.Azure.Functions.Worker
 {
     /// <summary>
     /// Converter to bind Blob Storage type parameters.
     /// </summary>
+    [SupportsDeferredBinding]
+    [SupportsJsonDeserialization]
```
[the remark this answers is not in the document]
surgupta-msft: @jviau Blob extension does not support some types like `BinaryData`, so it does not support everything.

src/DotNetWorker.Core/Context/Features/DefaultInputConversionFeature.cs
```diff
@@ -133,5 +187,58 @@ private async ValueTask<ConversionResult> AwaitAndReturnConversionTaskResult(Val
...
+            {
+                return converterSupports == true && IsJsonDeserializedObject(targetType.GetGenericArguments().FirstOrDefault());
+            }
+
+            return false;
+        }
+
+        private bool IsJsonDeserializedObject(Type type)
+        { 
+            return !type.GetConstructors().Any(a => a.GetParameters().Any());
```
jviau, line 241: Is this trying to say only types with parameterless contrustors are JSON deserializable? If so, this is not true - types can be deserialized from JSON with parametered constructors. Here is how it works for STJ: https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/immutability?pivots=dotnet-7-0

For Newtonsoft it is the same, but implicit - the attribute is not needed. Essentially, you can use constructors with parameters if the params match the property names (pascal to camel case conversion allowed).

Even then, the type could have a custom JSON converter which constructs the object in any number of ways.

What I am getting at is that it is impossible to identify JSON deserialization support statically, and even at runtime we have no idea what the entire set of JSON converters that may be used by the end `InputConverter`.

src/DotNetWorker.Core/Properties/AssemblyInfo.cs
jviau: Even for unit tests I would not suggest we add `InternalsVisibleTo`. This would indicate our design is not unit test friendly for other extensions and we should re-evaluate that.

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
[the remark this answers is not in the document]
jviau: As soon as we support POCO deserialization I don't think we can perform any static analysis, even for analyzers. Support can only be determined at runtime. As for BinaryData, then that will either need support added (that is possible I believe) or we will live with that being a runtime failure.

src/DotNetWorker.Grpc/Definition/GrpcFunctionDefinition.cs
surgupta-msft: Done

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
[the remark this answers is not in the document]
surgupta-msft: Discussed offline to remove attribute SupportsJsonDeserialization

extensions/Worker.Extensions.Abstractions/src/SupportsJsonDeserializationAttribute.cs
```diff
@@ -0,0 +1,16 @@
...
+using System;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Abstractions
+{
+    /// <summary>
+    /// Specifies if a converter supports Json Deserialization when generating function metadata.
+    /// This is to be used on converters that support Json Deserialization.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class)]
+    public class SupportsJsonDeserializationAttribute : Attribute
```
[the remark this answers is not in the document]
surgupta-msft: As discussed offline, removed this attribute

src/DotNetWorker.Core/Converters/Converter/EnableConvertersFallbackAttribute.cs
surgupta-msft: Done

src/DotNetWorker.Grpc/Definition/GrpcFunctionDefinition.cs
surgupta-msft: Done

src/DotNetWorker.Core/Converters/Converter/SupportedConverterType.cs
surgupta-msft: There is another attribute with similar name `InputConverterAttribute`. With the name `SupportedConverterTypeAttribute` we are trying to say that Converter supports this type. I am open to change it to a better name. Thoughts? cc @liliankasem @jviau

src/DotNetWorker.Core/Context/Features/DefaultInputConversionFeature.cs
surgupta-msft: Removed this logic

src/DotNetWorker.Core/Converters/Converter/EnableConvertersFallbackAttribute.cs
surgupta-msft: Added

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
surgupta-msft: Done - https://github.com/Azure/azure-functions-dotnet-worker/issues/1512

samples/WorkerBindingSamples/WorkerBindingSamples.csproj
surgupta-msft: no, removed

src/DotNetWorker.Core/Properties/AssemblyInfo.cs
surgupta-msft: With the current code we have, dont need this anymore. Removed it

samples/WorkerBindingSamples/WorkerBindingSamples.csproj
```diff
@@ -1,29 +1,31 @@
...
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
+    <ItemGroup>
+        <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.9.0-preview1" />
+        <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+        <PackageReference Include="System.Net.NameResolution" Version="4.3.0" />
+    </ItemGroup>
+    <ItemGroup>
+        <ProjectReference Include="..\..\extensions\Worker.Extensions.Abstractions\src\Worker.Extensions.Abstractions.csproj" />
```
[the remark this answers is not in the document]
surgupta-msft: I tried to make this same as FunctionApp.csproj as I also added a PublishTest for sdkbinding project. I think it is useful because we are doing a lot of changes in Metadata generation as per deferred binding logic. Want to make sure we have tests guarding that logic.

src/DotNetWorker.Core/Converters/Converter/SupportedConverterType.cs
liliankasem: I think we should leave it as it is. `InputConversionTargetType` could be confusing as it implies it's only for input and these supported type attributes can go on any converter

extensions/Worker.Extensions.CosmosDB/src/CosmosDBInputAttribute.cs
```diff
@@ -2,11 +2,13 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System.Collections.Generic;
+using Microsoft.Azure.Functions.Worker.Converters;
 using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
 
 namespace Microsoft.Azure.Functions.Worker
 {
-    [SupportsDeferredBinding]
+    [AllowConverterFallback(true)]
```
liliankasem, line 10: I think we want this to be false?

extensions/Worker.Extensions.Storage.Blobs/src/BlobInputAttribute.cs
```diff
@@ -1,11 +1,13 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using Microsoft.Azure.Functions.Worker.Converters;
 using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
 
 namespace Microsoft.Azure.Functions.Worker
 {
-    [SupportsDeferredBinding]
+    [AllowConverterFallback(true)]
```
liliankasem, line 9: Same here, should be false?

samples/WorkerBindingSamples/NuGet.Config
```diff
@@ -0,0 +1,7 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
+    <add key="azure_app_service" value="https://www.myget.org/F/azure-appservice/api/v2" />
```
liliankasem, line 5: make me a little nervous if this is in our sample; I don't want customers to be pulling prereleases packages from our  myget

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -804,8 +806,78 @@ private static bool SupportsDeferredBinding(CustomAttribute attribute)
...
+                if (string.Equals(attribute.AttributeType.FullName, Constants.SupportedConverterTypeAttributeType, StringComparison.Ordinal))
+                {
+                    foreach (CustomAttributeArgument element in attribute.ConstructorArguments)
+                    {
+                        if (element.Type.FullName == typeof(Type).FullName)
+                        {
+                            var supportedType = element.Value as TypeReference;
+
+                            if (supportedType is not null
+                                && supportedType != null && string.Equals(supportedType.FullName, bindingType.FullName, StringComparison.Ordinal))
```
liliankasem, line 871: `supportedType is not null && supportedType != null `

Aren't these doing the exact same check?

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -804,8 +806,78 @@ private static bool SupportsDeferredBinding(CustomAttribute attribute)
...
+
+                    if (!converterAdvertisesTypes)
+                    {
+                        // If a converter advertises deferred binding but does not explictly advertise any types then DeferredBinding will be supported for all the types
+                        return true;
+                    }
+                    else if (DoesConverterSupportTargetType(typeReferenceCustomAttributes, bindingType))
+                    {
+                        return true;
+                    }
```
liliankasem, line 850: Could this be simplified to:

````
if (!converterAdvertisesTypes)
{
    // If a converter advertises deferred binding but does not explicitly advertise any types then DeferredBinding will be supported for all the types
    return true;
}

return DoesConverterSupportTargetType(typeReferenceCustomAttributes, bindingType);
````

File: sdk/Sdk/FunctionMetadataGenerator.cs [hunk 27]
```diff
@@ -804,8 +806,78 @@ private static bool SupportsDeferredBinding(CustomAttribute attribute)
...
+        private static bool DoesConverterSupportTargetType(Collection<CustomAttribute> customAttributes, TypeReference bindingType)
+        {
+            // Parse attributes advertised by converter
+            foreach (CustomAttribute attribute in customAttributes)
+            {
+                if (string.Equals(attribute.AttributeType.FullName, Constants.SupportedConverterTypeAttributeType, StringComparison.Ordinal))
+                {
+                    foreach (CustomAttributeArgument element in attribute.ConstructorArguments)
+                    {
+                        if (element.Type.FullName == typeof(Type).FullName)
```
liliankasem, line 866: nit: use string.Equals

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -804,8 +806,78 @@ private static bool SupportsDeferredBinding(CustomAttribute attribute)
...
+            foreach (CustomAttribute bindingAttribute in typeDefinition.CustomAttributes)
+            {
+                if (string.Equals(bindingAttribute.AttributeType.FullName, Constants.InputConverterAttributeType, StringComparison.Ordinal))
+                {
+                    // InputConverterAttribute will have supported converter type
+                    foreach (var customAttribute in bindingAttribute.ConstructorArguments)
+                    {
+                        if (DoesConverterSupportDeferredBinding(customAttribute, bindingType))
+                        {
+                            return true;
```
liliankasem, line 819: nit: `return DoesConverterSupportDeferredBinding(customAttribute, bindingType)`

I don't think we need an if-statement here if we are returning true if DoesConverterSupportDeferredBinding is true

src/DotNetWorker.Core/Context/Features/DefaultFunctionInputBindingFeature.cs
```diff
@@ -69,18 +70,29 @@ public async ValueTask<FunctionInputBindingResult> BindFunctionInputAsync(Functi
...
                         {
-                            { PropertyBagKeys.ConverterType, converterTypeAssemblyFullName }
-                        };
+                            properties.Add(PropertyBagKeys.BindingAttributeSupportedConverters, converters);
                         }
 
-                        var converterContext = _converterContextFactory.Create(param.Type, source, context, properties);
+                        var converterContext = _converterContextFactory.Create(param.Type, source, context, properties.Count() != 0 
+                                                                                                ? properties.ToImmutableDictionary()
+                                                                                                : ImmutableDictionary<string, object>.Empty);
```
liliankasem, line 95: nit: love having the ? and : aligned with the =  (i think it makes it more readable :) )

```
var converterContext = _converterContextFactory.Create(param.Type, source, context, properties.Count() != 0 
                     ? properties.ToImmutableDictionary()
                     : ImmutableDictionary<string, object>.Empty);
```

sdk/Sdk/FunctionMetadataGenerator.cs
jviau: That suggested change would have behavioral differences then what is here when there is multiple `InputConverterAttribute`'s set, right?

sdk/Sdk/FunctionMetadataGenerator.cs
jviau: I can't find what `StringComparison` pattern matching uses, but this can also be written as:

``` CSharp
if (supportedType is { FullName: bindingType.FullName })
```

sdk/Sdk/FunctionMetadataGenerator.cs
surgupta-msft: Exactly.

sdk/Sdk/FunctionMetadataGenerator.cs
surgupta-msft: Done

src/DotNetWorker.Grpc/Definition/GrpcFunctionDefinition.cs
```diff
@@ -79,8 +84,83 @@ public GrpcFunctionDefinition(FunctionLoadRequest loadRequest, IMethodInfoLocato
...
+                {
+                    foreach (CustomAttributeTypedArgument supportedType in converterAttribute.ConstructorArguments)
+                    {
+                        if (supportedType.ArgumentType is not null
+                            && supportedType.Value is not null
+                            && supportedType.ArgumentType == typeof(Type))
+                        {
+                            Type? supportedTypeValue = supportedType.Value as Type;
+
+                            if (supportedTypeValue is not null)
```
jviau, line 154:
```suggestion
                            if (supportedType.Value is Type type)
```
Can remove the `as` check above.

src/DotNetWorker.Grpc/Definition/GrpcFunctionDefinition.cs
```diff
@@ -79,8 +84,83 @@ public GrpcFunctionDefinition(FunctionLoadRequest loadRequest, IMethodInfoLocato
...
+        {
+            var types = new List<Type>();
+
+            foreach (CustomAttributeData converterAttribute in converter.CustomAttributes)
+            {
+                if (converterAttribute.AttributeType == typeof(SupportedConverterTypeAttribute))
+                {
+                    foreach (CustomAttributeTypedArgument supportedType in converterAttribute.ConstructorArguments)
+                    {
+                        if (supportedType.ArgumentType is not null
```
jviau, line 148:
``` CSharp
if (supportedType is { ArgumentType: not null, Value: not null } && supportedType.ArgumentType == typeof(Type))
```

extensions/Worker.Extensions.Storage.Blobs/src/BlobInputAttribute.cs
surgupta-msft: Had a chat with Jacob. For these 2 converters (BlobStorage and Cosmos), both true and false will have same behavior because converter will handle everything (and return fail if Json deserialization throws exception). But for other extensions which will actually use BypassDeferredBinding, it depends how much restrictive they want their converter to be and then decide if they want fallback option or not - something to keep in our extension author guide.

src/DotNetWorker.Core/Converters/Converter/DefaultInputConverterProvider.cs
```diff
@@ -50,12 +50,30 @@ public IEnumerable<IInputConverter> RegisteredInputConverters
...
+        public IInputConverter GetOrCreateConverterInstance(Type converterType)
+        {
+            if (converterType is null)
+            {
+                throw new InvalidOperationException($"Could not create an instance of {(nameof(converterType))}.");
+            }
+
+            EnsureTypeCanBeAssigned(converterType);
+
+            return (IInputConverter)ActivatorUtilities.CreateInstance(_serviceProvider, converterType);
```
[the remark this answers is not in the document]
surgupta-msft: I did not completely follow the change being recommended here.

src/DotNetWorker.Core/Context/Features/DefaultInputConversionFeature.cs
```diff
@@ -111,6 +134,36 @@ private async ValueTask<ConversionResult> AwaitAndReturnConversionTaskResult(Val
...
+            var result = new Dictionary<IInputConverter, List<Type>>();
+
+            if (context.Properties.TryGetValue(PropertyBagKeys.BindingAttributeSupportedConverters, out var converterTypes))
+            {
+                if (converterTypes is not null && converterTypes.GetType() == typeof(Dictionary<Type, List<Type>>))
+                {
+                    var converters = converterTypes as Dictionary<Type, List<Type>>;
+                    var interfaceType = typeof(IInputConverter);
+
+                    foreach (var (converterTypesPair, converterType) in from converterTypesPair in converters
```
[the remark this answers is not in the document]
surgupta-msft: done

samples/WorkerBindingSamples/NuGet.Config
surgupta-msft: It worked without staging also

extensions/Worker.Extensions.Storage.Blobs/src/BlobInputAttribute.cs
jviau: @lilieankasem - if you have reason you want this to be `false` I am not against it. Functionally I don't think it will have any impact (at least that I see) to the end user between `false` and `true`.

liliankasem: I was thinking `false` for the same reasoning mentioned above, because these two converters are doing everything, fallback is not an option and I felt that false was a better representation of that. Alternatively, if it won't have impact for the scenario where it's all in (like blob and cosmos right now) - then we shouldn't have to provide this attribute at all?

liliankasem (approved):
just had one note about the `AllowConverterFallback` attribute for "all-in" scenarios (see comment response above)

src/DotNetWorker.Core/Converters/Converter/DefaultInputConverterProvider.cs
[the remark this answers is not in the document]
surgupta-msft: Okay I see what you are saying. I have created this method `GetOrCreateConverterInstance(Type converterType)` as an overload of another method in this class `GetOrCreateConverterInstance(string converterTypeName)` (just below this method) and trying to do the exact same thing in this method.

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
surgupta-msft: For trigger we will need Container client - https://github.com/Azure/azure-functions-dotnet-worker/blob/cb93c716190b75dea3d85edb4344b06a156ee267/test/E2ETests/E2EApps/E2EApp/Blob/BlobTriggerBindingFunctions.cs#L75

surgupta-msft: Resolving it as design is updated here and we won't advertise specific types for this converter now.

src/DotNetWorker.Core/Converters/Converter/AllowConverterFallbackAttribute.cs
```diff
@@ -0,0 +1,34 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.Functions.Worker.Converters
+{
+    /// <summary>
+    /// An attribute that specifies if Converter fallback is allowed
+    /// </summary>
+    [AttributeUsage(
```
mathewc, line 11: Isn't this attribute only applied to binding attribute Types (i.e. classes)? Why the long list of targets?

src/DotNetWorker.Core/Converters/Converter/AllowConverterFallbackAttribute.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// <summary>
+        /// Gets the value of whether Converter fallback is allowed.
+        /// </summary>
+        public bool AllowConverterFallback { get; }
+
+        /// <summary>
+        /// Creates a new instance of <see cref="AllowConverterFallbackAttribute"/>
+        /// </summary>
+        /// <param name="allowConverterFallback">The value to indicate if converter fallback is allowed.</param>
+        public AllowConverterFallbackAttribute(bool allowConverterFallback)
```
mathewc, line 28: It looks like the default if this attribute isn't specified is to allow fallback? I.e. the attribute isn't required. Then I wonder why you have an attribute that allows specification of true/false, and instead don't just have an attribute to disable fallback when needed (default being true). E.g. [DisableConverterFallback].

src/DotNetWorker.Core/Converters/Converter/DefaultInputConverterProvider.cs
```diff
@@ -50,12 +50,30 @@ public IEnumerable<IInputConverter> RegisteredInputConverters
...
+        /// Gets an instance of the converter for the type requested.
+        /// </summary>
+        /// <param name="converterType">The type for which we are requesting an IInputConverter instance.</param>
+        /// <exception cref="InvalidOperationException">Throws when the converterType param is null.</exception>
+        /// <returns>IConverter instance of the requested type.</returns>
+        public IInputConverter GetOrCreateConverterInstance(Type converterType)
+        {
+            if (converterType is null)
+            {
+                throw new InvalidOperationException($"Could not create an instance of {(nameof(converterType))}.");
```
mathewc, line 69: This should be an ArgumentNullException not IOE

src/DotNetWorker.Core/Converters/Converter/DefaultInputConverterProvider.cs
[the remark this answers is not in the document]
mathewc: Shouldn't the other overload be refactored to call this one then - after resolving its type name to a Type, call this overload?

src/DotNetWorker.Core/Converters/Converter/SupportedConverterType.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.Functions.Worker.Converters
+{
+    /// <summary>
+    /// An attribute that can specify a type of <see cref="Type"/> supported by function input conversion.
```
mathewc, line 9: nit: grammar "a type of Type"?

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -804,8 +806,75 @@ private static bool SupportsDeferredBinding(CustomAttribute attribute)
...
 
-            return typeDefinition.CustomAttributes
-                                 .Any(a => string.Equals(a.AttributeType.FullName, Constants.SupportsDeferredBindingAttributeType, StringComparison.Ordinal));
+            // checking attributes advertised by the binding attribute
+            foreach (CustomAttribute bindingAttribute in typeDefinition.CustomAttributes)
+            {
+                if (string.Equals(bindingAttribute.AttributeType.FullName, Constants.InputConverterAttributeType, StringComparison.Ordinal))
+                {
+                    // InputConverterAttribute will have supported converter type
+                    foreach (var customAttribute in bindingAttribute.ConstructorArguments)
```
mathewc, line 815: InputConverterAttribute ctor just takes a single parameter "converterType" right? The fact that you're naming this "customAttribute" here and foreaching confuses me.

test/FunctionMetadataGeneratorTests/FunctionMetadataGeneratorTests.cs
```diff
@@ -303,6 +305,142 @@ void ValidateBlobOutput(ExpandoObject b)
...
+                b => ValidateBlobTriggerForBlobClient(b),
+                b => ValidateBlobInput(b),
+                b => ValidateBlobOutput(b));
+
+
+            void ValidateBlobTriggerForBlobClient(ExpandoObject b)
+            {
+                AssertExpandoObject(b, new Dictionary<string, object>
+                {
+                    { "Name", "blob" },
```
mathewc, line 322: This single test function is validating like 20 different scenarios across all the storage binding types. Is there a reason why these aren't broken apart into smaller test cases?

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -804,8 +806,75 @@ private static bool SupportsDeferredBinding(CustomAttribute attribute)
                 return false;
             }
 
-            return typeDefinition.CustomAttributes
-                                 .Any(a => string.Equals(a.AttributeType.FullName, Constants.SupportsDeferredBindingAttributeType, StringComparison.Ordinal));
+            // checking attributes advertised by the binding attribute
+            foreach (CustomAttribute bindingAttribute in typeDefinition.CustomAttributes)
```
kshyju, line 810: Do we have an item tracking the change on the src gen version?

kshyju (approved):
The PR title suggests that this change is to skip the new binding behavior (so things will work as hot it is today - populating binding data in host). But this PR has more changes than that (Ex: using converter types advertised by trigger, using supported types published by converter etc.) Consider updating title as needed.

src/DotNetWorker.Core/Converters/Converter/DefaultInputConverterProvider.cs
jviau, line 74: @surgupta-msft, what I was meaning is this change:
```suggestion
            return (IInputConverter)ActivatorUtilities.GetOrCreateInstance(_serviceProvider, converterType);
```

jviau (approved):

src/DotNetWorker.Core/Converters/Converter/AllowConverterFallbackAttribute.cs
jviau: We have the `true`/`false` because we want to add an analyzer to suggest all input binding implementations that use `SupportsDeferredBinding` to be explicit about their `AllowConverterFallback` behavior. Thus, we need to be able to specify both `true` or `false`.

sdk/Sdk/FunctionMetadataGenerator.cs
surgupta-msft: Yes, updated the name to converterType instead of customAttribute. Foreaching only because at this point, we don't know that this attribute has how many constructor parameters.

src/DotNetWorker.Core/Converters/Converter/AllowConverterFallbackAttribute.cs
surgupta-msft: We did have discussion for both AllowConverterFallback and DisableConverterFallback in our design syncs. From last meeting, I remember we wanted to use AllowConverterFallback, unless I am missing anything. @jviau do you have more details/preference to share on this?

samples/WorkerBindingSamples/NuGet.Config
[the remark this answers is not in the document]
surgupta-msft: We have this [PublishTests.cs](https://github.com/Azure/azure-functions-dotnet-worker/blob/feature/sdk-type-binding/test/SdkE2ETests/PublishTests.cs) which test Metadata generation E2E for a sample App. The test for [FunctionApp](https://github.com/Azure/azure-functions-dotnet-worker/tree/main/samples/FunctionApp) was already there and I tried to setup a similar E2E test for [Sdk bindings samples](https://github.com/Azure/azure-functions-dotnet-worker/tree/surgupta/bypass-deferredbinding/samples/WorkerBindingSamples) as well. The test is failing if I remove Nuget.config (package restore errors)

src/DotNetWorker.Core/Converters/Converter/DefaultInputConverterProvider.cs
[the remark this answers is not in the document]
surgupta-msft: I made the change so that string overload will call the Type overload and I agree with the change to remove string overload. But it's kind of making me change 4-5 files. Let me know if I should update this PR or refactor it separately (I can start another PR with only those changes quickly.)

sdk/Sdk/FunctionMetadataGenerator.cs
surgupta-msft: I am discussing this with Sarah. Will create one.

test/FunctionMetadataGeneratorTests/FunctionMetadataGeneratorTests.cs
surgupta-msft: refactored it

src/DotNetWorker.Core/Context/Features/DefaultInputConversionFeature.cs
```diff
@@ -47,17 +48,39 @@ public async ValueTask<ConversionResult> ConvertAsync(ConverterContext converter
                 }
             }
 
-            // Use the registered converters. The first converter which can handle the conversion wins.
-            foreach (var converter in _inputConverterProvider.RegisteredInputConverters)
-            {
-                var conversionResult = await ConvertAsyncUsingConverter(converter, converterContext);
+            // Get all converters advertised by the Binding Attribute along with supported types by each converter
+            Dictionary<IInputConverter, List<Type>>? advertisedConverterTypes = GetExplicitConverterTypes(converterContext);
```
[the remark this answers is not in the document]
surgupta-msft: I have made a note of it for part2 changes (will include changes in Source gen + few refactoring suggested here).

PR #1467, opened 2023-04-13
Synchronizing scopes across middleware pipelines

brettsam:
Some renames and folder moves involved. Also some cleanup on a a couple of areas that I've noted.


extensions/Worker.Extensions.Http.AspNetCore/src/Coordinator/DefaultHttpCoordinator.cs
```diff
@@ -0,0 +1,66 @@
...
+
+            contextReference.StartFunction();
+            return contextReference.FunctionCompletionTask.Task;
+        }
+
+        // TODO:See about making this not public
+        public void CompleteFunctionInvocation(string invocationId)
+        {
+            // This is the last step; remove the context reference
+            if (_contextReferenceList.TryRemove(invocationId, out var contextRef))
```
brettsam, line 54: Note -- we were never removing this before.

extensions/Worker.Extensions.Http.AspNetCore/src/Coordinator/DefaultHttpCoordinator.cs
```diff
@@ -0,0 +1,66 @@
...
+        private readonly ConcurrentDictionary<string, ContextReference> _contextReferenceList;
+
+        public DefaultHttpCoordinator()
+        {
+            _contextReferenceList = new ConcurrentDictionary<string, ContextReference>();
+        }
+
+        public Task<FunctionContext> SetHttpContextAsync(string invocationId, HttpContext context)
+        {
+            var contextRef = _contextReferenceList.GetOrAdd(invocationId, static id => new ContextReference(id));
```
brettsam, line 22: Simplified these to be GetOrAdd() and to use a static factory. AddOrUpdate() has unique race conditions that meant we couldn't be guaranteed of success with the way we were doing it (see Remarks): https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.addorupdate?view=net-7.0

extensions/Worker.Extensions.Http.AspNetCore/src/Coordinator/ContextReference.cs
```diff
@@ -0,0 +1,81 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore.Coordinator
```
fabiocav, line 9: This namespace seems off. Can we drop `Coordinator` here?

extensions/Worker.Extensions.Http.AspNetCore/src/AspNetMiddleware/ApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Microsoft.AspNetCore.Builder;
+using Microsoft.Azure.Functions.Worker.Core.Pipeline;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore
+{
+    internal static class ApplicationBuilderExtensions
+    {
+        public static IApplicationBuilder UseSetServiceProviderMiddleware(this IApplicationBuilder builder) =>
```
fabiocav, line 8: Do we need this methods? Can we inline the middleware registration and remove this class?

extensions/Worker.Extensions.Http.AspNetCore/src/Coordinator/ContextReference.cs
```diff
@@ -0,0 +1,81 @@
...
+            _tokenRegistration = _token.Register(() =>
+            {
+                _functionStartTask.TrySetCanceled();
+                _functionCompletionTask.TrySetCanceled();
+                _functionContextValueSource.TrySetCanceled();
+                _httpContextValueSource.TrySetCanceled();
+            });
+        }
+
+        internal void StartFunction()
```
fabiocav, line 49: Can we make this an awaitable/`Task` returning method and return the `FunctionCompletionTask.Task` to simplify consumption?

extensions/Worker.Extensions.Http.AspNetCore/src/Coordinator/ContextReference.cs
brettsam: Yes -- VS auto-updated these when I organized them into folders and I didn't notice. @satvu mentioned the same thing.

extensions/Worker.Extensions.Http.AspNetCore/src/AspNetMiddleware/ApplicationBuilderExtensions.cs
brettsam: We really don't -- the one already existed so I added the new one. I'll inline them and remove this class.

extensions/Worker.Extensions.Http.AspNetCore/src/Coordinator/ContextReference.cs
brettsam: The caller of this shouldn't have to await anything though -- it's the consumer of the FunctionStartTask that is waiting. This call signals that "anyone awaiting the start of the function can now continue"

fabiocav (approved):

satvu (approved):

extensions/Worker.Extensions.Http.AspNetCore/src/Coordinator/ContextReference.cs
```diff
@@ -0,0 +1,80 @@
...
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore
+{
+    internal class ContextReference : IDisposable
+    {
+        private readonly TaskCompletionSource<bool> _functionStartTask = new();
+        private readonly TaskCompletionSource<bool> _functionCompletionTask = new();
+
+        private TaskCompletionSource<HttpContext> _httpContextValueSource = new();
+        private TaskCompletionSource<FunctionContext> _functionContextValueSource = new();
```
danielmarbach, line 17: @brettsam @fabiocav Is it safe to not use TaskContinuation.RunAsynchronously here given that by default some parts of continuation will now by default be executed on the thread that called the completion source set methods? I'm currently looking into an obscure bug of a client and started wondering whether this was deliberate or an oversight.

PR #1470, opened 2023-04-14
Queue storage extension support for QueueMessage & BinaryData

### LINKED ISSUE #1444: [SDK-Bindings - Queue] Implement QueueMessage converter
#### Problem Statement

Implement `QueueMessageConverter` to enable binding to the `QueueMessage`, `BinaryData` and `JObject` types so that we have parity with in-proc.

#### Tasks

- [x] Implement converter that enables binding to a single `QueueMessage`, `BinaryData` and `JObject`
- [x] Write unit tests for converter and support parts
- [x] Write E2E tests for queue trigger function with `QueueMessage` binding
- [ ] Release Queue worker extension package

#### Definition of Done

1. Worker Storage.Queue extension package is released (as a preview package) with support for binding to a `QueueMessage`, `BinaryData`, and `JObject`
2. Unit & E2E tests written and checked-in
3. Microsoft documentation updated to detail support for the new SDK-bindings

liliankasem:

resolves #1444

Implementing QueueMessageConverter to support binding to QueueMessage and BinaryData.

### Pull request checklist

### Additional information

Additional PR information


extensions/Worker.Extensions.Storage.Queues/src/QueueMessageConverter.cs
```diff
@@ -0,0 +1,130 @@
...
+
+            try
+            {
+                JsonSerializerOptions options = new() { Converters = { new QueueMessageJsonConverter() } };
+                return modelBindingData.Content.ToObjectFromJson<QueueMessage>(options);
+            }
+            catch (JsonException ex)
+            {
+                // Easy to have the queue payload not deserialize properly. So give a useful error.
+                string msg = String.Format(CultureInfo.CurrentCulture,
```
liliankasem, line 100: Using same message as the extension: https://github.com/Azure/azure-sdk-for-net/blob/056f7071a4b309a370c2ce7d52b2ef256f8dccf1/sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Queues/src/Triggers/UserTypeArgumentBindingProvider.cs#L74

extensions/Worker.Extensions.Storage.Queues/src/StorageQueuesExtensionStartup.cs
```diff
@@ -0,0 +1,30 @@
...
+        public override void Configure(IFunctionsWorkerApplicationBuilder applicationBuilder)
+        {
+            if (applicationBuilder == null)
+            {
+                throw new ArgumentNullException(nameof(applicationBuilder));
+            }
+
+            applicationBuilder.Services.Configure<WorkerOptions>((workerOption) =>
+            {
+                workerOption.InputConverters.RegisterAt<QueueMessageConverter>(0);
```
jviau, line 26: Can you remind me why we need to register this explicitly if we already have `InputConverterAttribute` on `QueryTriggerAttribute`?

samples/WorkerBindingSamples/Program.cs
```diff
@@ -9,6 +10,8 @@ public class Program
     {
         public static void Main()
         {
+            Console.WriteLine($"Azure Functions .NET Worker (PID: { Environment.ProcessId }) initialized in debug mode. Waiting for debugger to attach...");
```
jviau, line 13: Is this necessary? Technically it isn't always debug mode.

extensions/Worker.Extensions.Storage.Queues/src/QueueMessageConverter.cs
```diff
@@ -0,0 +1,130 @@
...
+            }
+
+            return true;
+        }
+
+        private object? ToTargetType(Type targetType, QueueMessage queueMessage) => targetType switch
+        {
+            Type _ when targetType == typeof(QueueMessage) => queueMessage,
+            Type _ when targetType == typeof(BinaryData) => ConvertMessageContentToBinaryData(queueMessage),
+            Type _ when targetType == typeof(JObject) => ConvertMessageContentToJObject(queueMessage),
```
jviau, line 81: `JObject` is from `Newtonsoft.Json` right? Is that intentional? I find it weird we are mixing STJ and Newtonsoft here.

Not sure if this helps/you are aware of this behavior. But STJ and Newtonsoft (and maybe other serialization libraries) have a behavior that when you deserialize as just `object`, they will deserialize to their object representation. I think that is `JsonElement` for STJ and `JObject` (or something else) for Newtonsoft. Their may be an opportunity to fit that behavior in here somehow?

extensions/Worker.Extensions.Storage.Queues/src/QueueMessageJsonConverter.cs
```diff
@@ -0,0 +1,44 @@
...
+        public override bool CanConvert(Type objectType)
+        {
+            return objectType == typeof(QueueMessage);
+        }
+
+        public override QueueMessage? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            JsonElement root = JsonDocument.ParseValue(ref reader).RootElement;
+
+            root.TryGetProperty("MessageId", out var messageId);
```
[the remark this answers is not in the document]
liliankasem: I think the only values that are nullable are the dates, outside of that we should expect every other value to exist

```csharp
        public System.DateTimeOffset? NextVisibleOn { get; internal set; }
        public System.DateTimeOffset? InsertedOn { get; internal set; }
        public System.DateTimeOffset? ExpiresOn { get; internal set; }
```

https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Azure.Storage.Queues/src/Models/QueueMessage.cs

extensions/Worker.Extensions.Storage.Queues/src/QueueMessageJsonConverter.cs
```diff
@@ -0,0 +1,44 @@
...
+    internal class QueueMessageJsonConverter : JsonConverter<QueueMessage>
+    {
+        public override bool CanConvert(Type objectType)
+        {
+            return objectType == typeof(QueueMessage);
+        }
+
+        public override QueueMessage? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            JsonElement root = JsonDocument.ParseValue(ref reader).RootElement;
```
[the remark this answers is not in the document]
liliankasem: I think perf is always a worth accounting for, happy to make the suggested changed here :)

extensions/Worker.Extensions.Storage.Queues/src/StorageQueuesExtensionStartup.cs
liliankasem: We do not need to register, I think this is just left over from the previous implementation - will clean this up!

samples/WorkerBindingSamples/Program.cs
liliankasem: Nope, this is just for me whilst I'm still testing E2E - will remove before merge

extensions/Worker.Extensions.Storage.Queues/src/QueueMessageConverter.cs
liliankasem: Yeah this one I'm not 100% sure about - I was trying to make sure we have parity with in-proc, and with in-proc we allow binding directly to `JObject` ([example](https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Queues#binding-queue-message-to-jobject)):

```csharp
public static class QueueTriggerFunction_JObject
{
    [FunctionName("QueueTriggerFunction")]
    public static void Run(
        [QueueTrigger("sample-queue")] JObject message,
        ILogger logger)
    {
        logger.LogInformation("Received message from sample-queue, content={content}", message["content"]);
    }
}
```

So I was trying to enable the exact capability here.

I wasn't aware of the STJ/Newtonsoft behaviour you mentioned, I'm happy to go down that route to make sure we support both?

extensions/Worker.Extensions.Storage.Queues/src/QueueMessageJsonConverter.cs
```diff
@@ -17,23 +17,74 @@ public override bool CanConvert(Type objectType)
...
+                }
+
+                var propertyName = reader.GetString();
+                reader.Read();
+
+                switch (propertyName)
+                {
+                    // We're not expecting these three values to be null. If they are, should we throw or just return an empty string?
+                    case "MessageId":
+                        messageId = reader.GetString() ?? throw new ArgumentNullException("MessageId");
```
jviau, line 62: Maybe a `JsonException`?

jviau (commented):
converter changes look good to me.

extensions/Worker.Extensions.Storage.Queues/src/QueueMessageConverter.cs
jviau: I understand in-proc supports `JObject`, but we have moved towards STJ as the default and I don't think it is right for this extension to bring in Newtonsoft.Json. I think it should be something intentional by the customer to enable `JObject` being supported - like they intentionally switch out the `WorkerOptions.ObjectSerializer` to a Newtonsoft implementation, then JObject is supported.

This would work by having this implementation import that serializer and use it when the type is any POCO.

surgupta-msft:
LGTM!

extensions/Worker.Extensions.Storage.Queues/src/QueueMessageConverter.cs
liliankasem: I think POCOs are already supported by the extension so we don't need to handle that here, but I replaced JObject with JsonElement - let me know what you think of the new changes

test/DotNetWorkerTests/GrpcFunctionDefinitionTests.cs
```diff
@@ -173,5 +214,11 @@ private class MyBlobFunctionClass
             }
         }
 
+        private class MyQueueFunctionClass
```
surgupta-msft, line 217: Adding here as a reminder. We will need test in `FunctionMetadataGenerator.cs` for queue similar to Blob as a number of conditions decide if `SupportsDeferredBinding` is `true`

https://github.com/Azure/azure-functions-dotnet-worker/blob/cb93c716190b75dea3d85edb4344b06a156ee267/test/FunctionMetadataGeneratorTests/FunctionMetadataGeneratorTests.cs#L247

extensions/Worker.Extensions.Storage.Queues/src/QueueMessageJsonConverter.cs
```diff
@@ -0,0 +1,92 @@
...
+                        break;
+                }
+            }
+
+            throw new JsonException("JSON payload expected to start with EndObject token.");
+        }
+
+        public override void Write(Utf8JsonWriter writer, QueueMessage value, JsonSerializerOptions options)
+        {
+            throw new NotImplementedException();
```
jviau, line 89: Is this the recommended exception to throw here? Or is `JsonException` more fitting?

Not saying it is, just want to understand for myself.

extensions/Worker.Extensions.Storage.Queues/src/QueueMessageJsonConverter.cs
```diff
@@ -0,0 +1,92 @@
...
+                    continue;
+                }
+
+                var propertyName = reader.GetString();
+                reader.Read();
+
+                switch (propertyName)
+                {
+                    // We're not expecting these three values to be null. If they are, should we throw or just return an empty string?
+                    case "MessageId":
```
jviau, line 58: Do we need case insensitive support?

extensions/Worker.Extensions.Storage.Queues/src/QueueMessageJsonConverter.cs
```diff
@@ -0,0 +1,92 @@
...
+                        messageId = reader.GetString() ?? throw new JsonException("MessageId");
```
jviau, line 59: Should we have more descriptive exception messages?

extensions/Worker.Extensions.Storage.Queues/src/QueueMessageJsonConverter.cs
```diff
@@ -0,0 +1,92 @@
...
+                        break;
+                    case "ExpiresOn":
+                        expiresOn = reader.GetDateTime();
+                        break;
+                    default:
+                        break;
+                }
+            }
+
+            throw new JsonException("JSON payload expected to start with EndObject token.");
```
jviau, line 84: I think this message needs to be update: "... **start** with **EndObject** token." <- doesn't sound right to me?

extensions/Worker.Extensions.Storage.Queues/src/Properties/AssemblyInfo.cs
```diff
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
-﻿using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
+using System.Runtime.CompilerServices;
+using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
 
-[assembly: ExtensionInformation("Microsoft.Azure.WebJobs.Extensions.Storage.Queues", "5.0.0")]
+[assembly: ExtensionInformation("Microsoft.Azure.WebJobs.Extensions.Storage.Queues", "5.3.0-alpha.20230329.1")]
+[assembly: InternalsVisibleTo("Microsoft.Azure.Functions.WorkerExtension.Tests, PublicKey=00240000048000009400000006020000002400005253413100040000010001005148be37ac1d9f58bd40a2e472c9d380d635b6048278f7d47480b08c928858f0f7fe17a6e4ce98da0e7a7f0b8c308aecd9e9b02d7e9680a5b5b75ac7773cec096fbbc64aebd429e77cb5f89a569a79b28e9c76426783f624b6b70327eb37341eb498a2c3918af97c4860db6cdca4732787150841e395a29cfacb959c1fd971c1")]
```
jviau, line 8: nit: new lines at end of file (and some other files)

extensions/Worker.Extensions.Storage.Queues/src/TypeConverters/QueueConverterBase.cs
```diff
@@ -0,0 +1,49 @@
...
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (context.TargetType != typeof(T))
+            {
+                return false;
+            }
+
+            if (!(context.Source is ModelBindingData bindingData))
```
jviau, line 34:
```suggestion
            if (context.Source is not ModelBindingData bindingData)
```

extensions/Worker.Extensions.Storage.Queues/src/TypeConverters/QueueConverterBase.cs
```diff
@@ -0,0 +1,49 @@
...
+using Microsoft.Azure.Functions.Worker.Storage.Queues;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    internal abstract class QueueConverterBase<T>  : IInputConverter
+    {
+        private readonly ILogger<QueueConverterBase<T>> _logger;
+
+        public QueueConverterBase(ILogger<QueueConverterBase<T>> logger)
```
jviau, line 17: For base classes, I recommend taking in the plain `ILogger`. `ILogger<T>` is purely for the DI scenario - it has no use beyond that.

```suggestion
        public QueueConverterBase(ILogger logger)
```

but also it isn't used, so can be removed?

extensions/Worker.Extensions.Storage.Queues/src/TypeConverters/QueueConverterBase.cs
```diff
@@ -0,0 +1,49 @@
...
+using Microsoft.Azure.Functions.Worker.Converters;
+using Microsoft.Azure.Functions.Worker.Core;
+using Microsoft.Azure.Functions.Worker.Storage.Queues;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    internal abstract class QueueConverterBase<T>  : IInputConverter
+    {
+        private readonly ILogger<QueueConverterBase<T>> _logger;
```
jviau, line 15: This is not used. Can we remove it?

extensions/Worker.Extensions.Storage.Queues/src/TypeConverters/QueueMessageBinaryDataConverter.cs
```diff
@@ -0,0 +1,69 @@
...
+                var result = new BinaryData(messageText);
+
+                return ConversionResult.Success(result);
+            }
+            catch (JsonException ex)
+            {
+                string msg = String.Format(CultureInfo.CurrentCulture,
+                    @"Binding parameters to complex objects uses Json.NET serialization.
+                    1. Bind the parameter type as 'string' instead to get the raw values and avoid JSON deserialization, or
+                    2. Change the queue payload to be valid json. The JSON parser failed: {0}",
```
jviau, line 44: New line maybe?
```suggestion
                    2. Change the queue payload to be valid json. 
                    The JSON parser failed: {0}",
```

extensions/Worker.Extensions.Storage.Queues/src/TypeConverters/QueueMessageConverter.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            try
+            {
+                var modelBindingData = (ModelBindingData)context.Source!;
+                QueueMessage queueMessage = ExtractQueueMessage(modelBindingData);
+                return new ValueTask<ConversionResult>(ConversionResult.Success(queueMessage));
+            }
+            catch (JsonException ex)
+            {
+                string msg = String.Format(CultureInfo.CurrentCulture,
```
jviau, line 43: There is a fair bit of common logic between this `ConvertAsync` and `QueueMessageBinaryDataConverter`'s implementation. You could extract the common logic to the base class you already have, then add `protected abstract` methods for the specialized logic.

``` CSharp
abstract class QueueMessageConverterBase
{
    public ValueTask<ConversionResult> ConvertAsync(...)
    {
        // common logic
    }

    protected abstract ValueTask<T> ConvertCoreAsync(ModelBindingData data);
}
```

extensions/Worker.Extensions.Storage.Queues/src/QueueMessageJsonConverter.cs
liliankasem: I'm not 100% sure. My initial though is no, this should always be CamelCase - but I don't think it hurts to account for insensitive casing?

extensions/Worker.Extensions.Storage.Queues/src/QueueMessageJsonConverter.cs
liliankasem: I think you're right, JsonException is more fitting since it's specifically designed for handling JSON-related errors; plus it lets us catch and handle the exception specifically related to JSON processing which is better than just throwing unimplemented - thanks for catching this!

extensions/Worker.Extensions.Storage.Queues/src/TypeConverters/QueueConverterBase.cs
liliankasem: Noted for future reference! But removing for now as unused

extensions/Worker.Extensions.Storage.Queues/src/TypeConverters/QueueMessageConverter.cs
liliankasem: great suggestion - thanks!

surgupta-msft (approved):

PR #1475, opened 2023-04-17
[Enhanced HTTP] Add handling for IActionResult

satvu:

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/extensions/Worker.Extensions.Http.AspNetCore/src/FunctionsMiddleware/FunctionsHttpProxyingMiddleware.cs
+++ b/extensions/Worker.Extensions.Http.AspNetCore/src/FunctionsMiddleware/FunctionsHttpProxyingMiddleware.cs
@@ -38,7 +38,7 @@ namespace Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore
             }
 
             // allows asp.net middleware to continue
-            _coordinator.CompleteFunctionInvocation(invocationId, context);
+            _coordinator.CompleteFunctionInvocation(invocationId);
         }
 
         private static void AddHttpContextToFunctionContext(FunctionContext funcContext, HttpContext httpContext)
--- a/extensions/Worker.Extensions.Http.AspNetCore/src/FunctionsMiddleware/FunctionsHttpProxyingMiddleware.cs
+++ b/extensions/Worker.Extensions.Http.AspNetCore/src/FunctionsMiddleware/FunctionsHttpProxyingMiddleware.cs
@@ -3,6 +3,9 @@
 
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.Abstractions;
+using Microsoft.AspNetCore.Routing;
 using Microsoft.Azure.Functions.Worker.Middleware;
 
 namespace Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore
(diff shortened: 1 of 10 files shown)


extensions/Worker.Extensions.Http.AspNetCore/src/HttpUriProvider.cs
```diff
@@ -21,7 +21,9 @@ public static Uri GetHttpUri()
             // TODO: replace local host string
             var uriString = "http://localhost:" + HttpPort.Value.ToString();
 
-            return new Uri(uriString);
+            _httpUri = new Uri(uriString);
```
satvu, line 24: bug, unrelated to IActionResult but value wasn't being cached

extensions/Worker.Extensions.Http.AspNetCore/src/WorkerBuilderExtensions.cs
```diff
@@ -31,7 +31,7 @@ public static IFunctionsWorkerApplicationBuilder UseAspNetCoreIntegration(this I
 
             builder.Services.AddSingleton<IHttpCoordinator, DefaultHttpCoordinator>();
 
-            var port = Utilities.GetUnusedTcpPort().ToString();
```
satvu, line 34: wasn't being used

extensions/Worker.Extensions.Http.AspNetCore/src/HttpUriProvider.cs
brettsam: I noticed this as well. But another idea is to use a Lazy<string> instead of Lazy<int> and to the entire calculation and ToString() there -- then the value is cached in the Lazy instead of having to call ToString() multiple times.

extensions/Worker.Extensions.Http.AspNetCore/src/FunctionsMiddleware/FunctionsHttpProxyingMiddleware.cs
```diff
@@ -27,6 +30,13 @@ public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next
 
             await next(context);
 
+            if (context.GetInvocationResult()?.Value is IActionResult actionResult)
+            {
+                ActionContext actionContext = new ActionContext(httpContext, httpContext.GetRouteData(), new ActionDescriptor());
```
jviau, line 35: Do we have plans to fill out `ActionDescriptor`?

satvu: Discussed at sync yesterday and no, we do not. Looking at [similar code](https://github.com/Azure/azure-functions-host/blob/76591f722931f651caddeb0f6080d934545246dd/src/WebJobs.Script.WebHost/Middleware/FunctionInvocationMiddleware.cs#L73) in the host side and we do not fill it out either.

jviau: If we don't do it in host code I am fine leaving this empty then.

I do wonder what functionality we are excluding here? `ActionDescriptor` essentially holds a breakdown the MVC endpoint that is being ran. We should consider populating it in the future.

fabiocav: There's no action associated with the context and we have a terminating middleware, no action selection will be performed. As @satvu pointed out, that's the same pattern used by the host.

extensions/Worker.Extensions.Http.AspNetCore/src/HttpUriProvider.cs
```diff
@@ -7,26 +7,18 @@ namespace Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore
...
     internal static class HttpUriProvider
     {
-        private static Uri? _httpUri;
+        public static Lazy<Uri> HttpUri = new Lazy<Uri>(() => GetHttpUri());
 
         public static Lazy<int> HttpPort = new Lazy<int>(() => Utilities.GetUnusedTcpPort());
 
+        public static Lazy<string> HttpUriString = new Lazy<string>(() => HttpUri.Value.ToString());
+
         public static Uri GetHttpUri()
```
brettsam, line 16: let's simplify this API:
- one public member -- `public static string GetHttpUri()` (since we never use the `Uri` -- only the string)
- one private lazy `HttpUriString` that does all the magic of calling `GetUnusedTcpPort()` and building the uri in its factory
- have `GetHttpUri()` return  `HttpUriString.Value`

brettsam (approved):

PR #1476, opened 2023-04-18
Fix null reference error if there's no built in retry support

### LINKED ISSUE #1438: Accessing to RetryContext.RetryCount and MaxRetryCount throw exception
In an isolated function for EventGridTrigger accessing FunctionContext.RetryContext.RetryCount and MaxRetryCount throw nullreferenceexception 

context.RetryContext?.RetryCount -> nullreferenceexception

aishwaryabh:
`context.RetryCount` and `context.MaxRetryCount` throw null exceptions since the `RetryContext` passed into `GrpcRetryContext` is set to null. This PR addresses that bug and adds a unit test.

resolves #1438

### Pull request checklist


src/DotNetWorker.Core/Context/RetryContext.cs
```diff
@@ -12,12 +12,12 @@ public abstract class RetryContext
         /// <summary>
         /// The the retry count for the current event.
         /// </summary>
-        public abstract int RetryCount { get; }
+        public abstract int? RetryCount { get; }
```
jviau, line 15: Unfortunately, this is a breaking change (as well as the `MaxRetryCount` change) - the type is going from `int` to `Nullable<int>`. We will need to find a way to solve this without breaking changes to the public API surface.

https://learn.microsoft.com/en-us/dotnet/core/compatibility/library-change-rules

src/DotNetWorker.Grpc/GrpcRetryContext.cs
```diff
@@ -14,8 +14,8 @@ public GrpcRetryContext(RetryContextMessage retryContext)
             _retryContext = retryContext;
         }
 
-        public override int RetryCount => _retryContext.RetryCount;
+        public override int? RetryCount => (_retryContext != null)?_retryContext.RetryCount: null;
 
-        public override int MaxRetryCount => _retryContext.MaxRetryCount;
-    }
+        public override int? MaxRetryCount => (_retryContext != null) ? _retryContext.MaxRetryCount: null;
+    }
```
jviau, line 20: nit: remove whitespace

```suggestion
    }
```

src/DotNetWorker.Grpc/GrpcRetryContext.cs
```diff
@@ -14,8 +14,8 @@ public GrpcRetryContext(RetryContextMessage retryContext)
             _retryContext = retryContext;
         }
 
-        public override int RetryCount => _retryContext.RetryCount;
+        public override int? RetryCount => (_retryContext != null)?_retryContext.RetryCount: null;
```
jviau, line 17: nit: parenthesis is not necessary, also spacing consistency

```suggestion
        public override int? RetryCount => _retryContext != null ?_retryContext.RetryCount : null;
```

jviau (changes requested):
See my comment regarding breaking changes. We can only make breaking changes between major version releases, and even then, they have to be considered carefully and go through an approval process. They are _very_ annoying to customers.

src/DotNetWorker.Core/Context/RetryContext.cs
aishwaryabh: I changed the value to be -1 if the RetryContext is null.

src/DotNetWorker.Grpc/GrpcRetryContext.cs
```diff
@@ -14,8 +14,8 @@ public GrpcRetryContext(RetryContextMessage retryContext)
             _retryContext = retryContext;
         }
 
-        public override int RetryCount => _retryContext.RetryCount;
+        public override int RetryCount => _retryContext != null? _retryContext.RetryCount: -1;
```
jviau, line 17: Can also use null-coalesce:

```suggestion
        public override int RetryCount => _retryContext?.RetryCount ?? -1;
```

jviau (approved):
Approved from my end, but I'd like to have others weigh in on if `-1` is the right approach here.

fabiocav (approved):

fabiocav (changes requested):
The value is not used to drive any logic in the worker. This exposes the retry information to be used in user code logic.

I had a couple of other ideas for the default value, but thinking more about this, the approach here should be to have parity with the in-proc experience to avoid surprises when migrating between the models. 

-1 has a specific documented meaning for the maximum retries (infinite) so likely not what we want to be using here.

I believe the in-process model will actually have a `null` `RetryContext` when policies are not supported. @aishwaryabh , can you please validate that and if that is the case, let’s decide how we want to apply changes here.

something we also want to do as part of the change, once we decide what we want to do. is update the public XML docs to make sure the behavior is clear.

aishwaryabh:
@fabiocav I validated that the RetryContext is set to null in the in-proc experience. I'm assuming we'd also want to set the `RetryContext` to be null for the out of proc experience as well then

fabiocav (approved):

src/DotNetWorker.Grpc/GrpcFunctionInvocation.cs
```diff
@@ -22,6 +22,6 @@ public GrpcFunctionInvocation(InvocationRequest invocationRequest)
 
         public override TraceContext TraceContext { get; }
 
-        public RetryContext Context => _retryContext ??= new GrpcRetryContext(_invocationRequest.RetryContext);
+        public RetryContext Context => _retryContext ??= (_invocationRequest.RetryContext == null ? null : new GrpcRetryContext(_invocationRequest.RetryContext));
```
fabiocav, line 25: Minor update that will simplify and remove the null check issue:
``` csharp
public RetryContext Context => _retryContext ??= _invocationRequest.RetryContext == null ? null! : new GrpcRetryContext(_invocationRequest.RetryContext);
```

PR #1489, opened 2023-04-25
[SignalRService] Add Serverless Hub

Y-Sindo:

#771 

### How to use serverless hub:
See this [commit ](https://github.com/aspnet/AzureSignalR-samples/commit/af330990c169aecc27572e0bba3af4e6d57bb588)for an example. It shows the code changes after applying `ServerlessHub<T>`.

### Pull request checklist

  * [x] Otherwise: [Documentation issue linked to PR](https://github.com/MicrosoftDocs/azure-docs/issues/108715)
* [ ] I have added all required tests (Unit tests, E2E tests)  **I don't know where to add tests for extensions**


extensions/Worker.Extensions.SignalRService/src/Hubs/ServiceHubContextProvider.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        protected ServiceManager CreateServiceManager()
+        {
+            var optionsSetup = new ServiceManagerOptionsSetup(_configuration, _azureComponentFactory, typeof(THub).GetCustomAttribute<ServerlessHub.SignalRConnectionAttribute>(true)?.ConnectionName ?? Constants.AzureSignalRConnectionStringName);
+            var serviceManagerBuilder = new ServiceManagerBuilder()
+                .WithOptions(optionsSetup.Configure)
+                .WithLoggerFactory(_loggerFactory)
+                .WithConfiguration(_configuration)
+                .WithCallingAssembly()
+                .AddUserAgent($" [{Constants.FunctionsWorkerProductInfoKey}={Constants.DotnetIsolatedWorker}");
```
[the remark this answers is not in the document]
Y-Sindo: Yes, the whitespace is used to separate the key value pairs and we now depend on it to parse the data.

extensions/Worker.Extensions.SignalRService/src/Configuration/IConfigurationExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Azure.SignalR;
+using Microsoft.Extensions.Azure;
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.Azure.Functions.Worker.SignalRService
+{
+    internal static class IConfigurationExtensions
```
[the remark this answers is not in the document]
Y-Sindo: fixed.

extensions/Worker.Extensions.SignalRService/src/Configuration/IConfigurationExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        public static IEnumerable<ServiceEndpoint> GetNamedEndpointsFromConnectionString(this IConfigurationSection section)
+        {
+            var endpointName = section.Key;
+            if (section.Value != null)
+            {
+                yield return new ServiceEndpoint(section.Key, section.Value);
+            }
+
+            if (section["primary"] != null)
```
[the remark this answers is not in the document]
Y-Sindo: Fixed.

extensions/Worker.Extensions.SignalRService/src/Configuration/ServiceManagerOptionsSetup.cs
```diff
@@ -0,0 +1,59 @@
...
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Azure.Functions.Worker.SignalRService
+{
+    internal class ServiceManagerOptionsSetup
+    {
+        private readonly IConfiguration _configuration;
+        private readonly Action<ServiceManagerOptions> _configureServiceManagerOptions;
+
+        public ServiceManagerOptionsSetup(IConfiguration configuration, AzureComponentFactory azureComponentFactory, string connectionStringKey)
```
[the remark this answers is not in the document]
jviau: `IConfigureOptions` is not to be added directly to `ServiceManagerBuilder`. You use `IConfigureOptions<ServiceManagerOptions>` to configure your `ServiceManagerOptions`, and then import `IOptions<ServiceManagerOptions>` and provide the value of that to your `ServiceManagerBuilder`.

extensions/Worker.Extensions.SignalRService/src/Hubs/ServiceHubContextProviderOfT.cs
```diff
@@ -0,0 +1,43 @@
...
+            ServiceHubContext = await serviceManager.CreateHubContextAsync<T>(typeof(THub).Name, cancellationToken);
+        }
+
+        public override async Task StopAsync(CancellationToken cancellationToken)
+        {
+            if (ServiceHubContext != null)
+            {
+                await ServiceHubContext.DisposeAsync();
+            }
+            else
```
jviau, line 37: nit: `else` statement can be removed entirely.

``` CSharp
public override async Task StopAsync(CancellationToken cancellationToken)
{
    if (ServiceHubContext != null)
    {
        await ServiceHubContext.DisposeAsync();
    }
}
```

extensions/Worker.Extensions.SignalRService/src/Hubs/ServerlessHubOfT.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            get
+            {
+                var type = GetType();
+                var hubContextCache = _serviceProvider.GetRequiredService<HubContextCache>();
+                if (hubContextCache.TryGetValue(type, out var hubContext))
+                {
+                    return (ServiceHubContext<T>)hubContext;
+                }
+                var hubContextProviderType = typeof(ServiceHubContextProvider<,>).MakeGenericType(type, typeof(T));
```
jviau, line 31: Curious what the point of this reflection is? Is it to allow for `ServerlessHub<T>` to be derived and still retrieve the exact type from the cache?

I think this can be achieved without reflection, instead since this is all internal and you exclusively control this code, we can rely on returning a lower type and then casting up to the expected type.

``` CSharp
var provider = (ServiceHubContextProvider)_serviceProvider.GetRequiredService(hubContextProviderType);
return (ServiceHubContext<T>)provider.ServiceHubContext;
```

Would then need to also address the lack of common inheritance chain of `ServiceHubContext` and `ServiceHubContext<T>`. Again, since it is all `internal` you could just change the contract to `public object ServiceHubContext { get; }` and rely on all usage to upcast to their expected type.

extensions/Worker.Extensions.SignalRService/src/Hubs/ServerlessHubOfT.cs
```diff
@@ -0,0 +1,85 @@
...
+namespace Microsoft.Azure.Functions.Worker.SignalRService
+{
+    public class ServerlessHub<T> where T : class
+    {
+        private static readonly ObjectSerializer JsonObjectSerializer = new JsonObjectSerializer(new(JsonSerializerDefaults.Web));
+        private static readonly NegotiationOptions DefaultNegotiateOptiosn = new();
+
+        private readonly IServiceProvider _serviceProvider;
+
+        private ServiceHubContext<T> HubContext
```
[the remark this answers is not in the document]
Y-Sindo: fixed

extensions/Worker.Extensions.SignalRService/src/Hubs/ServerlessHubOfT.cs
```diff
@@ -0,0 +1,85 @@
...
+using Azure.Core.Serialization;
+using Microsoft.AspNetCore.SignalR;
+using Microsoft.Azure.SignalR.Management;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Azure.Functions.Worker.SignalRService
+{
+    public class ServerlessHub<T> where T : class
+    {
+        private static readonly ObjectSerializer JsonObjectSerializer = new JsonObjectSerializer(new(JsonSerializerDefaults.Web));
```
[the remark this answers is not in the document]
Y-Sindo: fixed

extensions/Worker.Extensions.SignalRService/src/Hubs/ServiceHubContextProviderOfT.cs
Y-Sindo: fixed

extensions/Worker.Extensions.SignalRService/src/Hubs/ServerlessHubOfT.cs
Y-Sindo: As there might be multiple hubs, there might be multiple `ServiceHubContextProvider` instances in the service provider. The usage of generic parameters of `ServiceHubContextProvider` is to allow multiple such instances coexist in the service provider. Therefore, `ServiceHubContextProvider` has to be generic. However, I think it's still doable to remove the reflection. The updated code always retrieves the hub contexts from `HubContextCache`, whose content are filled by `ServiceHubContextProvider` during the `IHostedService.StartAsync`. 

I have a new question now. Should we make the initialization of `ServiceHubContext` lazy and initialize it when the users first request it? Users don't need to register their serverless hubs with `services.AddServerlessHub`. However, the problem is the initialization of `ServiceHubContext` is async, and it seems that we have to either force the async to be sync or wrap all the properties depending on hub context with `Task<>`. If there are other mechanisms in dotnet worker helping to solve the async problem more elegantly that'd be great.

extensions/Worker.Extensions.SignalRService/src/Configuration/ServiceManagerOptionsSetup.cs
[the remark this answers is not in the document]
Y-Sindo: Get it. Will update later.

extensions/Worker.Extensions.SignalRService/src/Hubs/ServerlessHubOfT.cs
jviau: In my suggestion you are still importing the closed-generic type `hubContextProviderType` (this type can be saved as an instance field, so you only need to call `MakeGenericType` once). But the change you made recently addresses the reflection problem well.

As for your other question. Yes, the `IHostedService` approach, and async initialization, is not ideal. But there isn't any single prescribed pattern.  The main drawback of the `IHostedService` approach is if anyone tries to use `ServiceHubContext` before application is fully started, as they may not be initialized yet. If you are fine with that, then you can keep the `IHostedService` approach.

An alternative would be to keep `IHostedService` approach, but then have all the accesses become sync-over-async if they try to access the context before it has been initialized in the background.

extensions/Worker.Extensions.SignalRService/src/Hubs/ServerlessHubOfT.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        /// <summary>
+        /// Gets an object that can be used to invoke methods on the clients connected to this hub.
+        /// </summary>
+        protected new IHubClients<T> Clients => HubContext.Clients;
+
+        /// <summary>
+        /// Get the group manager of this hub.
+        /// </summary>
+        protected new GroupManager Groups => HubContext.Groups;
```
jviau, line 44: For all of these properties where the types line up between base class and this class, can `virtual` and `override` be used instead?

extensions/Worker.Extensions.SignalRService/src/Hubs/ServerlessHub.cs
```diff
@@ -0,0 +1,84 @@
...
+
+#nullable enable
+
+namespace Microsoft.Azure.Functions.Worker.SignalRService
+{
+    public abstract class ServerlessHub
+    {
+        protected ServerlessHub(IServiceProvider serviceProvider)
+        {
+            ServiceProvider = serviceProvider;
```
jviau, line 20: Can `HubContextProvider` be retrieved from `IServiceProvider` directly here?

extensions/Worker.Extensions.SignalRService/src/Hubs/ServerlessHubOfT.cs
```diff
@@ -0,0 +1,65 @@
...
+using System;
+using System.Threading.Tasks;
+using Azure.Core.Serialization;
+using Microsoft.AspNetCore.SignalR;
+using Microsoft.Azure.SignalR.Management;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Azure.Functions.Worker.SignalRService
+{
+    public abstract class ServerlessHub<T> : ServerlessHub where T : class
```
jviau, line 13: Who is meant to derive from this type and `ServerlessHub`? If it is customers, it won't be unit testable since they will never be able to setup `IServiceProvider` correctly with `HubContextProvider` (since it is internal).

extensions/Worker.Extensions.SignalRService/src/Configuration/ServiceManagerOptionsSetup.cs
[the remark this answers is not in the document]
jviau: When/if you try this, I think you will want to use named options and use `connectionStringKey` as the name. Then you can import `IOptionsMonitor<ServiceManagerOptions>` and call `.Get(connectionStringKey)` or `.CurrentValue` for the default `connectionStringKey`.

extensions/Worker.Extensions.SignalRService/src/Hubs/ServerlessHubOfT.cs
Y-Sindo: Sure, `virtual` also allows the users to mock the impl and unit test it.

extensions/Worker.Extensions.SignalRService/src/Hubs/ServerlessHubOfT.cs
Y-Sindo: It is customers to derive from this type. Now the `HubContextProvider` is optional, and they can override those properties such as `Clients`, `GroupManager` for unit tests.

extensions/Worker.Extensions.SignalRService/src/Configuration/ServiceManagerOptionsSetup.cs
```diff
@@ -0,0 +1,73 @@
...
+
+namespace Microsoft.Azure.Functions.Worker.SignalRService
+{
+    internal class ServiceManagerOptionsSetup
+    {
+        private readonly IConfiguration _configuration;
+        private readonly AzureComponentFactory _azureComponentFactory;
+        private readonly IEnumerable<IConfigureOptions<ServiceManagerOptions>> _configureOptions;
+
+        public ServiceManagerOptionsSetup(IConfiguration configuration, AzureComponentFactory azureComponentFactory, IEnumerable<IConfigureOptions<ServiceManagerOptions>> configureOptions)
```
jviau, line 21: You shouldn't need to import `IConfigureOptions` yourself. You import `IOptionsMonitor<ServiceManagerOptions>`, and the applicable configurations will be ran for you when retrieving the options object.

https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-7.0

This is for aspnetcore, but applies to generic host as well.

extensions/Worker.Extensions.SignalRService/src/SignalRServiceDependencyInjectionExtensions.cs
```diff
@@ -0,0 +1,55 @@
...
+using Microsoft.Azure.Functions.Worker.SignalRService;
+using Microsoft.Extensions.Azure;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Hosting;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    public static class SignalRServiceDependencyInjectionExtensions
+    {
+        public static IServiceCollection AddServerlessHub<THub>(this IServiceCollection services) where THub : ServerlessHub
```
jviau, line 16: Something to consider is also adding a `AddServerlessHub(this IServiceCollection services, Type hubType)` - and verifying that `hubType` is an expected `ServerlessHub` implementation. This is a small convenience API for when customers are doing some more advanced scenario where they are resolving their serverless hub types at app startup.

extensions/Worker.Extensions.SignalRService/src/Hubs/ServiceHubContextInitializerOfT.cs
```diff
@@ -0,0 +1,38 @@
...
+#nullable enable
+
+namespace Microsoft.Azure.Functions.Worker.SignalRService
+{
+    internal class ServiceHubContextInitializer<THub, T> : ServiceHubContextInitializer<THub> where THub : ServerlessHub<T>
+        where T : class
+    {
+        private ServiceHubContext<T>? _serviceHubContext;
+
+        public ServiceHubContextInitializer(IConfiguration configuration, ILoggerFactory loggerFactory, HubContextProvider hubContextProvider, ServiceManagerOptionsSetup optionSetup) : base(configuration, loggerFactory, hubContextProvider, optionSetup)
```
jviau, line 19: nit: put `base` on new line
```suggestion
        public ServiceHubContextInitializer(IConfiguration configuration, ILoggerFactory loggerFactory, HubContextProvider hubContextProvider, ServiceManagerOptionsSetup optionSetup)
            : base(configuration, loggerFactory, hubContextProvider, optionSetup)
```

extensions/Worker.Extensions.SignalRService/src/Hubs/ServiceHubContextInitializerOfT.cs
```diff
@@ -0,0 +1,38 @@
...
+            using var serviceManager = CreateServiceManager();
+            _serviceHubContext = await serviceManager.CreateHubContextAsync<T>(typeof(THub).Name, cancellationToken);
+            HubContextProvider.Add(typeof(THub), _serviceHubContext);
+        }
+
+        public override async Task StopAsync(CancellationToken cancellationToken)
+        {
+            if (_serviceHubContext != null)
+            {
+                await _serviceHubContext.DisposeAsync();
```
jviau, line 34: I believe you should also make this type implement `IAsyncDisposable` and dispose `_serviceHubContext` if one is set.

There is an analyzer for this, but we may have it turned off. The gist is, if you ever hold disposables as fields, you must also be disposable yourself.

extensions/Worker.Extensions.SignalRService/src/Hubs/ServerlessHubOfT.cs
jviau: Okay - if customers are to derive from this type, I recommend (if you haven't already) you to see what the experience is yourself. Particularly around how easy is it to create a derived type and then unit test that type. Particularly without having access to any of the `internal` types.

extensions/Worker.Extensions.SignalRService/src/Hubs/ServerlessHubOfT.cs
Y-Sindo: Make the `IHostedService` `IDisposable`, as `ServiceHubContext` is not `IAsyncDisposable`( The return type of `DisposeAsync` is Task instead of ValueTask due to historical reason).

extensions/Worker.Extensions.SignalRService/src/Configuration/ServiceManagerOptionsSetup.cs
Y-Sindo: `IConfigureOptions` is used instead of `IOptionsMonitor` because `ServiceManagerOptions` is not a typical options structure. It has a method called `UseJsonObjectSerializer(ObjectSerializer)`, where the object serializer passed in the parameter is recorded in an internal field and we cannot get it back from outside. Therefore, I have to import the `ICongureOptions` myself.

extensions/Worker.Extensions.SignalRService/src/Hubs/ServiceHubContextInitializerOfT.cs
Y-Sindo: updated

extensions/Worker.Extensions.SignalRService/src/Hubs/ServiceHubContextInitializerOfT.cs
Y-Sindo: `ServiceHubContext` is not `IAsyncDisposable`( The return type of `DisposeAsync` is Task instead of ValueTask due to historical reason), so I implement `IDisposable`. Also refactor a little bit.

Y-Sindo:
@jviau
Do we have any pending problems?

jviau:
@Y-Sindo apologies, forgot to come back to this. I will make sure to take a look early next week. I think I wanted to take a closer look at the `IOptions` approach, see if there is an opportunity to more closely align with the established patterns.

extensions/Worker.Extensions.SignalRService/src/Hubs/ServerlessHub.cs
```diff
@@ -0,0 +1,69 @@
...
+
+#nullable enable
+
+namespace Microsoft.Azure.Functions.Worker.SignalRService
+{
+    public abstract class ServerlessHub
+    {
+        internal static readonly ObjectSerializer ObjectSerializer = new JsonObjectSerializer(new(JsonSerializerDefaults.Web));
+        internal static readonly NegotiationOptions DefaultNegotiateOptiosn = new();
+        internal readonly object? _hubContext;
```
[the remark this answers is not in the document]
Y-Sindo: I'm afraid no, because `ServiceHubContext<T>` isn't derived from `ServiceHubContext`.

extensions/Worker.Extensions.SignalRService/src/Hubs/ServerlessHub.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        protected ServerlessHub(IServiceProvider serviceProvider)
+        {
+            serviceProvider.GetService<HubContextProvider>()?.TryGetValue(GetType(), out _hubContext);
```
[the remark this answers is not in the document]
Y-Sindo: This line won't throw exception if the `HubContextProvider` is absent. In testing scenario, the users don't need to add `HubContextProvider` to the services, they just need to override the implementation of those protected members such as `Clients`, `Groups`.

extensions/Worker.Extensions.SignalRService/src/Configuration/ServiceManagerOptionsSetup.cs
```diff
@@ -0,0 +1,54 @@
...
+                }
+                else if (!string.IsNullOrWhiteSpace(serviceTransportTypeStr))
+                {
+                    throw new InvalidOperationException($"Invalid service transport type: {serviceTransportTypeStr}.");
+                }
+                //make connection more stable
+                options.ConnectionCount = 3;
+            };
+        }
+        public Action<ServiceManagerOptions> Configure(string connectionStringKey) =>
```
jviau, line 51: nit: new line

```suggestion

        public Action<ServiceManagerOptions> Configure(string connectionStringKey) =>
```

extensions/Worker.Extensions.SignalRService/src/SignalRServiceDependencyInjectionExtensions.cs
```diff
@@ -0,0 +1,74 @@
...
+        public static IServiceCollection AddServerlessHub<THub>(this IServiceCollection services, Action<ServiceManagerBuilder> configure)
+            where THub : ServerlessHub
+            => services.AddServerlessHub(typeof(THub), configure);
+
+        public static IServiceCollection AddServerlessHub(this IServiceCollection services, Type hubType)
+           => services.AddServerlessHub(hubType, _ => { });
+
+        public static IServiceCollection AddServerlessHub(this IServiceCollection services, Type hubType, Action<ServiceManagerBuilder> configure)
+        {
+            var allBaseTypes = hubType.AllBaseTypes().ToArray();
```
jviau, line 30: I like to use this API for runtime type checks:

``` CSharp
typeof(ServerlessHub).IsAssignableFrom(hubType)
```

This will check that `hubType` derives from `ServerlessHub`

jviau (approved):

extensions/Worker.Extensions.SignalRService/src/SignalRServiceDependencyInjectionExtensions.cs
Y-Sindo: Updated

extensions/Worker.Extensions.SignalRService/src/Configuration/ServiceManagerOptionsSetup.cs
```diff
@@ -0,0 +1,55 @@
...
+                if (Enum.TryParse<ServiceTransportType>(serviceTransportTypeStr, out var transport))
+                {
+                    options.ServiceTransportType = transport;
+                }
+                else if (!string.IsNullOrWhiteSpace(serviceTransportTypeStr))
+                {
+                    throw new InvalidOperationException($"Invalid service transport type: {serviceTransportTypeStr}.");
+                }
+                //make connection more stable
+                options.ConnectionCount = 3;
```
satvu, line 48: Could you expand on this comment since it seems to explain the hard-coded value?

satvu (approved):

extensions/Worker.Extensions.SignalRService/src/Configuration/ServiceManagerOptionsSetup.cs
Y-Sindo: Hi @satvu , I have added more explanations in the comment.

release_notes.md
```diff
@@ -3,6 +3,8 @@
 <!-- Please add your release notes in the following format:
 - My change description (#PR/#issue)
 -->
+### Microsoft.Azure.Functions.Worker.Extensions.SignalRService <version>
```
jviau, line 6: Extensions are released independently and have their own release notes. Please create a `extensions/Worker.Extensions.SignalRService/release_notes.md` and move these changes there.

See an example here: https://github.com/Azure/azure-functions-dotnet-worker/blob/main/extensions/Worker.Extensions.CosmosDB/release_notes.md

liliankasem: SignalR should already have its own release note file

liliankasem (approved):

release_notes.md
Y-Sindo: Updated

jviau (approved):

PR #1505, opened 2023-05-02
[Analyzer] Warning for extension constrained types

### LINKED ISSUE #1144: Implement analyzer to validate what types are supported by extensions
Implement the analyzer which validates what types are supported by the extensions.

Related design issue - https://github.com/Azure/azure-functions-dotnet-worker/issues/1036

Design doc - [Link](https://dev.azure.com/msazure/One/_git/AAPT-Antares-Docs?version=GBmaster&path=/TeamDocs/FunctionTeamDocs/Design/OOPWorkers/SdkTypeBindings/ConverterTypesValidationAnalyzer.md)

liliankasem:

resolves #1144

An analyzer that gives a warning if a function is binding to a type that is not supported by the binding being used. This rule only applies if 

a) the converter on the attribute is using the supported converter types attribute to advertise supported types and
b) the binding being used does not allow fallback to built-in converters

Outside of the above scenario, this analyzer does not come into play

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/docs/analyzer-rules/AZFW0010.md
+++ b/docs/analyzer-rules/AZFW0010.md
@@ -1,0 +1,24 @@
+# AZFW0010: Invalid binding type
+
+| | Value |
+|-|-|
+| **Rule ID** |AZFW00010|
+| **Category** |[Usage]|
+| **Severity** |Warning|
+
+## Cause
+
+This rule is triggered when a function is binding to a type that is not supported
+by the binding attribute being used.
+
+## Rule description
+
+Some bindings advertise the types that they support. If a binding advertises supported types
+and does not support falling back to built-in converters, then this rule will will be flagged.
+
+For example if you're binding to a `QueueTrigger` and the only supported type is `string` and
+your function is binding to a `bool`, then you will see this warning.
+
+## How to fix violations
+
+Change the binding type to a type that is supported by the binding your function is using.
--- a/docs/analyzer-rules/AZFW0010.md
+++ b/docs/analyzer-rules/AZFW0010.md
@@ -22,3 +22,21 @@ your function is binding to a `bool`, then you will see this warning.
 ## How to fix violations
 
 Change the binding type to a type that is supported by the binding your function is using.
+
+### Supported Types
+
+You can refer to the [public documentation](https://learn.microsoft.com/azure/azure-functions/functions-bindings-storage-blob?tabs=in-process%2Cextensionv5%2Cextensionv3&pivots=programming-language-csharp) to see which types are supported.
+
+| Binding | Supported Types |
+| ------- | --------------- |
+| BlobTrigger | POCO, String, Stream, Byte[], BlobClient*, BlobContainerClient |
+| BlobInput | POCO, IEnumerable*<POCO>, String,  IEnumberable<String>, Stream, IEnumberable<Stream>, Byte[], BlobClient, IEnumerable<BlobClient>, BlobContainerClient |
+| CosmosDBTrigger | POCO, IEnumerable<POCO> |
+| CosmosDBInput | POCO, IEnumerable<POCO>, CosmosClient, Database, Container |
+| EventGridTrigger | String, IEnumerable<String>, CloudEvent, IEnumerable<CloudEvent>, EventGridEvent, IEnumerable<EventGridEvent> |
+| EventHubTrigger | POCO, POCO[], String, String[] |
+| QueueTrigger | POCO, String, BinaryData, QueueMessage |
+| ServiceBusTrigger | String, String[], ServiceBusReceivedMessage, ServiceBusReceivedMessage[] |
+| TableInput | POCO, IEnumerable<POCO>, TableEntity, IEnumerable<TableEntity>, TableClient |
+
+> \* Where `BlobClient` is supported AppendBlobClient, BaseBlobClient, BlockBlobClient and PageBlobClient are also supported
--- a/sdk/Sdk.Analyzers/BindingTypeNotSupported.cs
+++ b/sdk/Sdk.Analyzers/BindingTypeNotSupported.cs
@@ -1,0 +1,133 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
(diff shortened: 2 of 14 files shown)


sdk/Sdk.Analyzers/BindingTypeNotSupported.cs
```diff
@@ -0,0 +1,96 @@
...
+                        var converter = context.Compilation.GetTypeByMetadataName(converterName);
+
+                        var converterAttributes = converter.GetAttributes();
+
+                        supportedTypes.AddRange(converterAttributes
+                                                    .Where(a => a.AttributeClass.Name == Constants.Names.SupportedConverterTypeAttribute)
+                                                    .SelectMany(a => a.ConstructorArguments.Select(arg => arg.Value))
+                                                    .ToList());
+
+                        var converterSupportsJsonDeserialization = converterAttributes.Any(a => a.AttributeClass.Name == Constants.Names.SupportsJsonDeserializationAttribute);
```
liliankasem, line 72: Unsure of how to also check for IEnumerable as well? Just because a converter supports JsonDeserialization, it doesn't mean it supports lists?

sdk/Sdk.Analyzers/BindingTypeNotSupported.cs
```diff
@@ -0,0 +1,133 @@
...
+
+                ReportDiagnostic(context, parameter, attributeType);
+            }
+        }
+
+        private static List<AttributeData> GetInputConverterAttributes(SymbolAnalysisContext context, ITypeSymbol attributeType)
+        {
+            var inputConverterAttributeType = context.Compilation.GetTypeByMetadataName(Constants.Types.InputConverterAttribute);
+            return attributeType.GetAttributes()
+                .Where(attr => attr.AttributeClass.Equals(inputConverterAttributeType))
```
kshyju, line 84: Switch to SymbolEqualityComparer instead of Equals.

Example: https://github.com/Azure/azure-functions-dotnet-worker/blob/a625cc1efc49d37611cbc35e88dbd376b073e49a/sdk/Sdk.Generators/FunctionsUtil.cs#L34

kshyju: Same comment for line 91

sdk/Sdk.Analyzers/BindingTypeNotSupported.cs
```diff
@@ -0,0 +1,133 @@
...
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.Azure.Functions.Worker.Sdk.Analyzers
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class BindingTypeNotSupported : DiagnosticAnalyzer
+    {
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get { return ImmutableArray.Create(DiagnosticDescriptors.BindingTypeNotSupported); } }
```
jviau, line 15:
```suggestion
        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(DiagnosticDescriptors.BindingTypeNotSupported);
```

What you have now will create a new `ImmutableArray` on every `get` on this property - which is potentially a lot of extra objects and GC. The above syntax I suggested will instantiate the array once and return it for every `get`.

sdk/Sdk.Analyzers/BindingTypeNotSupported.cs
```diff
@@ -0,0 +1,133 @@
...
+
+                var converterHasSupportedTypeAttribute = converterAttributes.Any(a => a.AttributeClass.Name == Constants.Names.SupportedConverterTypeAttribute);
+                if (!converterHasSupportedTypeAttribute)
+                {
+                    // If a converter does not have the `SupportedConverterTypeAttribute`, we don't need to check for supported types
+                    continue;
+                }
+
+                supportedTypes.AddRange(converterAttributes
+                    .Where(a => a.AttributeClass.Name == Constants.Names.SupportedConverterTypeAttribute)
```
jviau, line 114: I haven't written an analyzer before but is it okay to be comparing on just name. Should namespace be included in the comparison also?

jviau (approved):

sdk/Sdk.Analyzers/BindingTypeNotSupported.cs
kshyju: Good catch. 

I think our existing analyzers are already doing that.

docs/analyzer-rules/AZFW0010.md
```diff
@@ -0,0 +1,24 @@
+# AZFW0010: Invalid binding type
+
+| | Value |
+|-|-|
+| **Rule ID** |AZFW00010|
```
kshyju, line 5: Why not AZFW0009? https://github.com/Azure/azure-functions-dotnet-worker/tree/main/docs/analyzer-rules

docs/analyzer-rules/AZFW0010.md
```diff
@@ -0,0 +1,24 @@
...
+
+Some bindings advertise the types that they support. If a binding advertises supported types
+and does not support falling back to built-in converters, then this rule will will be flagged.
+
+For example if you're binding to a `QueueTrigger` and the only supported type is `string` and
+your function is binding to a `bool`, then you will see this warning.
+
+## How to fix violations
+
+Change the binding type to a type that is supported by the binding your function is using.
```
kshyju, line 24: Should we also list the types supported by each input bindings here to help the user? I think that might be helpful for the user.

kshyju (approved):

docs/analyzer-rules/AZFW0010.md
liliankasem: 9 already exists in our feature branch: https://github.com/Azure/azure-functions-dotnet-worker/tree/feature/sdk-type-binding/docs/analyzer-rules

sdk/Sdk.Analyzers/BindingTypeNotSupported.cs
liliankasem: Yeah I copied this from the other analyzers - will make a PR to update them all to fix this

sdk/Sdk.Analyzers/BindingTypeNotSupported.cs
liliankasem: I'll switch to checking the named symbol type to be safe

Issue #1474, opened 2023-04-17
[SDK-Bindings - Tables] Implement converter in Tables worker extension to support TableClient +


liliankasem:
#### Problem Statement

Implement `TablesConverter` to enable binding to the `TableClient`, `TableEntity` and `IEnumerable<TableEntity>` types so that we have parity with in-proc.

- [Tables worker extension project](https://github.com/Azure/azure-functions-dotnet-worker/tree/feature/sdk-type-binding/extensions/Worker.Extensions.Tables/src)

#### Tasks

- [x] Implement converter that enables binding to a single `TableClient`, `TableEntity` and multiple `TableEntity`'s
    - Being able to bind to multiple table entities looks like this: `IEnumerable<TableEntity>`, `IAsyncCollector<TableEntity>`*
- [x] Implement managed identity support via Options and AzureComponentFactory
- [x] Write unit tests for converter
- [x] Write unit tests for any other classes implemented (such as the config and options classes)
- [x] Write E2E tests for:
    - Function with Tables input binding to `TableClient`
    - Function with Tables input binding to `TableEntity`
    - Function with Tables input binding to `IEnumerable<TableEntity>`

* We may not need to support `IAsyncCollector<TableEntity>` so start with all of the other types first and we can come back to this at the end

#### Definition of Done

1. Worker Tables extension package is released (as a preview package) with support for binding to a `TableClient`, `TableEntity`, and `IEnumerable<TableEntity>`
2. Unit & E2E tests written and checked-in
3. Microsoft documentation updated to detail support for the new SDK-bindings

liliankasem:
First, utilize the author guide to implement this changes:

- [Dotnet extension author guide](https://dev.azure.com/msazure/One/_git/AAPT-Antares-Docs?path=/TeamDocs/FunctionTeamDocs/EngineeringHowTos/DeferredBinding/DotnetIsolatedExtensionGuide.md&version=GBmaster&_a=preview)

If you're stuck, use the Cosmos DB and Blob extensions as examples of how to do this; they both have converters implemented with managed identity support:

- [Blob worker extension](https://github.com/Azure/azure-functions-dotnet-worker/tree/feature/sdk-type-binding/extensions/Worker.Extensions.Storage.Blobs/src)
- [Cosmos worker extension(https://github.com/Azure/azure-functions-dotnet-worker/tree/feature/sdk-type-binding/extensions/Worker.Extensions.CosmosDB/src)

PR #1506, opened 2023-05-03
Table converter

aishwaryabh:
Creating a table converter to enable binding to the `TableClient`, `TableEntity` and `IEnumerable<TableEntity>` types so that we have parity with in-proc.
resolves https://github.com/Azure/azure-functions-dotnet-worker/issues/1474

### Pull request checklist

### Additional information

Additional PR information


extensions/Worker.Extensions.Tables/src/Config/TablesBindingOptions.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        public string? ConnectionString { get; set; }
+
+        public Uri? ServiceUri { get; set; }
+
+        public TokenCredential? Credential { get; set; }
+
+        public TableClientOptions? TableClientOptions { get; set; }
+
+        internal TableServiceClient CreateClient()
```
liliankasem, line 21: Let's implement a client cache similar to what we have in CosmosDBBindingOptions.cs

extensions/Worker.Extensions.Tables/src/Config/TablesBindingOptionsSetup.cs
```diff
@@ -0,0 +1,97 @@
...
+            }
+            else
+            {
+                if (TryGetServiceUri(connectionSection, out Uri serviceUri))
+                {
+                    options.ServiceUri = serviceUri;
+                }
+            }
+
+            options.TableClientOptions = (TableClientOptions)_componentFactory.CreateClientOptions(typeof(TableClientOptions), null, connectionSection);
```
liliankasem, line 59: Why do we need to create the client options here if they are passed into the function as a parameter?

extensions/Worker.Extensions.Tables/src/Config/TablesBindingOptionsSetup.cs
```diff
@@ -0,0 +1,97 @@
...
+            options.Credential = _componentFactory.CreateTokenCredential(connectionSection);
+        }
+
+        /// <summary>
+        /// Either constructs the serviceUri from the provided accountName
+        /// or retrieves the serviceUri for the specific resource (i.e. blobServiceUri or queueServiceUri)
+        /// </summary>
+        private bool TryGetServiceUri(IConfiguration configuration, out Uri serviceUri)
```
liliankasem, line 67: Since this function and FormatServiceUri are both used by the Tables and Blob extensions, move these two methods to the Shared extension project so that both extensions can use them + reduce code clones

extensions/Worker.Extensions.Tables/src/TableConverter.cs
```diff
@@ -0,0 +1,178 @@
...
+            TableServiceClient tableServiceClient = tableOptions.CreateClient();
+            return tableServiceClient.GetTableClient(tableName);
+        }
+
+        internal virtual TableEntity GetTableEntity(string? connection, string tableName, string? partitionKey, string? rowKey)
+        {
+            var tableOptions = _tableOptions.Get(connection);
+            TableServiceClient tableServiceClient = tableOptions.CreateClient();
+            var tableClient = tableServiceClient.GetTableClient(tableName);
+            return tableClient.GetEntity<TableEntity>(partitionKey, rowKey);
```
liliankasem, line 169: `GetTableEntity` and `GetTableClient` have the exact same code except for entity we do `tableClient.GetEntity<TableEntity>(partitionKey, rowKey);` - this can be consolidated into one function either using generics (or a simply if-statement checking the type) or via a 3rd function that will create and return the table client

extensions/Worker.Extensions.Tables/src/TableInputAttribute.cs
```diff
@@ -60,7 +72,7 @@ public TableInputAttribute(string tableName, string partitionKey, string rowKey)
         /// <summary>
         /// Used with filter. RowKey should be null. 
         /// </summary>
-        public int? Take
+        public int Take
```
liliankasem, line 75: Why was the null operator remove here? This may be a breaking change

samples/WorkerBindingSamples/Table/TableInputBindingSamples.cs
```diff
@@ -0,0 +1,97 @@
...
+        public TableInputBindingSamples(ILogger<TableInputBindingSamples> logger)
+        {
+            _logger = logger;
+        }
+
+        [Function(nameof(TableClientFunction))]
+        public async Task<HttpResponseData> TableClientFunction(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "get", "post")] HttpRequestData req,
+            [TableInput("TableName")] TableClient table)
+
```
liliankasem, line 29: nit: remove empty line

samples/WorkerBindingSamples/WorkerBindingSamples.csproj
```diff
@@ -7,16 +7,28 @@
...
+
+            <ProjectReference Include="..\..\extensions\Worker.Extensions.Http\src\Worker.Extensions.Http.csproj" />
+
+            <ProjectReference Include="..\..\extensions\Worker.Extensions.Storage\src\Worker.Extensions.Storage.csproj" />
+
+            <ProjectReference Include="..\..\src\DotNetWorker.ApplicationInsights\DotNetWorker.ApplicationInsights.csproj" />
+
+            <ProjectReference Include="..\..\src\DotNetWorker\DotNetWorker.csproj" />
+
+
```
liliankasem, line 30: nit: remove all the whitespace between every property

test/E2ETests/E2EApps/E2EApp/E2EApp.csproj
```diff
@@ -42,8 +43,8 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.WebUtilities" Version="2.2.0" />
-    <PackageReference Condition="$(TestBuild) == 'true'" Include="Microsoft.Azure.Functions.Worker" Version="1.12.1-preview1" />
-    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.9.0-preview1" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.13.0-preview1-local202305031031" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.10.0-local202305031056" />
```
liliankasem, line 47: This change shouldn't be checked in; it's automated by the e2e setup script right?

test/E2ETests/E2EApps/E2EApp/E2EApp.csproj
```diff
@@ -24,6 +24,7 @@
     <ProjectReference Include="..\..\..\..\extensions\Worker.Extensions.EventHubs\src\Worker.Extensions.EventHubs.csproj" />
     <ProjectReference Include="..\..\..\..\extensions\Worker.Extensions.Http\src\Worker.Extensions.Http.csproj" />
     <ProjectReference Include="..\..\..\..\extensions\Worker.Extensions.Timer\src\Worker.Extensions.Timer.csproj" />
+      <ProjectReference Include="..\..\..\..\extensions\Worker.Extensions.Tables\src\Worker.Extensions.Tables.csproj" />
```
liliankasem, line 27:
```suggestion
     <ProjectReference Include="..\..\..\..\extensions\Worker.Extensions.Tables\src\Worker.Extensions.Tables.csproj" />
```

test/E2ETests/E2EApps/E2EApp/Table/TableInputBindingFunctions.cs
```diff
@@ -0,0 +1,92 @@
...
+        public TableInputBindingFunction(ILogger<TableInputBindingFunction> logger)
+        {
+            _logger = logger;
+        }
+
+        [Function(nameof(TableClientFunction))]
+        public async Task<HttpResponseData> TableClientFunction(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "get", "post")] HttpRequestData req,
+            [TableInput("TableName")] TableClient table)
+
```
liliankasem, line 26:
```suggestion
```

test/E2ETests/E2ETests/E2ETests.csproj
```diff
@@ -19,6 +19,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="Azure.Data.Tables" Version="12.8.0" />
```
liliankasem, line 22: You shouldn't have to do this; the worker extension will bring in this package

test/E2ETests/E2ETests/Helpers/TableHelpers.cs
```diff
@@ -0,0 +1,52 @@
...
+        private static readonly TableClient _tableClient;
+        static TableHelpers()
+        {
+            var builder = new System.Data.Common.DbConnectionStringBuilder
+            {
+                ConnectionString = Constants.Tables.TablesConnectionStringSetting
+            };
+            var tableName = Constants.Tables.TableName;
+            _tableClient = new TableClient(Constants.Tables.TablesConnectionStringSetting, tableName);
+
```
liliankasem, line 25:
```suggestion
```

test/E2ETests/E2ETests/Helpers/TableHelpers.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        public async static Task CreateTable()
+        {
+            _ = await _tableClient.CreateIfNotExistsAsync();
+        }
+
+        public async static Task DeleteTable()
+        {
+            _ = await _tableClient.DeleteAsync();
```
liliankasem, line 35:
```suggestion
            await _tableClient.DeleteAsync();
```

don't need to assign if not used

test/E2ETests/E2ETests/Helpers/TableHelpers.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using Azure.Data.Tables;
+using Microsoft.Azure.Documents;
+using Microsoft.Azure.Documents.Client;
+using Microsoft.Azure.Functions.Tests.E2ETests;
+
+namespace Microsoft.Azure.Functions.Worker.E2ETests.Helpers
+{
+    public static class TableHelpers
```
liliankasem, line 14: Make sure to include a Dispose method to dispose of the client when done

test/E2ETests/E2ETests/Tables/TablesEndToEndTests.cs
```diff
@@ -0,0 +1,140 @@
...
+    {
+        private readonly IDisposable _disposeLog;
+        private FunctionAppFixture _fixture;
+        private const string partitionKey = "Partition";
+        private const string firstRowKey = "FirstRowKey";
+        private const string firstValue = "FirstValue";
+        private const string secondRowKey = "SecondRowKey";
+        private const string secondValue = "SecondValue";
+        private const string thirdRowKey = "ThirdRowKey";
+        private const string thirdValue = "ThirdValue";
```
liliankasem, line 30: nit: I would prefer if we define expected values that we are asserting against in the test method that they are used in. It's okay to be repetitive in tests. It makes tests easier to follow and safer in a sense as we have all the test items self contained

test/WorkerExtensionTests/Table/TableConverterTests.cs
```diff
@@ -0,0 +1,221 @@
...
+using Microsoft.Azure.Functions.Worker.Extensions.Tables.Config;
+using Microsoft.Azure.Functions.Worker.Extensions.Tables;
+using Azure.Data.Tables;
+using System.Linq;
+
+namespace Microsoft.Azure.Functions.WorkerExtension.Tests.Table
+{
+    public class TableConverterTests
+    {
+        private Mock<TableConverter> _mockTableConverter;
```
liliankasem, line 28: We should write the unit tests without mocking the converter; instead mock the external source (the table client). We have this done in the cosmos db converter tests so I would reuse the work in that (we have a GH issue to update the blob tests to be the same way).

In general, we shouldn't be making things internal just to test them, you should be able to test all of the code paths via the main entry point (in this case, ConvertAsync).

https://github.com/Azure/azure-functions-dotnet-worker/blob/feature/sdk-type-binding/test/WorkerExtensionTests/Cosmos/CosmosDBConverterTests.cs

test/WorkerExtensionTests/WorkerExtensionTests.csproj
```diff
@@ -24,7 +24,23 @@
...
+
+            <ProjectReference Include="..\..\extensions\Worker.Extensions.Http\src\Worker.Extensions.Http.csproj" />
+
+            <ProjectReference Include="..\..\extensions\Worker.Extensions.Storage\src\Worker.Extensions.Storage.csproj" />
+
+            <ProjectReference Include="..\..\src\DotNetWorker.ApplicationInsights\DotNetWorker.ApplicationInsights.csproj" />
+
+            <ProjectReference Include="..\..\src\DotNetWorker\DotNetWorker.csproj" />
+
+
```
liliankasem, line 43: This looks like a duplicate item group for project references, we already have a lot of these imported on lines 25-28. Remove this section and update the one above it to add what you're missing if anything

samples/WorkerBindingSamples/WorkerBindingSamples.csproj
liliankasem: We already have an item group for project references - you should keep them combined and remove any dups

extensions/Worker.Extensions.Tables/src/TableConverter.cs
liliankasem, line 169: Could anything go wrong here if partition key and/or row key are null or empty? What happens if you pass one without the other?

samples/WorkerBindingSamples/Table/TableInputBindingSamples.cs
```diff
@@ -0,0 +1,97 @@
...
+using Azure.Storage.Blobs;
+using Azure.Storage.Blobs.Models;
+using Microsoft.Azure.Cosmos;
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Azure.Functions.Worker.Http;
+using Microsoft.Extensions.Logging;
+
+namespace WorkerBindingSamples.Table
+{
+    public class TableInputBindingSamples
```
liliankasem, line 16: Did you test all of these scenarios with user and system managed identity?

extensions/Worker.Extensions.Tables/src/Config/TablesBindingOptionsSetup.cs
aishwaryabh: It's not passed in as a parameter. It's a private member of the TablesBindingOptions class, so it just reads the value.

aishwaryabh: I believe Blob is doing the same thing. Should I follow that pattern?

extensions/Worker.Extensions.Tables/src/TableInputAttribute.cs
aishwaryabh: I get an error when I try to use Take like this:
image

The error is `Take is not a valid named attribute argument because it is not a valid attribute parameter type`

aishwaryabh: The error goes away when I remove the nullable. Any ideas for why that happens?

test/E2ETests/E2ETests/E2ETests.csproj
aishwaryabh: I see [here](https://github.com/Azure/azure-functions-dotnet-worker/pull/1506/files/eda9939a70e1f0358eae19309ac72fcc25286867#diff-40f748aa75dcc9c5ab0e6e6354ca57e661c9d577a042ef0f2d383cac67ce1685R11) we're bringing in blob and it's not coming through the worker extension. Should table follow that pattern as well?

test/E2ETests/E2ETests/Helpers/TableHelpers.cs
aishwaryabh: I have a dispose method[ here](https://github.com/Azure/azure-functions-dotnet-worker/pull/1506/files/eda9939a70e1f0358eae19309ac72fcc25286867#diff-f119006f256a571974deb81d3bd8372fdb030f44357611442cdad9bcc29b8cd6R133). Does that work?

surgupta-msft:
The target branch of QueueConverter [PR](https://github.com/Azure/azure-functions-dotnet-worker/pull/1470) is `surgupta/bypass-deferredbinding`. Do we want the same target branch for this PR? If yes, you can click on Edit button near PR title and change the base branch.

Also, feel free to rebase this PR with bypass deferred binding changes. We have everything in place needed for Table extension. Let me know if you have any questions.

extensions/Worker.Extensions.Tables/src/TableConverter.cs
```diff
@@ -0,0 +1,178 @@
...
+using Microsoft.Extensions.Options;
+using Microsoft.Extensions.Logging;
+using Microsoft.Azure.Functions.Worker.Extensions.Tables;
+using Azure.Data.Tables;
+using Microsoft.Azure.Functions.Worker.Extensions.Tables.Config;
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    /// <summary>
+    /// Converter to bind Blob Storage type parameters.
```
surgupta-msft, line 24: Update the comment for table extension

extensions/Worker.Extensions.Tables/src/TableExtensionStartup.cs
```diff
@@ -0,0 +1,41 @@
...
+                throw new ArgumentNullException(nameof(applicationBuilder));
+            }
+
+            applicationBuilder.Services.AddAzureClientsCore(); // Adds AzureComponentFactory
+            applicationBuilder.Services.AddOptions<TablesBindingOptions>();
+            applicationBuilder.Services.AddSingleton<IConfigureOptions<TablesBindingOptions>, TablesBindingOptionsSetup>();
+
+            applicationBuilder.Services.Configure<WorkerOptions>((workerOption) =>
+            {
+                workerOption.InputConverters.RegisterAt<TableConverter>(0);
```
surgupta-msft, line 37: Once you rebase with bypass deferred binding changes, we won't need this registration anymore.

extensions/Worker.Extensions.Tables/src/TableConverter.cs
aishwaryabh: Yeah it would return a NRE if either the partition key and/or row key are null. If it's empty and `GetEntity` can't find the appropriate value, it will throw an exception, which will be caught be the outer try catch in `ConvertAsync`

extensions/Worker.Extensions.Tables/src/Config/TablesBindingOptions.cs
aishwaryabh: done!

extensions/Worker.Extensions.Tables/src/TableConverter.cs
aishwaryabh: done!

test/WorkerExtensionTests/Table/TableConverterTests.cs
aishwaryabh: Done!

extensions/Worker.Extensions.Tables/src/TableConverter.cs
aishwaryabh: I added a unit test to address this scenario and explicitly check if partition key and/or row key are null

extensions/Worker.Extensions.Tables/src/TableConverter.cs
```diff
@@ -0,0 +1,181 @@
...
+        internal virtual TableClient GetTableClient(string? connection, string tableName)
+        {
+            var tableOptions = _tableOptions.Get(connection);
+            TableServiceClient tableServiceClient = tableOptions.CreateClient();
+            return tableServiceClient.GetTableClient(tableName);
+        }
+
+        internal virtual async Task<TableEntity> GetTableEntity(string? connection, string tableName, string? partitionKey, string? rowKey)
+        {
+            if (partitionKey == null || rowKey == null)
```
jviau, line 167: nit: validate these separately so the exception is clear which one is null.

extensions/Worker.Extensions.Tables/src/Worker.Extensions.Tables.csproj
```diff
@@ -14,6 +14,19 @@
...
 
+  <ItemGroup>
+    <PackageReference Include="Azure.Data.Tables" Version="12.8.0" />
+    <PackageReference Include="Azure.Identity" Version="1.8.0" />
+    <PackageReference Include="Azure.Storage.Blobs" Version="12.13.0" />
+    <PackageReference Include="Microsoft.Extensions.Azure" Version="1.6.3" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+  </ItemGroup>
+
+    <ItemGroup>
```
jviau, line 28: nit: indentation is inconsistent when above.

samples/WorkerBindingSamples/Table/TableInputBindingSamples.cs
aishwaryabh: Yes tested in all scenarios!

extensions/Worker.Extensions.Tables/src/TableExtensionStartup.cs
aishwaryabh: Updated!

extensions/Worker.Extensions.Storage.Blobs/src/Config/BlobStorageBindingOptionsSetup.cs
```diff
@@ -61,39 +61,5 @@ public void Configure(string name, BlobStorageBindingOptions options)
             options.BlobClientOptions = (BlobClientOptions)_componentFactory.CreateClientOptions(typeof(BlobClientOptions), null, connectionSection);
             options.Credential = _componentFactory.CreateTokenCredential(connectionSection);
         }
-
-        /// <summary>
-        /// Either constructs the serviceUri from the provided accountName
-        /// or retrieves the serviceUri for the specific resource (i.e. blobServiceUri or queueServiceUri)
-        /// </summary>
-        private bool TryGetServiceUri(IConfiguration configuration, out Uri serviceUri)
```
surgupta-msft, line 69: Are we moving this logic to shared project because it is common between Blobs and Tables? We can bring this up in our feature sync that do we want to let each extension define its own authentication logic or have it in shared project. For cosmos, this is happening inside [cosmos extension](https://github.com/Azure/azure-functions-dotnet-worker/blob/feature/sdk-type-binding/extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBBindingOptionsSetup.cs). I understand this will cause duplication of logic in Blobs and Tables but then this may not be useful for remaining extensions. So, lets discuss this one.

extensions/Worker.Extensions.Tables/src/Config/TablesBindingOptions.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        public TokenCredential? Credential { get; set; }
+
+        public TableClientOptions? TableClientOptions { get; set; }
+
+        internal ConcurrentDictionary<string, TableServiceClient> ClientCache { get; } = new ConcurrentDictionary<string, TableServiceClient>();
+
+        internal virtual TableServiceClient CreateClient()
+        {
+            if (ConnectionString == null && ServiceUri == null)
```
surgupta-msft, line 25: using _is null_

extensions/Worker.Extensions.Tables/src/Config/TablesBindingOptions.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
```
surgupta-msft, line 1: We need copyright statement in all the new files added

samples/WorkerBindingSamples/Table/TableInputBindingSamples.cs
```diff
@@ -0,0 +1,92 @@
...
+            {
+                val.TryGetValue("Text", out var text);
+                _logger.LogInformation("Value of text: " + text);
+
+                await response.WriteStringAsync(text?.ToString() ?? "");
+            }
+
+            return response;
+        }
+
```
surgupta-msft, line 40: Consider adding sample functions for some types which bypass deferred binding and get handled on Extension side, if exist.

test/FunctionMetadataGeneratorTests/FunctionMetadataGeneratorTests.cs
```diff
@@ -957,6 +957,16 @@ private class SDKTypeBindings_BlobStorage
...
+            public object BlobClientToBlobStringFunction(
+                [BlobTrigger("container2/%file%")] BlobClient blob,
+                [BlobInput("container2/%file%")] string blobinput)
+            {
+                throw new NotImplementedException();
+            }
+
+
             [Function("BlobClientToBlobStringFunction")]
             [BlobOutput("container1/hello.txt", Connection = "MyOtherConnection")]
```
surgupta-msft, line 971: Add tests for Table extension also similar to Blobs which validate if SupportsDeferredBinding is being assigned correctly by `FunctionMetadataGenerator` based on `binding type` for both the cases - types that support and types that bypass deferred binding

extensions/Worker.Extensions.Storage.Blobs/src/Config/BlobStorageBindingOptionsSetup.cs
aishwaryabh: Yeah I moved the logic to the shared project since the code was very similar. The only difference is the subdomain being passed in, so I added that in as a parameter

extensions/Worker.Extensions.Tables/src/TableConverter.cs
```diff
@@ -0,0 +1,181 @@
...
+                return ConversionResult.Failed(ex);
+            }
+
+            return ConversionResult.Unhandled();
+        }
+
+        internal virtual async ValueTask<ConversionResult> ConvertFromCollectionBindingDataAsync(ConverterContext context, CollectionModelBindingData collectionModelBindingData)
+        {
+            var tableCollection = new List<object>(collectionModelBindingData.ModelBindingDataArray.Length);
+            Type elementType = context.TargetType.IsArray ? context.TargetType.GetElementType() : context.TargetType.GenericTypeArguments[0];
```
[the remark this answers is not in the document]
aishwaryabh: I'm actually removing that method since the tables webjob extension only returns of type `ParameterBindingData`, and not a collection type.

extensions/Worker.Extensions.Tables/src/TableInputAttribute.cs
aishwaryabh: Added a new constructor

extensions/Worker.Extensions.Storage.Blobs/src/Config/BlobStorageBindingOptionsSetup.cs
liliankasem: For any shared auth logic, we should have them in the shared project. If an extension is doing something different, that's fine. I don't see the downside to this

test/E2ETests/E2EApps/E2EApp/E2EApp.csproj
aishwaryabh: for the worker.sdk package I needed to specify @surgupta-msft 's bypass deferred binding changes. Is there a way to get around this?

samples/WorkerBindingSamples/Table/TableInputBindingSamples.cs
aishwaryabh: Added

test/E2ETests/E2ETests/Tables/TablesEndToEndTests.cs
```diff
@@ -0,0 +1,150 @@
...
+            HttpStatusCode expectedStatusCode = HttpStatusCode.OK;
+
+            Assert.Equal(expectedStatusCode, response.StatusCode);
+            Assert.Equal($"{firstValue},{secondValue}", actualMessage);
+
+            // Delete table
+            await TableHelpers.DeleteTable();
+        }
+
+        public async void Dispose()
```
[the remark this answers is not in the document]
aishwaryabh: Removed async void

test/FunctionMetadataGeneratorTests/FunctionMetadataGeneratorTests.cs
aishwaryabh: Added tests for both files!

extensions/Worker.Extensions.Tables/src/Properties/AssemblyInfo.cs
```diff
@@ -1,6 +1,10 @@
...
+using System.Runtime.CompilerServices;
+using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
 
-[assembly: ExtensionInformation("Microsoft.Azure.WebJobs.Extensions.Tables", "1.0.0")]
+// TODO: Change this once web jobs extension has been released
+[assembly: ExtensionInformation("Microsoft.Azure.WebJobs.Extensions.Tables", "1.2.0-alpha.20230517.1")]
+[assembly: InternalsVisibleTo("Microsoft.Azure.Functions.WorkerExtension.Tests, PublicKey=00240000048000009400000006020000002400005253413100040000010001005148be37ac1d9f58bd40a2e472c9d380d635b6048278f7d47480b08c928858f0f7fe17a6e4ce98da0e7a7f0b8c308aecd9e9b02d7e9680a5b5b75ac7773cec096fbbc64aebd429e77cb5f89a569a79b28e9c76426783f624b6b70327eb37341eb498a2c3918af97c4860db6cdca4732787150841e395a29cfacb959c1fd971c1")]
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7")]
```
liliankasem, line 10: I'm not 100% sure that we need the `DynamicProxyGenAssembly2` assembly reference so it might be safe to remove this

test/E2ETests/E2EApps/E2EApp/E2EApp.csproj
aishwaryabh: Will update this once packages have been released on public feeds

extensions/Worker.Extensions.Tables/src/Config/TablesBindingOptions.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        public string? ConnectionString { get; set; }
+
+        public Uri? ServiceUri { get; set; }
+
+        public TokenCredential? Credential { get; set; }
+
+        public TableClientOptions? TableClientOptions { get; set; }
+
+        internal ConcurrentDictionary<string, TableServiceClient> ClientCache { get; } = new ConcurrentDictionary<string, TableServiceClient>();
```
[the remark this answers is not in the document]
aishwaryabh: @brettsam we added caching for cosmos DB due to the caching client present in the webjobs extension worker [here](https://github.com/Azure/azure-webjobs-sdk-extensions/blob/0af7d172090d1b698b13ebcaef628177a8c3605f/src/WebJobs.Extensions.CosmosDB/Config/CosmosDBExtensionConfigProvider.cs#L4). Would we also need this for tables?

extensions/Worker.Extensions.Tables/src/Properties/AssemblyInfo.cs
aishwaryabh: Removed

extensions/Worker.Extensions.Tables/src/TableConverter.cs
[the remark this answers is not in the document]
aishwaryabh: Added this issue: https://github.com/Azure/azure-functions-dotnet-worker/issues/1552

extensions/Worker.Extensions.Tables/src/Config/TablesBindingOptions.cs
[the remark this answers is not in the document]
aishwaryabh: I had an offline discussion with Josh and learned that the TableClient is a thin wrapper aronnd the HttpClient in the webjobs extension, so there isn't much benefit to caching it. I removed the cache

extensions/Worker.Extensions.Shared/Configuration/ConfigurationExtensions.cs
```diff
@@ -1,7 +1,9 @@
-// Copyright (c) .NET Foundation. All rights reserved.
+﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using Microsoft.Extensions.Configuration;
+using System;
```
liliankasem, line 5: nit: order imports (Microsoft. should come after System - we usually have System imports first, followed by everything else in alphabetically order)

extensions/Worker.Extensions.Tables/src/Config/TablesBindingOptions.cs
```diff
@@ -0,0 +1,39 @@
...
+        private TableServiceClient? TableServiceClient { get; set; }
+
+        internal virtual TableServiceClient CreateClient()
+        {
+            if (ConnectionString is null && ServiceUri is null)
+            {
+                throw new ArgumentNullException(nameof(ConnectionString) + " " + nameof(ServiceUri));
+            }
+
+            if (TableServiceClient != null)
```
liliankasem, line 29: nit: 
```suggestion
            if (TableServiceClient is not null)
```

extensions/Worker.Extensions.Tables/src/TableConverter.cs
```diff
@@ -0,0 +1,175 @@
...
+                }
+            }
+            return bindingDataContent;
+        }
+
+        internal virtual async Task<TableEntity> GetTableEntity(string? connection, string tableName, string? partitionKey, string? rowKey)
+        {
+            if (partitionKey == null)
+            {
+                throw new ArgumentNullException($"Partition key {partitionKey} cannot be null");
```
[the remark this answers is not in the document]
liliankasem: This applies to all places `ArgumentNullException` is used

extensions/Worker.Extensions.Tables/src/TableConverter.cs
```diff
@@ -0,0 +1,175 @@
...
+            }
+            if (rowKey == null)
```
liliankasem, line 167: super nit: space out code i.e. new lines around the if-statement

samples/WorkerBindingSamples/Table/TableInputBindingSamples.cs
```diff
@@ -0,0 +1,123 @@
...
+        {
+            var response = req.CreateResponse(HttpStatusCode.OK);
+            List<string> tableList = new();
+
+            foreach (TableEntity tableEntity in tables)
+            {
+                tableEntity.TryGetValue("Text", out var text);
+                _logger.LogInformation("Value of text: " + text);
+                tableList.Add((text?.ToString()) ?? "");
+
```
liliankasem, line 87:
```suggestion
```

samples/WorkerBindingSamples/Table/TableInputBindingSamples.cs
```diff
@@ -0,0 +1,123 @@
...
+            return response;
+        }
+
+        [Function(nameof(PocoFunction))]
+        public async Task<HttpResponseData> PocoFunction(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "get","post", Route = null)] HttpRequestData req,
+            [TableInput("TableName")] IEnumerable<MyEntity> entities,
+            FunctionContext executionContext)
+        {
+            var logger = executionContext.GetLogger(nameof(PocoFunction));
```
liliankasem, line 100: Let's use a class level logger (see cosmos or blob samples)

samples/WorkerBindingSamples/Table/TableInputBindingSamples.cs
```diff
@@ -0,0 +1,123 @@
...
+            List<string> entityList = new();
+
+            foreach (MyEntity entity in entities)
+            {
+                logger.LogInformation($"Text: {entity.Text}");
+                entityList.Add((entity.Text ?? "").ToString());
+            }
+
+            await response.WriteStringAsync(string.Join(",", entityList));
+            return response;
```
liliankasem, line 112:
```suggestion
            return await response.WriteStringAsync(string.Join(",", entityList));
```

samples/WorkerBindingSamples/WorkerBindingSamples.csproj
```diff
@@ -18,6 +18,7 @@
         <ProjectReference Include="..\..\extensions\Worker.Extensions.Storage\src\Worker.Extensions.Storage.csproj" />
         <ProjectReference Include="..\..\src\DotNetWorker.ApplicationInsights\DotNetWorker.ApplicationInsights.csproj" />
         <ProjectReference Include="..\..\src\DotNetWorker\DotNetWorker.csproj" />
+		<ProjectReference Include="..\..\extensions\Worker.Extensions.Tables\src\Worker.Extensions.Tables.csproj" />
```
liliankasem, line 21:
```suggestion
        <ProjectReference Include="..\..\extensions\Worker.Extensions.Tables\src\Worker.Extensions.Tables.csproj" />
```

test/E2ETests/E2ETests/Helpers/TableHelpers.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            await _tableClient.CreateIfNotExistsAsync();
+        }
+
+        public async static Task DeleteTable()
+        {
+            await _tableClient.DeleteAsync();
+        }
+
+        // keep
```
liliankasem, line 26: nit:
```suggestion
```

extensions/Worker.Extensions.Tables/src/TableConverter.cs
```diff
@@ -0,0 +1,175 @@
...
+        public async ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
+        {
+            return context?.Source switch
+            {
+                ModelBindingData binding => await ConvertFromBindingDataAsync(context, binding),
+                _ => ConversionResult.Unhandled()
+            };
+        }
+
+        public virtual async ValueTask<ConversionResult> ConvertFromBindingDataAsync(ConverterContext context, ModelBindingData modelBindingData)
```
liliankasem, line 45:
```suggestion
        private async ValueTask<ConversionResult> ConvertFromBindingDataAsync(ConverterContext context, ModelBindingData modelBindingData)
```

extensions/Worker.Extensions.Tables/src/TableConverter.cs
```diff
@@ -0,0 +1,175 @@
...
+
+            if (string.IsNullOrEmpty(tableName?.ToString()))
+            {
+                throw new ArgumentNullException("'TableName' cannot be null or empty");
+            }
+
+            return await ToTargetTypeAsync(targetType, connection?.ToString() ?? null, tableName!.ToString(), partitionKey?.ToString() ?? null, rowKey?.ToString() ?? null, Convert.ToInt32(take?.ToString()), filter?.ToString() ?? null);
+        }
+
+        internal Dictionary<string, object> GetBindingDataContent(ModelBindingData bindingData)
```
liliankasem, line 97: I think all methods should be private here, only ConvertAsync (as the access point of the function) should be public

extensions/Worker.Extensions.Tables/src/TableConverter.cs
```diff
@@ -0,0 +1,175 @@
...
+        {
+            return bindingData?.ContentType switch
+            {
+                Constants.JsonContentType => new Dictionary<string, object>(bindingData?.Content?.ToObjectFromJson<Dictionary<string, object>>(), StringComparer.OrdinalIgnoreCase),
+                _ => throw new NotSupportedException($"Unexpected content-type. Currently only '{Constants.JsonContentType}' is supported.")
+            };
+        }
+
+        internal virtual async Task<object?> ToTargetTypeAsync(Type targetType, string? connection, string tableName, string? partitionKey, string? rowKey, int take, string? filter) => targetType switch
```
liliankasem, line 106: There are a lot of parameters here, it might be worth using an object/model instead. i.e. instead of using a dictionary in `GetBindingDataContent` - you could deserilize into an object reference we create so that it's cleaner

extensions/Worker.Extensions.Tables/src/Worker.Extensions.Tables.csproj
```diff
@@ -14,6 +14,18 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\Worker.Extensions.Abstractions\src\Worker.Extensions.Abstractions.csproj" />
+    <ProjectReference Include="..\..\..\src\DotNetWorker.Core\DotNetWorker.Core.csproj" />
```
liliankasem, line 17: Need to update the version to match the WebJobs one, it should be `1.2.0-preview1`

    <VersionPrefix>1.2.0</VersionPrefix>
    <VersionSuffix>-preview1</VersionSuffix>

extensions/Worker.Extensions.Tables/src/TableConverter.cs
```diff
@@ -0,0 +1,175 @@
...
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    /// <summary>
+    /// Converter to bind Table type parameters.
+    /// </summary>
+    [SupportsDeferredBinding]
+    [SupportedConverterType(typeof(TableClient))]
+    [SupportedConverterType(typeof(TableEntity))]
+    [SupportedConverterType(typeof(IEnumerable<TableEntity>))]
```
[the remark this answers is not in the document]
aishwaryabh: Discussed offline and decided to create 3 separate converters with ConvertAsync being abstract in the base class

samples/WorkerBindingSamples/Table/TableInputBindingSamples.cs
aishwaryabh: This needs to be seperate lines otherwise it doesn't compile. WriteStringAsync is void so we need to return response itself after the call has been done

extensions/Worker.Extensions.Tables/src/TableConverter.cs
aishwaryabh: done!

extensions/Worker.Extensions.Tables/src/Config/TablesBindingOptions.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        public string? ConnectionString { get; set; }
+
+        public Uri? ServiceUri { get; set; }
+
+        public TokenCredential? Credential { get; set; }
+
+        public TableClientOptions? TableClientOptions { get; set; }
+
+        private TableServiceClient? TableServiceClient { get; set; }
```
jviau, line 20: For `private`, I typically use fields (as opposed to properties). But no harm in keeping it a property.

extensions/Worker.Extensions.Tables/src/TypeConverters/TableClientConverter.cs
```diff
@@ -0,0 +1,67 @@
...
+            }
+            catch (Exception ex)
+            {
+                return ConversionResult.Failed(ex);
+            }
+
+            return ConversionResult.Unhandled();
+        }
+
+        private Task<TableClient> ConvertModelBindingData(TableData content)
```
jviau, line 57: Does this need to be `Task<T>` returning? Can it be just:
```suggestion
        private TableClient ConvertModelBindingData(TableData content)
```

extensions/Worker.Extensions.Tables/src/TypeConverters/TableConverterBase.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            // Parse through dictionary
+            content.TryGetValue(Constants.Connection, out var connection);
+            content.TryGetValue(Constants.TableName, out var tableName);
+            content.TryGetValue(Constants.PartitionKey, out var partitionKey);
+            content.TryGetValue(Constants.RowKey, out var rowKey);
+            content.TryGetValue(Constants.Take, out var take);
+            content.TryGetValue(Constants.Filter, out var filter);
+
+            TableData tableData = new TableData();
```
jviau, line 80: nit: use object initializer
```suggestion
            TableData tableData = new()
            {
                Connection = connection?.ToString(),
                TableName = tableName?.ToString(),
                ... etc
            };
```

extensions/Worker.Extensions.Tables/src/TypeConverters/TableEntityConverter.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            }
+            catch (Exception ex)
+            {
+                return ConversionResult.Failed(ex);
+            }
+
+            return ConversionResult.Unhandled();
+        }
+        private async Task<TableEntity> ConvertModelBindingData(TableData content)
```
jviau, line 58: nit: add newline
```suggestion

        private async Task<TableEntity> ConvertModelBindingData(TableData content)
```

extensions/Worker.Extensions.Tables/src/TypeConverters/TableConverterBase.cs
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        internal TableClient GetTableClient(string? connection, string tableName)
+        {
+            var tableOptions = _tableOptions.Get(connection);
+            TableServiceClient tableServiceClient = tableOptions.CreateClient();
+            return tableServiceClient.GetTableClient(tableName);
+        }
+        
+        internal TableData GetTableData(Dictionary<string, object> content)
```
jviau, line 70: Can this be `protected`? I didn't notice any external uses:

```suggestion
        protected TableData GetTableData(Dictionary<string, object> content)
```

extensions/Worker.Extensions.Tables/src/TypeConverters/TableConverterBase.cs
```diff
@@ -0,0 +1,91 @@
...
+        internal Dictionary<string, object> GetBindingDataContent(ModelBindingData? bindingData)
+        {
+            return bindingData?.ContentType switch
+            {
+                Constants.JsonContentType => new Dictionary<string, object>(bindingData?.Content?.ToObjectFromJson<Dictionary<string, object>>(), StringComparer.OrdinalIgnoreCase),
+                _ => throw new NotSupportedException($"Unexpected content-type. Currently only '{Constants.JsonContentType}' is supported.")
+            };
+        }
+
+        internal TableClient GetTableClient(string? connection, string tableName)
```
jviau, line 63: Can this be `protected`? I didn't see any external uses.

```suggestion
        protected TableClient GetTableClient(string? connection, string tableName)
```

extensions/Worker.Extensions.Tables/src/TypeConverters/TableData.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Tables.TypeConverters
+{
+    internal class TableData
```
jviau, line 6: is this used outside the `TableConverterBase` at all? If not, consider moving it to be a nested `protected` class of that type.

jviau (approved):
Meant to approve my last review - all my comments are non-blocking.

extensions/Worker.Extensions.Tables/src/TypeConverters/TableClientConverter.cs
aishwaryabh: `ConvertAsync` is an abstract member in the base class, and the other converters that override `ConvertAsync` need to be async so it needs to return `ValueTask<ConversionResult>` . That's why I kept ConvertModelBindingData as returning type `Task<TableClient>` since there are no async calls in this specific instance of ConvertAsync. Does that make sense?

extensions/Worker.Extensions.Tables/src/TypeConverters/TableConverterBase.cs
```diff
@@ -0,0 +1,103 @@
...
+                PartitionKey = partitionKey?.ToString(),
+                RowKey = rowKey?.ToString(),
+                Take = Convert.ToInt32(take?.ToString()),
+                Filter = filter?.ToString()
+            };
+
+            return tableData;
+        }
+
+        protected class TableData
```
liliankasem, line 93: Actually I just remembered that `https://github.com/Azure/azure-functions-dotnet-worker/blob/main/extensions/Worker.Extensions.Tables/src/TableInputAttribute.cs` has all the same properties as this so you might be able to just deserialize directly to TableInputAttribute. If not, let's move this to it's own file i.e. `TableBindingData.cs`

liliankasem (commented):
Looks great! Just one more comment about the use and creation of TableData then ready to sign off

extensions/Worker.Extensions.Tables/src/TypeConverters/TableClientConverter.cs
jviau: You can still remove the async here. You can also remove `async` keyword from `ConvertAsync` above, and change all the return values to be wrapped in a `new(ConversionResult...)` (which is shorthand for `new ValueTask<ConversionResult>(ConversionResult)`

liliankasem (approved):

extensions/Worker.Extensions.Tables/src/TypeConverters/TableConverterBase.cs
liliankasem: tabling for later (pun intended)

extensions/Worker.Extensions.Tables/src/TypeConverters/TableClientConverter.cs
aishwaryabh: Removed async!

extensions/Worker.Extensions.Tables/src/TypeConverters/TableConverterBase.cs
aishwaryabh: Synced offline and kept TableData in same file. I tried to serialize with TableInputAttribute but that didn't work since there was an issue with TableName not properly getting parsed

PR #1526, opened 2023-05-09
[Source Gen]Fix - function executor generator produces code with syntax error

### LINKED ISSUE #1525: FunctionExecutorGenerator produces incorrect code when there are non static functions from multiple classes
The current version of function execution generator is producing incorrect source code when there are multiple function classes. 

For example, the below code

```csharp
public class Function1
{
    [Function("Foo")]
    public HttpResponseData Foo([HttpTrigger(AuthorizationLevel.Anonymous, "get")] HttpRequestData req)
    {

        return response;
    }
}
public class Function2
{
    [Function("Bar")]
    public HttpResponseData Bar([HttpTrigger(AuthorizationLevel.Anonymous, "get")] HttpRequestData req)
    {

        return response;
    }
}
```
is producing the below code. The relevant portion of the error is the `types` directory declaration which is missing a comma between the dictionary entries in the collection initializer.

```csharp
    {
        private readonly IFunctionActivator _functionActivator;
        private readonly Dictionary<string, Type> types = new()
        {
            { "FunctionApp44.Function1", Type.GetType("FunctionApp44.Function1")! }
 { "FunctionApp44.Function2", Type.GetType("FunctionApp44.Function2")! },
        };

        public DirectFunctionExecutor(IFunctionActivator functionActivator)
        {
            _functionActivator = functionActivator ?? throw new ArgumentNullException(nameof(functionActivator));
        }

        {
           // Excluded. 
          // Code for this method is correct.
        }
    }
```

kshyju:
Fixes #1525 

### Pull request checklist

### DIFF
--- a/sdk/Sdk.Generators/Sdk.Generators.csproj
+++ b/sdk/Sdk.Generators/Sdk.Generators.csproj
@@ -10,7 +10,7 @@
     <IncludeBuildOutput>false</IncludeBuildOutput>
     <EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>
     <MinorProductVersion>1</MinorProductVersion>
-    <VersionSuffix>-preview2</VersionSuffix>
+    <VersionSuffix>-preview3</VersionSuffix>
     <IsRoslynComponent>true</IsRoslynComponent>
   </PropertyGroup>
 
--- a/sdk/Sdk.Generators/FunctionExecutor/FunctionExecutorGenerator.Emitter.cs
+++ b/sdk/Sdk.Generators/FunctionExecutor/FunctionExecutorGenerator.Emitter.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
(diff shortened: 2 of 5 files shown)


sdk/Sdk.Generators/Sdk.Generators.csproj
```diff
@@ -10,7 +10,7 @@
     <IncludeBuildOutput>false</IncludeBuildOutput>
     <EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>
     <MinorProductVersion>1</MinorProductVersion>
-    <VersionSuffix>-preview2</VersionSuffix>
+    <VersionSuffix>-preview3</VersionSuffix>
```
kshyju, line 13: May update the version based on feedback from the team.

satvu (approved):
lgtm - just the versioning question needs to be resolved.

jviau (approved):

sdk/Sdk.Generators/Sdk.Generators.csproj
jviau: Is there an alternative version you are considering? If this is to remain in preview, then the change you made looks good to me.

kshyju: Yea, I was discussing with the folks whether we should move this package out of preview (since the SDK GA package relies on this). But that is not a blocker for checking in this PR at the moment. So going to check in this version for now.

Issue #1513, opened 2023-05-05
Worker SDK build failed when moving file.


YuiTH:
@fabiocav 
Our build pipeline seems randomly failing with reference multiple packages related to Microsoft.Azure.Functions.Worker.Extensions.*.
Error messages has multiple formats, but it always related to some kind of file lock or file move failed.

Here are some [logs we collected](https://microsoftapc-my.sharepoint.com/:f:/g/personal/yaobolun_microsoft_com/EvlR8fY9SBxErPW5kQ2-BYcBL8xU0yNPyU5eZp1epJsfxw?e=bqWJew)

And here are some insights from Rainer Sigwald: 
> In this case, I see two operations on the file that failed:
> Copying file from "C:\Users\xxx\AppData\Local\Temp\ymu02bvj.ere\publishout\Microsoft.Azure.WebJobs.Extensions.dll" to "C:\Users\xxx\AppData\Local\Temp\ymu02bvj.ere\publishout\bin\Microsoft.Azure.WebJobs.Extensions.dll".
> 
> Moving file from "C:\Users\xxx\AppData\Local\Temp\ymu02bvj.ere\publishout\Microsoft.Azure.WebJobs.Extensions.dll" to "C:\Users\xxx\AppData\Local\Temp\ymu02bvj.ere\publishout\bin\Microsoft.Azure.WebJobs.Extensions.dll".
> 
> The latter is the one that failed, the former seemed fine. The Move task is generally perfectly happy to overwrite files, so I’m not totally sure what’s going wrong here.
> 
> I wonder if antivirus software may be interfering with a deletion operation on a just-created-in-the-TEMP-folder DLL because it’s doing a scan operation.
> 
[...]

PR #1532, opened 2023-05-10
Add retries to generate function metadata on IOException

jviau:

addresses part of #1513

This PR simply adds retries when trying to write to `function.metadata` as part of a function apps build. Typically, this file would not be locked, but if a function app is built twice in quick succession, this file may be locked by something like antivirus. This is a low-risk change to try and let the build succeed in scenarios where it might fail.

### Pull request checklist

### DIFF
--- a/sdk/Sdk/FunctionMetadataJsonWriter.cs
+++ b/sdk/Sdk/FunctionMetadataJsonWriter.cs
@@ -1,4 +1,4 @@
-// Copyright (c) .NET Foundation. All rights reserved.
+﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System.Collections.Generic;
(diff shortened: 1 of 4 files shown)


sdk/Sdk/FunctionMetadataJsonWriter.cs
jviau, line 1: This file is just formatting / code style changes. No functional changes.

jviau:
Another fix worth considering for this is to fix incremental build. A function app built twice with no changes should be a no-op, however our SDK always triggers full builds.

sdk/Sdk/Tasks/GenerateFunctionMetadata.cs
```diff
@@ -60,6 +62,32 @@ public override bool Execute()
...
+                    FunctionMetadataJsonWriter.WriteMetadata(functions, OutputPath!);
+                    break;
+                }
+                catch (IOException ex)
+                {
+                    attempt++;
+                    if (attempt == 10)
+                    {
+                        Log.LogErrorFromException(ex);
+                        throw new FunctionsMetadataGenerationException();
```
kshyju, line 81: Should we use the FunctionsMetadataGenerationException constructor which takes the message and inner exception?

kshyju (approved):

kshyju (approved):

kshyju (approved):

kshyju:
Mondays!
image

sdk/Sdk/Tasks/GenerateFunctionMetadata.cs
jviau: I can if you want - however it will have no affect.

1. I already log the exception on line 80. `Log.LogErrorFromException(ex);`.
2. Our catch block lines 56 to 62 make no use of the inner exception or message.

jviau: I'll leave it for now. If/when we adjust the catch block then we can add it as an inner exception. I don't want to add it now, have the catch block be changed, and we end up double logging the inner error message (because the line 80 is left in also)

PR #1534, opened 2023-05-11
Create FunctionContext extension method to retrieve HttpContext

### LINKED ISSUE #1531: Create extension method to retrieve HttpContext
An extension method for `FunctionContext` should be created to enable retrieval of the `HttpContext`.

satvu:

resolves #1531 

### Pull request checklist

### DIFF
--- a/extensions/Worker.Extensions.Http.AspNetCore/src/FunctionContextExtensions.cs
+++ b/extensions/Worker.Extensions.Http.AspNetCore/src/FunctionContextExtensions.cs
@@ -16,7 +16,7 @@ namespace Microsoft.Azure.Functions.Worker
         /// Retrieve the <see cref="HttpContext"/> for the function execution.
         /// </summary>
         /// <param name="context">The <see cref="FunctionContext"/>.</param>
-        /// <returns></returns>
+        /// <returns>The <see cref="HttpContext"/> for the function execution or null if it does not exist.</returns>
         public static HttpContext? GetHttpContext(this FunctionContext context)
         {
             if (context.Items.TryGetValue(Constants.HttpContextKey, out var requestContext)
--- a/extensions/Worker.Extensions.Http.AspNetCore/src/FunctionContextExtensions.cs
+++ b/extensions/Worker.Extensions.Http.AspNetCore/src/FunctionContextExtensions.cs
@@ -17,17 +17,15 @@ namespace Microsoft.Azure.Functions.Worker
         /// </summary>
         /// <param name="context">The <see cref="FunctionContext"/>.</param>
         /// <returns></returns>
-        public static HttpContext GetHttpContext(this FunctionContext context)
+        public static HttpContext? GetHttpContext(this FunctionContext context)
         {
             if (context.Items.TryGetValue(Constants.HttpContextKey, out var requestContext)
                 && requestContext is HttpContext httpContext)
             {
                 return httpContext;
             }
-            else
-            {
-                throw new InvalidOperationException($"{nameof(context)} does not have an {nameof(httpContext)} item.");
-            }
+
+            return null;
         }
     }
 }
--- a/extensions/Worker.Extensions.Http.AspNetCore/src/FunctionContextExtensions.cs
+++ b/extensions/Worker.Extensions.Http.AspNetCore/src/FunctionContextExtensions.cs
@@ -1,0 +1,33 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore;
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    /// <summary>
+    /// Provides extension methods to work with <see cref="FunctionContext"/>.
+    /// </summary>
+    public static class FunctionContextExtensions
+    {
+        /// <summary>
+        /// Retrieve the <see cref="HttpContext"/> for the function execution.
+        /// </summary>
+        /// <param name="context">The <see cref="FunctionContext"/>.</param>
+        /// <returns></returns>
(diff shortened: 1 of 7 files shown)


satvu:
Manually tested but looking to add to the test suite

extensions/Worker.Extensions.Http.AspNetCore/src/FunctionContextExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+        public static HttpContext GetHttpContext(this FunctionContext context)
+        {
+            if (context.Items.TryGetValue(Constants.HttpContextKey, out var requestContext)
+                && requestContext is HttpContext httpContext)
+            {
+                return httpContext;
+            }
+            else
+            {
+                throw new InvalidOperationException($"{nameof(context)} does not have an {nameof(httpContext)} item.");
```
kshyju, line 29: Should we throw or return `null`?

extensions/Worker.Extensions.Http.AspNetCore/src/FunctionContextExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// Retrieve the <see cref="HttpContext"/> for the function execution.
+        /// </summary>
+        /// <param name="context">The <see cref="FunctionContext"/>.</param>
+        /// <returns></returns>
+        public static HttpContext GetHttpContext(this FunctionContext context)
+        {
+            if (context.Items.TryGetValue(Constants.HttpContextKey, out var requestContext)
+                && requestContext is HttpContext httpContext)
+            {
+                return httpContext;
```
kshyju, line 25: We also have another `GetHttpRequestDataAsync` extension method which returns an `HttpRequestData` instance. Wonder we need to keep that in sync if users are updating http props in this HttpContext. Ex: If user adds another request header to the HttpContext.Request, should that be reflected in the HttpRequestData instance? Or customers won't be using the HttpRequestData type in this new world?

extensions/Worker.Extensions.Http.AspNetCore/src/HttpContextConverter.cs
```diff
@@ -18,7 +18,7 @@ public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
 
             if (context.TargetType == typeof(HttpRequest))
             {
-                if (context.FunctionContext.Items.TryGetValue("HttpRequestContext", out var requestContext)
+                if (context.FunctionContext.Items.TryGetValue(Constants.HttpContextKey, out var requestContext)
```
kshyju, line 21: The converter name made me think that this will populate a parameter of type `HttpContext`. Should we rename it to HttpRequestConverter ?

extensions/Worker.Extensions.Http.AspNetCore/src/FunctionContextExtensions.cs
satvu: I don't believe the two models should be combined - so `HttpRequestData` shouldn't be used with `HttpRequest`.

extensions/Worker.Extensions.Http.AspNetCore/src/HttpContextConverter.cs
satvu: Makes sense to me - will change if there are no objections.

extensions/Worker.Extensions.Http.AspNetCore/src/FunctionsMiddleware/FunctionsHttpProxyingMiddleware.cs
```diff
@@ -43,7 +43,7 @@ public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next
 
         private static void AddHttpContextToFunctionContext(FunctionContext funcContext, HttpContext httpContext)
         {
-            funcContext.Items.Add("HttpRequestContext", httpContext);
+            funcContext.Items.Add(Constants.HttpContextKey, httpContext);
```
jviau, line 46: Consider using `IInvocationFeatures` instead. (`FunctionContext.Features`) - as I see `HttpContext` being a feature of this invocation.

extensions/Worker.Extensions.Http.AspNetCore/src/FunctionContextExtensions.cs
```diff
@@ -17,17 +17,15 @@ public static class FunctionContextExtensions
         /// </summary>
         /// <param name="context">The <see cref="FunctionContext"/>.</param>
         /// <returns></returns>
-        public static HttpContext GetHttpContext(this FunctionContext context)
+        public static HttpContext? GetHttpContext(this FunctionContext context)
```
kshyju, line 20: nit: Update the returns part of method summary since this is going to be a pubic extns method.

extensions/Worker.Extensions.Http.AspNetCore/src/FunctionContextExtensions.cs
fabiocav: The methods exist, so we'd either need to make it very clear it's not a supported scenario, or make sure those are in sync (i.e., return an `HttpRequestData` implementation backed by the `HttpContext`). Doesn't need to be part of this PR, but something we should track and flag in the initial preview.

extensions/Worker.Extensions.Http.AspNetCore/src/FunctionsMiddleware/FunctionsHttpProxyingMiddleware.cs
fabiocav: Agreed. Doesn't have to be part of this PR, so we can track this separately.

extensions/Worker.Extensions.Http.AspNetCore/src/HttpContextConverter.cs
fabiocav: I think the change makes sense, but let's do it in a separate PR.

fabiocav (commented):
Minor comments.

fabiocav (approved):

PR #1539, opened 2023-05-11
Remove Worker.Core assembly from DOTNET_STARTUP_HOOKS at startup

### LINKED ISSUE #1318: StartupHook is preventing launching of child .NET processes
Similar to the issue with Hot Reload (https://github.com/dotnet/runtime/issues/58000) when [StarHostAction](https://github.com/Azure/azure-functions-core-tools/blob/v4.x/src/Azure.Functions.Cli/Actions/HostActions/StartHostAction.cs) in the Azure Functions CLI sets DOTNET_STARTUP_HOOKS it should be cleared in [StartupHook](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/src/DotNetWorker.Core/StartupHook.cs) to prevent all child processes from inheriting the environment variable and failing to run.

This is currently blocking the ability to use the [VisualStudioCredential](https://learn.microsoft.com/en-us/dotnet/api/azure.identity.visualstudiocredential?view=azure-dotnet) from [Azure.Identity](https://www.nuget.org/packages/Azure.Identity) in some scenarios as it relies on starting a child process.

jviau:

resolves #1318

### Pull request checklist

### DIFF
--- a/src/DotNetWorker.Core/StartupHook.cs
+++ b/src/DotNetWorker.Core/StartupHook.cs
@@ -20,8 +20,8 @@ using System.Threading;
 internal class StartupHook
 {
     const string StartupHooksEnvVar = "DOTNET_STARTUP_HOOKS";
-    private static readonly char s_startupSeparator = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? ';' : ':';
-    private static readonly string? s_assemblyName = typeof(StartupHook).Assembly.GetName().Name;
+    private static readonly string _startupSeparator = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? ";" : ":";
+    private static readonly string? _assemblyName = typeof(StartupHook).Assembly.GetName().Name;
 
     public static void Initialize()
     {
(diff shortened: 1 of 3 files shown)


src/DotNetWorker.Core/StartupHook.cs
```diff
@@ -62,4 +70,22 @@ static bool WaitOnDebugger(int cycle)
...
+            // If this call happened, we are clearly part of this environment variable.
+            // This is mostly to make strict-nulls happy.
+            return;
+        }
+
+        // netstandard2.0 has no StringSplitOptions overload.
+        IEnumerable<string> parts = startupHooks.Split(s_startupSeparator).Where(x => !string.IsNullOrEmpty(x));
+        string newValue = string.Join(
+            s_startupSeparator.ToString(), // netstandard2.0 only accepts string here.
+            parts.Where(x => !x.Equals(s_assemblyName, StringComparison.Ordinal)));
```
fabiocav, line 88: Do we need to handle null or empty strings here? I'm inclined to say we shouldn't modify the input beyond removing our assembly. 

Wondering if we could simplify to something like:

``` csharp
            IEnumerable<string> parts = startupHooks.Split(s_startupSeparator[0])
                                                    .Where(x => !string.Equals(x, s_assemblyName, StringComparison.Ordinal));
            string newValue = string.Join(s_startupSeparator, parts);
```

The above relies on switching the separator from a `char` to a `string`, makes it a bit more concise/simpler and preserves all other input.

fabiocav (approved):

liliankasem (approved):

PR #1548, opened 2023-05-18
[SourceGen] Add retry options to worker-indexing scenarios

### LINKED ISSUE #1050: Add Retry Policy attributes to source generator
Similar to [this issue](https://github.com/Azure/azure-functions-dotnet-worker/issues/971) for the current function-indexing methods, we want to add retry policy attributes for source-generated work-driven function indexing.

satvu:

resolves #1050 

### Pull request checklist

### Additional information

5/18 update: Leaving in draft, need to finish diagnostics errors, add handling in GrpcWorker to send retry info over, complete tests. This iteration only has the first step (writing out retry options to generated file).

6/5 update: Validated e2e.

Diagnostic descriptors and error handling will be added in a different PR to make the changes smaller/easier to review. Tracking that work in #1592.


sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Emitter.cs
```diff
@@ -110,6 +119,52 @@ private string AddBindingInfo(string bindingListVariableName, IList<IDictionary<
 
                 return builder.ToString();
             }
+
+            private bool TryBuildRetryOptions(GeneratorRetryOptions? retry, out string? retryOptionsAsString)
```
jviau, line 123: Consider passing in the existing `StringBuilder` and appending directly. Would be a good idea to follow that practice everywhere. Also don't need a `bool` return, because if it already has the `StringBuilder` it can just return without appending anything.

Generators can be run in a tight loop in your editor, so bad perf will impact developer experience. The more objects you allocate, the more GC needs to run, which impacts perf.

src/DotNetWorker.Core/FunctionMetadata/DefaultFunctionMetadata.cs
```diff
@@ -47,6 +47,9 @@ public class DefaultFunctionMetadata : IFunctionMetadata
         /// <inheritdoc/>
         public string? ScriptFile { get => _scriptFile; set => ClearIdAndSet(value, ref _scriptFile); }
 
+        /// <inheritdoc/>
+        public IRetryOptions? Retry{ get; set; }
```
jviau, line 51:
```suggestion
        public IRetryOptions? Retry { get; set; }
```

src/DotNetWorker.Core/FunctionMetadata/DefaultRetryOptions.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.Functions.Worker.Core.FunctionMetadata
+{
+    /// <summary>
+    ///
```
jviau, line 9: nit: fill out summary.

src/DotNetWorker.Core/FunctionMetadata/IRetryOptions.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        /// <summary>
+        /// The maximum number of retries allowed per function execution. -1 means to retry indefinitely.
+        /// </summary>
+        public int? MaxRetryCount { get; }
+
+        /// <summary>
+        /// The delay used between retries when using a fixed delay strategy.
+        /// </summary>
+        public string? DelayInterval { get; }
```
jviau, line 21: Can we use `TimeSpan` instead of `string`?

src/DotNetWorker.Core/FunctionMetadata/IRetryOptions.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        /// <summary>
+        /// The maximum retry delay when using an exponential backoff strategy.
+        /// </summary>
+        public string? MaximumInterval { get; }
+
+        /// <summary>
+        /// The retry strategy being used (fixed delay or exponential backoff).
+        /// </summary>
+        public string? Strategy { get; }
```
jviau, line 36: Can we use an enum here?

src/DotNetWorker.Core/FunctionMetadata/IRetryOptions.cs
```diff
@@ -0,0 +1,38 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.Functions.Worker.Core.FunctionMetadata
+{
+    /// <summary>
+    /// Function execution retry policy to use on invocation failures.
+    /// </summary>
+    public interface IRetryOptions
```
jviau, line 11: For the exponential backoff, do we need to also include a backoff coefficient? Or is that something we have historically not included?

src/DotNetWorker.Core/FunctionMetadata/IRetryOptions.cs
satvu: The grpc message uses an enum that's in the protofile, and I was thinking of waiting until the grpcworker layer to use that enum to parse the string here. The pattern being used to couple the grpc and core worker types is having them both implement an interface but for this particular property I was unsure how to go about it. 

I *think* we can still use the same pattern (have two enums, one in protofile and one in core implementing an interface defined in core) but wasn't 100% it was good practice.

src/DotNetWorker.Core/FunctionMetadata/IRetryOptions.cs
satvu: Historically has not been included - here's the related webjobs type [here](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.Abstractions/Description/RetryOptions.cs).

src/DotNetWorker.Core/FunctionMetadata/DefaultRetryOptions.cs
```diff
@@ -0,0 +1,28 @@
...
+        public string? DelayInterval { get; set; }
+
+        /// <inheritdoc/>
+        public string? MinimumInterval { get; set; }
+
+        /// <inheritdoc/>
+        public string? MaximumInterval { get; set; }
+
+        /// <inheritdoc/>
+        public string? Strategy { get; set; }
```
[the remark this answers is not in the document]
satvu: Yes, but I set this value based on the attribute type found during parsing. The flow in the generator and the host for creation/validation is check strategy then go fill out or check the other necessary properties.

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -82,12 +89,13 @@ public IReadOnlyList<GeneratorFunctionMetadata> GetFunctionMetadataInfo(List<Met
...
+            private bool TryGetBindings(MethodDeclarationSyntax method, SemanticModel model, out IList<IDictionary<string, object>>? bindings, out bool hasHttpTrigger, out GeneratorRetryOptions? validatedRetryOptions)
             {
                 hasHttpTrigger = false;
+                validatedRetryOptions = null;
 
-                if (!TryGetMethodOutputBinding(method, model, out bool hasOutputBinding, out IList<IDictionary<string, object>>? methodOutputBindings)
-                    || !TryGetParameterInputAndTriggerBindings(method, model, out hasHttpTrigger, out IList<IDictionary<string, object>>? parameterInputAndTriggerBindings)
+                if (!TryGetMethodOutputBinding(method, model, out bool hasOutputBinding, out GeneratorRetryOptions? retryOptions, out IList<IDictionary<string, object>>? methodOutputBindings)
+                    || !TryGetParameterInputAndTriggerBindings(method, model, out bool supportsRetryOptions, out hasHttpTrigger, out IList<IDictionary<string, object>>? parameterInputAndTriggerBindings)
```
satvu, line 98: I don't love this, but this is a result of retry attributes being used on methods and trigger attributes being used on parameters. After retry options information is retrieved and stored from analyzing method attributes, we need to wait for the parsing of parameter attributes to validate that the trigger supports retry options.

Worried there will be similar scenarios that arise in the future and we continue adding variables like this ...

jviau: ah yeah it is getting unwieldy fast. Maybe add a TODO comment to refactor/clean it up?

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -157,11 +186,63 @@ private bool TryGetMethodOutputBinding(MethodDeclarationSyntax method, SemanticM
...
+                    // Would not expect this to fail since MaxRetryCount is a required value of a retry policy attribute
+                    if (attrProperties!.TryGetValue(Constants.RetryConstants.MaxRetryCountKey, out object? maxRetryCount))
+                    {
+                        retryOptions.MaxRetryCount = maxRetryCount!.ToString();
+                    }
+
+                    // Check which strategy is being used by checking the attribute class
+                    if (SymbolEqualityComparer.Default.Equals(attribute.AttributeClass, _knownFunctionMetadataTypes.FixedDelayRetryAttribute))
+                    {
+                        retryOptions.Strategy = Constants.RetryConstants.FixedDelayRetryName;
```
jviau, line 206: small nit: while this is all internal and is not a big concern. An enum here is more fitting still. Primarily because it is easier and cheaper to compare it later on (enum equality is much cheaper than string equality).

No need to fix this if you aren't bothered, but something to keep in mind for later. Don't be afraid to create multiple internal types to drive your scenario.

jviau (approved):

PR #1578, opened 2023-05-30
Update blob converter & refactor blob unit tests 

liliankasem:
- Refactoring the blob converter to make everything exception ConvertAsync public
- Updating all unit tests so that we are testing through one entry point


test/WorkerExtensionTests/Blob/BlobStorageConverterTests.cs
```diff
@@ -22,375 +24,773 @@
 using Moq;
 using Xunit;
 
-using Constants = Microsoft.Azure.Functions.Worker.Extensions.Storage.Blobs.Constants;
-
+// Tests missing for BlobBaseClient; BlockBlobClient; PageBlobClient; AppendBlobClient
+// Moq does not support mocking extension methods directly
```
liliankasem, line 28: @fabiocav open to suggestions on how to address this. We could always introduce a factory but I figured that might be overkill for right now? Thoughts?

test/WorkerExtensionTests/Blob/BlobStorageConverterTests.cs
```diff
@@ -22,375 +24,773 @@
...
+            var clientResult = (BlobClient)conversionResult.Value;
+
+            // Assert
+            Assert.Equal(ConversionStatus.Succeeded, conversionResult.Status);
+            Assert.Equal("MyBlob", clientResult.Name);
         }
 
-        [Fact]
-        public async Task ConvertAsync_SourceAsModelBindingData_ReturnsSuccess()
+        [Fact (Skip = "Fails: ChangeType doesn't work for mock objects 'Object must implement IConvertible'")]
```
liliankasem, line 82: @fabiocav There are some collection tests that I am skipping for now because I'm assuming your changes will remove the use of ChangeType. Would expect these tests to pass with your changes.

test/WorkerExtensionTests/Blob/BlobStorageConverterTests.cs
```diff
@@ -22,375 +24,773 @@
...
+            Assert.Equal(ConversionStatus.Succeeded, conversionResult.Status);
+            Assert.Equal(expectedStream, streamResult);
         }
 
-        [Theory]
-        [InlineData(Constants.BlobExtensionName, true)]
-        [InlineData(" ", false)]
-        [InlineData("incorrect-value", false)]
-        public void IsBlobExtension_MatchesExpectedOutput(string sourceVal, bool expectedResult)
+        [Fact (Skip = "Fails - broken scenario. ChangeType fails trying to convert MemoryStream to Stream) - 'Object must implement IConvertible'")]
```
liliankasem, line 314: @fabiocav same situation as above but the difference here is that **this is a real broken scenario** (the others only fail because we are mocking them). Also hoping your changes will fix this with the removal of ChangeType

extensions/Worker.Extensions.Storage.Blobs/src/Config/BlobStorageBindingOptions.cs
```diff
@@ -17,7 +17,7 @@ internal class BlobStorageBindingOptions
 
         public BlobClientOptions? BlobClientOptions { get; set; }
 
-        public BlobServiceClient CreateClient()
+        internal virtual BlobServiceClient CreateClient()
```
jviau, line 20: should this options object cache the client? Can then also use that to set the client during a unit test instead of mocking.

```suggestion
        internal BlobServiceClient? Client { get; set; }

        internal BlobServiceClient CreateClient()
        {
            if (this.Client is not null)
            {
                return this.Client;
            }
            
            // create client logic
        }
```

test/WorkerExtensionTests/Blob/BlobStorageConverterTests.cs
```diff
@@ -22,375 +24,773 @@
...
+            var conversionResult = await _blobStorageConverter.ConvertAsync(context);
+            var clientResult = (IEnumerable<BlobClient>)conversionResult.Value;
+
+            // Assert
+            Assert.Equal(ConversionStatus.Succeeded, conversionResult.Status);
+            Assert.IsAssignableFrom<IEnumerable<BlobClient>>(clientResult);
+            Assert.Equal("MyBlob", clientResult.First().Name);
+        }
+
+        [Fact (Skip = "Fails: ChangeType doesn't work for mock objects 'Object must implement IConvertible'")]
```
jviau, line 107: You can add extra interfaces and associated implementations to mock objects.

``` CSharp
TMock mock = new Mock<TMock>();
Mock<IConvertable> convertable = mock.As<IConvertable>();
convertable.Setup(...);
```

jviau (approved):

extensions/Worker.Extensions.Storage.Blobs/src/Config/BlobStorageBindingOptions.cs
liliankasem: Yup! I think there was an issue to do this but will fix here since a) needs to be done and b) cleans up the  unit test!

test/WorkerExtensionTests/Blob/BlobStorageConverterTests.cs
liliankasem: I'm not sure I understand how this works xD Do you mean that TMock is BlobClient i.e.

```csharp
    var mockBlobClient = new Mock<BlobClient>();
    Mock<IConvertible> convertible = mockBlobClient.As<IConvertible>();
```

Or that I should be creating something like TMock and using that to setup the blobClient?

jviau: Sorry - `TMock` is just a placeholder. So yes, your example is correct. `BlobClient` would be `TMock` in this case.

fabiocav (approved):
Let's get this in. We have good coverage with E2E tests as well. I'll rebase on this work and we can iterate on tweaks if needed.

test/WorkerExtensionTests/Blob/BlobStorageConverterTests.cs
```diff
@@ -43,9 +43,7 @@ public BlobStorageConverterTests()
             _mockBlobServiceClient = new Mock<BlobServiceClient>();
 
             var mockBlobOptions = new Mock<BlobStorageBindingOptions>();
-            mockBlobOptions
-                .Setup(m => m.CreateClient())
-                .Returns(_mockBlobServiceClient.Object);
+            mockBlobOptions.Object.Client = _mockBlobServiceClient.Object;
```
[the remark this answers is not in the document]
liliankasem: Ah gotcha, no need to mock the options - understood

liliankasem:
> Let's get this in. We have good coverage with E2E tests as well. I'll rebase on this work and we can iterate on tweaks if needed.

Sounds good, we can address the remainder of Jacob's comments in your refactor PR

Issue #1565, opened 2023-05-25
[SDK-Bindings - EventGrid] Implement converter in EventGrid worker extension


aishwaryabh:
#### Problem Statement
Implement `EventGridConverter` trigger binding types `CloudEvent`, `CloudEvent[]`, `EventGridEvent`, `EventGridEvent[]`, `BinaryData`, `BinaryData[]`, and `String[]` so that we have parity with in-proc.

#### Tasks

- [x] Implement converter that enables binding to a single `CloudEvent`, `EventGridEvent`, `string` and `BinaryData` and multiple `CloudEvent`'s, `EventGridEvent`, `string`, and `BinaryData`
    - Being able to bind to multiple table entities looks like this: `CloudEvent[]`, `EventGridEvent[]`, `BinaryData[]`, `String[]`
- [x] Write unit tests for converters

#### Definition of Done

1. Worker EventGrid extension package is released (as a preview package)
2. Unit & E2E tests written and checked-in
3. Microsoft documentation updated to detail support for the new SDK-bindings

PR #1589, opened 2023-06-02
[SDK Type-Binding] EventGrid Converter

aishwaryabh:
Implement `EventGridConverter` trigger binding types `CloudEvent`, `CloudEvent[]`, `EventGridEvent`, `EventGridEvent[]`, `BinaryData`, `BinaryData[]`, and `String[]` so that we have parity with in-proc.

resolves https://github.com/Azure/azure-functions-dotnet-worker/issues/1565

> Note: We don't want to support deferred binding for EventGrid since the payload that is already passed in from the WebJobs extension can be directly consumed by the dotnet-worker. So it doesn't need to go through converting the EventGrid payload into `ParameterBindingData`, so we wouldn't need to support deferred binding

### Pull request checklist


extensions/Worker.Extensions.EventGrid/src/Worker.Extensions.EventGrid.csproj
```diff
@@ -16,6 +17,12 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\Worker.Extensions.Abstractions\src\Worker.Extensions.Abstractions.csproj" />
+    <ProjectReference Include="..\..\..\src\DotNetWorker.Core\DotNetWorker.Core.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Azure.Messaging.EventGrid" Version="4.16.0" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
```
liliankasem, line 25: We should be using System.Text, or whatever the serializer is set to with worker options; so this import should not be required

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataArrayConverter.cs
```diff
@@ -0,0 +1,62 @@
...
+            if (!CanConvert(context))
+            {
+                return new(ConversionResult.Unhandled());
+            }
+            try
+            {
+                var contextSource = context?.Source as string;
+
+                if (contextSource is not null)
+                {
```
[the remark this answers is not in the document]
aishwaryabh: Should this be failed or unhandled? Failed requires an exception and it looks like right now Unhandled would be returned, not failed

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        public override ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
+        {
+            if (!CanConvert(context))
+            {
+                return new(ConversionResult.Unhandled());
+            }
+            try
+            {
+                var contextSource = context?.Source as string;
```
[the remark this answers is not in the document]
aishwaryabh: Yup I believe so. Every situation I tested converted context.Source into string

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridMultipleCloudEventConverter.cs
```diff
@@ -0,0 +1,49 @@
...
+using Microsoft.Azure.Functions.Worker.Converters;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.EventGrid.TypeConverters
+{
+    /// <summary>
+    /// Converter to bind to CloudEvent parameter.
+    /// </summary>
+    [SupportedConverterType(typeof(CloudEvent[]))]
+    internal class EventGridMultipleCloudEventConverter : EventGridConverterBase<CloudEvent[]>
```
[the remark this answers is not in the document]
aishwaryabh: I combined the converters so that EventGridCloudEventConverter covers both `CloudEvent` and `CloudEvent[]`

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridConverterBase.cs
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.EventGrid.TypeConverters
+{
+    internal abstract class EventGridConverterBase<T> : IInputConverter
+    {
+        private readonly ILogger<EventGridConverterBase<T>> _logger;
+
+        public EventGridConverterBase(ILogger<EventGridConverterBase<T>> logger)
+        {
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
```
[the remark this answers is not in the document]
aishwaryabh: Discussed offline and decided to combine converters for single/collection entities and also remove the base class and add a utility helper class instead.

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridMultipleCloudEventConverter.cs
```diff
@@ -0,0 +1,49 @@
...
+            try
+            {
+                var contextSource = context?.Source as string;
+
+                if (contextSource is not null)
+                {
+                    var cloudEvents = JsonSerializer.Deserialize<CloudEvent[]>(contextSource);
+                    
+                    return new(ConversionResult.Success(cloudEvents));
+                }
```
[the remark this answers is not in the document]
aishwaryabh: Added utility class

extensions/Worker.Extensions.EventGrid/src/Worker.Extensions.EventGrid.csproj
```diff
@@ -6,7 +6,8 @@
     <Description>Azure Event Grid extensions for .NET isolated functions</Description>
 
     <!--Version information-->
-    <VersionPrefix>3.2.1</VersionPrefix>
+    <VersionPrefix>3.3.0</VersionPrefix>
```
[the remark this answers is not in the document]
aishwaryabh: Updated assembly.info

extensions/Worker.Extensions.EventGrid/src/Worker.Extensions.EventGrid.csproj
aishwaryabh: We use JsonConvert in `EventGridBinaryDataConverter` so I think we still do need Newtonsoft.Json

extensions/Worker.Extensions.EventGrid/src/Properties/AssemblyInfo.cs
```diff
@@ -1,6 +1,9 @@
-// Copyright (c) .NET Foundation. All rights reserved.
+﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using System.Runtime.CompilerServices;
 using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
 
 [assembly: ExtensionInformation("Microsoft.Azure.WebJobs.Extensions.EventGrid", "3.2.1")]
```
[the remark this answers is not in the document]
aishwaryabh: where do you update it? I don't see a reference in the EventGrid project so just wondering

extensions/Worker.Extensions.EventGrid/src/EventGridTriggerAttribute.cs
```diff
@@ -1,11 +1,30 @@
...
+using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
+using Microsoft.Azure.Functions.Worker.Extensions.EventGrid.TypeConverters;
 
 namespace Microsoft.Azure.Functions.Worker
 {
+    [AllowConverterFallback(true)]
+    [InputConverter(typeof(EventGridCloudEventConverter))]
+    [InputConverter(typeof(EventGridEventConverter))]
+    [InputConverter(typeof(EventGridBinaryDataConverter))]
+    [InputConverter(typeof(EventGridStringArrayConverter))]
```
surgupta-msft, line 14: I am not seeing `[SupportsDeferredBinding]` attribute on any of these converters. Is this the expected behavior?

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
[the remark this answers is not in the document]
surgupta-msft: Source should be `ModelBindingData` or `CollectionModelBindingData` - [Link](https://github.com/Azure/azure-functions-dotnet-worker/blob/0cc8e2ac342b227cbf78fa2ffca3582bac306ac7/extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs#LL43C17-L43C17) unless I am interpreting this PR incorrectly.

extensions/Worker.Extensions.EventGrid/src/EventGridTriggerAttribute.cs
surgupta-msft: Discussed offline

liliankasem: Yeah we are not using deferred binding for this extension - it is not required as we already get all the information we need to do this without having to use ParameterBindingData

extensions/Worker.Extensions.EventGrid/src/Properties/AssemblyInfo.cs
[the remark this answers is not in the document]
liliankasem: Looks like latest is 3.2.1 so no change here: https://www.nuget.org/packages/Microsoft.Azure.WebJobs.Extensions.EventGrid/3.2.1

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataArrayConverter.cs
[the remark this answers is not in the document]
liliankasem: Unhandled should be okay for now

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
```diff
@@ -0,0 +1,83 @@
...
+                                }
+                            }
+                            return new(ConversionResult.Success(binaryDataList.ToArray()));
+                        }
+                    }
+                }
+            }
+            catch (JsonException ex)
+            {
+                string msg = String.Format(CultureInfo.CurrentCulture,
```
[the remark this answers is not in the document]
liliankasem: An exception message I copied from the queue extension and works here as well

extensions/Worker.Extensions.EventGrid/src/Worker.Extensions.EventGrid.csproj
liliankasem: Right but we get to choose what we use, we don't have to use `JsonConvert.SerializeObject(item)` - we can use the System.Text equivalent so we don't have to bring in a new dependency. As for using the work options instead, here is an example of that:

https://github.com/Azure/azure-functions-dotnet-worker/blob/0cc8e2ac342b227cbf78fa2ffca3582bac306ac7/extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs#L164

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridHelper.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Globalization;
+using System.Text.Json;
+using System.Threading.Tasks;
+using Microsoft.Azure.Functions.Worker.Converters;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.EventGrid.TypeConverters
+{
+    internal static class EventGridHelper
+    {
+        internal static ValueTask<ConversionResult> ConvertHelper(ConverterContext context)
```
liliankasem, line 14: I think we can rename this to be more specific to what this function is doing i.e `DeserilizeToTargetType`

I wonder if we want to leave each converter in control of the ConversionResult and we should just have this helper return whatever `item` is?

extensions/Worker.Extensions.EventGrid/src/Worker.Extensions.EventGrid.csproj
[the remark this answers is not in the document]
liliankasem: What did you change? The WebJobs extension is still 3.2.1 so I'm a little nervous about us changing this to be ahead of WebJobs.  I wonder if 3.2.2 is okay for now? Might be worth asking Fabio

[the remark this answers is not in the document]
aishwaryabh: ahh ok I'll change to 3.2.2. I wasn't sure what to change to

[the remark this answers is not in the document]
aishwaryabh: ooo ok so after trying to compile 3.2.2 doesn't work. Should we just leave the version as 3.1.1? Do we still need preview appended?

extensions/Worker.Extensions.EventGrid/src/Worker.Extensions.EventGrid.csproj
aishwaryabh: Removed Newtonsoft.Json dependency!

samples/WorkerBindingSamples/WorkerBindingSamples.csproj
```diff
@@ -7,16 +7,18 @@
         <Nullable>enable</Nullable>
     </PropertyGroup>
     <ItemGroup>
+        <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.EventGrid" Version="2.1.0" />
```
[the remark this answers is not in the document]
aishwaryabh: removed dependency

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridHelper.cs
aishwaryabh: Discussed offline and decided to keep structure

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
```diff
@@ -0,0 +1,97 @@
...
+        public EventGridBinaryDataConverter(IOptions<WorkerOptions> workerOptions)
+        {
+            _workerOptions = workerOptions ?? throw new ArgumentNullException(nameof(workerOptions));
+        }
+        public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
+        {
+            if (context is null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
```
liliankasem, line 35: context should never be null at this point so you can probably just remove this check from all converters; + you're already using context? so that works fine

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
```diff
@@ -0,0 +1,97 @@
...
+
+            if (context.TargetType != typeof(BinaryData) && context.TargetType != typeof(BinaryData[]))
```
liliankasem, line 37: nit:
```suggestion
            if (!(context.TargetType is BinaryData || context.TargetType is BinaryData[]))
```

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
```diff
@@ -0,0 +1,97 @@
...
+                return new(ConversionResult.Unhandled());
+            }
+
+            try
+            {
+                var contextSource = context?.Source as string;
+
+                if (contextSource is null)
+                {
+                    return new(ConversionResult.Unhandled());
```
liliankasem, line 48: I think we can drop this if-statement; `context?.Source` shouldn't be null, and if there is a failed cast we will catch the exception and fail there (for all the converters)

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
```diff
@@ -0,0 +1,97 @@
...
+                    {
+                        foreach (var item in jsonData)
+                        {
+                            if (item is not null)
+                            {
+                                var binaryData = new BinaryData(item);
+                                binaryDataList.Add(binaryData);
+                            }
+                        }
+                        return new(ConversionResult.Success(binaryDataList.ToArray()));
```
liliankasem, line 74: nit on this if-else-staement:

I think this would be cleaner as a switch with a method for  ConvertToBinaryDataArray i.e.

```csharp
var result = context!.TargetType switch
{
    Type t when t == typeof(BinaryData) => BinaryData.FromObjectAsJson<string>(contextSource),
    Type t when t == typeof(BinaryData[]) => ConvertToBinaryDataArray(contextSource),
    _ => null
};

if (result is not null)
{
    return new(ConversionResult.Success(result));
}
```

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
```diff
@@ -0,0 +1,97 @@
...
+                {
+                    var binaryData = BinaryData.FromObjectAsJson<string>(contextSource);
+                    return new(ConversionResult.Success(binaryData));
+                }
+                else if (context!.TargetType == typeof(BinaryData[]))
+                {
+                    byte[] byteArray = Encoding.UTF8.GetBytes(contextSource);
+                    MemoryStream stream = new MemoryStream(byteArray);
+
+                    var jsonData = (List<object>?)_workerOptions?.Value?.Serializer?.Deserialize(stream, typeof(List<object>), CancellationToken.None);
```
liliankasem, line 61: I think this could be simplified a bit:

1. we probably don't need all of the worker options so we can update the ctor and grab just the Serializer

`_serializer = workerOptions?.Value?.Serializer;`

2. we might be able to do something like this ( i haven't tested the code)

```csharp
        private BinaryData[]? ConvertToBinaryDataArray(string contextSource)
        {
            var contentStream = new MemoryStream(Encoding.UTF8.GetBytes(contextSource ?? ""));
            var binaryDataList = _serializer.Deserialize(contentStream, typeof(BinaryData[]), CancellationToken.None);

            return binaryDataList as BinaryData[] ?? null;
        }
```

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridStringArrayConverter.cs
```diff
@@ -0,0 +1,88 @@
...
+                {
+                    foreach (var item in jsonData)
+                    {
+                        if (item is not null)
+                        {
+                            var data = item.ToString();
+                            stringList.Add(data);
+                        }
+                    }
+                    return new(ConversionResult.Success(stringList.ToArray()));
```
liliankasem, line 66: Same comments here as `EventGridBinaryDataConverter`

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridHelper.cs
```diff
@@ -0,0 +1,46 @@
...
+            {
+                var contextSource = context?.Source as string;
+
+                if (contextSource is null)
+                {
+                    return new(ConversionResult.Unhandled());
+                }
+                
+                var targetType = context!.TargetType;
+                var item = JsonSerializer.Deserialize(contextSource, targetType);
```
liliankasem, line 26: I wonder if we should be consistent and use the worker options sterilizer for all

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
aishwaryabh: If we do #2, we run into this error `Deserialization of types without a parameterless constructor, a singular parameterized constructor, or a parameterized constructor annotated with 'JsonConstructorAttribute' is not supported. Type 'System.BinaryData'`. That's why I just deserialized the object to a List<object> to get all the values from the list and then loop through each one to create a new BinaryData object. Does that make sense?

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
aishwaryabh: Some unit tests fail if I remove this statement (if contextSource is an empty object for example, it will still serialize that into BinaryData and return success even if it shouldn't be). Is that ok to keep this if statement?

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridStringArrayConverter.cs
```diff
@@ -0,0 +1,88 @@
...
+using Microsoft.Azure.Functions.Worker.Converters;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.EventGrid.TypeConverters
+{
+    /// <summary>
+    /// Converter to bind to string[] parameter.
+    /// </summary>
+    [SupportedConverterType(typeof(string[]))]
+    internal class EventGridStringArrayConverter : IInputConverter
```
JoshLove-msft, line 21: This is already handled in the extension - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventgrid/Microsoft.Azure.WebJobs.Extensions.EventGrid/src/Config/EventGridExtensionConfigProvider.cs#L74-L75

Why do we want to implement it here as well?

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
```diff
@@ -0,0 +1,97 @@
...
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.EventGrid.TypeConverters
+{
+    /// <summary>
+    /// Converter to bind to BinaryData or BinaryData[] parameter.
+    /// </summary>
+    [SupportedConverterType(typeof(BinaryData))]
+    [SupportedConverterType(typeof(BinaryData[]))]
+    internal class EventGridBinaryDataConverter : IInputConverter
```
JoshLove-msft, line 22: This is already handled in the extension - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventgrid/Microsoft.Azure.WebJobs.Extensions.EventGrid/src/Config/EventGridExtensionConfigProvider.cs#L82-L83

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
```diff
@@ -0,0 +1,97 @@
...
+                            }
+                        }
+                        return new(ConversionResult.Success(binaryDataList.ToArray()));
+                    }
+                }
+            }
+            catch (JsonException ex)
+            {
+                string msg = String.Format(CultureInfo.CurrentCulture,
+                    @"Binding parameters to complex objects uses Json.NET serialization.
```
JoshLove-msft, line 81: This exception isn't correct because we are using System.Text.Json serializer, not JSON.NET.

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
JoshLove-msft: System.Text.Json cannot be used to deserialize into BinaryData. Here is how it is handled in the extension - https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventgrid/Microsoft.Azure.WebJobs.Extensions.EventGrid/src/Config/EventGridExtensionConfigProvider.cs#L83

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridHelper.cs
aishwaryabh: I think it might be worth keeping the JsonSerializer for some, since we'd have to convert the contextSource into a byte array and then convert it to a Stream just so we can use the worker options serializer. Using this method makes it much more concise imo.

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
```diff
@@ -0,0 +1,97 @@
...
+                var contextSource = context?.Source as string;
+
+                if (contextSource is null)
+                {
+                    return new(ConversionResult.Unhandled());
+                }
+
+                if (context!.TargetType == typeof(BinaryData))
+                {
+                    var binaryData = BinaryData.FromObjectAsJson<string>(contextSource);
```
JoshLove-msft, line 53: This would end up reserializing the JSON object into a JSON string, instead of wrapping the JSON object into a BinaryData instance directly (which is what we want).

```c#
var json = "{ \"name\": \"John\" }";
Console.WriteLine(BinaryData.FromObjectAsJson<string>(json));
Console.WriteLine(BinaryData.FromString(json));
```

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
```diff
@@ -0,0 +1,97 @@
...
+                        return new(ConversionResult.Success(binaryDataList.ToArray()));
+                    }
+                }
+            }
+            catch (JsonException ex)
+            {
+                string msg = String.Format(CultureInfo.CurrentCulture,
+                    @"Binding parameters to complex objects uses Json.NET serialization.
+                    1. Bind the parameter type as 'string' instead to get the raw values and avoid JSON deserialization, or
+                    2. Change the queue payload to be valid json.
```
JoshLove-msft, line 83:
```suggestion
                    2. Change the event payload to be valid JSON.
```

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
```diff
@@ -0,0 +1,97 @@
...
+                    The JSON parser failed: {0}",
+                    ex.Message);
+
+                return new(ConversionResult.Failed(new InvalidOperationException(msg)));
```
JoshLove-msft, line 87: Should we include the underlying JSON exception?

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridHelper.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Globalization;
+using System.Text.Json;
+using System.Threading.Tasks;
+using Microsoft.Azure.Functions.Worker.Converters;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.EventGrid.TypeConverters
+{
+    internal static class EventGridHelper
+    {
+        internal static ValueTask<ConversionResult> DeserilizeToTargetType(ConverterContext context)
```
JoshLove-msft, line 14:
```suggestion
        internal static ValueTask<ConversionResult> DeserializeToTargetType(ConverterContext context)
```

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridHelper.cs
JoshLove-msft: I think we want to use the default JsonSerializer. The EventGridEvent and CloudEvent types have JsonConverters defined that will handle the serialization. If we passed in options that specified separate converters it's possible that those converters would override the standard ones.

extensions/Worker.Extensions.EventGrid/src/Worker.Extensions.EventGrid.csproj
```diff
@@ -16,6 +17,11 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\Worker.Extensions.Abstractions\src\Worker.Extensions.Abstractions.csproj" />
+    <ProjectReference Include="..\..\..\src\DotNetWorker.Core\DotNetWorker.Core.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Azure.Messaging.EventGrid" Version="4.16.0" />
```
JoshLove-msft, line 24:
```suggestion
    <PackageReference Include="Azure.Messaging.EventGrid" Version="4.17.0" />
```

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
aishwaryabh: @JoshLove-msft Does the original way for deserializing into BinaryData[] look ok?

JoshLove-msft: I don't think we need to convert the string into bytes and then into a stream. We can deserialize the string into List<string> and then iterate and create BD instances using  BD.FromString.

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridStringArrayConverter.cs
liliankasem: If it's handled in the extension we don't need to repeat - can probably remove the string and string[] converters 

@aishwarya can you validate that string bindings work without these new converters?

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridHelper.cs
aishwaryabh: Sounds good I'll keep the default JsonSerializer then

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
aishwaryabh: yup it's included already by adding in ex.message!

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
aishwaryabh: Changed!

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
JoshLove-msft: It's included in the message, but should we also include it as an inner exception for the InvalidOperationException?

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
aishwaryabh: Changed!

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
```diff
@@ -21,66 +17,32 @@ namespace Microsoft.Azure.Functions.Worker.Extensions.EventGrid.TypeConverters
...
+                        return new(ConversionResult.Success((BinaryData.FromString(contextSource))));
+                    case Type t when t == typeof(BinaryData[]):
+                        return new(ConversionResult.Success(ConvertToBinaryDataArray(contextSource)));
                 }
             }
             catch (JsonException ex)
             {
                 string msg = String.Format(CultureInfo.CurrentCulture,
-                    @"Binding parameters to complex objects uses Json.NET serialization.
+                    @"Binding parameters to complex objects uses System.Text.Json serialization.
```
JoshLove-msft, line 43: I would just leave it as JSON serialization.

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridStringArrayConverter.cs
```diff
@@ -0,0 +1,69 @@
...
+            try
+            {
+                var contextSource = context?.Source as string;
+
+                if (contextSource is null)
+                {
+                    return new(ConversionResult.Unhandled());
+                }
+
+                var jsonData = JsonSerializer.Deserialize(contextSource, typeof(List<object>)) as List<object>;
```
JoshLove-msft, line 34: Why not just Deserialize directly to `List<string>` using the generic Deserialize overload?

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridStringArrayConverter.cs
aishwaryabh: Synced offline and came to the conclusion to keep the StringArrayConverter, since the JSONConverter doesnt' work as expected

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridStringArrayConverter.cs
aishwaryabh: I get a JSON exception if I do `JsonSerializer.Deserialize(contextSource, typeof(string[]))`. Is that what you meant?

liliankasem (approved):

extensions/Worker.Extensions.EventGrid/src/release_notes.md
```diff
@@ -0,0 +1,3 @@
+### Microsoft.Azure.Functions.Worker.Extensions.EventGrid <version>
+
+- Add ability to bind a event grid trigger to CloudEvent, CloudEvent[], EventGridEvent, EventGridEvent[], BinaryData, BinaryData[], and String[]
```
JoshLove-msft, line 3:
```suggestion
- Add ability to bind a event grid trigger to CloudEvent, CloudEvent[], EventGridEvent, EventGridEvent[], BinaryData, BinaryData[], and string[]
```

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridStringArrayConverter.cs
JoshLove-msft: Ah ok - we are dealing with JSON objects so that won't work. We can still use an object array instead in case System.Text.Json is able to optimize that and we avoid the subsequent ToArray call on L47.

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridStringArrayConverter.cs
```diff
@@ -0,0 +1,69 @@
...
+                List<string> stringList = new List<string>();
+
+                if (jsonData is not null)
+                {
+                    foreach (var item in jsonData)
+                    {
+                        if (item is not null)
```
JoshLove-msft, line 41: Why do we explicitly skip null items? I don't think it is possible given how Event Grid works, but in theory if there was a null JSON object in the payload, it would be reasonable to include that as a null in the string array.

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridStringArrayConverter.cs
```diff
@@ -0,0 +1,69 @@
...
+                        {
+                            var data = item.ToString();
+                            stringList.Add(data);
+                        }
+                    }
+                    return new(ConversionResult.Success(stringList.ToArray()));
```
JoshLove-msft, line 47: We could use a Linq expression, like:
```c#
ConversionResult.Success(jsonData.Select(d => d.ToString()).ToArray())
```
and avoid the need for the for loop.

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridStringArrayConverter.cs
```diff
@@ -0,0 +1,69 @@
...
+            {
+                return new(ConversionResult.Unhandled());
+            }
+            try
+            {
+                var contextSource = context?.Source as string;
+
+                if (contextSource is null)
+                {
+                    return new(ConversionResult.Failed(new InvalidOperationException("Context source cannot be null")));
```
JoshLove-msft, line 31: The context source wasn't necessarily null - it may just not have been a string.

JoshLove-msft, line 31:
```suggestion
                    return new(ConversionResult.Failed(new InvalidOperationException("Context source must be a non-null string")));
```

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
```diff
@@ -0,0 +1,79 @@
...
+        private BinaryData[]? ConvertToBinaryDataArray(string contextSource)
+        {
+            var jsonData = JsonSerializer.Deserialize(contextSource, typeof(List<object>)) as List<object>;
+            List<BinaryData> binaryDataList = new List<BinaryData>();
+
+            if (jsonData is not null)
+            {
+                foreach (var item in jsonData)
+                {
+                    if (item is not null)
```
JoshLove-msft, line 68: Similar question about the null checking here as for the string array method.

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
```diff
@@ -0,0 +1,79 @@
...
+            }
+            catch (Exception ex)
+            {
+                return new(ConversionResult.Failed(ex));
+            }
+
+            return new(ConversionResult.Unhandled());
+        }
+
+        private BinaryData[]? ConvertToBinaryDataArray(string contextSource)
```
JoshLove-msft, line 59: Doesn't look like the returned array would ever be null, though it probably should be if `jsonData` was null.

aishwaryabh: Kept this since I don't check if `jsonData` is null anymore

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
aishwaryabh: Removed null check

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridStringArrayConverter.cs
aishwaryabh: I included null items now

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
```diff
@@ -0,0 +1,76 @@
...
+        private BinaryData?[]? ConvertToBinaryDataArray(string contextSource)
+        {
+            var jsonData = JsonSerializer.Deserialize(contextSource, typeof(List<object>)) as List<object>;
+            List<BinaryData?> binaryDataList = new List<BinaryData?>();
+
+            if (jsonData is not null)
+            {
+                foreach (var item in jsonData)
+                {
+                    var binaryData = BinaryData.FromString(item.ToString());
```
JoshLove-msft, line 68: if item is null we should not create the BD instance but just pass null to the list.

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridHelper.cs
```diff
@@ -0,0 +1,46 @@
...
+            catch (JsonException ex)
+            {
+                string msg = String.Format(CultureInfo.CurrentCulture,
+                    @"Binding parameters to complex objects uses JSON serialization.
+                    1. Bind the parameter type as 'string' instead to get the raw values and avoid JSON deserialization, or
+                    2. Change the event payload to be valid json.
+                    The JSON parser failed: {0}",
+                    ex.Message);
+
+                return new(ConversionResult.Failed(new InvalidOperationException(msg)));
```
jviau, line 38: Should the `JsonException` be passed as `InnerException`?

jviau (approved):

extensions/Worker.Extensions.EventGrid/src/Worker.Extensions.EventGrid.csproj
```diff
@@ -6,7 +6,8 @@
     <Description>Azure Event Grid extensions for .NET isolated functions</Description>
 
     <!--Version information-->
-    <VersionPrefix>3.2.1</VersionPrefix>
+    <VersionPrefix>3.2.2</VersionPrefix>
```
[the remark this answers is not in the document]
JoshLove-msft: Should this be a preview still?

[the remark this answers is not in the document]
JoshLove-msft: Nevermind I see that it is a preview.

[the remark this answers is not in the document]
aishwaryabh: So I initially changed the version to 3.3.0 but @liliankasem suggested it should be 3.2.2[ here](https://github.com/Azure/azure-functions-dotnet-worker/pull/1589#discussion_r1227228716) since we shouldn't be out of sync for major/minor version when comparing the webjobs extension.

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
aishwaryabh: I ended up adding null if item is null to address this scenario

extensions/Worker.Extensions.EventGrid/src/Worker.Extensions.EventGrid.csproj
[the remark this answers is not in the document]
fabiocav: There's no requirement for the versions to be aligned with the WebJobs extension version. It helps avoid confusion in many cases, but semver should be followed for the worker extension and takes precedence over that convenience.

In this case, a minor version bump would be appropriate.

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
```diff
@@ -0,0 +1,76 @@
...
+    {
+        public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
+        {
+            try
+            {
+                var contextSource = context?.Source as string;
+
+                if (contextSource == null)
+                {
+                    return new(ConversionResult.Failed(new InvalidOperationException("Context source cannot be null")));
```
fabiocav, line 28: The message here is not quite accurate. `contextSource` would be `null` here if `Source` is not a string, and the message wouldn't be correct.

Something like:
`"The input source should be a non-null string." would be more appropriate.

Extra:  You may also want to add the source type there so we know something came in with a different type if that was ever the case.

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
```diff
@@ -0,0 +1,76 @@
...
+                }
+                var targetType = context!.TargetType;
```
fabiocav, line 30: A parameter `null` check would simplify that and provide better diagnostics instead failing with a null reference exception here if this was ever `null` for any reason.

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
fabiocav: Flagging this as unresolved as it doesn't look like the change was applied.

The exception should indeed be added as the inner exception for the `InvalidOperationException` being created. We can also remove the message from the one you're crafting

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridHelper.cs
```diff
@@ -0,0 +1,46 @@
...
+{
+    internal static class EventGridHelper
+    {
+        internal static ValueTask<ConversionResult> DeserializeToTargetType(ConverterContext context)
+        {
+            try
+            {
+                var contextSource = context?.Source as string;
+
+                if (contextSource is null)
```
fabiocav, line 20:
```suggestion
                if (context?.Source is not string contextSource)
```

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
```diff
@@ -0,0 +1,76 @@
...
+    [SupportedConverterType(typeof(BinaryData[]))]
+    internal class EventGridBinaryDataConverter : IInputConverter
+    {
+        public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
+        {
+            try
+            {
+                var contextSource = context?.Source as string;
+
+                if (contextSource == null)
```
fabiocav, line 26:
```suggestion
                if (context?.Source is not string contextSource)
```

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridHelper.cs
```diff
@@ -0,0 +1,46 @@
...
+                {
+                    return new(ConversionResult.Failed(new InvalidOperationException("Context source cannot be null")));
```
fabiocav, line 22: Same comment about the message here.

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridHelper.cs
```diff
@@ -0,0 +1,46 @@
...
+                var item = JsonSerializer.Deserialize(contextSource, targetType);
+                return new(ConversionResult.Success(item));
+            }
+            catch (JsonException ex)
+            {
+                string msg = String.Format(CultureInfo.CurrentCulture,
+                    @"Binding parameters to complex objects uses JSON serialization.
+                    1. Bind the parameter type as 'string' instead to get the raw values and avoid JSON deserialization, or
+                    2. Change the event payload to be valid json.
+                    The JSON parser failed: {0}",
```
fabiocav, line 35: The exception is passed in as the inner exception. Should be fine to remove the message.

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridStringArrayConverter.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            if (context.TargetType != typeof(string[]))
+            {
+                return new(ConversionResult.Unhandled());
+            }
+            try
+            {
+                var contextSource = context?.Source as string;
+
+                if (contextSource is null)
```
fabiocav, line 30:
```suggestion
                if (context?.Source is not string contextSource)
```

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridStringArrayConverter.cs
```diff
@@ -0,0 +1,62 @@
...
+                {
+                    return new(ConversionResult.Failed(new InvalidOperationException("Context source must be a non-null string")));
```
fabiocav, line 32: Good message here! :)

samples/WorkerBindingSamples/EventGrid/EventGridTriggerBindingSamples.cs
```diff
@@ -0,0 +1,105 @@
...
+    {
+        private readonly ILogger _logger;
+
+        public EventGridTriggerBindingSamples(ILoggerFactory loggerFactory)
+        {
+            _logger = loggerFactory.CreateLogger<EventGridTriggerBindingSamples>();
+        }
+
+        [Function("MyEventFunction")]
+        public void MyEventFunction([EventGridTrigger] MyEvent input)
```
fabiocav, line 21: Great to have samples using this!

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
aishwaryabh: I initially had the param check but Lilian had mentioned that context is never going to be null. Is it ok to keep it this way?

extensions/Worker.Extensions.EventGrid/src/Worker.Extensions.EventGrid.csproj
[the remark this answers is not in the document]
aishwaryabh: Updated to 3.3.0

extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
fabiocav: Being defensive here and having the parameter check helps us identify issues in cases where, even if we don't expect (and particularly those), this is passed in as `null` as we'd get an exception with the detailed information/parameter, making diagnostics simpler. Been hit too many times with null reference exceptions in places where the assumption was that it would never happen.

There are a few scenarios that could potentially invalidate the assumption as the code evolves (including scenarios where the converter may not be directly called by the binding infrastructure).

It's unlikely this would ever be `null`, I'm not against keeping this as-is (principally in this case as it wouldn't be super difficult to identify the issue with a stack), but wanted to share the context above as a defensive approach, even in cases like this, help make the potential issues easier to diagnose.

aishwaryabh: Added check for if context is null!

fabiocav (approved):

PR #1594, opened 2023-06-06
Adding support for linux cold start

kshyju:


host/azure-pipelines.yml
```diff
@@ -15,31 +15,111 @@ pr:
...
-      command: "pack"
-      packagesToPack: "$(Build.SourcesDirectory)/host/tools/build/Microsoft.Azure.Functions.DotnetIsolatedNativeHost.nuspec"
-      versioningScheme: "off"
-      packDestination: "$(Build.ArtifactStagingDirectory)/host/build/nuget"
-      basePath: "$(Build.SourcesDirectory)/host/tools/build"
-
-  # Publish artifacts.
-  - publish: $(Build.ArtifactStagingDirectory)/host/build/nuget
-    artifact: drop
+stages:
```
kshyju, line 18: Example build using this pipeline: https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=133946&view=results

host/src/FunctionsNetHost/Logger.cs
```diff
@@ -17,43 +16,30 @@ static Logger()
...
 
-        internal static bool IsDebugLogEnabled => IsDebugLevelLogEnabled;
-
-        /// <summary>
-        /// Logs a debug message if "AZURE_FUNCTIONS_FUNCTIONSNETHOST_TRACE" environment variable value is set to "1"
-        /// For optimal performance (when building the log message needs execute some code (Ex: someObj.ToString() or so),
-        /// consider checking "IsDebugLogEnabled" before calling LogDebug.
-        /// </summary>
-        internal static void LogDebug(string message)
+        internal static bool IsTraceLogEnabled
```
kshyju, line 21: Simplified this Logger to only 2 methods and one property to expose whether trace level logging is enabled.

In the previous version I cached the environment variable value (to determine whether trace logging should be enabled or not). That would not work if customer wants to enable this setting via portal (configuration->settings). If we observe an issue with calls to environment variable read, we can revisit.

host/src/FunctionsNetHost/AppLoader/NetHost.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        [DllImport("nethost", CharSet = CharSet.Auto)]
+        private static extern int get_hostfxr_path(
+        [Out] char[] buffer,
+        [In] ref int buffer_size,
+        IntPtr reserved);
+
+        internal static string GetHostFxrPath()
+        {
+            char[] buffer = new char[500];
```
kshyju, line 18: Hopefully 500 is enough.
Example paths: 
C:\Program Files\dotnet\host\fxr\7.0.5\hostfxr.dll
C:\Program Files\dotnet\host\fxr\8.0.0-preview.4.23259.5\hostfxr.dll
/usr/lib/dotnet/host/fxr/7.0.5/libhostfxr.so

May be 200 may be enough.

host/src/FunctionsNetHost/AppLoader/NetHost.cs
```diff
@@ -0,0 +1,31 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Runtime.InteropServices;
+
+namespace FunctionsNetHost
+{
+    internal class NetHost
+    {
+        [DllImport("nethost", CharSet = CharSet.Auto)]
+        private static extern int get_hostfxr_path(
```
kshyju, line 11: https://github.com/dotnet/runtime/blob/main/docs/design/features/native-hosting.md#new-host-binary-for-finding-hostfxr

Found that this method always return the latest (If a preview version exists, that will be returned). In one machine I had net6,7 and 8 preview and I got 8 preview path.

jviau (commented):
Please go over your logger changes and:

1. Ensure they all use explicit verbositry overloads - `Log{Verbosity}(...)`
2. Use structured logging as appropriate: ❌  `logger.LogTracr($"message {variable}")` -> ✅ `logger.LogTrace("message {Variable}", variable)`
3. Use exception overloads, do not pass in exception as a template argument: ❌ `logger.LogError($"message {ex}")` -> ✅ `logger.LogError(ex, "message")`
    - If you really want to, you can format the exception message into the template. But it typically is not needed, log providers should have special handling for the exception argument.
4. Nice to have / maybe a later PR: consider using high performance logging. Such as logging source gen: https://learn.microsoft.com/en-us/dotnet/core/extensions/logger-message-generator

host/src/FunctionsNetHost/Native/NativeExports.cs
```diff
@@ -35,24 +35,24 @@ public static int GetApplicationProperties(NativeHostData nativeHostData)
                                                 delegate* unmanaged<byte**, int, IntPtr, IntPtr> requestCallback,
             IntPtr grpcHandler)
         {
-            Logger.LogDebug("NativeExports.RegisterCallbacks method invoked.");
-            
+            Logger.LogTrace("NativeExports.RegisterCallbacks method invoked.");
+
             try
             {
                 NativeHostApplication.Instance.SetCallbackHandles(requestCallback, grpcHandler);
```
fabiocav, line 42: Can we use the application handle provided instead of going through a static here?

host/src/FunctionsNetHost/Native/NativeExports.cs
```diff
@@ -35,24 +35,24 @@ public static int GetApplicationProperties(NativeHostData nativeHostData)
...
         }
 
         [UnmanagedCallersOnly(EntryPoint = "send_streaming_message")]
         public static unsafe int SendStreamingMessage(IntPtr pInProcessApplication, byte* streamingMessage, int streamingMessageSize)
         {
-            Logger.LogDebug($"NativeExports.SendStreamingMessage method invoked.");
+            Logger.LogTrace($"NativeExports.SendStreamingMessage method invoked.");
 
             try
             {
```
fabiocav, line 58: Same comment here (and on all methods taking the handle) on using the handle vs the static approach.

src/DotNetWorker.Grpc/NativeHostIntegration/NativeMethods.cs
```diff
@@ -8,14 +8,29 @@
...
-            _ = get_application_properties(out var hostData);
-            return hostData;
+            var result = get_application_properties(out var hostData);
+            if (result == 1)
+            {
+                return hostData;
+            }
+
+            throw new InvalidOperationException(
+                $"Invoking get_application_properties method failed. Expected result:1, Actual result:{result}");
```
fabiocav, line 33: nit: Change to `$"Invalid result returned from get_application_properties: {result}"`

host/src/FunctionsNetHost/Native/NativeExports.cs
kshyju: @fabiocav  I had to revert the iteration I pushed with this change earlier today, as found an issue with that. approach I am going to address that change as a separate PR. This current version is working and can be for used for our tests and measurements while I work on the switch to using the handle.

fabiocav: As discussed offline, let's not block on this. Get the changes in with this as-is and we'll continue to iterate.

kshyju:
> Please go over your logger changes and:
> 
> 1. Ensure they all use explicit verbositry overloads - `Log{Verbosity}(...)`
> 2. Use structured logging as appropriate: ❌  `logger.LogTracr($"message {variable}")` -> ✅ `logger.LogTrace("message {Variable}", variable)`
> 3. Use exception overloads, do not pass in exception as a template argument: ❌ `logger.LogError($"message {ex}")` -> ✅ `logger.LogError(ex, "message")`
>    
>    * If you really want to, you can format the exception message into the template. But it typically is not needed, log providers should have special handling for the exception argument.
> 4. Nice to have / maybe a later PR: consider using high performance logging. Such as logging source gen: https://learn.microsoft.com/en-us/dotnet/core/extensions/logger-message-generator

This Logger.Log method here is a single place to direct all log messages to std output. I originally had LogInfo, But in a later iteration I changed it back to Log as we treat all log messages as same (forward them to std out after appending the prefix) - we just send it to stdout, so I decided to keep it to a single/simple Log message. LogTrace was added so that callers can do a one liner instead of Logger.IsDebugLogEnabled { Log().

host/tools/build/Microsoft.Azure.Functions.DotnetIsolatedNativeHost.nuspec
```diff
@@ -17,7 +17,8 @@
     </contentFiles>
   </metadata>
   <files>
-    <file src="..\..\build\win-x64\FunctionsNetHost\*" target="contentFiles/any/any/workers/dotnet-isolated/bin" />
+    <file src="..\..\dist\windows\*" target="contentFiles/any/any/workers/dotnet-isolated/bin" />
+    <file src="..\..\dist\linux\*" target="contentFiles/any/any/workers/dotnet-isolated/bin" />
```
jviau, line 21: This is putting both windows and linux files in the same folder. Will they not overwrite each other?

Curious why you are not using packaging from `.csproj` as well? You can control all the specific runtime stuff from there.

jviau:
> This Logger.Log method here is a single place to direct all log messages to std output. I originally had LogInfo, But in a later iteration I changed it back to Log as we treat all log messages as same (forward them to std out after appending the prefix) - we just send it to stdout, so I decided to keep it to a single/simple Log message. LogTrace was added so that callers can do a one liner instead of Logger.IsDebugLogEnabled { Log().

Ah, completely missed this is not an `ILogger`, but a static class. Is there a reason we only use std out here? Is this the final logging design for this? Will customers be able to suppress logs from the native host?

host/tools/build/Microsoft.Azure.Functions.DotnetIsolatedNativeHost.nuspec
kshyju: The file extensions are different for windows & Linux. So we are good.

Originally we did not have this C# project (so no csporj), hence we opted for .nuspec file. Yea. We could consider switching to that. Perhaps in a follow up PR.

kshyju:
> > This Logger.Log method here is a single place to direct all log messages to std output. I originally had LogInfo, But in a later iteration I changed it back to Log as we treat all log messages as same (forward them to std out after appending the prefix) - we just send it to stdout, so I decided to keep it to a single/simple Log message. LogTrace was added so that callers can do a one liner instead of Logger.IsDebugLogEnabled { Log().
> 
> Ah, completely missed this is not an `ILogger`, but a static class. Is there a reason we only use std out here? Is this the final logging design for this? Will customers be able to suppress logs from the native host?

Customers won't be able to suppres this. These are more like system logs. 
Writing to std out is how the worker communicates to host before establishing grpc connection. We do have an opportunity to switch to grpc logs after grpc handshake is done. Will explore that (in a future PR).

kshyju:
Used bits from this PR for validation in win & linux consumption stage0

host/tools/build/worker.config.json
```diff
@@ -5,5 +5,21 @@
     "defaultExecutablePath": "%FUNCTIONS_WORKER_DIRECTORY%/bin/FunctionsNetHost.exe",
     "defaultWorkerPath": "bin/FunctionsNetHost.exe",
     "workerIndexing": "true"
-  }
+  },
+  "profiles": [
+    {
+      "profileName": "DotnetIsolatedLinux",
```
fabiocav, line 11: 👍🏼

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
```diff
@@ -127,7 +127,7 @@ public override HttpResponseData CreateResponse()
             return new GrpcHttpResponseData(FunctionContext, System.Net.HttpStatusCode.OK);
         }
 
-#if NET5_0
+#if NET5_0_OR_GREATER
```
fabiocav, line 130: Should this change be done on `main`, outside of this PR?

src/DotNetWorker.Grpc/NativeHostIntegration/NativeMethods.cs
```diff
@@ -31,15 +41,39 @@ public static void SendStreamingMessage(NativeSafeHandle nativeApplication, Stre
...
+        /// <summary>
+        /// Custom import resolve callback.
+        /// When trying to resolve "FunctionsNetHost", we return the handle using GetMainProgramHandle API in this callback.
+        /// </summary>
+        private static IntPtr ImportResolver(string libraryName, System.Reflection.Assembly assembly, DllImportSearchPath? searchPath)
+        {
+            if (libraryName == NativeWorkerDll)
+            {
+#if NET6_0
+                if (OperatingSystem.IsLinux())
```
fabiocav, line 64: 👍🏼

fabiocav (approved):
Nothing major, so approving so we don't block on bringing this to the feature branch. Still have some comments open and want to revisit the flow to use the handle resolved and provided from managed -> native

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
kshyju: Yes. I will create a PR to fix that in main.

host/src/FunctionsNetHost/FunctionsNetHost.csproj
```diff
@@ -16,7 +16,13 @@
   <PropertyGroup>
     <ExportsFile>$(MSBuildThisFileDirectory)exports.def</ExportsFile>
   </PropertyGroup>
-  
+
+  <!--Below is needed to tell linker to include our exported functions in net7.0.-->
+  <!--Use IlcExportUnmanagedEntrypoints when adopting .net8.0-->
+  <ItemGroup>
+    <CustomLinkerArg Include="-Wl,--export-dynamic" Condition=" '$([MSBuild]::IsOSPlatform(`Linux`))' == 'true' " />
```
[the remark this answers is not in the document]
kshyju: I removed this and switched to a msbuild prop to handle this situation.

src/DotNetWorker.Grpc/Http/GrpcHttpRequestData.cs
kshyju: PR to fix this in main: https://github.com/Azure/azure-functions-dotnet-worker/pull/1662

PR #1595, opened 2023-06-06
Roranjan/application insights filtering

RohitRanjanMS:
Updated sample to include setting ApplicationInsights default filter.

### DIFF
--- a/samples/FunctionApp/Program.cs
+++ b/samples/FunctionApp/Program.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Azure.Functions.Worker;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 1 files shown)


samples/FunctionApp/Program.cs
```diff
@@ -21,9 +23,16 @@ static async Task Main(string[] args)
...
                     builder
-                        .AddApplicationInsights()
+                        .AddApplicationInsights(options =>
+                        {
+                            // Configure the underlying ApplicationInsightsServiceOptions
+                            options.EnableAdaptiveSampling = false;
+                        })
                         .AddApplicationInsightsLogger();
                 })
+                // Application Insights collects these ILogger logs, with a severity of Warning or above by default, and dependencies.
```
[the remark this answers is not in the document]
RohitRanjanMS: yeah, this is not optimal, but it will make things obvious. I have updated the comment.

samples/FunctionApp/Program.cs
```diff
@@ -21,9 +23,17 @@ static async Task Main(string[] args)
...
+                        .AddApplicationInsights(options =>
+                        {
+                            // Configure the underlying ApplicationInsightsServiceOptions
+                            options.EnableAdaptiveSampling = false;
+                        })
                         .AddApplicationInsightsLogger();
                 })
+                // The Application Insights SDK adds a default logging filter that instructs ILogger to capture only Warning and more severe logs. Application Insights requires an explicit override.
+                // Log levels can also be configured using appsettings.json. For more information, see https://learn.microsoft.com/en-us/azure/azure-monitor/app/worker-service#ilogger-logs
+                .ConfigureLogging(logging => logging
```
jviau, line 35: I prefer removing the app insights filter as opposed to updating it. The difference is this now lets app insights take on default log filters:

``` CSharp
services.Configure<LoggerFilterOptions>(options =>
            {
                LoggerFilterRule toRemove = options.Rules.FirstOrDefault(rule => rule.ProviderName
                    == "Microsoft.Extensions.Logging.ApplicationInsights.ApplicationInsightsLoggerProvider");

                if (toRemove is not null)
                {
                    options.Rules.Remove(toRemove);
                }
            });
```

Removing the filter will allows the general `Logging:LogLevel` filter values to apply to app insights. Without removing it you have to explicitly configure `Logging:ApplicationInsights:LogLevel`.

samples/FunctionApp/Program.cs
```diff
@@ -21,9 +23,17 @@ static async Task Main(string[] args)
                 .ConfigureFunctionsWorkerDefaults(builder =>
                 {
                     builder
-                        .AddApplicationInsights()
+                        .AddApplicationInsights(options =>
+                        {
+                            // Configure the underlying ApplicationInsightsServiceOptions
+                            options.EnableAdaptiveSampling = false;
```
jviau, line 29: Is this something we want to recommend to customers? It can cause *a lot* of telemetry to be sent up. Depending on what their app does, they may quickly hit their daily cap without it.

In my experience when I disabled it for an aspnetcore app, I hit 100GB in only a few hours. We should at least add a comment explaining the considerations and tradeoffs the customer should make before disabling this.

RohitRanjanMS: Changed the property to a less significant one. My intent is just to make it obvious and show how to configure options.

jviau: If that is the case, can you add a comment making that explicit?

``` CSHarp
// The below property configuration is purely for demonstration purposes. It is not required you set the same property to the same value in your application.
```

jviau (approved):

samples/FunctionApp/Program.cs
RohitRanjanMS: Added comment.

RohitRanjanMS:
Updated sample to include setting ApplicationInsights default filter.

PR #1600, opened 2023-06-07
Add DefaultValue attribute to IsBatched

### LINKED ISSUE #1598: ServiceBus trigger does not have DefaultValue attribute for IsBatched prop
On the IsBatched prop on the service bus trigger [here](https://github.com/Azure/azure-functions-dotnet-worker/blob/a625cc1efc49d37611cbc35e88dbd376b073e49a/extensions/Worker.Extensions.ServiceBus/src/ServiceBusTriggerAttribute.cs#L76), it is missing a DefaultValue attribute which is used on all other extensions implementing IsBatched.

This value is used by generators to decide what value to use when a customer does not provide a value for IsBatched. Example from EventHubs [here](https://github.com/Azure/azure-functions-dotnet-worker/blob/a625cc1efc49d37611cbc35e88dbd376b073e49a/extensions/Worker.Extensions.EventHubs/src/EventHubTriggerAttribute.cs#L44), one difference is that EH default val is `true` and SB is `false`. There's no consistent default across extensions.

satvu:

resolves #1598

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/extensions/Worker.Extensions.ServiceBus/src/ServiceBusTriggerAttribute.cs
+++ b/extensions/Worker.Extensions.ServiceBus/src/ServiceBusTriggerAttribute.cs
@@ -73,6 +73,7 @@ namespace Microsoft.Azure.Functions.Worker
         /// <summary>
         /// Gets or sets the configuration to enable batch processing of events. Default value is "false".
         /// </summary>
+        [DefaultValue(false)]
         public bool IsBatched
         {
             get => _isBatched;
--- a/extensions/Worker.Extensions.ServiceBus/release_notes.md
+++ b/extensions/Worker.Extensions.ServiceBus/release_notes.md
@@ -6,4 +6,4 @@
 
 ### Microsoft.Azure.Functions.Worker.Extensions.ServiceBus <version>
 
-- <entry>
+- Add `DefaultValue` attribute to the `IsBatched` prop of the ServiceBus trigger.


extensions/Worker.Extensions.ServiceBus/src/ServiceBusTriggerAttribute.cs
```diff
@@ -73,6 +73,7 @@ public ServiceBusTriggerAttribute(string topicName, string subscriptionName)
         /// <summary>
         /// Gets or sets the configuration to enable batch processing of events. Default value is "false".
         /// </summary>
+        [DefaultValue(false)]
```
satvu, line 76: Since line 10 has `private bool _isBatched = false;` we know to set the DefaultValue to false.

surgupta-msft: Will you be also releasing this package? or are you aware of any plans for that?

surgupta-msft (approved):

extensions/Worker.Extensions.ServiceBus/src/ServiceBusTriggerAttribute.cs
satvu: Unless there is other work we are waiting on for this extension, then yes, I will bump the version in another PR and release it.

PR #1602, opened 2023-06-08
[SourceGen]  Bug fix for when DefaultValue is not present on an IsBatched prop

### LINKED ISSUE #1599: [Source Gen] Handle cases where IsBatched does not have a default value
Related to #1544 and #1598. 

Source Generator does not handle cases where IsBatched is missing a default value - the legacy generator defaults to false.

satvu:

resolves #1599

### Pull request checklist

### Additional information

Additional PR information


sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -594,17 +594,26 @@ private bool IsCardinalityValid(IParameterSymbol parameterSymbol, TypeSyntax? pa
...
+                        var defaultVal = defaultValAttr!.ConstructorArguments.SingleOrDefault().Value!.ToString(); // there is only one constructor arg for the DefaultValue attribute (the default value)
+
+                        if (!bool.TryParse(defaultVal, out bool b) || !b)
+                        {
+                            return true;
+                        }
+                    }
+                    else
+                    {
+                        return false; // If DefaultValue attribute not found, default to false. This behavior is in sync with the legacy generator.
```
satvu, line 614: This matches with legacy generator behavior, but I wonder if we want to throw a diagnostic warning here or log something since all triggers with `IsBatched` should have a DefaultValue attribute set. 

For some triggers, DefaultValue is true, others are set to false. I think it is true that most are set to false, but do we want to automatically fall back to that without a warning? Since it think it's a conceptual bug to not have DefaultValue set on a trigger.

surgupta-msft: If DefaultValue attribute is not present, then should the `false` at the end of this method (line 668) return? or am I missing any scenario for which we need to return `false` here?

I feel having tests for true/false cases will help to make sure we are not changing the behavior.

The main issue I encountered was getting null exception if defaultValueAttribute is not present which will be solved by null check added in line 603. I can also help with testing by taking these changes in my branch.

jviau (approved):

surgupta-msft (approved):

PR #1313, opened 2023-02-06
Add ability to bind to SBReceivedMessage

JoshLove-msft:
### Issue describing the changes in this PR

resolves #1535

### Pull request checklist

* [ ] My changes **do not** require documentation changes
  * [ ] Otherwise: Documentation issue linked to PR
* [ ] My changes **should not** be added to the release notes for the next release
  * [x] Otherwise: I've added my notes to `release_notes.md`
* [x] My changes **do not** need to be backported to a previous version
  * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [x] I have added all required tests (Unit tests, E2E tests)

### Additional information

Additional PR information

PR #1609, opened 2023-06-09
Support binding to EventData

JoshLove-msft:

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

Additional PR information

Also fixed an issue in the SB extension where the Constants type was inadvertently made public in the previous preview release.

SB extension PR for reference - https://github.com/Azure/azure-functions-dotnet-worker/pull/1313


extensions/Worker.Extensions.ServiceBus/src/Constants.cs
```diff
@@ -1,10 +1,9 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
-
 namespace Microsoft.Azure.Functions.Worker.Extensions.ServiceBus
 {
-    public static class Constants
```
JoshLove-msft, line 7: This was inadvertently made public in a preview version, so it should be fine to change it back.

extensions/Worker.Extensions.EventHubs/release_notes.md
```diff
@@ -4,6 +4,6 @@
 - My change description (#PR/#issue)
 -->
 
-### Microsoft.Azure.Functions.Worker.Extensions.EventHubs <version>
+### Microsoft.Azure.Functions.Worker.Extensions.EventHubs 5.4.0-preview1
 
-- Add `BindingCapabilities` attribute to EventHubTrigger to express function-level retry capabilities. (#1457)
\ No newline at end of file
+- Add support for binding to `EventData`(#<PR>)
```
jsquire, line 9: Is that meant to be this PR# or the one where we added the constructor?

extensions/Worker.Extensions.ServiceBus/release_notes.md
```diff
@@ -4,6 +4,6 @@
 - My change description (#PR/#issue)
 -->
 
-### Microsoft.Azure.Functions.Worker.Extensions.ServiceBus <version>
+### Microsoft.Azure.Functions.Worker.Extensions.ServiceBus 5.10.0-preview3
 
-- <entry>
+- Make Constants class internal (#<PR>)
```
jsquire, line 9: Is this a convention that the Functions build system fills the PR slug for?

jsquire (approved):

extensions/Worker.Extensions.EventHubs/release_notes.md
JoshLove-msft: I need to update to include this PR.

liliankasem (approved):

extensions/Worker.Extensions.EventHubs/src/EventDataConverter.cs
```diff
@@ -0,0 +1,45 @@
...
+                _ => ConversionResult.Unhandled()
+            };
+            return new ValueTask<ConversionResult>(result);
+        }
+
+        private EventData ConvertToEventData(ModelBindingData binding)
+        {
+            if (binding.ContentType != Constants.BinaryContentType)
+            {
+                throw new InvalidOperationException(
```
liliankasem, line 38: I think I forgot to mention this in the service bus PR as well, but we would like all the converters that use ParameterBindingData to confirm the source of the data to determine if this is the converter that should be handling the request:

```csharp
// e.g. this but with AzureEventHubsEventData
if (bindingData?.Source is not Constants.CosmosExtensionName)
{
    _logger.LogTrace("Source '{source}' is not supported by {converter}", bindingData?.Source, nameof(CosmosDBConverter));
    return ConversionResult.Unhandled();
}
```

surgupta-msft (approved):

extensions/Worker.Extensions.EventHubs/src/EventHubExtensionStartup.cs
```diff
@@ -0,0 +1,25 @@
...
+    public class EventHubExtensionStartup : WorkerExtensionStartup
+    {
+        public override void Configure(IFunctionsWorkerApplicationBuilder applicationBuilder)
+        {
+            if (applicationBuilder == null)
+            {
+                throw new ArgumentNullException(nameof(applicationBuilder));
+            }
+
+            applicationBuilder.Services.AddAzureClientsCore(); // Adds AzureComponentFactory
```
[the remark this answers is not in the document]
JoshLove-msft: Ok will remove from here and SB.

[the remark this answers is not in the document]
JoshLove-msft: In that case, should the entire type be removed since this was the only configuration happening?

extensions/Worker.Extensions.EventHubs/src/EventDataConverter.cs
JoshLove-msft: Does the conversion result need to be unhandled or can we throw like we do for an invalid content type?

liliankasem: In this case it should be unhandled; we throw with content type because at that point we know we can handle the conversion but were not provided the correct content-type. Now that I'm writing all that out, I can see how content type could also be unhandled ☠️

JoshLove-msft: I changed it to use ConversionResult.Failed for both cases.

extensions/Worker.Extensions.EventHubs/src/EventHubExtensionStartup.cs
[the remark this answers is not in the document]
liliankasem: That's a good question, we originally had the startup because we had to register the converter, then later added AddAzureClientsCore for MI - so without the need for either of those, I think it's safe to remove the type but probably best to test E2E and confirm it still works

extensions/Worker.Extensions.EventHubs/src/EventDataConverter.cs
liliankasem: I'm fine with that for now, I can discuss with the team if we want to change things across all converters for consistency and how want want to deal with all those cases

extensions/Worker.Extensions.EventHubs/src/EventHubExtensionStartup.cs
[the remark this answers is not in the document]
JoshLove-msft: Looks like it is required:

System.TypeLoadException: Could not load type 'Microsoft.Azure.Functions.Worker.EventHubExtensionStartup' from assembly 'Microsoft.Azure.Functions.Worker.Extensions.EventHubs, Version=5.4.0.0, Culture=neutral, PublicKeyToken=551316b6919f366c'.
   at Microsoft.Azure.Functions.Worker.WorkerExtensionStartupCodeExecutor.Configure(IFunctionsWorkerApplicationBuilder applicationBuilder)
   at Microsoft.Extensions.DependencyInjection.ServiceCollectionExtensions.RunExtensionStartupCode(IFunctionsWorkerApplicationBuilder builder) in C:\Users\jolov\src\r\azure-functions-dotnet-worker\src\DotNetWorker.Core\Hosting\ServiceCollectionExtensions.cs:line 144
   at Microsoft.Extensions.DependencyInjection.ServiceCollectionExtensions.AddFunctionsWorkerCore(IServiceCollection services, Action`1 configure) in C:\Users\jolov\src\r\azure-functions-dotnet-worker\src\DotNetWorker.Core\Hosting\ServiceCollectionExtensions.cs:line 98
   at Microsoft.Extensions.DependencyInjection.ServiceCollectionExtensions.AddFunctionsWorkerDefaults(IServiceCollection services, Action`1 configure) in C:\Users\jolov\src\r\azure-functions-dotnet-worker\src\DotNetWorker\Hosting\ServiceCollectionExtensions.cs:line 39
   at Microsoft.Extensions.Hosting.WorkerHostBuilderExtensions.<>c__DisplayClass5_0.<ConfigureFunctionsWorkerDefaults>b__2(HostBuilderContext context, IServiceCollection services) in C:\Users\jolov\src\r\azure-functions-dotnet-worker\src\DotNetWorker\Hosting\WorkerHostBuilderExtensions.cs:line 166
   at Microsoft.Extensions.Hosting.HostBuilder.CreateServiceProvider()
   at Microsoft.Extensions.Hosting.HostBuilder.Build()
   at Program.<Main>$(String[] args) in C:\Users\jolov\RiderProjects\Company.FunctionApp7\Company.FunctionApp7\Program.cs:line 3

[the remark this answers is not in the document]
JoshLove-msft: Actually, nevermind. After I cleaned all the local nuget packages and rebuilt all dependencies it worked. I guess there was something that cached that attribute locally.

extensions/Worker.Extensions.EventHubs/src/EventDataConverter.cs
JoshLove-msft: What is the difference in behavior between Unhandled and Failure? Does Unhandled mean we will fall back to other converter (potentially defined in the host extension)? If so, I don't think we would want to do that. 
Or does it just influence the exception reported back to the user?

liliankasem: I believe that unhandled will lead to falling back to other converters, however we now we have the `[AllowConverterFallback(true)]` attribute which also prompts us to fall back to converters; so I don't actually know what the behaviour looks like if we return unhandled today and AllowConverterFallback = false - will this just lead to a fail?

@surgupta-msft do you have any insights? Wondering if you tested a case like this when building the bypass deferred binding feature?

surgupta-msft: If conversion is unhandled and AllowConverterFallback = false, then the behavior will be exactly same as it was before my changes of bypass deferred binding (if none of the registered converters can handle conversion). So, below line will be returned.

https://github.com/Azure/azure-functions-dotnet-worker/blob/54b185f0c7ed74c0db7b569021defb3dfbf05bab/src/DotNetWorker.Core/Context/Features/DefaultInputConversionFeature.cs#L86


Below is the flow of conversion logic if it helps

```csharp

// Convert using advertised converters. Return if Succeeded/Failed
// if (AllowConverterFallback == true)
       // Convert using registered converters. Return if Succeeded/Failed
// return unhandled
```

Issue #1329, opened 2023-02-09
[SDK Binding - Blob] Improve error logging when deserialization fails for POCO scenarios


liliankasem:
If you bind to a POCO i.e.

`[BlobInput("book-container", IsBatched = true)] Book[] books)`

The format of the blob needs to be json, for example:

book.json -> `{ "id": "1", "name": "mybook" }`

If you don't do that and use a blob with string content, the deserialization will fail with an error that looks like this:

```log
Executed 'Functions.BlobInputBookArrayFunction' (Failed, Id=501b3023-ee56-41f4-a99d-332bba664544, Duration=66ms)

[2023-02-09T21:45:01.210Z] System.Private.CoreLib: Exception while executing function: Functions.BlobInputBookArrayFunction. System.Private.CoreLib: Result: Failure

[2023-02-09T21:45:01.210Z] Exception: Microsoft.Azure.Functions.Worker.FunctionInputConverterException: Error converting 1 input parameters for Function 'BlobInputBookArrayFunction': Cannot convert input parameter 'books' to type 'SampleApp.Book[]' from type 

'Microsoft.Azure.Functions.Worker.Grpc.Messages.GrpcCollectionModelBindingData'. Error:System.Text.Json.JsonException: 'h' is an invalid start of a value. Path: $ | LineNumber: 0 | BytePositionInLine: 0.
```

We should consider improving the logging here. This exception is currently caught in our try-catch so this would be a good place to display a better error message:

https://github.com/Azure/azure-functions-dotnet-worker/blob/f6cf04adfcac6af3c822d1e950cc071594373db5/extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs#L87-L103

fabiocav:
Flagging to investigate what, if any, enhancements we can make to this.

fabiocav:
Assigning to sprint 148 for validation

aishwaryabh:
This error still repros, so I'll create a fix for displaying a better display message.

PR #1639, opened 2023-06-15
[SDK TypeBinding] Improve error logging for failed deserialization for POCO scenarios

aishwaryabh:
This PR adds specific error logging if deserialization fails since the POCO was expecting a json instead of a string.
resolves https://github.com/Azure/azure-functions-dotnet-worker/issues/1329

### Pull request checklist

### Additional information

Additional PR information


jviau (approved):

test/WorkerExtensionTests/Blob/BlobStorageConverterTests.cs
```diff
@@ -771,6 +771,40 @@ public async Task ConvertAsync_POCO_InvalidJson_ReturnsFailed()
             Assert.IsType<JsonException>(conversionResult.Error);
         }
 
+        [Fact]
+        public async Task ConvertAsync_ValidModelBindingData_POCOCollection_Array_ReturnsFailed()
```
kshyju, line 775: nit: The test is confirming incorrect JSON causes conversion failure. correct? The testname has "ValidModelBindingData". Should that be improved?

kshyju (approved):

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -105,6 +107,17 @@ private async ValueTask<ConversionResult> ConvertFromCollectionBindingDataAsync(
 
                 return ConversionResult.Success(result);
             }
+            catch (JsonException ex)
```
kshyju, line 110: Are we only making this change to Blob extension converter? We have a built-in JSonPocoConverter as well. Wonder we should do similar thing there as well.

surgupta-msft (approved):

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
kshyju: That probably is a breaking change. So may be something we can discuss with the team and address later in a separate PR.

aishwaryabh: Yeah this PR specifically focuses on the use case for blob for POCO serialization. We can address this in another PR

test/WorkerExtensionTests/Blob/BlobStorageConverterTests.cs
aishwaryabh: Changed!

PR #1648, opened 2023-06-19
Merging HTTP feature branch

### LINKED ISSUE #1387: HTTP model enhancements
This epic tracks the issues/work items that are part of the out-of-proc (and more specifically, .NET Isolated) HTTP model enhancements that will bring support for more advanced HTTP scenarios, HTTP performance improvements and deeper integration with ASP.NET.

## The current state: what we're solving

The HTTP model exposed by the host to out-of-proc workers relies on a representation of the HTTP request being passed over the worker gRPC channel as part of the function invocation. The HTTP response representation is returned to the host using the same channel in the invocation result.

This is the model all out-of-proc workers (Node, Java, Python, PowerShell and .NET) use, and some of its key limitations are:

- The worker has no access to the request/response streams.
- Integration with HTTP frameworks in out-of-proc scenarios is difficult.
    - Although possible for simple cases, integrations based on the existing model are very fragile and have several limitations, leading to reliability problems, surprising/inconsistent behavior and poor development experiences.
- Significant performance impact
- Higher resource utilization
- Inability to support common use cases supported by the in-proc model (e.g., file/large blob transfers, streaming scenarios, etc.)

The HTTP model supported by the .NET Isolated worker operates within those constraints, utilizing HTTP request and response representation types (`HttpRequestData` and `HttpResposeData`) that expose APIs reflecting what the current HTTP model supports. 

We decided to take this approach, instead of providing implementations of `HttpRequest` and `HttpResponse` on top of the existing ASP.NET model for two main reasons:

1) Because of the limitations described above, several of the commonly used APIs and features supported by the ASP.NET model would either not work or, worse yet, not work as expected.
2) From the beginning, the plan was to proceed with the plans outlined in this epic, which would leave us with two different implementations, with different underlying models, based on the same set of abstractions. Because of previous experiences with in-proc model, we wanted to avoid that as it leads to confusion and difficult documentation, support and development experiences. 

## Where we are headed: The new HTTP model

With the new HTTP model, the host will become an intermediary and and act as a reverse proxy, proxying the actual HTTP request to the worker for handling.

Workers supporting this new model will be enhanced to listen on an HTTP endpoint, in addition to the gRPC channel used for invocations, allowing them to maintain the existing support for input and output bindings, with the existing execution model, while exposing a parallel pipeline used for the HTTP request payload.

The following diagram illustrates, at a high level, the flow proposed with the new HTTP model:
```mermaid
flowchart LR
    A[Host]--> |Request| B[HTTP]
    A[Host]--> |Invocation| C[gRPC]
    subgraph Worker
    B-->D
    C-->D
    D[Dispatcher] --> Function
    end
```

With the flow above, the host will split the HTTP request and function invocation message, creating a correlation between the two. The HTTP request goes through the HTTP pipeline while the function invocation, with all associated input binding data and context, goes through the existing dispatching pipeline. Eventually, the pipelines merge, where the worker augments the `FunctionContext` with the `HttpRequest` information coming from the HTTP pipeline and exposing that to the Function (both, in the context and/or in automatic parameter binding).

With the new model, a simple HTTP triggered function would change from this:

```csharp
[Function("HttpFunction")]
public HttpResponseData Run([HttpTrigger(AuthorizationLevel.Anonymous, "get", "post", Route = null)] HttpRequestData req,
    FunctionContext executionContext)
{
 
    response.WriteString("Welcome to .NET isolated worker !!");

    return response;
}
```

To the following:
```csharp
[Function("HttpFunction")]
public IActionResult Run(
    [HttpTrigger(AuthorizationLevel.Anonymous, "get", "post", Route = null)] HttpRequest req, FunctionContext executionContext)
{
    return new OkObjectResult("Welcome to .NET isolated worker !!");
}
```

This model will also enable deeper integrations with web frameworks (e.g., ASP.NET), providing better control over the HTTP pipeline and enabling scenarios not currently supported in the in-proc experience, like the ability to register and manage ASP.NET middleware and better control over authentication/authorization.

## Work items

This epic will be updated with references to the work items associated with the HTTP model enhancements, as those are defined. 

### Host Items

Host items can be found under [this label](https://github.com/Azure/azure-functions-host/labels/feature%3A%20HTTP%20Types).

### Worker Items

Worker items will use the same label as this issue ([feature:HTTP types](https://github.com/Azure/azure-functions-dotnet-worker/labels/feature%3A%20HTTP%20types)).

fabiocav:

resolves #1387 

### Pull request checklist


samples/FunctionApp/HttpProxyTrigger/HttpProxyTrigger.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System.Threading;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Azure.Functions.Worker;
+
+namespace FunctionApp
+{
+    public class HttpProxyTrigger
+    {
+        [Function(nameof(HttpProxyTrigger))]
```
satvu, line 10: Outside of scope for this PR, but do we want to make a different sample app for the aspnet integration examples? Would make sense since the wire up is different + we don't want to interfere with the current http model being used in this app.

satvu (approved):

samples/FunctionApp/HttpProxyTrigger/HttpProxyTrigger.cs
fabiocav: We should

Issue #1337, opened 2023-02-13
[SDK Binding - Blob] Binding to BlobContainerClient should not require you to provide a blob path


liliankasem:
When binding to `BlobContainerClient` as an input binding, you still need to provide a blob name in the input binding i.e.  `BlobInput("input-container/sample1.txt"`

```csharp
[Function(nameof(BlobContainerClientInputFunction))]
public async Task<HttpResponseData> BlobContainerClientInputFunction(
    [HttpTrigger(AuthorizationLevel.Anonymous, "get", "post")] HttpRequestData req,
    [BlobInput("input-container/sample1.txt")] BlobContainerClient client)
```
	
But since we are binding to a container, it doesn't make sense to have to provide a blob name in the binding; instead I would expect that you could bind to just the container like this:

```csharp
[Function(nameof(BlobContainerClientInputFunction))]
public async Task<HttpResponseData> BlobContainerClientInputFunction(
    [HttpTrigger(AuthorizationLevel.Anonymous, "get", "post")] HttpRequestData req,
    [BlobInput("input-container")] BlobContainerClient client)
```

This does not work today and you get the following error message:

`Invalid blob path specified : 'input-container'. Blob identifiers must be in the format 'container/blob'`

> Note: this is different to a blob trigger which lets you bind to just the container, and it is not the same as IsBatched which also lets you bind to only a container but expects a collection binding i.e. BlobContainerClient[] which does not make sense for this scenario

### Proposed Change

1. On the WebJobs extension side we relax the rules to allow input binding with just the container name
2. On the worker side we will have more checks in place to make sure customers do not bind a string/byte[] or anything other than a BlobContainerClient to a container path

fabiocav:
Assigning to @liliankasem for additional context on the change we'll be proposing to the blob extension.

aishwaryabh:
The Blob WebJobs extensions has a property called `isContainerBinding` for determining if the passed in input can be parsed correctly. We basically need to pass in `isContainerBinding = true` [here](https://github.com/Azure/azure-sdk-for-net/blob/f9bb57eec632f15fd3e93ce26c007794214d8ea4/sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Blobs/src/Config/BlobsExtensionConfigProvider.cs#L221-L225), and as of today there is no way to do this other than adding a new attribute.

PR #1650, opened 2023-06-20
Dotnet worker: Allow binding to BlobContainerClient without blob

aishwaryabh:

resolves https://github.com/Azure/azure-functions-dotnet-worker/issues/1337

For binding `BlobContainerClient` as an input binding, we are required to provide blob name in the result such as `BlobInput("input-container/blob-name")`. We should be allowed to do this without providing the blob name, and currently we run into an error if you try to just put down the container name:
`Invalid blob path specified : 'input-container'. Blob identifiers must be in the format 'container/blob'`

I added a change to the webjobs extension which loosens the parsing and validating checks by adding a new parameter to `BlobPath.ParseAndValidate` called `isParameterBindingData`. This value is only set to true if we are binding to a parameter binding data type and then skips some of the checks which caused this to fail initially.

This PR adds a check in the dotnet worker to ensure that blob name cannot be null if we are not binding to a `BlobContainerClient`, which is the check we are skipping on the webjobs extension.

### Pull request checklist

### Additional information

Additional PR information


samples/WorkerBindingSamples/Blob/BlobInputBindingSamples.cs
```diff
@@ -114,5 +114,20 @@ public BlobInputBindingSamples(ILogger<BlobInputBindingSamples> logger)
 
             return req.CreateResponse(HttpStatusCode.OK);
         }
+
+        [Function(nameof(BlobContainerClientInputFunction))]
+        public HttpResponseData BlobContainerClientInputFunction(
```
liliankasem, line 119: nit: move to be above the BlobClient sample
nit2: rename to `BlobInputContainerClientFunction` to match the naming pattern of the rest of the functions

aishwaryabh: Changed!

.gitignore
```diff
@@ -4,6 +4,7 @@
 ## Get latest from https://github.com/github/gitignore/blob/master/VisualStudio.gitignore
 
 Azure.Functions.Cli
+azurite
```
liliankasem, line 7: This is odd, it already exists in the feature branch

https://github.com/Azure/azure-functions-dotnet-worker/blob/80a2659320b7b20cf9697a2b6a9cda75e780185c/.gitignore#L7

liliankasem: AH, you need to rebase. I updated feature/sdk-binding to be caught up with main

liliankasem (approved):

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -160,6 +160,11 @@ private bool IsBlobExtension(ModelBindingData bindingData)
                 throw new ArgumentNullException(nameof(containerName));
             }
 
+            if (targetType != typeof(BlobContainerClient) && string.IsNullOrEmpty(blobName))
```
surgupta-msft, line 163: @liliankasem Highlighting that this check may change when we remove `CollectionModelBindingData`. Below code is valid and it does not have target type as `BlobContainerClient`

`BlobInput("container-path") IEnumerable<string> blobs`

surgupta-msft (approved):

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
liliankasem: Ah yup I have that covered in my [PR](https://github.com/Azure/azure-functions-dotnet-worker/pull/1690)

liliankasem:
Going to abandon this as it will be covered in another PR: https://github.com/Azure/azure-functions-dotnet-worker/pull/1690

PR #1660, opened 2023-06-21
Add binding attribute to converter context

liliankasem:
Adding the binding attribute for a given parameter to the converter context, this enables converters to have all the information they need about the attribute a parameter is using without having to make a roundtrip to WebJobs.

### DIFF
--- a/src/DotNetWorker.Core/Context/Features/DefaultFunctionInputBindingFeature.cs
+++ b/src/DotNetWorker.Core/Context/Features/DefaultFunctionInputBindingFeature.cs
@@ -39,7 +39,7 @@ namespace Microsoft.Azure.Functions.Worker.Context.Features
                     // Return the cached value if BindFunctionInputAsync is called a second time during invocation.
                     return _inputBindingResult!;
                 }
-                
+
                 IFunctionBindingsFeature functionBindings = context.GetBindings();
                 var inputBindingCache = context.InstanceServices.GetService<IBindingCache<ConversionResult>>();
                 var inputConversionFeature = context.Features.Get<IInputConversionFeature>();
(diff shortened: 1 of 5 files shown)


src/DotNetWorker.Core/Context/Features/DefaultFunctionInputBindingFeature.cs
```diff
@@ -129,6 +116,14 @@ public async ValueTask<FunctionInputBindingResult> BindFunctionInputAsync(Functi
             }
         }
 
+        private void AddFunctionParameterPropertyIfPresent(IDictionary<string, object> properties, FunctionParameter param, string key)
+        {
+            if (param.Properties.TryGetValue(key, out var val))
```
jviau, line 121: small nit:
```suggestion
            if (param.Properties.TryGetValue(key, out object val))
```

jviau (approved):

fabiocav:
Flagging for review

fabiocav (commented):
Flagging for review

src/DotNetWorker.Core/Context/Features/DefaultFunctionInputBindingFeature.cs
fabiocav, line 1: Flagging for review

src/DotNetWorker.Core/Context/Features/DefaultFunctionInputBindingFeature.cs
```diff
@@ -72,25 +72,12 @@ public async ValueTask<FunctionInputBindingResult> BindFunctionInputAsync(Functi
...
-                        if (param.Properties.TryGetValue(PropertyBagKeys.BindingAttributeSupportedConverters, out var converters))
-                        {
-                            properties.Add(PropertyBagKeys.BindingAttributeSupportedConverters, converters);
-                        }
-
-                        var converterContext = _converterContextFactory.Create(param.Type, source, context, properties.Count() != 0 
+                        AddFunctionParameterPropertyIfPresent(properties, param, PropertyBagKeys.ConverterType);
+                        AddFunctionParameterPropertyIfPresent(properties, param, PropertyBagKeys.AllowConverterFallback);
+                        AddFunctionParameterPropertyIfPresent(properties, param, PropertyBagKeys.BindingAttributeSupportedConverters);
+                        AddFunctionParameterPropertyIfPresent(properties, param, PropertyBagKeys.BindingAttribute);
```
fabiocav, line 78: The existing properties are primarily there to support infrastructure, but since this one will be exposed to converters, we should create an extension method for `ConverterContext` so that is what is ultimately used by the converters, abstracting away the details of how this is added to the context and improving discoverability.

liliankasem: Just to clarify, are you suggesting a ConverterContext extension method that is used directly by the converter to grab the BindingAttribute i.e. `context.GetBindingAttribute()`. But no change to this file as we still need to pass that information down to the ConverterContext in the first place, or are you also suggesting that this extension method does what AddFunctionParameterPropertyIfPresent is doing here too?

fabiocav: No change to this file. Just the public API that would be consumed by the converters.

liliankasem: @fabiocav I think this is addressed, mind reviewing again? Thanks!

src/DotNetWorker.Core/Converters/Converter/ConverterContextExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// </summary>
+    public static class ConverterContextExtensions
+    {
+        /// <summary>
+        /// Tries to retrieve the binding attribute from the <see cref="ConverterContext"/>.
+        /// </summary>
+        /// <param name="context">The converter context.</param>
+        /// <param name="bindingAttribute">When this method returns, contains the binding attribute if found; otherwise, <c>null</c>.</param>
+        /// <returns><c>true</c> if the binding attribute is found in the converter context; otherwise, <c>false</c>.</returns>
+        public static bool TryGetBindingAttribute(this ConverterContext context, out object? bindingAttribute)
```
fabiocav, line 17: As defined, we should be able to simplify this to:

```csharp
        public static bool TryGetBindingAttribute(this ConverterContext context, out object? bindingAttribute)
            => context.Properties.TryGetValue(PropertyBagKeys.BindingAttribute, out bindingAttribute);
```

But what is the expected usage here? Would it make more sense to expose a generic public API that enables the consumer to get a typed attribute instance? (e.g. `TryGetBindingAttribute<T>`, potentially with a constraint)

liliankasem: Done!

PR #1676, opened 2023-06-23
Core deferred binding feature

### LINKED ISSUE #1702: Release deferred binding feature
- [ ] Merge core deferred binding feature in to main
 - [ ] Release stable worker and SDK packages
 - [ ] Release stable abstractions package

liliankasem:
Cherry picking the core pieces of the deferred binding feature, leaving out all extension changes.

- Introduce ModelBindingData models
- Introduce the deferred binding attributes
- Sdk generator changes to implement `supportsDeferredBinding` metadata property
- Creation of new unit test project for worker extensions

resolves #1702


surgupta-msft:
Did first pass on the PR to check if all the needed files are included here -- looks good. Will do another pass for reviewing the PR.

Question - what is the plan for bringing in sdk.analyzers and other tests - specially GrpcFunctionDefinitionTests and Metadata generator tests as all the functionality is here but we can't add their tests at this point?

liliankasem:
> Did first pass on the PR to check if all the needed files are included here -- looks good. Will do another pass for reviewing the PR.
> 
> Question - what is the plan for bringing in sdk.analyzers and other tests - specially GrpcFunctionDefinitionTests and Metadata generator tests as all the functionality is here but we can't add their tests at this point?

Thanks for confirming Surbhi, I mostly want to make sure everything we need is coming through, but doesn't hurt to do another PR pass just in case and we can address any additional feedback.

For analyzers, I'd like to bring those in with their own PR.
For those tests, a lot of them are tied to an extension so I would like to bring those in as we bring in each extension

jviau (approved):

src/DotNetWorker.Core/CollectionModelBindingData.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    /// <summary>
+    /// A representation of Collection of Microsoft.Azure.WebJobs.ParameterBindingData
+    /// </summary>
+    public abstract class CollectionModelBindingData
+    {
+        /// <summary>
+        /// Gets the array of ModelBindingData
+        /// </summary>
+        public abstract ModelBindingData[] ModelBindingDataArray { get; }
```
brettsam, line 14: Have we done an API review of these names? Having `Array` at the end of a property name jumped out at me...

brettsam (commented):
Mostly just a quick question about whether we need a public surface API review on this...

src/DotNetWorker.Core/CollectionModelBindingData.cs
liliankasem: We had some reviews but might have missed this; I also don't like the look/sound of array here. I also wonder if we need this type at all? Could we not just reuse `GrpcModelBindingData` and have a different constructor for collections? Or something similar?

I.e. instead of `TypedData.DataOneofCase.CollectionModelBindingData => new GrpcCollectionModelBindingData(typedData.CollectionModelBindingData),`

we could do:

`TypedData.DataOneofCase.CollectionModelBindingData => new GrpcModelBindingData(typedData.CollectionModelBindingData),`

cc: @fabiocav @surgupta-msft

Issue #1656, opened 2023-06-21
Create samples capturing commons scenarios


fabiocav:
Create issues for a set of samples capturing common scenarios using HTTP triggers.

danielniccoli:
Noteworthy? https://github.com/Azure/Azure-Functions/issues/2384

fabiocav:
@danielniccoli thank you for pointing that out! This issue is tracking HTTP samples (updated title for clarity), but we've transferred your issue to the worker repo for triage.

Thanks again

PR #1682, opened 2023-06-23
Create ASP.NET integration sample app

satvu:

resolves #issue_for_this_pr

### Pull request checklist


liliankasem (approved):

samples/AspNetIntegration/SimpleHttpTrigger/SimpleHttpTrigger.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace AspNetIntegration
+{
+    public class SimpleHttpTrigger
+    {
+        //<docsnippet_aspnet_http_trigger>
+        [Function("SimpleHttpTrigger")]
+        public IActionResult Run([HttpTrigger(AuthorizationLevel.Anonymous, "get")] HttpRequest req)
+        {
+            return new OkObjectResult("Welcome to Azure Functions!");
```
kshyju, line 16: Should we also have an example which show how to get HttpContext and use that? Ex: Read a header from httpcontext?

satvu: This is a smaller PR to create the app and add bits needed for initial docsnippets - more scenarios can be tracked through issues and added to this kind-of epic: https://github.com/Azure/azure-functions-dotnet-worker/issues/1656

kshyju: My bad. Did not realize we were using `HttpRequest` type here. I thought it was `HttpRequestData`

samples/AspNetIntegration/FooMiddleware.cs
```diff
@@ -0,0 +1,36 @@
...
+namespace AspNetIntegration
+{
+    internal class FooMiddleware : IFunctionsWorkerMiddleware
+    {
+        //<docsnippet_aspnet_middleware>
+        public Task Invoke(FunctionContext context, FunctionExecutionDelegate next)
+        {
+            // retrieve http context from function context
+            HttpContext httpContext = context.GetHttpContext() 
+                ?? throw new InvalidOperationException($"{nameof(context)} has no http context associated with it.");
```
kshyju, line 18: Question: Should we throw here or gracefully exit? Because there is nothing stopping from customers to add a non http trigger today this function app and it will throw here. 

May be leave a comment here about that or register the middleware only for http requests like [here](https://github.com/Azure/azure-functions-dotnet-worker/blob/a6322a3eb70bb6df93a9b22950c7bd50f8157e97/samples/CustomMiddleware/Program.cs#L23)

satvu: good callout - will register this only for http requests.

samples/AspNetIntegration/Program.cs
```diff
@@ -0,0 +1,23 @@
...
+using AspNetIntegration;
+using Microsoft.Extensions.Hosting;
+
+//<docsnippet_aspnet_registration>
+var host = new HostBuilder()
+    .ConfigureFunctionsWebApplication(builder =>
+    {
+        // can still register middleware and use this extension method the same way
+        // .ConfigureFunctionsWorkerDefaults() is used
+        builder.UseWhen<FooMiddleware>((context)=>
```
fabiocav, line 13: This code can't be used in a doc snippet as it isn't self-contained (i.e., it contains a reference to a class outside of the snippet, `FooMiddleware`), so when shown in isolation, customers can't copy/paste and compile.

samples/AspNetIntegration/FooMiddleware.cs
```diff
@@ -0,0 +1,36 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Azure.Functions.Worker.Middleware;
+
+namespace AspNetIntegration
+{
+    internal class FooMiddleware : IFunctionsWorkerMiddleware
```
fabiocav, line 11: Can we use a better name here?

fabiocav: Actually, for a simple example, is a middleware required? If this sample is primarily to show the ASP.NET Core integration, we should keep it simpler and move this to a different sample.

fabiocav (changes requested):

samples/AspNetIntegration/FooMiddleware.cs
satvu: I was thinking we would have multiple examples in this app (including the simple trigger). Should we have separate apps? At least for the middleware? Or do we want to leave that out of these samples in the repo and leave it for the docs?

satvu:
@fabiocav @mattchenderson The only docsnippets are currently the simple http trigger function and the asp net extension registration - let me know if there's any more you want to include from here.

fabiocav (approved):

samples/AspNetIntegration/local.settings.json
```diff
@@ -0,0 +1,8 @@
+{
+    "IsEncrypted": false,
+  "Values": {
+    "AzureWebJobsStorage": "UseDevelopmentStorage=true",
+    "FUNCTIONS_WORKER_RUNTIME": "dotnet-isolated",
+    "AzureWebJobsFeatureFlags": "EnableHttpProxying"
+  }
```
RussKie, line 7: This looks like a bad merge... This JSON is invalid.

PR #1690, opened 2023-06-27
Blob extension: Infer collection bindings using the blobPath

### LINKED ISSUE #1675: Update worker blob extension to infer IsBatched
- Update blob input attribute to remove IsBatched property
    - Rev major version of the blob extension as this is breaking change
- Update the blob converter to infer IsBatched based on the blob path

liliankasem:

resolves #1675

### Pull request checklist

### Additional information

Additional PR information


extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -160,47 +118,120 @@ private bool IsBlobExtension(ModelBindingData bindingData)
...
+            if (IsCollectionBinding(blobName, targetType))
+            {
+                Type elementType = targetType.IsArray ? targetType.GetElementType() : targetType.GenericTypeArguments[0];
+
+                if (elementType == typeof(BlobContainerClient))
+                {
+                    throw new InvalidOperationException("Collections of BlobContainerClient are not supported.");
+                }
+
+                var resultList = new List<object>();
```
liliankasem, line 132: @fabiocav so this used to be `IList result = Array.CreateInstance(elementType, collectionModelBindingData.ModelBindingDataArray.Length);`

But we don't know the length here and I would like to avoid iterating over `resultSegment` just to get a count. Do you have any suggestions for a clean way to do this?

Everything is working at the moment up till `result = (IList)Activator.CreateInstance(resultType, resultList);`

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -135,72 +93,143 @@ private bool IsBlobExtension(ModelBindingData bindingData)
...
+            Type elementType = targetType.IsArray ? targetType.GetElementType() : targetType.GenericTypeArguments[0];
+
+            if (elementType == typeof(BlobContainerClient))
+            {
+                throw new InvalidOperationException("Collections of BlobContainerClient are not supported.");
+            }
+
+            var blobCount = 0;
+            var resultType = typeof(List<>).MakeGenericType(elementType);
+            var result = (IList)Activator.CreateInstance(resultType);
```
liliankasem, line 170: @fabiocav so this used to be `IList result = Array.CreateInstance(elementType, collectionModelBindingData.ModelBindingDataArray.Length);`

But we don't know the length here and so I switched this around to creating a list and later converting to an array - is there a better way to do this?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -135,72 +93,143 @@ private bool IsBlobExtension(ModelBindingData bindingData)
...
-            Type _ when targetType == typeof(BlobBaseClient) => CreateBlobClient<BlobBaseClient>(connectionName, containerName, blobName),
-            Type _ when targetType == typeof(BlobClient) => CreateBlobClient<BlobClient>(connectionName, containerName, blobName),
-            Type _ when targetType == typeof(BlockBlobClient) => CreateBlobClient<BlockBlobClient>(connectionName, containerName, blobName),
-            Type _ when targetType == typeof(PageBlobClient) => CreateBlobClient<PageBlobClient>(connectionName, containerName, blobName),
-            Type _ when targetType == typeof(AppendBlobClient) => CreateBlobClient<AppendBlobClient>(connectionName, containerName, blobName),
-            Type _ when targetType == typeof(BlobContainerClient) => CreateBlobContainerClient(connectionName, containerName),
-            _ => await DeserializeToTargetObjectAsync(targetType, connectionName, containerName, blobName)
+        private bool IsCollectionBinding(Type type, string blobName)
+        {
+            bool isContainer = string.IsNullOrEmpty(blobName);
```
liliankasem, line 137: When the binding is `container/subdir`, blobName will be `subdir`.

We can either use the blobPath instead (will need this code change in first #1660). Or we can have our own rule where there has to be a trailing `/` i.e. `container/subdir/`

liliankasem: Second thoughts on the training `/` we would have to make another WebJobs change as this will fail the blob path validation

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
jviau: I can't think of anything better given the constraints.

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -135,72 +95,148 @@ private bool IsBlobExtension(ModelBindingData bindingData)
...
+            var result = (IList)Activator.CreateInstance(resultType);
+
+            await foreach (BlobItem blobItem in container.GetBlobsAsync(prefix: blobPath).ConfigureAwait(false))
+            {
+                var element = await ToTargetTypeAsync(elementType, container, blobItem.Name);
+
+                if (element is not null)
+                {
+                    result.Add(element);
+                    blobCount++;
```
jviau, line 188: Can you use `result.Length;`?

jviau (approved):

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
liliankasem: good call!

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -57,136 +59,166 @@ private async ValueTask<ConversionResult> ConvertFromBindingDataAsync(ConverterC
...
+
+        /// <summary>
+        /// Determines if the binding is a collection binding.
+        /// A collection binding is when the target type is an array or IEnumerable and
+        /// the blob name is either null or empty (meaning a container path is provided).
+        /// If a blob name is provided, it must be a directory path (no file extension provided).
+        /// </summary>
+        private bool IsCollectionBinding(Type type, string blobName)
+        {
+            if (type == typeof(string) || type == typeof(byte[]))
```
fabiocav, line 129: Trying to remember, didn't we support binding a blob to `string[]` to support the deserialization case? Would be good to have a test validating that, as I don't think this will currently support that flow.

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
fabiocav: Alternatives I'm thinking of would unnecessarily increase complexity here for very little benefit. Let's keep this.

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -57,136 +59,166 @@ private async ValueTask<ConversionResult> ConvertFromBindingDataAsync(ConverterC
...
+                    result.Add(element);
+                }
             }
 
-            if (string.IsNullOrEmpty(containerName))
+            if (targetType.IsArray)
             {
-                throw new ArgumentNullException(nameof(containerName));
+                var arrayResult = Array.CreateInstance(elementType, result.Count);
+                ((IList)result).CopyTo(arrayResult, 0);
```
fabiocav, line 177: nit: cast to `IList` here shouldn't be required.

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -57,136 +59,166 @@ private async ValueTask<ConversionResult> ConvertFromBindingDataAsync(ConverterC
...
-            if (string.IsNullOrEmpty(connectionName))
+            var resultType = typeof(List<>).MakeGenericType(elementType);
+            var result = (IList)Activator.CreateInstance(resultType);
+
+            await foreach (BlobItem blobItem in container.GetBlobsAsync(prefix: blobPath).ConfigureAwait(false))
             {
-                throw new ArgumentNullException(nameof(connectionName));
+                var element = await ToTargetTypeAsync(elementType, container, blobItem.Name);
+
+                if (element is not null)
```
fabiocav, line 168: Should we be ignoring `null` values here? Are you expecting the `null` cases here to be a result of deserialization? If so, shouldn't those results still be provided/exposed?

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
fabiocav: Using `string[]` as the example, but the concern here applies to any JSON array deserialization target.

surgupta-msft (approved):

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
```diff
@@ -57,136 +59,182 @@ private async ValueTask<ConversionResult> ConvertFromBindingDataAsync(ConverterC
...
+            return true;
         }
 
-        private async Task<object?> ConvertModelBindingDataAsync(IDictionary<string, string> content, Type targetType, ModelBindingData bindingData)
+        private async Task<object> BindToCollectionAsync(Type targetType, Type elementType, BlobContainerClient container, string blobPath)
         {
-            content.TryGetValue(Constants.Connection, out var connectionName);
-            content.TryGetValue(Constants.ContainerName, out var containerName);
-            content.TryGetValue(Constants.BlobName, out var blobName);
+            if (BlobIsFileRegex.IsMatch(blobPath))
```
fabiocav, line 173: Would it make sense to move this check to the `IsCollectionBinding` method, returning false if it's a file and causing the logic to fallback to single blob binding, which would remove the need to call `DeserializeToTargetObjectAsync` from multiple locations, avoiding duplication and the special casing in this path.

fabiocav (approved):
Leaving one comment, but not blocking.

extensions/Worker.Extensions.Storage.Blobs/src/BlobStorageConverter.cs
liliankasem: Ooo good call out!! I didn't go for that initially because I thought it would end up hitting the (for example) string single case, but we don't have a switch case for `string[]` 🤦🏻‍♀️ so this would hit DeserializeToTargetObjectAsync and work as expected - will fix :)

PR #1699, opened 2023-06-28
Remove explicit AppInsights.WorkerService reference and APIs

### LINKED ISSUE #1640: Update Application Insights service registration methods

jviau:

resolves #1640

### Pull request checklist

### DIFF
--- a/src/DotNetWorker.ApplicationInsights/FunctionsApplicationInsightsExtensions.cs
+++ b/src/DotNetWorker.ApplicationInsights/FunctionsApplicationInsightsExtensions.cs
@@ -2,10 +2,15 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.ApplicationInsights;
 using Microsoft.ApplicationInsights.Extensibility;
 using Microsoft.Azure.Functions.Worker.ApplicationInsights;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.Azure.Functions.Worker
 {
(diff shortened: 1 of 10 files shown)


src/DotNetWorker.ApplicationInsights/FunctionsApplicationInsightsExtensions.cs
```diff
@@ -3,65 +3,27 @@
...
     public static class FunctionsApplicationInsightsExtensions
     {
-        /// <summary>
-        /// Adds Application Insights support by internally calling <see cref="ApplicationInsightsExtensions.AddApplicationInsightsTelemetryWorkerService(IServiceCollection)"/>.
-        /// </summary>
-        /// <param name="builder">The <see cref="IFunctionsWorkerApplicationBuilder"/></param>
-        /// <param name="configureOptions">Action to configure ApplicationInsights services.</param>
-        /// <returns>The <see cref="IFunctionsWorkerApplicationBuilder"/></returns>
-        public static IFunctionsWorkerApplicationBuilder AddApplicationInsights(this IFunctionsWorkerApplicationBuilder builder, Action<ApplicationInsightsServiceOptions>? configureOptions = null)
+        public static IServiceCollection ConfigureFunctionsApplicationInsights(this IServiceCollection services)
```
brettsam, line 14: Didn't we also talk about some PostConfigureOptions call that would throw if an ITelemetryConfig wasn't registered with DI?

I don't recall the exact mechanism you'd mentioned... but just some way to make sure that the customer hasn't *only* called this API and expects it to work.

jviau: Ah right, will need to think about that a bit more. The issue with the `PostConfigure` is that something needs to resolve the imports for that validation to happen. So need a reliable startup hook that _isn't_ dependent on AppInsights being present.

There is a host startup, which may work.

jviau: Added!

brettsam (approved):

PR #1712, opened 2023-07-03
Adding support for HTTP body binding 

### LINKED ISSUE #164: HttpTrigger with complex type does not work

fabiocav:
### Changes in this PR

This PR introduces the ability to map custom attributes to converters, and leverages that capability to introduce a new attribute `FromBodyAttribute`) in the HTTP extension, adding binding support for function parameters using the request body as the source (e.g., ability to bind, `string`, `byte[]`, POCOs).

resolves #164

### Pull request checklist


fabiocav:
@mattchenderson this will require doc updates. It would be great to avoid blocking on that, so happy to create a separate item if can track this separately.

extensions/Worker.Extensions.Http/src/FromBodyConverter.cs
```diff
@@ -0,0 +1,34 @@
...
+using Microsoft.Azure.Functions.Worker.Converters;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Http.Converters
+{
+    internal class FromBodyConverter : IInputConverter
+    {
+        public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
+        {
+            var bodyConversionFeature = context.FunctionContext.Features.Get<IFromBodyConversionFeature>()
+                 ?? DefaultFromBodyConversionFeature.Instance;
```
fabiocav, line 15: Not 100% happy with the default not being present in the context and used as a fallback. It enables the behavior, but does mean the feature is only present when dealing with the non-default behavior.

I have some changes planned to address this (same pattern of already in place in different flows), but not tracking that for this PR

extensions/Worker.Extensions.Http/src/DefaultFromBodyConversionFeature.cs
```diff
@@ -0,0 +1,110 @@
...
+            {
+                Memory<byte> bytes = ReadBytes(requestData, context.CancellationToken);
+                result = bytes;
+            }
+            else if (HasJsonContentType(requestData))
+            {
+                ObjectSerializer serializer = requestData.FunctionContext.InstanceServices.GetService<IOptions<WorkerOptions>>()?.Value?.Serializer
+                 ?? throw new InvalidOperationException("A serializer is not configured for the worker.");
+
+                result = serializer.Deserialize(requestData.Body, targetType, context.CancellationToken);
```
kshyju, line 71: There is also a `DeserializeAsync` [method](https://learn.microsoft.com/en-us/dotnet/api/azure.core.serialization.objectserializer.deserializeasync?view=azure-dotnet) which returns a `ValueTask`.

extensions/Worker.Extensions.Http/src/DefaultFromBodyConversionFeature.cs
```diff
@@ -0,0 +1,110 @@
...
+using Microsoft.Azure.Functions.Worker.Extensions.Http.Converters;
+using Microsoft.Azure.Functions.Worker.Http;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Http
+{
+    internal class DefaultFromBodyConversionFeature : IFromBodyConversionFeature
+    {
+        public static IFromBodyConversionFeature Instance { get; } = new DefaultFromBodyConversionFeature();
```
kshyju, line 19: Can mark as `internal` ?

extensions/Worker.Extensions.Http/src/FromBodyConverter.cs
```diff
@@ -0,0 +1,34 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.Azure.Functions.Worker.Converters;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Http.Converters
+{
+    internal class FromBodyConverter : IInputConverter
```
kshyju, line 10: Are we missing the piece of code to register this new input converter?

kshyju (commented):
Looks like tests are failing.

extensions/Worker.Extensions.Http/src/DefaultFromBodyConversionFeature.cs
```diff
@@ -0,0 +1,110 @@
...
+using System.Threading.Tasks;
+using Azure.Core.Serialization;
+using Microsoft.Azure.Functions.Worker.Extensions.Http.Converters;
+using Microsoft.Azure.Functions.Worker.Http;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Http
+{
+    internal class DefaultFromBodyConversionFeature : IFromBodyConversionFeature
```
brettsam, line 17: Are we expecting a matching implementation for the AspNetCore integration?

extensions/Worker.Extensions.Http/src/FromBodyConverter.cs
brettsam: I think that because `[FromBody]` is an `InputConverterAttribute`, it gets auto-used.

fabiocav: Correct. This is automatically used as it is an `InputConverterAttribute`

extensions/Worker.Extensions.Http/src/DefaultFromBodyConversionFeature.cs
fabiocav: Yes. I'll follow up with that in a separate PR soon.

test/E2ETests/E2ETests/Fixtures/FunctionAppFixture.cs
```diff
@@ -55,7 +55,12 @@ public async Task InitializeAsync()
                     _funcProcess.StartInfo.ArgumentList.Add("HelloFromQuery");
                     _funcProcess.StartInfo.ArgumentList.Add("HelloFromJsonBody");
                     _funcProcess.StartInfo.ArgumentList.Add("HelloUsingPoco");
+                    _funcProcess.StartInfo.ArgumentList.Add("HelloWithNoResponse");
+                    _funcProcess.StartInfo.ArgumentList.Add("PocoFromBody");
+                    _funcProcess.StartInfo.ArgumentList.Add("PocoBeforeRouteParameters");
+                    _funcProcess.StartInfo.ArgumentList.Add("PocoAfterRouteParameters");
                     _funcProcess.StartInfo.ArgumentList.Add("ExceptionFunction");
+                    _funcProcess.StartInfo.ArgumentList.Add("PocoWithoutBindingSource");
```
satvu, line 63: More of a forward-looking question, but how are we going to test both non-ASP.NET functions _and_ ASP.NET functions with this e2e app as only one mode is really supported at a time?

extensions/Worker.Extensions.Http/src/FromBodyConverter.cs
kshyju: Oh yes, I forgot that.

extensions/Worker.Extensions.Http/src/DefaultFromBodyConversionFeature.cs
fabiocav: This is intentional. All code paths here are synchronous as  they're optimized for the mainline scenario using the default `HttpRequestData` implementation, where the body is loaded `MemoryStream`. 

Benchmarked this and the non-async method performed better for core scenarios.

test/E2ETests/E2ETests/Fixtures/FunctionAppFixture.cs
fabiocav: Will need to work through that. While we can make the logic conditional on what we pass into the process, we might want to consider a separate app.

brettsam (approved):

satvu (approved):

extensions/Worker.Extensions.Http/src/FromBodyConverter.cs
```diff
@@ -0,0 +1,34 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
```
kshyju, line 4: nit: removed unused using stmnt.

test/E2ETests/E2ETests/HttpEndToEndTests.cs
```diff
@@ -55,6 +56,30 @@ public async Task HttpTriggerTestsMediaTypeDoNotMatter(string functionName, stri
...
+        public async Task HttpTriggerTests_PocoFromBody(string functionName, string route, string body, string mediaType, HttpStatusCode expectedStatusCode, string expectedBody)
+        {
+            HttpResponseMessage response = await HttpHelpers.InvokeHttpTriggerWithBody($"{route}{functionName}", body, expectedStatusCode, mediaType);
+            string responseBody = await response.Content.ReadAsStringAsync();
+
+            Assert.Equal(expectedStatusCode, response.StatusCode);
+            Assert.Equal(expectedBody, responseBody);
+        }
+
+
```
kshyju, line 72: nit: extra line break

kshyju (approved):

extensions/Worker.Extensions.Http/src/FromBodyAttribute.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.Functions.Worker.Converters;
+using Microsoft.Azure.Functions.Worker.Extensions.Http.Converters;
+
+namespace Microsoft.Azure.Functions.Worker.Http
+{
+    /// <summary>
+    /// Specifies that a parameter should be bound using the request body.
```
kshyju, line 10: Should we also specify this is valid to use for HttpTrigger , in the method summary ?

PR #1727, opened 2023-07-10
Update and fix unit tests

liliankasem:
- Update a couple tests that were missed when converter behaviour was changed (in how we handle invalid content source and content type)
- Refactored Table unit tests
- Refactored EventGrid unit tests

### DIFF
--- a/extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
+++ b/extensions/Worker.Extensions.EventGrid/src/TypeConverters/EventGridBinaryDataConverter.cs
@@ -18,18 +18,20 @@ namespace Microsoft.Azure.Functions.Worker.Extensions.EventGrid.TypeConverters
     {
         protected override ConversionResult ConvertCore(Type targetType, string json)
         {
-            object result = targetType switch
+            ConversionResult result = targetType switch
             {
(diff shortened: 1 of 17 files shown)


test/FunctionMetadataGeneratorTests/FunctionMetadataGeneratorTests.cs
```diff
@@ -176,7 +176,7 @@ public void StorageFunctions()
             AssertDictionary(extensions, new Dictionary<string, string>
             {
                 { "Microsoft.Azure.WebJobs.Extensions.Storage.Queues", "5.1.3" },
-                { "Microsoft.Azure.WebJobs.Extensions.Storage.Blobs", "5.1.2" },
+                { "Microsoft.Azure.WebJobs.Extensions.Storage.Blobs", "" },
```
aishwaryabh, line 179: Is this intentionally an empty string for blob versioning?

aishwaryabh (approved):
LGTM! Thanks for refactoring the table and event grid tests :)

test/FunctionMetadataGeneratorTests/FunctionMetadataGeneratorTests.cs
liliankasem: yikes, definitely not - find and replace gone wrong :)

PR #1749, opened 2023-07-13
Worker Extension project incremental build support

### LINKED ISSUE #1740: SDK: Incremental build support, build in intermediate dir
## Overview

This issue is to improve our SDK and address some pain points with it. Primarily, customers are encountering file lock and anti-virus issues. To mitigate these further (first mitigation in #1532), we will perform two changes to our SDK:

## 1. In-memory extension project

Instead of emitting an actual .csproj to the temp directory as part of collecting host extensions, we will use the MSBuild APIs to create and evaluate a completely in-memory .csproj. This will allow us to avoid disk writes and triggering anti-virus scans when we write to the temp directory.

### Update

This does not seem feasible as the MSBuild construction & evaluation types we would need for this are defined in `Microsoft.Build` nuget package, which has no `netstandard2.0` support. Using this package will not be a stable approach. Will take a different approach instead.

### Alternative approach

Instead of using in-memory, we can improve the on-disk approach with one of two attempts:

1. Move the csproj to the intermediate output directory.
  - pro: keeps this out of the temp directory. Can keep the name consistent and reliably re-use the project between builds.
  - con: might pick up implicit imports by accident. But this is already a risk with even the tmp directory. We should suppress implicit imports anyways.
2. If the intermediate output directory does not work, we can atleast write to the intermediate output dir the path of any existing extension csproj in the TMP dir, so we can attempt to re-use it.

## 2. Incremental build support

We will refactor our SDK targets to have full incremental build support, meaning building the same dotnet isolated worker csproj twice will yield a no-op on the second build (provided `--no-incremental` is not supplied).

jviau:

resolves #1740 

This PR attempts to solve frequent pain points with our SDK build steps which cause file locked issues. I believe the issue lies with how we perform an inner-build with the `WorkerExtensions.csproj` to generate collect host extensions. To mitigate this inner-build causing file lock issues, two primary changes have been made:

1. The `WorkerExtension.csproj` is now generate & built in the intermediate output path of the outer-project (typically `obj/{Configuration}/{TargetFramework}`) under the `WorkerExtensions` folder.
2. Incremental build support has been improved
  - Copy tasks now use `SkipUnchanged` to avoid copying files which already exist and are up to date at the target destination
  - Generation of the csproj itself now checks for an existing csproj, and only writes if there are differences.

These changes combined reduces how much happens on re-builds of a functions worker project, which as a result means less file interactions and thus less file locked issues.

### Pull request checklist

TODO

### DIFF
--- a/samples/FunctionApp/FunctionApp.csproj
+++ b/samples/FunctionApp/FunctionApp.csproj
@@ -14,7 +14,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.ApplicationInsights.WorkerService" Version="2.21.0" />
-    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.7.0" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.13.0-preview1-local" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.2" />
     <PackageReference Include="System.Net.NameResolution" Version="4.3.0" />
   </ItemGroup>
--- a/sdk/Sdk/ExtensionsCsprojGenerator.cs
+++ b/sdk/Sdk/ExtensionsCsprojGenerator.cs
@@ -31,9 +31,23 @@ namespace Microsoft.Azure.Functions.Worker.Sdk
         {
             var extensionsCsprojFilePath = Path.Combine(_outputPath, ExtensionsProjectName);
 
-            RecreateDirectory(_outputPath);
+            string csproj = GetCsProjContent();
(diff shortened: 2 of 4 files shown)


release_notes.md
```diff
@@ -15,3 +15,9 @@
 ### Microsoft.Azure.Functions.Worker.Grpc <version>
 
 - <entry>
+
+### Microsoft.Azure.Functions.Worker.Sdk 1.13.0-preview1
```
liliankasem, line 19: probably should update to latest

liliankasem (approved):

PR #1750, opened 2023-07-13
SDK-binding support for CosmosDB worker extension 

### LINKED ISSUE #1751: Release CosmosDB worker extension with SDK-binding support
- [ ] Merge CosmosDB worker extension sdk-binding feature in to main
 - [ ] Release stable CosmosDB worker extension package

liliankasem:

resolves #1751

### Pull request checklist

### Additional information

Additional PR information


extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
```diff
@@ -0,0 +1,188 @@
...
+                FeedResponse<T> response = await iterator.ReadNextAsync();
+                documentList.AddRange(response.Resource);
+            }
+            return documentList;
+        }
+
+        private T CreateCosmosClient<T>(CosmosDBInputAttribute cosmosAttribute)
+        {
+            var cosmosDBOptions = _cosmosOptions.Get(cosmosAttribute?.Connection);
+            CosmosClient cosmosClient = cosmosDBOptions.GetClient(cosmosAttribute?.Connection!, cosmosAttribute?.PreferredLocations!);
```
jviau, line 175: We already get the cosmos options based on `Connection`, passing connection in to `GetClient` seems redundant and also incorrect. It looks to me you shouldn't be able to get options with connection "A" then call GetClient with connection "B" - that would provide an incorrect client.

Removing `Connection` from the `GetClient` call would fix this.

extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBBindingOptions.cs
```diff
@@ -0,0 +1,52 @@
...
+        internal virtual CosmosClient GetClient(string connection, string preferredLocations = "")
+        {
+            if (string.IsNullOrEmpty(connection))
+            {
+                throw new ArgumentNullException(nameof(connection));
+            }
+
+            string cacheKey = BuildCacheKey(connection, preferredLocations);
+
+            CosmosClientOptions cosmosClientOptions = new ()
```
jviau, line 32: Customers have no way of configuring CosmosDB JSON serialization behavior. We should consider adding that (can be post-GA if we need).

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
liliankasem: connection is being passed in just for the cache which we decided to remove anyway in favor of ` internal CosmosClient? Client { get; set; }` like we do in blob - seems like we never got round to removing that so I'll fix this now (and remove use of connection here as it will no longer be needed)

extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBBindingOptions.cs
liliankasem: Would a good way to do this be to use the worker options serializer here?

jviau: Might be possible. It won't be the same as Azure Track 2 SDKS, as they directly take `ObjectSerializer`. CosmosDB has a different serializer built into it, so would need to do whatever shimming necessary to integrate the worker options serializer there.

extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBBindingOptions.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        public string? ConnectionString { get; set; }
+
+        public string? AccountEndpoint { get; set; }
+
+        public TokenCredential? Credential { get; set; }
+
+        internal CosmosClient? Client { get; set; }
+
+        internal virtual CosmosClient GetClient(string preferredLocations = "")
```
jviau, line 22: I don't think a singular client is correct due to `preferredLocations`. We may still need the dictionary cache for these different clients.

liliankasem: Ahh fair point

liliankasem: @jviau updated again xD let me know what you think

extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBBindingOptions.cs
liliankasem: Is this what you had in mind: https://github.com/Azure/azure-functions-dotnet-worker/pull/1779/files

jviau (approved):

PR #1755, opened 2023-07-14
App Insights: improving property mappings; adding role details

### LINKED ISSUE #1634: App Insights telemetry role name not correct after slot swap
App Insights relies on two mechanisms to keep the role name correct: the `WEBSITE_HOSTNAME` env var and the `WAS-DEFAULT-HOSTNAME` header that App Service sends with every request. The header is important because during a slot swap, the host name can change... thus changing the role name on telemetry.

However, in the Functions worker, this does not work. The header is never received (as HTTP isn't truly HTTP -- and most triggers are not HTTP anyway). Then, the fallback to WEBSITE_HOSTNAME ends up being incorrect and you end up with a prod site reporting its role name as "staging" (or something).

In Functions/WebJobs we've had logic here for many years that creates the role name based on `WEBSITE_SITE_NAME` and `WEBSITE_SLOT_NAME`. This has worked for us very reliably, so we'll move to that.

It may be worth the host sending this along during initialization -- but notes in the WebJobs code seem to indicate that these values can change, so we'd need to ensure that is not true before we rely on them that way.

WebJobs initializer: 
- https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsRoleEnvironmentTelmetryInitializer.cs

References to App Insights initializers:
- https://github.com/microsoft/ApplicationInsights-dotnet/blob/main/WEB/Src/Web/Web/AzureAppServiceRoleNameFromHostNameHeaderInitializer.cs
- https://github.com/microsoft/ApplicationInsights-dotnet/blob/main/WEB/Src/WindowsServer/WindowsServer/AzureWebAppRoleEnvironmentTelemetryInitializer.cs

### LINKED ISSUE #1706: InvocationId and LogLevel Missing from customDimensions
Hello, we recently updated from an older version of **Microsoft.Azure.Functions.Worker.*** to the latest avaible version

- Microsoft.Azure.Functions.Worker 1.16.0
- Microsoft.Azure.Functions.Worker.Extensions.ServiceBus 5.11.0
- Microsoft.Azure.Functions.Worker.Sdk 1.11.0

After the update, for some reason, the **_InvocationId_** and the **_LogLevel_** values are missing from the **customDimensions** and because of this, most part of our logs are not displayed in the **Azure Function's Monitor** section:
image

The query which is generaterd when clicking the option _**Run query in Application Insighs**_  from the azure function invocation is added below and, as expected, displays the same result as in the image above:

```
union traces
| union exceptions
| where timestamp > ago(30d)
| where operation_Id == 'op_id_value_here'
| where customDimensions['InvocationId'] == 'invocation_id_value_here'
| order by timestamp asc
| project
    timestamp,
    message = iff(message != '', message, iff(innermostMessage != '', innermostMessage, customDimensions.['prop__{OriginalFormat}'])),
    logLevel = customDimensions.['LogLevel']
```

What we've noticed is that if we're removing the `customDimensions['InvocationId'] == 'invocation_id_value_here'` filter, we obtain all the information which is logged by the function (as before updating to latest worker.* version).
image

What worth mentioning is that for the logs which appear in the **_Monitor blade_** (for a particular invocation) we have in **_customDimensions_** object both the _**InvocationId**_ and **_LogLevel_** properties, as shown in the image below:
image

But for the logs which are not displayed in  the **_Monitor blade_** we do not have those properties, and instead of **_InvocationId_** property we have _**AzureFunctions_InvocationId**_ (which has the same value as InvocationId):
image

brettsam:
Fixes #1634 
Fixes #1706 

### Pull request checklist

### Additional information

Additional PR information


src/DotNetWorker.ApplicationInsights/Initializers/WebJobsRoleEnvironmentTelmetryInitializer.cs
```diff
@@ -0,0 +1,86 @@
...
+        /// </summary>
+        /// <param name="telemetry">The telemetry to initialize.</param>
+        public void Initialize(ITelemetry telemetry)
+        {
+            if (telemetry == null)
+            {
+                return;
+            }
+
+            var siteSlotName = new Lazy<string?>(() =>
```
jviau, line 39: I see this is copied over. But it doesn't serve a purpose? What difference does reading this environment variable a few lines later matter?

``` CSharp
string siteSlotName = GetAzureWebsiteUniqueSlotName();
```
Is just as effective. It will be read independently for each `Initialize` call.

But given this is copied over, leave it if you want. Although it is an extra unnecessary allocation.

brettsam: I was wondering the same thing... I went back and it looks like @mathewc did this in a perf improvement PR several years ago: https://github.com/Azure/azure-webjobs-sdk/pull/2633. 

I know that Environment reads are abnormally expensive in Antares so he was likely only trying to read the env if it was absolutely necessary, even if it meant an extra allocation of that Lazy?

But yeah -- I mostly just copied it over as-is and didn't want to go mucking with it. Maybe we can refine it again, though...

jviau (approved):

fabiocav (approved):

Issue #1566, opened 2023-05-25
Integrate Microsoft.Extensions.Azure and IAzureClientFactory<T> into Azure SDK extension packages


jviau:
## Overview

We have an opportunity to leverage `Microsoft.Extensions.Azure` package and its `IAzureClientFactory<TClient>` to enrich Azure SDK usage from the dotnet worker. This work will provide the following benefits:

1. Customers can use DI to get Azure SDK clients as an alternative to input binding.
    - This is possible today, but customers have to perform all the setup work themselves. I believe we already have enough information for our Azure SDK extension packages to perform do the setup work automatically.
2. We can internally consume `IAzureClientFactory<TClient>` to improve our own consumption of clients. Particularly in the SDK deferred binding scenarios.
    - Today we are performing a lot of manual client configuration and construction. Instead, we can perform it once at startup and then input bindings for a client can rely internally on `IAzureClientFactory<TClient>`.

## Implementation / Design Ideas

We already have our configuration structure in the shape that the Azure SDKs want. And we already refer to them by name via `"ConnectionName"`. I believe all we need to do is for the respective Azure SDK extension packages to scan all the configured `ConnectionName`'s available, and then use the APIs from `Microsoft.Extensions.Azure` to build & register a client by name. Then all these clients should be available to us (and customers) via `IAzureClientFactory<TClient>.CreateClient(<connectionName>)`

liliankasem:
So the pattern we're following right now is similar to what WebJobs.Extensions.Storage.Blobs is doing ([not using AzureClientFactory](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Blobs/src/StorageBlobsWebJobsBuilderExtensions.cs#L31)). From my understanding, for us to use IAzureClientFactory<TClient> we need to register this at startup i.e.

```csharp
clientBuilder.AddBlobServiceClient(<connectionstring/connectionname/serviceuri>).WithName("MyClientName");

// ... and later in use it our converter:

BlobServiceClient blobServiceClient = _blobServiceClientFactory.CreateClient("MyClientName");
```

[[docs](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/extensions/Microsoft.Extensions.Azure/samples/Startup.cs)]

[...]

PR #1764, opened 2023-07-20
Fix BlobStorageBindingOptions client cache

liliankasem:
Bug: We never set the Client value in BlobStorageBindingOptions


jviau:
Good catch. It just meant we were newing up clients constantly. Which would actually lead to SNAT exhaustion due to not re-using the underlying HTTP clients.

Speaking of which, that does make #1566 even more important, as it will integrate with `IHttpClientFactory`. We could do that ourselves also if we wanted.

jviau (approved):

PR #1776, opened 2023-07-21
SDK-binding support for Tables worker extension 

### LINKED ISSUE #1767: Release Tables worker extension with SDK-binding support
- [ ] Merge Tables worker extension sdk-binding feature in to main
 - [ ] Release stable Tables worker extension package

liliankasem:

resolves #1767 

### Pull request checklist

### Additional information

Additional PR information


extensions/Worker.Extensions.Tables/src/Properties/AssemblyInfo.cs
```diff
@@ -1,6 +1,9 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
-﻿using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
+using System.Runtime.CompilerServices;
+using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
 
-[assembly: ExtensionInformation("Microsoft.Azure.WebJobs.Extensions.Tables", "1.1.0")]
+[assembly: ExtensionInformation("Microsoft.Azure.WebJobs.Extensions.Tables", "1.2.0-beta.1")]
```
liliankasem, line 7: We're blocked on this WebJobs release

aishwaryabh (approved):
LGTM!

extensions/Worker.Extensions.Tables/src/Config/TablesBindingOptions.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        public TableClientOptions? TableClientOptions { get; set; }
+
+        private TableServiceClient? TableServiceClient;
+
+        internal virtual TableServiceClient CreateClient()
+        {
+            if (ConnectionString is null && ServiceUri is null)
+            {
+                throw new ArgumentNullException(nameof(ConnectionString) + " " + nameof(ServiceUri));
```
mathewc, line 26: Incorrect values being passed to the ANE ctor - should be a param name not an error message

extensions/Worker.Extensions.Tables/src/Config/TablesBindingOptionsSetup.cs
```diff
@@ -0,0 +1,59 @@
...
+            {
+                name = Constants.Storage; // default
+            }
+
+            IConfigurationSection connectionSection = _configuration.GetWebJobsConnectionStringSection(name);
+
+            if (!connectionSection.Exists())
+            {
+                // Not found
+                throw new InvalidOperationException($"Tables connection configuration '{name}' does not exist. " +
```
mathewc, line 39: remove this unnecessary string concat

extensions/Worker.Extensions.Tables/src/TypeConverters/TableEntityEnumerableConverter.cs
```diff
@@ -0,0 +1,103 @@
...
+
+            int countRemaining = content.Take;
+
+            var entities = tableClient.QueryAsync<TableEntity>(
+                            filter: filter,
+                            maxPerPage: maxPerPage).ConfigureAwait(false);
+
+            List<TableEntity> bindingDataContent = new List<TableEntity>();
+
+            await foreach (var entity in entities)
```
mathewc, line 90: Why are we materializing this ourselves instead of just letting the user enumerate themselves? The In-proc extension did this (looks like this code is a copy) because it was typed as JArray. But here the user is binding to an IEnumerable so we could just hand that back to them?

liliankasem: Just to clarify, you're saying that the code was copied from [CreateJArray](https://github.com/Azure/azure-sdk-for-net/blob/f738f6a6b3f7f95114cc92fa26ded100907b74de/sdk/tables/Microsoft.Azure.WebJobs.Extensions.Tables/src/TablesExtensionConfigProvider.cs#L135), and in that code they did this because they had to create a JArray. We do not have to do this because `entities` is already IEnumerable so we should be able to just return that?

I agree with you here; I missed this in the original [PR](https://github.com/Azure/azure-functions-dotnet-worker/pull/1506), but returning entities directly here makes the most sense - thanks for catching!

liliankasem: Though I wonder if we would still need to convert this to IEnumerable since technically `entities` is actually `ConfiguredCancelableAsyncEnumerable<TableEntity>`. Or if we should change the supported binding type to IAsyncEnumerable

liliankasem: @mathewc thoughts on this ^ ? I updated the code but still creating an IList

mathewc: You're pointing out that the entities is actually an `AsyncPageable<T>`? Then perhaps that's what we should be returning - there are methods on there related to paging that the user might care about. When we do a ToList, it ends up enumerating all the pages (possibly multiple service calls), right? Our goal with SDK bindings is to allow the user to bind to SDK natural types. So we need to decide if we want to interject ourselves and force enumerate/convert.

liliankasem: Discussed offline, continuing with IEnumerable for now as we need to take into account Take for parity. Future work for IAsyncPageable

mattchenderson (approved):
Approving as these changes were previously reviewed going into the branch, and we've discussed the `IEnumerable` hydration offline. I have not gone through and conducted another review of the changes themselves.

PR #1785, opened 2023-07-26
Implementation of the FromBody binding converter for the ASP.NET Core Integration

### LINKED ISSUE #1728: Implement a FromBodyConverter in ASP.NET Core integration
Create an implementation of a `FromBodyConverter` in the ASP.NET Core integration project, providing binding support using the ASP.NET Core functionality.

fabiocav:

resolves #1728 

### Pull request checklist


extensions/Worker.Extensions.Http.AspNetCore/src/FromBodyConverstionFeature.cs
```diff
@@ -0,0 +1,93 @@
...
+                       {
+                           BinderModelName = binderModelName,
+                           BindingSource = BindingSource.Body,
+                           BinderType = metadata.BinderType,
+                           PropertyFilterProvider = metadata.PropertyFilterProvider,
+                       },
+                       CacheToken = null
+                   });
+
+                if (modelBinder is not null)
```
brettsam, line 47: What if modelBinder is null? Should we log an error/ return failure? Or can that not happen?

extensions/Worker.Extensions.Http.AspNetCore/src/FromBodyConverstionFeature.cs
```diff
@@ -0,0 +1,93 @@
...
+                        }
+                    };
+
+                    await modelBinder.BindModelAsync(modelBindingContext);
+
+                    if (modelBindingContext.Result.IsModelSet)
+                    {
+                        return modelBindingContext.Result.Model;
+                    }
+                    else if (!modelBindingContext.ModelState.IsValid)
```
brettsam, line 67: what scenario would trigger this? the payload can't be bound to the object?

samples/AspNetIntegration/BodyBindingHttpTrigger.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Azure.Functions.Worker;
+using FromBodyAttribute = Microsoft.Azure.Functions.Worker.Http.FromBodyAttribute;
```
mattchenderson, line 7: This wasn't needed for the base implementation for the non-ASP.NET Core version of HttpTrigger, but it does stand out here. Even with documentation, I think a number of folks will miss this and end up using the version from ASP.NET Core.

fabiocav: Yes, this is present here to disambiguate the type reference as `Microsoft.AspNetCore.Mvc` is also being used. It is unfortunate, but there are a few things we can do to avoid confusion. We discussed some of those, but to capture here, we could also honor MVC attribute and leverage analyzers to help.

extensions/Worker.Extensions.Http.AspNetCore/src/FromBodyConverstionFeature.cs
fabiocav: Yes. A request without a body payload that can be deserialized into the target type or a validation failure are some examples.

extensions/Worker.Extensions.Http.AspNetCore/src/FromBodyConverstionFeature.cs
fabiocav: The binder is registered by default, and should always be present with current implementation, but once the ASP.NET integration becomes more customizable, there’s a possibility this could be hit. I’ll update to ensure this is validated and the error is clear.

brettsam (approved):

satvu (approved):

PR #1794, opened 2023-07-31
Auto register auto generated IHostBuilder extension methods

### LINKED ISSUE #1161: Auto-register GeneratedFunctionMetadataProvider
Today the generated provider needs to be manually added to the host builder. We should use `WorkerExtensionStartup` to automatically make this registration call for the user. Additionally, this behavior should be configurable by the user via MSBuild property. Something along the lines of `<GeneratedFunctionMetadataProvider_AutoRegister>true|false</GeneratedFunctionMetadataProvider_AutoRegister>` with a sensible default if not set in the project.

kshyju:
Fixes #1161 

Auto register the `IHostBuilder` extension methods emitted by our source generators for metadata provider and function executor. 

Currently we have 2 methods we want to explicitly register in main like below.

```csharp
var host = new HostBuilder()

    // Below 2 are the auto generated methods
    .ConfigureGeneratedFunctionMetadataProvider()
    .ConfigureGeneratedFunctionExecutor()

    .Build();
host.Run();
```

In this PR we are introducing a new interface `IAutoConfigureStartup` as below.

```csharp
public interface IAutoConfigureStartup
{
    /// <summary>
    /// Configures an <see cref="IHostBuilder"/> instance for startup.
    /// Implementing classes define the configuration logic within this method.
    /// This method will be invoked during the bootstrapping process of the function app.
    /// </summary>
    /// <param name="hostBuilder">The <see cref="IHostBuilder"/> instance to configure.</param>
    public void Configure(IHostBuilder hostBuilder);
}
```
Any class which wants to do auto configuration during startup should implement this interface. Those types will be discovered when the app starts and the Configure method will be invoked.

Updated both the source generators to emit an implementation of the above interface. The `Configure` method is simply calling the existing extension method as shown below. 

```csharp
public class FunctionMetadataProviderAutoStartup : IAutoConfigureStartup
{
    public void Configure(IHostBuilder hostBuilder)
    {
        hostBuilder.ConfigureGeneratedFunctionMetadataProvider();
    }
}
```

The behavior is disabled by default and one can enable them by setting the below MSBuild property.

```xml
<FunctionsAutoRegisterGeneratedMetadataProvider>True</FunctionsAutoRegisterGeneratedMetadataProvider>
```
For function executor registration, the default behavior is to auto register as long as the code is generated. But if user wishes to disable it, they can do it by doing the following. 

```xml
<FunctionsAutoRegisterGeneratedFunctionsExecutor>False</FunctionsAutoRegisterGeneratedFunctionsExecutor>
``` 

Sample csproj file content where auto registration of generated function metadata provider and generated executor.

```xml
<FunctionsEnableWorkerIndexing>True</FunctionsEnableWorkerIndexing>
<FunctionsAutoRegisterGeneratedMetadataProvider>True</FunctionsAutoRegisterGeneratedMetadataProvider>
```

### Pull request checklist

* [x] My changes **do not** require documentation changes  - May create documentation as we GA these features.


sdk/Sdk.Generators/Constants.cs
```diff
@@ -14,6 +14,8 @@ internal static class BuildProperties
         {
             internal const string EnableSourceGen = "build_property.FunctionsEnableMetadataSourceGen";
             internal const string EnablePlaceholder = "build_property.FunctionsEnableExecutorSourceGen";
+            internal const string AutoRegisterGeneratedFunctionsExecutor = "build_property.FunctionsAutoRegisterGeneratedFunctionsExecutor";
```
kshyju, line 17: Should we use the `FunctionsEnable` prefix to be consistent with the other 2 properties? Any suggestions for the names?

src/DotNetWorker.Core/Hosting/AutoGeneratedClassAttribute.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using System;
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    /// <summary>
+    /// Identifies class generated by Azure functions worker SDK.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class)]
+    public sealed class AutoGeneratedClassAttribute : Attribute
```
kshyju, line 12: I originally named this as `FunctionsWorkerSdkGeneratedClassAttribute`, but later felt that it is somewhat redundant with the namespace. So switched to `Auto` prefix. (Same for the method attribute as well). Let me know you think we should use a different prefix/name.

sdk/Sdk.Generators/Constants.cs
satvu: The name makes sense to me as is and I don't think enable would add much clarity - it would also make the property name very long.

satvu (approved):
lgtm! also like the naming choices as is.

sdk/Sdk.Generators/Constants.cs
liliankasem: Maybe `FunctionsEnableGeneratedExecutorRegistration`

mattchenderson:
Do we need to have both `FunctionsEnableExecutorSourceGen` and `FunctionsAutoRegisterGeneratedFunctionsExecutor`? I feel like the latter behavior can just be included in the former, assuming that basically a double call to `.ConfigureGeneratedFunctionExecutor()` is fine for any situation in which the existing behavior of `FunctionsEnableExecutorSourceGen` was used and that method was called explicitly.

Just running the permutations:
| Enable... | AutoRegister... | Behavior | could just be... |
|-|-|-|-|
| False | False | Existing default, without | FunctionsEnableExecutorSourceGen=False |
| False | True | Invalid case, right? | N/A |
| True | False | No action, unless you make code changes, but do we need that? | N/A |
| True| True | feature on without code changes| FunctionsEnableExecutorSourceGen=True |

I think we would be best off keeping it simple and confined to the MSBuild option.

sdk/Sdk/Sdk.csproj
```diff
@@ -1,8 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <MinorProductVersion>13</MinorProductVersion>
+    <MinorProductVersion>14</MinorProductVersion>
     <PatchProductVersion>0</PatchProductVersion>
+    <VersionSuffix>-preview1</VersionSuffix>
```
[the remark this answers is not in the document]
kshyju: I rolled back the version bump changes from all csproj files. When it is time to release, will bump the versions as needed.

liliankasem (approved):

src/DotNetWorker.Core/Hosting/CoreWorkerHostBuilderExtensions.cs
```diff
@@ -38,5 +40,48 @@ public static IHostBuilder ConfigureFunctionsWorker(this IHostBuilder builder, A
...
+        /// <returns>
+        /// true if the method is an extension method for <see cref="IHostBuilder"/> and has
+        /// the <see cref="AutoGeneratedHostBuilderConfigureMethodAttribute"/> decoration on it
+        /// otherwise, false.
+        /// </returns>
+        private static bool IsAutoGeneratedExtensionMethodForHostBuilder(MethodInfo method)
+        {
+            return method.IsDefined(typeof(System.Runtime.CompilerServices.ExtensionAttribute), inherit: false)
+                   && method.IsDefined(typeof(AutoGeneratedHostBuilderConfigureMethodAttribute))
+                   && method.GetParameters().Length > 0
```
jviau, line 83: should you check parameters has a length of exactly 1? And that is `IHostBuilder`?

jviau (approved):
The attribute route for discovering and invoking these extensions is fine. You could also create a new interface which has a `Configure(IHostBuilder builder)` on it, which you will discover, activate, and invoke. Both approaches are fine, so it is up to your preference.

aishwaryabh (approved):

src/DotNetWorker.Core/Hosting/CoreWorkerHostBuilderExtensions.cs
```diff
@@ -38,5 +40,48 @@ public static IHostBuilder ConfigureFunctionsWorker(this IHostBuilder builder, A
...
+        /// </summary>
+        /// <param name="method">The <see cref="MethodInfo"/> representing the method to be checked.</param>
+        /// <returns>
+        /// true if the method is an extension method for <see cref="IHostBuilder"/> and has
+        /// the <see cref="AutoGeneratedHostBuilderConfigureMethodAttribute"/> decoration on it
+        /// otherwise, false.
+        /// </returns>
+        private static bool IsAutoGeneratedExtensionMethodForHostBuilder(MethodInfo method)
+        {
+            return method.IsDefined(typeof(System.Runtime.CompilerServices.ExtensionAttribute), inherit: false)
```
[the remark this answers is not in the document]
liliankasem: > inherit: true to search this member's inheritance chain to find the attributes; otherwise, false. This parameter is ignored for properties and events.

So for the given `method` that we want to check, by setting inherit to "false" we are saying that we do not want to check this method's inheritance chain to check if the method is an extension method or not.

For future reference: when it comes to things coming from .NET (like MethodInfo), you can usually look up documentation for the methods and their parameters in the official Microsoft docs:

https://learn.microsoft.com/en-us/dotnet/api/system.reflection.memberinfo.isdefined?view=net-7.0#system-reflection-memberinfo-isdefined(system-type-system-boolean)

src/DotNetWorker.Core/Hosting/AutoGeneratedClassAttribute.cs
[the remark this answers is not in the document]
kshyju: @jviau  I originally named this as F`unctionsWorkerSdkGeneratedClassAttribute`, but later felt that it is somewhat redundant with the namespace. So switched to Auto prefix. (Same for the method attribute as well).  Feel free to suggest a better name.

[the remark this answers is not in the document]
jviau: I am thinking instead of naming based on it being primarily used for auto-generated classes, name based on what it does: auto startup hook. Maybe `HostBuilderStartupHookAttribute`. 

Also to my point for using a new interface vs attributes. The attribute route requires knowledge of using an attribute in _two_ places. The interface route requires single knowledge: implement the interface.

kshyju:
> The attribute route for discovering and invoking these extensions is fine. You could also create a new interface which has a `Configure(IHostBuilder builder)` on it, which you will discover, activate, and invoke. Both approaches are fine, so it is up to your preference.

@jviau 

I considered this route, but found a few issues.

Currently we generate an extension method `ConfigureGeneratedFunctionMetadataProvider` inside the static `WorkerHostBuilderFunctionMetadataProviderExtension` class. we cannot implement the new interface on this class (static methods/class decoration are not allowed when implementing the interface).

Another option is to not emit the static extension method and static class. Simply create a class which implements the said interface with `public void Configure(IHostBuilder hostBuilder)` method and our startup code can call this for all implementations of this interface, but then we are taking away the ability to manually call this extension methods if needed in main (as we do today). Also wouldn't that be a breaking change?

Third option is to have an additional class (which implements the said interface) along with existing extension method classes. But I thought that seems more code we are emitting. So I decided to not go that route. But I am happy to pivot if there is a better way to handle this.

Thoughts?

jviau:
> Third option is to have an additional class (which implements the said interface) along with existing extension method classes. But I thought that seems more code we are emitting. So I decided to not go that route. But I am happy to pivot if there is a better way to handle this.

This is what I was thinking if you _did_ decide to go that route. Yes, it is more code to emit, but that doesn't really cost anything besides authoring the generator itself.

I am fine with either route, interface or attributes. Another idea for attributes is to remove the method attribute and instead look for method by convention: static, public, named "Configure", accepts `IHostBuilder` as only parameter.

satvu (approved):

liliankasem (approved):

sdk/Sdk.Generators/FunctionExecutor/FunctionExecutorGenerator.Emitter.cs
```diff
@@ -13,8 +13,9 @@ public partial class FunctionExecutorGenerator
     {
         internal static class Emitter
         {
-            internal static string Emit(IEnumerable<ExecutableFunction> functions, CancellationToken cancellationToken)
+            internal static string Emit(IEnumerable<ExecutableFunction> functions, bool includeAutoGeneratedAttributes, CancellationToken cancellationToken)
```
jviau, line 16: small nit: update parameter name.

src/DotNetWorker.Core/Hosting/CoreWorkerHostBuilderExtensions.cs
```diff
@@ -38,5 +40,35 @@ public static IHostBuilder ConfigureFunctionsWorker(this IHostBuilder builder, A
...
+        /// <summary>
+        /// Invokes auto-generated configuration methods for a given <see cref="IHostBuilder"/>.
+        /// This method searches for classes that implement the <see cref="IAutoConfigureStartup"/> interface,
+        /// excluding interfaces and abstract classes. For each identified class, it locates the
+        /// <c>Configure</c> method with the signature <c>void Configure(IHostBuilder hostBuilder)</c>,
+        /// and executes the method using an instance of the class.
+        /// </summary>
+        /// <param name="builder">The <see cref="IHostBuilder"/> to configure.</param>
+        /// <returns>The same <see cref="IHostBuilder"/> after the auto-generated configuration methods are invoked.</returns>
+        public static IHostBuilder InvokeAutoGeneratedConfigureMethods(this IHostBuilder builder)
```
jviau, line 53: small suggestion: if this is going to be `public`, then consider adding a parameter to accept the `Assembly` that we will scan. This will make it slightly more flexible if devs start calling it directly. Our usage of it can then pass in the entry assembly.

src/DotNetWorker.Core/Hosting/CoreWorkerHostBuilderExtensions.cs
```diff
@@ -38,5 +40,35 @@ public static IHostBuilder ConfigureFunctionsWorker(this IHostBuilder builder, A
...
+        {
+            var autoConfigureStartupTypes = Assembly.GetEntryAssembly()!
```
jviau, line 55: There are some cases where `GetEntryAssembly()` returns null. Will that ever impact us? https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.getentryassembly?view=net-7.0

src/DotNetWorker.Core/Hosting/CoreWorkerHostBuilderExtensions.cs
```diff
@@ -38,5 +40,35 @@ public static IHostBuilder ConfigureFunctionsWorker(this IHostBuilder builder, A
...
+
+            foreach (var type in autoConfigureStartupTypes)
+            {
+                var configureMethod = type.GetMethod(nameof(IAutoConfigureStartup.Configure), new[] { typeof(IHostBuilder) });
+                if (configureMethod == null)
+                {
+                    continue;
+                }
+
+                var instance = Activator.CreateInstance(type) as IAutoConfigureStartup;
```
jviau, line 67: small preference for a hard cast - if this is somehow `null`, something has gone wrong.

jviau (approved):

src/DotNetWorker.Core/Hosting/CoreWorkerHostBuilderExtensions.cs
kshyju: That makes sense. Switched to direct cast.

src/DotNetWorker.Core/Hosting/CoreWorkerHostBuilderExtensions.cs
kshyju: Good to know. I do not think any of our cases are impacted by this. We already tested the previous version of the PR on cold start-placeholder scenarios where the FunctionsNetHost is loading the customer assembly and it seems to work.

src/DotNetWorker.Core/Hosting/CoreWorkerHostBuilderExtensions.cs
kshyju: Going to leave it as it is for now. If we have a need to use the code in this method in a second place, we shall consider that change.

sdk/Sdk.Generators/FunctionExecutor/FunctionExecutorGenerator.Emitter.cs
kshyju: Updated parameter name.

src/DotNetWorker.Core/Hosting/CoreWorkerHostBuilderExtensions.cs
kshyju: This is not valid anymore as we switched from this implementation.

PR #1811, opened 2023-08-09
Attribute-based middleware

xin9le:
# Issue describing the changes in this PR

This PR allows for the provision of Middleware that can be applied at the function level based on attributes. Currently, control is possible through the use of the `.UseWhen<TMiddleware>()` method, but because it is very primitive, SDK users suffer from implementation pains. This PR resolves that issue.

# Pull request checklist

# How to use

## Step.1 : Creates custom middleware

```cs
internal sealed class CustomMiddleware : IFunctionsWorkerMiddleware
{
    {
        // do something
        await next(context);
    }
}
```

## Step.2 : Registers middleware into DI containers
```cs
new HostBuilder()
    .ConfigureFunctionsWorkerDefaults(static (context, builder) =>
    {
        builder.UseWhenAttribute<CustomMiddleware>();
    })
    .Build()
    .Run();
```

## Step.3 : Annotate `FunctionMiddlewareAttribute<T>`
```cs
internal sealed class FooTimer
{
    [FunctionMiddleware<CustomMiddleware>]  // anotate if you want to use
    [Function("FooTimer")]
    public async Task EntryPoint(
        [TimerTrigger("%AzureWebJobs.FooTimer.Cron%")] TimerInfo timer,
        FunctionContext context)
    {
        // do something
    }
}
```

# Limitation

This feature uses [Generic Attributes](https://github.com/dotnet/csharplang/issues/124). Therefore, it cannot be applied unless C# 11 or later. It is possible to implement this feature without using generic attributes, but it would lose the ability to constrain to the `IFunctionsWorkerMiddleware` type. I have leaned towards adding the constraint in this trade-off, but I would like to hear the opinions of everyone on the SDK team.

# Typical scenario

Sometimes, we may want to set up middleware on a function-by-function basis. For example, you might want to add API key authentication only to certain functions for HttpTrigger. In such cases, it would be cumbersome to check for the API key at the entry point of each HttpTrigger, so you'd probably want to streamline this process with middleware. To achieve this, currently, the only solution is to use the `.UseWhen<TMiddleware>()` method as follows.

```cs
// Program.cs
new HostBuilder()
    .ConfigureFunctionsWorkerDefaults(static (context, builder) =>
    {
        builder.UseWhen<ApiKeyAuthMiddleware>(static context =>
        {
            return context.FunctionDefinition.Name is "Http_DoSomething";
        });
    });
```
```cs
// SampleFunctions.cs
internal sealed class SampleFunctions
{
    [Function("Http_DoSomething")]
    public async Task EntryPoint(
        [HttpTrigger(AuthorizationLevel.Function, "get", Route = "something")] HttpRequestData request,
        FunctionContext context)
    {
        // do something
    }
}
```

This implementation has the following pain points:

- You need to write the implementation for applying the middleware and the functions to be applied in different files.
- You need to compare the names of the target functions as `string`.
    - Although the pain can be alleviated by defining the function names as constants, in my experience, few developers can safely implement changes to function names.
- As the number of middleware and trigger to be applied increases, the predicate delegate becomes bloated and maintainability decreases.

Therefore, I believe that an attribute-based declarative approach, like the one in this PR, can solve the problems mentioned above. The method of `ActionFilterAttribute` in ASP.NET Core MVC has a long history, so I guess there are many developers who are familiar with it.

# Implementation approach

In this PR implementation, I'm scanning the assembly to get the custom attribute `FunctionMiddlewareAttribute<T>` applied to the entry point. As far as I know, there is no other way to get the custom attribute applied to the entry point via `FunctionContext`, which is the callback argument for `.UseWhen<TMiddleware>()`. If a way to get the custom attribute from the `FunctionContext` is provided, it would probably make the implementation simpler.

### DIFF
--- a/src/DotNetWorker.Core/Context/FunctionContextMetadataExtensions.cs
+++ b/src/DotNetWorker.Core/Context/FunctionContextMetadataExtensions.cs
@@ -1,0 +1,58 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Reflection;
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    /// <summary>
+    /// <see cref="FunctionContext"/> extension methods for metadata.
+    /// </summary>
+    internal static class FunctionContextMetadataExtensions
+    {
+        /// <summary>
+        /// Gets the <see cref="MethodInfo"/> of the function entry point.
+        /// </summary>
+        /// <param name="context"></param>
+        /// <returns></returns>
+        public static MethodInfo? GetMethodInfo(this FunctionContext context)
+        {
+            // note:
+            //  - Use caching strategy for MethodInfo if you want to speed up.
+            //  - Not cached at this time because of limited use.
+
+            var assembly = findAssembly(context.FunctionDefinition.PathToAssembly.AsSpan());
+            if (assembly is null)
+                return null;
+
+            var entryPoint = context.FunctionDefinition.EntryPoint;
+            return findMethod(assembly, entryPoint.AsSpan());
+
+
+            #region Local Functions
+            static Assembly? findAssembly(ReadOnlySpan<char> path)
+            {
+                foreach (var assembly in AppDomain.CurrentDomain.GetAssemblies())
+                {
+                    var location = assembly.Location.AsSpan();
+                    if (location.Equals(path, StringComparison.Ordinal))
+                        return assembly;
+                }
+                return null;
+            }
+
+
+            static MethodInfo? findMethod(Assembly assembly, ReadOnlySpan<char> entryPoint)
+            {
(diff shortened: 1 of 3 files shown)


xin9le:

xin9le:
@fabiocav
Thank you for your response, and I apologize for the slight delay.

I've added 2 points **'Typical scenario'** and **'Implementation approach'** to the details of the PR. Could you please check it out?

kshyju:
I am not in favor of using a middleware as an attribute on the function. It seems wrong and might cause further confusion. 

But I like the idea of supporting something similar to `ActionFilterAttribute`.  Feel free to open an issue and we can discuss/define details of that requirement and go from there.

xin9le:
I have opened an Issue #1949 to discuss this matter. I would appreciate it if we could discuss it there.

fabiocav:
Closing based on the discussion in #1949

PR #1843, opened 2023-08-17
App builder extension method to register extension options

### LINKED ISSUE #1826: [FSharp] Options not configured with new extension versions that use WorkerExtensionStartup
# Summary

After updating **Microsoft.Azure.Functions.Worker.Sdk** from **1.11.0** to **1.14.0**, functions written in F# started throwing exceptions about parameter binding. This affects at least the CosmosDB and Blob bindings. Most likely the problem is related to other bindings as well. The message is always in the form

Error converting 1 input parameters for Function '\<func name>': Cannot convert input parameter '\<name of the input>' to type \<any provided type> from type **'Microsoft.Azure.Functions.Worker.Grpc.Messages.GrpcModelBindingData'. Error:System.ArgumentNullException: Value cannot be null. (Parameter 'ConnectionName')**.

image

This problem does not occur with functions written in C#.

# Steps to reproduce
- [Use this repo](https://github.com/anielii/azure-functions-dotnet-worker-issue/tree/master)
- Downgrade **Microsoft.Azure.Functions.Worker.Sdk** to **1.11.0** to check that the project works fine with the older version.

liliankasem:

Related issue: #1826

### Pull request checklist

### DIFF
--- a/extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBBindingOptionsSetup.cs
+++ b/extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBBindingOptionsSetup.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using Microsoft.Azure.Cosmos;
 using Microsoft.Azure.Functions.Worker.Extensions;
 using Microsoft.Azure.Functions.Worker.Extensions.CosmosDB;
 using Microsoft.Extensions.Azure;
@@ -15,10 +16,18 @@ namespace Microsoft.Azure.Functions.Worker
         private readonly IConfiguration _configuration;
         private readonly AzureComponentFactory _componentFactory;
 
-        public CosmosDBBindingOptionsSetup(IConfiguration configuration, AzureComponentFactory componentFactory)
+        public CosmosDBBindingOptionsSetup(IOptions<WorkerOptions> workerOptions, IOptions<CosmosDBBindingOptions> cosmosOptions, IConfiguration configuration, AzureComponentFactory componentFactory)
         {
             _configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
             _componentFactory = componentFactory ?? throw new ArgumentNullException(nameof(componentFactory));
+
+            if (workerOptions is null)
+            {
+                throw new ArgumentNullException(nameof(workerOptions));
+            }
+
+            CosmosSerializer cosmosSerializer = new WorkerCosmosSerializer(workerOptions.Value?.Serializer);
+            cosmosOptions.Value.Serializer = cosmosSerializer;
         }
 
         public void Configure(CosmosDBBindingOptions options)
--- a/extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBBindingOptionsSetup.cs
+++ b/extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBBindingOptionsSetup.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
-using Microsoft.Azure.Cosmos;
 using Microsoft.Azure.Functions.Worker.Extensions;
 using Microsoft.Azure.Functions.Worker.Extensions.CosmosDB;
 using Microsoft.Extensions.Azure;
@@ -15,19 +14,13 @@ namespace Microsoft.Azure.Functions.Worker
     {
         private readonly IConfiguration _configuration;
         private readonly AzureComponentFactory _componentFactory;
+        private readonly IOptions<WorkerOptions> _workerOptions;
 
-        public CosmosDBBindingOptionsSetup(IOptions<WorkerOptions> workerOptions, IOptions<CosmosDBBindingOptions> cosmosOptions, IConfiguration configuration, AzureComponentFactory componentFactory)
+        public CosmosDBBindingOptionsSetup(IConfiguration configuration, AzureComponentFactory componentFactory, IOptions<WorkerOptions> workerOptions, )
         {
             _configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
             _componentFactory = componentFactory ?? throw new ArgumentNullException(nameof(componentFactory));
-
-            if (workerOptions is null)
-            {
-                throw new ArgumentNullException(nameof(workerOptions));
-            }
-
-            CosmosSerializer cosmosSerializer = new WorkerCosmosSerializer(workerOptions.Value?.Serializer);
-            cosmosOptions.Value.Serializer = cosmosSerializer;
+            _workerOptions = workerOptions ?? throw new ArgumentNullException(nameof(workerOptions));
         }
 
         public void Configure(CosmosDBBindingOptions options)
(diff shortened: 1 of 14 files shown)


extensions/Worker.Extensions.CosmosDB/src/Extensions/FunctionsWorkerApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+            if (!IsFSharp(Assembly.GetCallingAssembly()))
+            {
+                Console.WriteLine("FSharp not detected detected. Skipping CosmosDB extension configuration.");
+                return builder;
+                // Should we throw instead?
+                // throw new InvalidOperationException("The 'ConfigureCosmosDB' extension method is only for FSharp projects, please remove the call to 'ConfigureCosmosDB' from your startup code.");
+                // LOG: dotnet exited with code 134 (0x86). Unhandled exception. System.InvalidOperationException: The 'ConfigureCosmosDB' extension method is only for FSharp projects, please remove the call to 'ConfigureCosmosDB' from your startup code.
+            }
+
+            var cosmosdbStartup = new CosmosExtensionStartup();
```
jviau, line 41: We can avoid instantiating this if we flip the implementation. The static extension method contains all the logic, and the startup class calls the extension method.

extensions/Worker.Extensions.CosmosDB/src/Extensions/FunctionsWorkerApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                throw new System.ArgumentNullException(nameof(builder));
+            }
+
+            // TODO: How to handle scenarios where this method is used by a C# function?
+            // Avoid double calling?
+            // Could we write an analyzer to detect this?
+
+            // Tested and this is working. Almost all F# projects have a hidden reference to the FSharp.Core assembly
+            if (!IsFSharp(Assembly.GetCallingAssembly()))
```
jviau, line 32: Instead of checking for F#, there are several ways to make this safe to call redundantly (and it is great practice to do so for bootstrap code!)

First evaluate the impact of running multiple times. When I look at `CosmosExtensionStartup` we can break down what happens:

1. Calls `AddAzureClientsCore()`. This is safe to call multiple times, it uses `TryAdd*` methods. So repeated calls are no-op.
2. Options configuration. Redundant calls _will_ cause these to be re-registered, but what is the impact of that? The implementations are safe to call multiple times. So, this is just sub-optimal. We can probably live with that.

If you _really_ want to solve the second part, you can use `TryAddEnumerable` overload and move the `IOptions<WorkerOptions>` binding logic into `CosmosDBBindingOptionsSetup`.

liliankasem: Thanks for the insight! Does this look right? https://github.com/Azure/azure-functions-dotnet-worker/pull/1843/commits/86a04513d541c40ce27027482fe05aa7523b4c6e

samples/FSharpSample/FSharpSample.fsproj
```diff
@@ -0,0 +1,32 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
liliankasem, line 1: We don't need to check this sample in

liliankasem:
@mattchenderson we will need to update the docs to make sure folks are aware that if they use Cosmos, Tables or Blob extensions, they will need to register via the ext method:

```fsharp
module Program

open Microsoft.Extensions.Hosting
open Microsoft.Azure.Functions.Worker

module Program =
    [<EntryPoint>]
    let main _ =
        let hostBuilder = new HostBuilder()

        // FSharp applications need to configure the Cosmos, Blob and/or Table extensions manually.
        hostBuilder.ConfigureFunctionsWorkerDefaults(fun (context: HostBuilderContext) (appBuilder: IFunctionsWorkerApplicationBuilder) ->
            appBuilder.ConfigureCosmosDB() |> ignore
            appBuilder.ConfigureBlobStorage() |> ignore
            appBuilder.ConfigureTables() |> ignore
        ) |> ignore

        let host = hostBuilder.Build()
        host.Run()

        0
```

Probably calls for a templates update as well

extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBBindingOptionsSetup.cs
```diff
@@ -15,10 +16,18 @@ internal class CosmosDBBindingOptionsSetup : IConfigureNamedOptions<CosmosDBBind
         private readonly IConfiguration _configuration;
         private readonly AzureComponentFactory _componentFactory;
 
-        public CosmosDBBindingOptionsSetup(IConfiguration configuration, AzureComponentFactory componentFactory)
+        public CosmosDBBindingOptionsSetup(IOptions<WorkerOptions> workerOptions, IOptions<CosmosDBBindingOptions> cosmosOptions, IConfiguration configuration, AzureComponentFactory componentFactory)
```
jviau, line 19:
```suggestion
        public CosmosDBBindingOptionsSetup(IOptions<WorkerOptions> workerOptions, IConfiguration configuration, AzureComponentFactory componentFactory)
```

This type is configuring `CosmosDBBindingOptions`, it cannot import it. Save `IOptions<WorkerOptions>` to a field and set the serializer as part of the `Configure` call.

jviau (changes requested):

extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBBindingOptionsSetup.cs
liliankasem: That's the whole point of this class, how did I miss that 🤦🏻‍♀️ 🤦🏻‍♀️

extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBBindingOptionsSetup.cs
```diff
@@ -14,11 +14,13 @@ internal class CosmosDBBindingOptionsSetup : IConfigureNamedOptions<CosmosDBBind
     {
         private readonly IConfiguration _configuration;
         private readonly AzureComponentFactory _componentFactory;
+        private readonly IOptions<WorkerOptions> _workerOptions;
 
-        public CosmosDBBindingOptionsSetup(IConfiguration configuration, AzureComponentFactory componentFactory)
+        public CosmosDBBindingOptionsSetup(IConfiguration configuration, AzureComponentFactory componentFactory, IOptions<WorkerOptions> workerOptions, )
```
jviau, line 19:
```suggestion
        public CosmosDBBindingOptionsSetup(IConfiguration configuration, AzureComponentFactory componentFactory, IOptions<WorkerOptions> workerOptions)
```

You have an extra comma

jviau, line 19: This might not really have any impact, but just for accuracy:

Since this is being registered as `singleton`, it would be better to import `IOptionsMonitor<WorkerOptions>` and access it via `.CurrentValue`. I say this has no real impact because when you get down to it, we only access the `ObjectSerializer` and that is unlikely to change over the lifetime of this process. But it is good habit to ensure that options reacts to changes of config and other options.

jviau (approved):

PR #1847, opened 2023-08-21
Adding "global::" namespace prefix to avoid namespace conflict inside generated executor

### LINKED ISSUE #1816: Namespace conflict inside generated executor
Let's say we have a function like below

```csharp
[Function(nameof(Function))]
public void Run([QueueTrigger("myqueue-items", Connection = "Pi")] QueueMessage message)
{
    _logger.LogInformation($"C# Queue trigger function processed: {message.MessageText}");
}
```
The `QueueMessage` type is defined in the namespace `Azure.Storage.Queues.Models` which is inside `Azure.Storage.Queues.dll` assembly.

https://learn.microsoft.com/en-us/dotnet/api/azure.storage.queues.models.queuemessage?view=azure-dotnet

When we enable the source generation for the function executor by adding this build property
```xml
```

It will generate the below class. Note that the class is generated inside `Microsoft.Azure.Functions.Worker` namespace.

```csharp
{
    {
        private readonly IFunctionActivator _functionActivator;
        private readonly Dictionary<string, Type> types = new()
        {
            { "SrcGenNamespaceConflict.Function", Type.GetType("SrcGenNamespaceConflict.Function")! }
        };

        public DirectFunctionExecutor(IFunctionActivator functionActivator)
        {
            _functionActivator = functionActivator ?? throw new ArgumentNullException(nameof(functionActivator));
        }

        {
            var inputBindingFeature = context.Features.Get<IFunctionInputBindingFeature>()!;
            var inputBindingResult = await inputBindingFeature.BindFunctionInputAsync(context)!;
            var inputArguments = inputBindingResult.Values;

            if (string.Equals(context.FunctionDefinition.EntryPoint, "SrcGenNamespaceConflict.Function.Run", StringComparison.OrdinalIgnoreCase))
            {
                var instanceType = types["SrcGenNamespaceConflict.Function"];
                var i = _functionActivator.CreateInstance(instanceType, context) as SrcGenNamespaceConflict.Function;
                i.Run((Azure.Storage.Queues.Models.QueueMessage)inputArguments[0]);
            }
        }
    }
}
```

We use the fully qualified name of the type `Azure.Storage.Queues.Models.QueueMessage` (inside the Run method call). But this creates a compile error.
> The type or namespace name 'Storage' does not exist in the namespace 'Microsoft.Azure'

This is because the "Azure" part is interpreted as our namespace. So it is showing our types (from the worker is showing) in intellisense.

image

One solution I can think of is, instead of using `Microsoft.Azure.Functions.Worker` namespace, we could consider using some other namespace (Use the namespace where the main is defined in or the namespace of one of the function classes?)

kshyju:
Fixes #1816 

### Pull request checklist

### Additional information

Additional PR information


sdk/Sdk.Generators/FunctionExecutor/FunctionExecutorGenerator.Parser.cs
```diff
@@ -48,7 +48,8 @@ internal ICollection<ExecutableFunction> GetFunctions(List<MethodDeclarationSynt
                             continue;
                         }
 
-                        methodParameterList.Add(parameterSymbol.Type.ToDisplayString());
+                        var fullyQualifiedTypeName = parameterSymbol.Type.ToDisplayString(SymbolDisplayFormat.FullyQualifiedFormat);
```
kshyju, line 51: > Formats the names of all types and namespaces in a fully qualified style (including the global alias).

https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.symboldisplayformat.fullyqualifiedformat?view=roslyn-dotnet-4.6.0

jviau (approved):

jviau:
This is an aside to the problem at hand, but I wonder if we should use `RootNamespace` from the project for this files namespace?

liliankasem (approved):

kshyju:
> This is an aside to the problem at hand, but I wonder if we should use `RootNamespace` from the project for this files namespace?

Yes. some thing we can consider for all src generators. IIRC, I tried that route and found out that root namespace is not always a non null value. Let me see I can go back and check my repro and we can sync on this.

kshyju:
The fix is released in 1.14.1

https://github.com/Azure/azure-functions-dotnet-worker/releases/tag/sdk-1.14.1

PR #1861, opened 2023-08-24
Refactor SDK targets to address multiple writes and flaky builds

### LINKED ISSUE #1834: Unable to publish an Isolated Azure Function
This is related to https://github.com/dotnet/sdk/issues/30321

An internal team (represented by @gangaken here) is publishing an Isolated Azure Function and experiencing problems during the publish. I can't upload full details here, but Gangadhar and myself are both available on Teams internally.

The main symptom of the failure is that during the publish of their single functions app, the publish of the 'internal' WorkerExtensions.csproj project fails during the `_GenerateFunctionsAndCopyContentFiles` target while `Move`ing  the runtime assets of the WorkerExtensions project to the `publishout/bin` directory. According to the binlog, all of the contents of the `publishout/bin` directory are DoubleWrites - this to me points to a potential source of the problem. Looking at the binlog for one of these DoubleWrite files (for example System.Security.Cryptography.ProtectedData.dll), I see two copies:

* One during the `_CopyResolveDFilesToPublishPreserverNewest` Task from the .NET SDK, whose role is to copy publishable outputs to the publish directory
* One during the `_GenerateFunctionsAndCopyContentFiles` Task from this project.

That's about as far as I've gotten.

jviau:

resolves #1834

### Pull request checklist

### Additional information

This PR is a near complete overhaul of our SDK targets. The aim of the changes is to integrate more closely with existing dotnet SDK targets to eliminate redundant work.

1. Most `Copy` tasks have been removed and instead we augment specific item groups so that `CopyToOutputDirectory` and `CopyToPublishDirectory` tasks will copy our files for us.
2. Incremental build support has been improved, we now use `Inputs` and `Outputs`.
  - Some incremental build optimizations are thwarted by certain targets from the extension inner build. For example, `extensions.json` is always written to, so we always see it as changed and also write to it.
3. Targets have been shifted to perform all incremental steps into the `IncrementalOutputPath`. Only copying the final files in one go as part of the existing `CopyToDirectoryDirectory` (or publish dir).
4. For `publish`, we never actually call `publish` on the inner `WorkerExtensions.csproj`. Instead, we can rely on the existing build targets and collecting of files to accomplish the same thing more efficiently.

# Concerns

~This is a breaking change for anyone that has customized our targets. Hence the draft state. I am going to go through this and see if I can rename targets back to existing ones to minimize or eliminate the breaking change.~ See comment below

### DIFF
--- a/DotNetWorker.sln
+++ b/DotNetWorker.sln
@@ -134,6 +134,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Worker.Extensions.Tests", "
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "AspNetIntegration", "samples\AspNetIntegration\AspNetIntegration.csproj", "{D2F67410-9933-42E8-B04A-E17634D83A30}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Net7Worker", "samples\Net7Worker\Net7Worker.csproj", "{EC1A321B-70F6-420B-85D4-56C7869BB71B}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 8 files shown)


jviau:
Actually, after reviewing the changes technically this is not breaking. The convention in MSBuild is any property, target, or item starting with `_` is considered private. All of our targets start with `_` so nothing public is changing.

liliankasem (approved):

brettsam (approved):
Reviewing these msbuild targets is tough! I think my only comment is that I'd like to see some comments sprinkled throughout the new Targets explaining what they're doing so it's easier for other folks to find the right place to update in the future.

Issue #1651, opened 2023-06-21
NullReferenceException thrown when query string value is missing


paulbatum:
Start with the following working function definition:

```csharp
        [Function("helloworld")]
        public HttpResponseData HelloWorld(
            [HttpTrigger(AuthorizationLevel.Anonymous, "get", "post")] HttpRequestData req)
        {
            _logger.LogInformation("C# HTTP trigger function processed a request.");
            var response = req.CreateResponse(HttpStatusCode.OK);
            return response;
        }
```

Now, add a parameter that will come from the query string, with a default value:
```csharp
        [Function("helloworld")]
        public HttpResponseData HelloWorld(
            [HttpTrigger(AuthorizationLevel.Anonymous, "get", "post")] HttpRequestData req,
            int foo=100)
        {
            _logger.LogInformation("C# HTTP trigger function processed a request.");
            var response = req.CreateResponse(HttpStatusCode.OK);
            return response;
        }
```

Next, invoke the function but do NOT pass the query string param:

GET http://localhost:7157/api/helloworld

Expected behavior: The default value is used. Or, failing that, at least provide an error message that is more meaningful to the user, saying that this scenario is not supported. 

Actual behavior: a NullReferenceException is thrown, and the error message/stack do not make the problem obvious:

```
[2023-06-21T00:11:02.831Z] Function 'helloworld', Invocation id '32806cc9-5bbc-44cd-82af-9834fb926f35': An exception was thrown by the invocation.
[2023-06-21T00:11:02.832Z] Result: Function 'helloworld', Invocation id '32806cc9-5bbc-44cd-82af-9834fb926f35': An exception was thrown by the invocation.
Exception: System.NullReferenceException: Object reference not set to an instance of an object.
[2023-06-21T00:11:02.833Z]    at lambda_method1(Closure , Function1 , Object[] )
[2023-06-21T00:11:02.834Z]    at Microsoft.Azure.Functions.Worker.Invocation.MethodInvokerWithReturnValue`2.InvokeAsync(TReflected instance, Object[] arguments) in D:\a\_work\1\s\src\DotNetWorker.Core\Invocation\MethodInvokerWithReturnValue.cs:line 22
[...]

PR #1868, opened 2023-08-29
Adding optional parameter support

kshyju:
Adding optional parameter support.

With this change, when a function has an optional parameter with a default value defined, and no corresponding input data is available for this parameter during invocation, we will use the default value of the parameter.

This will fix issues such as #1651 


liliankasem (approved):

src/DotNetWorker.Core/Definition/FunctionParameter.cs
```diff
@@ -22,6 +22,16 @@ public FunctionParameter(string name, Type type)
         {
         }
 
+        /// <summary>
+        /// Creates an instance of the <see cref="FunctionParameter"/> class.
+        /// </summary>
+        /// <param name="name">The parameter name.</param>
+        /// <param name="type">The <see cref="System.Type"/> of the parameter.</param>
+        /// <param name="defaultValue">Default value of the parameter.</param>
+        public FunctionParameter(string name, Type type, object? defaultValue): this(name, type, defaultValue, ImmutableDictionary<string, object>.Empty)
```
fabiocav, line 31: nit: follow the existing formatting. Change to:
``` csharp
public FunctionParameter(string name, Type type, object? defaultValue) 
    : this(name, type, defaultValue, ImmutableDictionary<string, object>.Empty)
```

kshyju: Fixed formatting as suggested.

src/DotNetWorker.Core/Context/Features/DefaultFunctionInputBindingFeature.cs
```diff
@@ -97,6 +100,21 @@ public async ValueTask<FunctionInputBindingResult> BindFunctionInputAsync(Functi
                         errors.Add(
                             $"Cannot convert input parameter '{param.Name}' to type '{param.Type.FullName}' from type '{source.GetType().FullName}'. Error:{bindingResult.Error}");
                     }
+                    else if (bindingResult.Status == ConversionStatus.Unhandled)
+                    {
+                        // If still unhandled after going through all converters,check an explicit default value was provided for the parameter.
+                        if (param.DefaultValue is not null)
```
jviau, line 106: This looks like it won't support `null` default values, ie: `MyFunction(string str = null)`. Is this intentional?

satvu (approved):

src/DotNetWorker.Core/Context/Features/DefaultFunctionInputBindingFeature.cs
kshyju: Yea, the original plan was to throw an error in that case. But I realize it will be confusing, I pushed a new iteration which handles this case. So if you have "string foo = null" , it will be populated as null if we cannot populate any values.

satvu (approved):

PR #1882, opened 2023-09-01
[Source Gen] Add diagnostic logs for errors when parsing attribute arguments

### LINKED ISSUE #1878: [Source Gen] Add diagnostic logs for errors when parsing attribute arguments
Follow up to https://github.com/Azure/azure-functions-dotnet-worker/pull/1877 - need to add diagnostic logs for these scenarios.

satvu:

resolves #1878

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/docs/analyzer-rules/AZFW0013.md
+++ b/docs/analyzer-rules/AZFW0013.md
@@ -1,0 +1,25 @@
+# AZFW0013: Unable to parse binding argument
+
+| | Value |
+|-|-|
+| **Rule ID** |AZFW0013|
+| **Category** |[AzureFunctionsSyntax]|
+| **Severity** |Error|
+
+## Cause
+
+This rule is triggered when a binding attribute argument is an invalid or null value.
+
+## Rule description
+
+[Attributes](https://learn.microsoft.com/en-us/dotnet/csharp/advanced-topics/reflection-and-attributes/) are used to define bindings, and the function metadata generator parses the arguments passed into these attributes to generate binding information during start up.
+
+If the arguments passed in are invalid for any reason (such as being null), this rule is enforced.
+
+## How to fix violations
+
+Review the binding attribute.
+
+## When to suppress warnings
+
+This rule should not be suppressed because this error will prevent your functions from running.
--- a/sdk/Sdk.Generators/DiagnosticDescriptors.cs
+++ b/sdk/Sdk.Generators/DiagnosticDescriptors.cs
@@ -62,5 +62,12 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
                     messageFormat: "Invalid use of a retry attribute. Check that the attribute is used on a trigger that supports function-level retry.",
                     category: "FunctionMetadataGeneration",
                     severity: DiagnosticSeverity.Error);
+        public static DiagnosticDescriptor InvalidBindingAttributeArgument { get; }
+                    = Create(id: "AZFW0013",
+                    title: "Invalid argument in binding attribute.",
+                    messageFormat: "Invalid argument passed in binding attribute. Check that the argument is not null.",
+                    category: "FunctionMetadataGeneration",
+                    severity: DiagnosticSeverity.Error);
+
     }
 }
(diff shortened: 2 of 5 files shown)


sdk/Sdk.Generators/DiagnosticDescriptors.cs
```diff
@@ -62,5 +62,12 @@ private static DiagnosticDescriptor Create(string id, string title, string messa
                     messageFormat: "Invalid use of a retry attribute. Check that the attribute is used on a trigger that supports function-level retry.",
                     category: "FunctionMetadataGeneration",
                     severity: DiagnosticSeverity.Error);
+        public static DiagnosticDescriptor InvalidBindingAttributeArgument { get; }
```
satvu, line 65: Named arguments can be null ([example](https://github.com/Azure/azure-functions-dotnet-worker/blob/b4e4ba87f09ce6e7d157d9ed02661bc9db4f4d52/sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs#L700)).

This error deals specifically when an arg can't be parsed using [this method](https://github.com/Azure/azure-functions-dotnet-worker/blob/b4e4ba87f09ce6e7d157d9ed02661bc9db4f4d52/sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs#L700). Possibilities are that arg is not of a supported type, arg is null, or it's an enum that couldn't be parsed.

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -632,7 +632,8 @@ private bool TryGetAttributeProperties(AttributeData attributeData, Location? at
                             }
                             else
                             {
-                                // TODO: Log diagnostic error
+                                _context.ReportDiagnostic(Diagnostic.Create(DiagnosticDescriptors.InvalidBindingAttributeArgument, attribLocation));
```
kshyju, line 635: So is this preventing user to explicitly specify null value for a property when using a binding? Is that something really we should throw an error for?

kshyju: Also synced offline. If we don't think we can think of a scenario where this will happen, probably we should not add this change. The less code we ship, the better.

jviau (approved):
I don't really see any harm in adding a diagnostic message for an edge case. It is better than having it silently no-op. Now if it was more logic than just a diagnostic message I would reconsider - but it isn't.

docs/analyzer-rules/AZFW0013.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Cause
+
+This rule is triggered when a binding attribute argument is an invalid or null value.
+
+## Rule description
+
+[Attributes](https://learn.microsoft.com/en-us/dotnet/csharp/advanced-topics/reflection-and-attributes/) are used to define bindings, and the function metadata generator parses the arguments passed into these attributes to generate binding information during start up.
+
+If the arguments passed in are invalid for any reason (such as being null), this rule is enforced.
```
liliankasem, line 17: What else would be considered an invalid value?

liliankasem (approved):

docs/analyzer-rules/AZFW0013.md
satvu: Anything that can't be parsed [here](https://github.com/Azure/azure-functions-dotnet-worker/blob/b4e4ba87f09ce6e7d157d9ed02661bc9db4f4d52/sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs#L700). Shyju and I tried to think through a scenario where this could happen and couldn't come up with one, since you'll get a compiler error before even getting this diagnostic log.

PR #1896, opened 2023-09-11
[Source Gen] Refactor to walk dependent assemblies

### LINKED ISSUE #1160: Generate function metadata for functions from dependent assemblies
# Overview

There is currently a gap with function metadata generator and the existing `function.metadata` approach with functions from dependent assemblies. The `function.metadata` approach will include functions from referenced projects/assemblies, whereas the source generator will not. To achieve behavior parity between the two methods, we need to generate function metadata for referenced assemblies as well. We do not want to mix host-provided metadata from `function.metadata` and worker-provided metadata from the generated provider. As such, we will need to also include metadata for referenced assemblies in the generated approach.

# Requirements:

1. Function metadata is generated for functions from dependent assemblies that **do not** have their own "default" function metadata provider (see default vs auxiliary provider below).
2. This metadata is registered along with the metadata from the current assembly.

## Default vs Auxiliary Provider

Part of this work will require identifying the difference between a "default" provider and an "auxiliary" provider.

### Default Provider

- Expected to provide function metadata for all of the "traditional" functions in this assembly.
- Can include metadata for other assemblies as well.
- A given assembly should be covered by a maximum of 1 default provider.
    - `DefaultFunctionMetadataProvider` is a "default" provider which covers _all_ assemblies referenced by the project.
    - `GeneratedFunctionMetadataProvider` is a "default" provider which covers the current assembly (and after this work, referenced assemblies _which are not themselves covered by a "default" provider_).
    - Customers may be able to define their own "default" provider.
- We could have an analyzer which warns when an assembly has more than 1 default provider.
- See if we can do any runtime validation. However, this may not be possible as we are unsure what assemblies a default provider covers.

### Auxiliary provider

- An assembly can have any number of auxiliary providers.
- These are expected to provide function metadata in _addition_ to the "default" provider. As in, auxiliary providers _should not_ provide metadata for 'traditional' functions.

# Concerns

Ideally this work will be part of the existing source generator. However, I am unsure if we can walk referenced assemblies as part of a source generator. If this is not possible, we can go with an MSBuild task which generates and includes a .cs file at build time. This generated file from MSBuild can still be incorporated into the source generated file by making the type "partial" and combining the logic of the two together.

satvu:

resolves #1160 

### Pull request checklist


sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.cs
```diff
@@ -68,5 +74,62 @@ private static bool ShouldIncludeAutoGeneratedAttributes(GeneratorExecutionConte
...
+                }
+            }
+
+            return entryAssemblyFuncs.Count > 0 ? entryAssemblyFuncs : ImmutableList<IMethodSymbol>.Empty;
+        }
+
+        /// <summary>
+        /// Collect methods with Function attributes on them from dependent/referenced assemblies.
+        /// </summary>
+        private IList<IMethodSymbol> GetDependentAssemblyFunctions(GeneratorExecutionContext context)
```
jviau, line 99: nit, use `IEnumerable<IMethodSymbol>` and `yield return` - it will be more efficient as you already `.ToList()` the results.

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.cs
```diff
@@ -68,5 +74,62 @@ private static bool ShouldIncludeAutoGeneratedAttributes(GeneratorExecutionConte
 
             return string.Equals(value, bool.TrueString, System.StringComparison.OrdinalIgnoreCase);
         }
+
+        private IList<IMethodSymbol> GetEntryAssemblyFunctions(List<MethodDeclarationSyntax> candidateMethods, GeneratorExecutionContext context)
```
jviau, line 78: same here, `IEnumerable<IMethodSymbol>` and `yield return`.

sdk/Sdk.Generators/FunctionsUtil.cs
```diff
@@ -28,19 +29,43 @@ internal static class FunctionsUtil
...
             return false;
         }
 
+        internal static bool TryGetFunctionName(ISymbol symbol, Compilation compilation, out string? functionName) 
+        {
+            functionName = null;
+
+            var functionAttribute = symbol.GetAttributes().Where(a => SymbolEqualityComparer.Default.Equals(a.AttributeClass, compilation.GetTypeByMetadataName(Constants.Types.FunctionName)));
+            
+            if (functionAttribute.Count() > 0)
```
jviau, line 60: When using LINQ, use `.Any()` instead of `.Count() > 0`, as `.Count()` will need to expand the entire resulting enumerable. `Any()` will return soon as 1 element is yielded.

With that said, looks like your code can be further optimized to:

``` CSharp
var functionAttribute = symbol.GetAttributes()
    .Where(a => SymbolEqualityComparer.Default.Equals(a.AttributeClass, compilation.GetTypeByMetadataName(Constants.Types.FunctionName)))
    .FirstOrDefault();
if (functionAttribute is not null)
{
    functionName = (string)functionAttribute.ConstructorArguments.First().Value!;
    return true;
}

return false;
```

sdk/Sdk.Generators/FunctionsUtil.cs
```diff
@@ -53,5 +78,16 @@ internal static string GetFullyQualifiedMethodName(MethodDeclarationSyntax metho
...
+
+        internal static string GetFullyQualifiedMethodName(IMethodSymbol method)
+        {
+            var fullyQualifiedClassName = method.ContainingSymbol.ToDisplayString();
+            return $"{fullyQualifiedClassName}.{method.Name}";
+        }
+
+        internal static string GetFullyQualifiedMetadataName(IMethodSymbol method)
+        {
+            throw new NotImplementedException();
```
jviau, line 90: Can this be removed?

test/Sdk.Generator.Tests/FunctionMetadataProviderGeneratorTests/DependentAssemblyTest.cs
```diff
@@ -0,0 +1,138 @@
...
+                            Function1RawBindings.Add(@"{""name"":""req"",""type"":""HttpTrigger"",""direction"":""In"",""authLevel"":""Anonymous"",""methods"":[""get"",""post""]}");
+                            Function1RawBindings.Add(@"{""name"":""$return"",""type"":""http"",""direction"":""Out""}");
+                
+                            var Function1 = new DefaultFunctionMetadata
+                            {
+                                Language = "dotnet-isolated",
+                                Name = "DependencyFunc",
+                                EntryPoint = "DependentAssemblyWithFunctions.DependencyFunc.Run",
+                                RawBindings = Function1RawBindings,
+                                ScriptFile = "TestProject.dll"
```
jviau, line 104: Is `ScriptFile` supposed to be `TestProject.dll` here? Wondering if it should be `DependentAssemblyWithFunctions.dll`

Can you compare what it is from the JSON generator?

satvu: It is supposed to be `DependentAssemblyWithFunctions.dll` - fixing

test/DependentAssemblyWithFunctions/DependentAssemblyWithFunctions.csproj
```diff
@@ -0,0 +1,14 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net7.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.18.0" />
```
kshyju, line 10: Should we use 1.19.0 which is the latest stable?

jviau (approved):

test/DependentAssemblyWithFunctions/Class1.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            _logger.LogInformation("C# HTTP trigger function processed a request.");
+
+            var response = req.CreateResponse(HttpStatusCode.OK);
+            response.Headers.Add("Content-Type", "text/plain; charset=utf-8");
+
+            response.WriteString("Welcome to Azure Functions!");
+
+            return response;
+        }
```
kshyju, line 29: A function marked with `internal` visibility is ignored? Might be better to add that to this class and verify that in the test. 
Next time I have the same question, I can look at the tests to get an idea about the behavior.

kshyju: Also add a Static function.

sdk/Sdk.Generators/FunctionsUtil.cs
```diff
@@ -28,19 +29,45 @@ internal static class FunctionsUtil
...
             return false;
         }
 
+        internal static bool TryGetFunctionName(ISymbol symbol, Compilation compilation, out string? functionName) 
+        {
+            functionName = null;
+
+            var functionAttribute = symbol.GetAttributes()
+                .Where(a => SymbolEqualityComparer.Default.Equals(a.AttributeClass, compilation.GetTypeByMetadataName(Constants.Types.FunctionName)))
+                .FirstOrDefault();
```
kshyju, line 60: you can replace the Where with `FirstOrDefault` and remove the final call to `FirstOrDefault` at line 60

sdk/Sdk.Generators/FunctionsUtil.cs
```diff
@@ -53,5 +80,11 @@ internal static string GetFullyQualifiedMethodName(MethodDeclarationSyntax metho
 
             return $"{fullyQualifiedClassName}.{method.Identifier.ValueText}";
         }
+
+        internal static string GetFullyQualifiedMethodName(IMethodSymbol method)
```
kshyju, line 84: Should we update the `GetFullyQualifiedMethodName` method above to use this new one?

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -244,29 +240,22 @@ private bool TryGetRetryOptionsFromAttribute(AttributeData attribute, Location l
...
                 supportsRetryOptions = false;
                 hasHttpTrigger = false;
                 bindingsList = new List<IDictionary<string, object>>();
 
-                foreach (ParameterSyntax parameter in method.ParameterList.Parameters)
+                foreach (IParameterSymbol parameter in method.Parameters)
                 {
                     // If there's no attribute, we can assume that this parameter is not a binding
-                    if (parameter.AttributeLists.Count == 0)
+                    if (parameter.GetAttributes().Count() == 0)
```
kshyju, line 252:
```suggestion
                   if (!parameter.GetAttributes().Any())
```

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -276,12 +265,12 @@ private bool TryGetParameterInputAndTriggerBindings(MethodDeclarationSyntax meth
...
+                            DataType dataType = _dataTypeParser.GetDataType(parameter.Type);
 
                             bool cardinalityValidated = false;
                             bool supportsDeferredBinding = false;
 
-                            if (SupportsDeferredBinding(attribute, parameterSymbol.Type.ToString()))
+                            if (SupportsDeferredBinding(attribute, parameter.Type.ToString()))
                             {
                                 supportsDeferredBinding = true;
                             }
```
kshyju, line 276: Can be `bool supportsDeferredBinding = SupportsDeferredBinding(attribute, parameter.Type.ToString());` ?

sdk/Sdk.Generators/FunctionsUtil.cs
satvu: Looks like there's no more references to the original one, so removed.

test/DependentAssemblyWithFunctions/Class1.cs
satvu: Tested legacy and functions marked with internal are NOT ignored. Same behavior with source-gen, test added.

kshyju (approved):

Issue #1055, opened 2022-09-20
[AOT] Replace usage of IConfiguration.Bind


kshyju:
`IConfiguration.Bind` is not trimmer friendly.

See https://github.com/dotnet/runtime/discussions/59230

We use this method to bind to[ GrpcWorkerStartupOptions](https://github.com/Azure/azure-functions-dotnet-worker/blob/1da3c27d3b24e648a20356fb3180abfcdc7f6cff/src/DotNetWorker.Grpc/GrpcStartupOptions.cs#L7) type, which is to represent the data from the command line arguments sent from host

https://github.com/Azure/azure-functions-dotnet-worker/blob/1da3c27d3b24e648a20356fb3180abfcdc7f6cff/src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs#L92-L96

jviau:
Also relevant: https://github.com/dotnet/runtime/issues/44493

kshyju:
This issue is fixed with #1897

PR #1897, opened 2023-09-11
Handle new command line arguments with "functions-" prefix

### LINKED ISSUE #1547: Fix command line arguments for worker getting overridden
PR #583 changed the order or registering configuration sources so that environment variables will be registered before command line arguments. This change fixed issues where the environment variables were overriding the command line argument values (ex: port number for GRPC communication) and that causes worker failing to start.

Looks like this issue was impacting our Linux use cases.

https://github.com/Azure/azure-functions-host/issues/7555
https://github.com/Azure/azure-functions-dotnet-worker/issues/524

The fix was made to the worker code. But this does not prevent the issue completely because customer can add environment variables as a configuration source after the "ConfigureFunctionsWorkerDefaults" method call in the worker main code. 

```csharp
var host = new HostBuilder()
.ConfigureAppConfiguration(builder => builder.AddEnvironmentVariables())
.Build();
```

This code causes worker to fail when trying to GRPC handshake. (in Linux + VNet use case)

The workaround is to swap the order as below
```csharp
var host = new HostBuilder()
.ConfigureAppConfiguration(builder => builder.AddEnvironmentVariables())
.Build();
```
#1500 is an issue where customers ran into this situation. 

There were few ideas discussed in the hallway conversations, updating the worker code to check for command line arguments with FUNC prefix (Ex: `FUNCTIONS_GRPC_PORT`) and use that value if present, else fallback to existing implementation (using `PORT` argument). Another approach is to pass this information worker needs via environment arguments.

kshyju:
Part of #1900  
Resolves #1547 

This PR adds the relevant changes to read the new command line arguments we plan to send from the host.  See the host issue https://github.com/Azure/azure-functions-host/issues/9504 for understanding the need for this change.

Example of command line argument string host will send to worker process.
> -host localhost --port 80 --workerId testWorkerId --requestId testId --grpcMaxMessageLength 2147483647 --functions-uri http://localhost:80 --functions-workerid testWorkerId --functions-requestid testId --functions-grpcmaxmessagelength 2147483647

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/extensions/Worker.Extensions.Rpc/src/ConfigurationExtensions.cs
+++ b/extensions/Worker.Extensions.Rpc/src/ConfigurationExtensions.cs
@@ -32,7 +32,7 @@ namespace Microsoft.Azure.Functions.Worker.Extensions.Rpc
             }
             else
             {
-                string uriString = $"http://{configuration["HOST"]}:{configuration["PORT"]}";
+                var uriString = $"http://{configuration["HOST"]}:{configuration["PORT"]}";
                 if (!Uri.TryCreate(uriString, UriKind.Absolute, out grpcUri))
                 {
                     throw new InvalidOperationException($"The gRPC channel URI '{uriString}' could not be parsed.");
--- a/extensions/Worker.Extensions.Rpc/src/ConfigurationExtensions.cs
+++ b/extensions/Worker.Extensions.Rpc/src/ConfigurationExtensions.cs
@@ -6,10 +6,7 @@ using Microsoft.Extensions.Configuration;
 
 namespace Microsoft.Azure.Functions.Worker.Extensions.Rpc
 {
-    /// <summary>
-    /// contains extension methods for IConfiguration.
-    /// </summary>
-    public static class ConfigurationExtensions
+    internal static class ConfigurationExtensions
     {
         /// <summary>
         /// Gets the functions host gRPC address from <see cref="IConfiguration"/>.
--- a/extensions/Worker.Extensions.Rpc/src/ConfigurationExtensions.cs
+++ b/extensions/Worker.Extensions.Rpc/src/ConfigurationExtensions.cs
@@ -13,15 +13,28 @@ namespace Microsoft.Azure.Functions.Worker.Extensions.Rpc
         /// </summary>
         /// <param name="configuration">The configuration to get the host URI from.</param>
         /// <returns>The URI representing the functions host.</returns>
+        /// <exception cref="UriFormatException">If <paramref name="configuration"/> contain a value for 'functions-uri', but is not a valid URL.</exception>
         /// <exception cref="InvalidOperationException">
         /// If <paramref name="configuration"/> does not contain the required values.
         /// </exception>
         public static Uri GetFunctionsHostGrpcUri(this IConfiguration configuration)
         {
-            string uriString = $"http://{configuration["HOST"]}:{configuration["PORT"]}";
-            if (!Uri.TryCreate(uriString, UriKind.Absolute, out Uri? grpcUri))
+            Uri? grpcUri;
+            var functionsUri = configuration["functions-uri"];
+            if (functionsUri is not null)
             {
-                throw new InvalidOperationException($"The gRPC channel URI '{uriString}' could not be parsed.");
+                if (!Uri.TryCreate(functionsUri, UriKind.Absolute, out grpcUri))
+                {
+                    throw new UriFormatException($"'{functionsUri}' is not a valid value for 'functions-uri'. Value should be a valid URL.");
+                }
+            }
+            else
+            {
+                string uriString = $"http://{configuration["HOST"]}:{configuration["PORT"]}";
+                if (!Uri.TryCreate(uriString, UriKind.Absolute, out grpcUri))
+                {
+                    throw new InvalidOperationException($"The gRPC channel URI '{uriString}' could not be parsed.");
+                }
             }
 
             return grpcUri;
--- a/extensions/Worker.Extensions.Rpc/src/ConfigurationExtensions.cs
+++ b/extensions/Worker.Extensions.Rpc/src/ConfigurationExtensions.cs
@@ -6,26 +6,28 @@ using Microsoft.Extensions.Configuration;
 
 namespace Microsoft.Azure.Functions.Worker.Extensions.Rpc
 {
-    internal static class ConfigurationExtensions
+    /// <summary>
+    /// contains extension methods for IConfiguration.
+    /// </summary>
+    public static class ConfigurationExtensions
     {
         /// <summary>
         /// Gets the functions host gRPC address from <see cref="IConfiguration"/>.
         /// </summary>
         /// <param name="configuration">The configuration to get the host URI from.</param>
         /// <returns>The URI representing the functions host.</returns>
-        /// <exception cref="UriFormatException">If <paramref name="configuration"/> contain a value for 'functions-uri', but is not a valid URL.</exception>
         /// <exception cref="InvalidOperationException">
         /// If <paramref name="configuration"/> does not contain the required values.
         /// </exception>
         public static Uri GetFunctionsHostGrpcUri(this IConfiguration configuration)
         {
             Uri? grpcUri;
-            var functionsUri = configuration["functions-uri"];
+            var functionsUri = configuration["Functions:Worker:HostEndpoint"];
             if (functionsUri is not null)
             {
                 if (!Uri.TryCreate(functionsUri, UriKind.Absolute, out grpcUri))
                 {
-                    throw new UriFormatException($"'{functionsUri}' is not a valid value for 'functions-uri'. Value should be a valid URL.");
+                    throw new InvalidOperationException($"The gRPC channel URI '{functionsUri}' could not be parsed.");
                 }
             }
             else
--- a/src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
+++ b/src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
@@ -40,7 +40,7 @@
     we use the legacy Grpc.Core package instead, due to limitations with the current 
     Grpc.Net.Client implementation-->
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
-    <PackageReference Include="Grpc.Core" Version="2.46.6" />
+    <PackageReference Include="Grpc.Core" Version="2.45.0" />
     <Compile Remove="NativeHostIntegration/**" />
   </ItemGroup>
 
@@ -55,7 +55,6 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\extensions\Worker.Extensions.Abstractions\src\Worker.Extensions.Abstractions.csproj" />
-    <ProjectReference Include="..\..\extensions\Worker.Extensions.Rpc\src\Worker.Extensions.Rpc.csproj" />
     <ProjectReference Include="..\DotNetWorker.Core\DotNetWorker.Core.csproj" />
   </ItemGroup>
 
--- a/src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
+++ b/src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
@@ -40,7 +40,7 @@
     we use the legacy Grpc.Core package instead, due to limitations with the current 
     Grpc.Net.Client implementation-->
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
-    <PackageReference Include="Grpc.Core" Version="2.45.0" />
+    <PackageReference Include="Grpc.Core" Version="2.46.6" />
     <Compile Remove="NativeHostIntegration/**" />
   </ItemGroup>
 
@@ -55,6 +55,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\extensions\Worker.Extensions.Abstractions\src\Worker.Extensions.Abstractions.csproj" />
+    <ProjectReference Include="..\..\extensions\Worker.Extensions.Rpc\src\Worker.Extensions.Rpc.csproj" />
     <ProjectReference Include="..\DotNetWorker.Core\DotNetWorker.Core.csproj" />
   </ItemGroup>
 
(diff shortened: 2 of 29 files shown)


kshyju:
Relevant host change (still in draft because waiting for some lang workers to fix their packages to handle this): https://github.com/Azure/azure-functions-host/pull/9514/files

src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs
```diff
@@ -68,9 +68,14 @@ public static IServiceCollection AddGrpc(this IServiceCollection services)
...
 
             services.AddOptions<GrpcWorkerStartupOptions>()
-                .Configure<IConfiguration>((arguments, config) =>
+                .Configure<IConfiguration>((grpcWorkerStartupOption, config) =>
                 {
-                    config.Bind(arguments);
+                    grpcWorkerStartupOption.Host = config["host"];
+                    grpcWorkerStartupOption.Port = config.GetValue<int>("port");
+                    grpcWorkerStartupOption.Uri = config["functions-uri"];
+                    grpcWorkerStartupOption.RequestId = config["functions-requestid"] ?? config["requestId"];
```
jviau, line 76: An alternative is to use `switchMappings` when adding command line args. This way we can map `--functions-requestid` to be `RequestId` in `IConfiguration`, and so on.

https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Configuration.CommandLine/src/CommandLineConfigurationExtensions.cs#L155

jviau (commented):
While we are enhancing this, couple comments:

1. Technically command-line semantics is to start with `--`, then kebab-case the argument. So it would be `--functions-request-id` and not `--functions-requestid`.
2. Should we map these into a specific configuration section so they don't get accidentally overriden by something unrelated? Maybe everything under `Functions:Worker` section? (`Functions:Worker:HostUri`, `Functions:Worker:RequestId`, `Functions:Worker:WorkerId`)

src/DotNetWorker.Grpc/GrpcStartupOptions.cs
```diff
@@ -6,6 +6,12 @@ namespace Microsoft.Azure.Functions.Worker
 {
     internal class GrpcWorkerStartupOptions
     {
+        /// <summary>
+        /// Absolute URI of the grpc server including port.
+        /// Use this value to connect if not null, else use <see cref="Host"/> and <see cref="Port"/> to build the URI for connecting to the grpc server.
+        /// </summary>
+        public string? Uri { get; set; }
```
jviau, line 13: We should remove `Host` and `Port`. Update options configuration to combine them onto this property if the explicit uri is not present, but host & port are.

kshyju: We do use the `Hos`t and `Port` [in one place](https://github.com/Azure/azure-functions-dotnet-worker/blob/e358ca5b804dc282e1c98b48ab6097caca0fc0fb/src/DotNetWorker.Grpc/GrpcWorkerClientFactory.cs#L129) when we are creating an instance of `GrpcCore.Channel`. That class does not have a constructor which takes a full URI, but there is one which takes a target string. The constructor we are calling will build the target string and call the other constructor. The targetString is built by `string.Format("{0}:{1}", host, port)` and further down the line passed to a native API. I do not see a constructor which takes a full Uri.

We could convert our `string? Uri` to `Uri` type which will allow us to get the Host and Port.  Thoughts?

kshyju:
@jviau Pushed an update to switch to kebab case.

src/DotNetWorker.Grpc/GrpcStartupOptions.cs
jviau: Yeah that works - I believe there are even analyzers that will tell you to use `Uri` type over `string`.

kshyju: Added the Uri type property and removed Host and Port.

extensions/Worker.Extensions.Rpc/src/ConfigurationExtensions.cs
```diff
@@ -13,15 +13,28 @@ internal static class ConfigurationExtensions
...
+        /// <exception cref="UriFormatException">If <paramref name="configuration"/> contain a value for 'functions-uri', but is not a valid URL.</exception>
         /// <exception cref="InvalidOperationException">
         /// If <paramref name="configuration"/> does not contain the required values.
         /// </exception>
         public static Uri GetFunctionsHostGrpcUri(this IConfiguration configuration)
         {
-            string uriString = $"http://{configuration["HOST"]}:{configuration["PORT"]}";
-            if (!Uri.TryCreate(uriString, UriKind.Absolute, out Uri? grpcUri))
+            Uri? grpcUri;
+            var functionsUri = configuration["functions-uri"];
```
jviau, line 23: As mentioned in my other comment, I think we should use command line switch mapping, so `--functions-uri` maps to `FunctionsUri`* in `IConfiguration`.

Additionally, it may be worth considering moving this all under a config section to reduce risk of collisions - but this can be done in a separate PR if you want.

fabiocav: I would be in favor of making any configuration adjustments in this PR to avoid changing this too much just in case anyone ends up using it. Moving this to a configuration section makes sense to me.

extensions/Worker.Extensions.Rpc/src/ConfigurationExtensions.cs
```diff
@@ -13,15 +13,28 @@ internal static class ConfigurationExtensions
...
+            if (functionsUri is not null)
             {
-                throw new InvalidOperationException($"The gRPC channel URI '{uriString}' could not be parsed.");
+                if (!Uri.TryCreate(functionsUri, UriKind.Absolute, out grpcUri))
+                {
+                    throw new UriFormatException($"'{functionsUri}' is not a valid value for 'functions-uri'. Value should be a valid URL.");
```
fabiocav, line 28: If we do map the switch to another configuration name, this validation would need to occur in a different location, as this code wouldn't know anything about the switch name. We need to use the configuration key here instead.

src/DotNetWorker.Grpc/GrpcStartupOptions.cs
```diff
@@ -2,13 +2,13 @@
...
+using System;
+
 namespace Microsoft.Azure.Functions.Worker
 {
     internal class GrpcWorkerStartupOptions
     {
-        public string? Host { get; set; }
-
-        public int Port { get; set; }
+        public Uri? Uri { get; set; }
```
fabiocav, line 11: nit: name the property `HostEndpoint`

src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs
```diff
@@ -68,9 +67,30 @@ public static IServiceCollection AddGrpc(this IServiceCollection services)
 #endif
 
             services.AddOptions<GrpcWorkerStartupOptions>()
-                .Configure<IConfiguration>((arguments, config) =>
+                .Configure<IConfiguration>((grpcWorkerStartupOption, config) =>
                 {
-                    config.Bind(arguments);
+                    Uri? grpcUri;
+                    var functionsUri = config["functions-uri"];
```
fabiocav, line 73: Let's avoid the code duplication here and reuse the logic used int he `Worker.Extensions.Grpc` project

src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
```diff
@@ -40,7 +40,7 @@
     we use the legacy Grpc.Core package instead, due to limitations with the current 
     Grpc.Net.Client implementation-->
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
-    <PackageReference Include="Grpc.Core" Version="2.45.0" />
+    <PackageReference Include="Grpc.Core" Version="2.46.6" />
```
kshyju, line 43: Even though a project reference to "**Worker.Extensions.Rpc.csproj**" was added to this project (line 58 below), we still need to keep these explicit package references for Grpc packages (Grpc.Core, Grpc.Net.Client, Grpc.Net.ClientFactory) because DotNetWorker.Grpc project still have net5 support where Worker.Extensions.Rpc does not.

kshyju:
@fabiocav @jviau This is ready for another round of review.

src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs
```diff
@@ -68,9 +68,12 @@ public static IServiceCollection AddGrpc(this IServiceCollection services)
 #endif
 
             services.AddOptions<GrpcWorkerStartupOptions>()
-                .Configure<IConfiguration>((arguments, config) =>
+                .Configure<IConfiguration>((grpcWorkerStartupOption, config) =>
                 {
-                    config.Bind(arguments);
+                    grpcWorkerStartupOption.HostEndpoint = config.GetFunctionsHostGrpcUri();
+                    grpcWorkerStartupOption.RequestId = config["Functions:Worker:RequestId"] ?? config["requestId"];
+                    grpcWorkerStartupOption.WorkerId = config["Functions:Worker:WorkerId"] ?? config["workerId"];
```
kshyju, line 75: We could consider removing the fallback code after a few host releases goes out. Opened an issue to track that.

[Evaluate removing the fallback command line argument reading code(HOST, PORT)](https://github.com/Azure/azure-functions-dotnet-worker/issues/1908)

src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
```diff
@@ -55,6 +55,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\extensions\Worker.Extensions.Abstractions\src\Worker.Extensions.Abstractions.csproj" />
+    <ProjectReference Include="..\..\extensions\Worker.Extensions.Rpc\src\Worker.Extensions.Rpc.csproj" />
```
jviau, line 58: hmm I am not entirely sold on having this reference. At least not yet, the RPC package is still preview and has not yet been published to nuget.

Even then, the goal of the Extensions.Rpc project is to assist other extensions with using custom RPC back to the host. If we do want DotnetWorker.Grpc to use this - which it could - it would be a larger effort to have Extensions.Rpc to have a public API surface that meets its needs better.

extensions/Worker.Extensions.Rpc/src/ConfigurationExtensions.cs
```diff
@@ -6,7 +6,10 @@
 
 namespace Microsoft.Azure.Functions.Worker.Extensions.Rpc
 {
-    internal static class ConfigurationExtensions
+    /// <summary>
+    /// contains extension methods for IConfiguration.
+    /// </summary>
+    public static class ConfigurationExtensions
```
jviau, line 12: I'd rather we keep this internal. It isn't the worst thing to duplicate in `DotNetWorker.Grpc`. We can look into better ways to make it DRY later.

jviau (commented):
Looks great! Only concern is the reference to Extensions.Rpc. I suggest we just duplicate code for now. We can follow up later and see if there is an opportunity to make Extensions.Rpc appropriate for DotNetWorker.Grpc to leverage.

src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs
```diff
@@ -68,9 +68,12 @@ public static IServiceCollection AddGrpc(this IServiceCollection services)
 #endif
 
             services.AddOptions<GrpcWorkerStartupOptions>()
-                .Configure<IConfiguration>((arguments, config) =>
+                .Configure<IConfiguration>((grpcWorkerStartupOption, config) =>
```
jviau, line 71: This would be a good candidate for configuration binding:

``` CSharp
services.AddOptions<GrpcWorkerStartupOptions>()
    .BindConfiguration("Functions:Worker")
    .Configure<IConfiguration>((opt, config) =>
    {
        opt.HostEndpoint ??= // host + port check
        opt.RequestId ??= config["requestId"];
        // etc...
    });
    
```
 A drawback to be aware of is `.BindConfiguration` uses reflection, so it will have a startup impact (but how bad?). In NET8 they are introducing a source gen which will remove the reflection.

Also, what you have now is totally fine if you want to keep it that way.

src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
kshyju: Sounds good to me. @fabiocav let me know you have concerns because t[his comment ](https://github.com/Azure/azure-functions-dotnet-worker/pull/1897#discussion_r1325152768)of yours will not be addressed.

src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs
kshyju: `IConfiguration.Bind` is one of the methods [we identified last year as non AOT friendly](https://github.com/Azure/azure-functions-dotnet-worker/issues/1055) when we investigated AOT readiness of dotnet worker.  May be this is something we can revisit as part of that AOT readiness effort later.

extensions/Worker.Extensions.Rpc/src/ConfigurationExtensions.cs
kshyju: The new itearation is using switch mapping and using a configuration section for the grpc channel information.

src/DotNetWorker.Grpc/GrpcStartupOptions.cs
kshyju: Switched property name as suggested.

src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs
kshyju: This is still not addressed based on the [latest comment ](https://github.com/Azure/azure-functions-dotnet-worker/pull/1897#discussion_r1326259085)from @jviau

src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
kshyju: Closing this comment as thus change is not valid anymore.

src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
kshyju: Pushed an update to address this.

jviau (approved):

Issue #1900, opened 2023-09-12
Update dotnet worker to handle the additional command line args with "functions-" prefix


kshyju:
Update dotnet worker to handle the additional command line args with "functions-" prefix. See https://github.com/Azure/azure-functions-host/issues/9504 for details.

Example of command line argument string host will send to worker process.

-host localhost --port 80 --workerId testWorkerId --requestId testId --grpcMaxMessageLength 2147483647 --functions-uri [http://localhost:80](http://localhost/) --functions-workerid testWorkerId --functions-requestid testId --functions-grpcmaxmessagelength 2147483647

Need to update managed code and FunctionsNetHost.

PR #1901, opened 2023-09-12
Updating FunctionsNetHost to handle the new "functions-" prefix cmdline args.

kshyju:
Updating FunctionsNetHost to handle the new "functions-" prefix cmdline args.

Part of #1900  

Also bumped up the global JSON .net version to rc1.

### Pull request checklist

### DIFF
--- a/host/src/FunctionsNetHost/Program.cs
+++ b/host/src/FunctionsNetHost/Program.cs
@@ -29,40 +29,35 @@ namespace FunctionsNetHost
 
         private static async Task<GrpcWorkerStartupOptions> GetStartupOptionsFromCmdLineArgs(string[] args)
         {
-            var hostOption = new Option<string>("--host");
-            var portOption = new Option<int>("--port");
-            var workerOption = new Option<string>("--workerId");
-            var grpcMsgLengthOption = new Option<int>("--grpcMaxMessageLength");
-            var requestIdOption = new Option<string>("--requestId");
-            var funcUriOption = new Option<string>("--functions-uri");
-            var funcRequestIdOption = new Option<string>("--functions-request-id");
-            var funcWorkerIdOption = new Option<string>("--functions-worker-id");
-            var funcGrpcMsgLengthOption = new Option<int?>("--functions-grpc-max-message-length");
+            var uriOption = new Option<string>("--functions-uri") { IsRequired = true };
+            var requestIdOption = new Option<string>("--functions-request-id") { IsRequired = true };
+            var workerIdOption = new Option<string>("--functions-worker-id") { IsRequired = true };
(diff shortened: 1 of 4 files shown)


host/src/FunctionsNetHost/Program.cs
```diff
@@ -35,25 +34,36 @@ private static async Task<GrpcWorkerStartupOptions> GetStartupOptionsFromCmdLine
...
+            rootCommand.SetHandler((context) =>
             {
-                workerStartupOptions.Host = host;
-                workerStartupOptions.Port = port;
-                workerStartupOptions.WorkerId = workerId;
-                workerStartupOptions.GrpcMaxMessageLength = grpcMsgLength;
-                workerStartupOptions.RequestId = requestId;
-            },
-                hostOption, portOption, workerOption, grpcMsgLengthOption, requestIdOption);
+                var serverEndpoint = context.ParseResult.GetValueForOption(funcUriOption) ?? $"http://{context.ParseResult.GetValueForOption(hostOption)}:{context.ParseResult.GetValueForOption(portOption)}";
```
kshyju, line 60: Synced with Jon from the .net command line team and he suggested I should switch to the new approach which uses the `InvocationContext` and calling the `GetValueForOption` method. This is something they have introduced in the newer version.

> This is closer to the pattern we're arriving at in the next version. Handler parameter binding is going away as a concept.

host/src/FunctionsNetHost/Program.cs
```diff
@@ -35,25 +34,36 @@ private static async Task<GrpcWorkerStartupOptions> GetStartupOptionsFromCmdLine
...
+                workerStartupOptions.ServerUri = new Uri(serverEndpoint);
```
fabiocav, line 61: Can we use `TryCreate` and provide a clear error if we have an invalid URI?

host/src/FunctionsNetHost/Program.cs
```diff
@@ -35,25 +34,36 @@ private static async Task<GrpcWorkerStartupOptions> GetStartupOptionsFromCmdLine
...
+
+                workerStartupOptions.GrpcMaxMessageLength = context.ParseResult.GetValueForOption(funcGrpcMsgLengthOption) ?? context.ParseResult.GetValueForOption(grpcMsgLengthOption);
```
fabiocav, line 63: As discussed, shouldn't need the fallback checks.

fabiocav (approved):

host/src/FunctionsNetHost/Program.cs
kshyju: The new iteration uses TryCreate to validate and throw a more meaningful message.

host/src/FunctionsNetHost/Program.cs
kshyju: Good point. Removed fallback.

jviau (approved):

PR #1203, opened 2022-12-01
Updating samples and tests targets and dependencies

fabiocav:
This PR aligns with Azure SDK changes and updates all test and sample application TFMs to the latest supported (net7.0)

All applicable test and sample dependency updates are also applied.

### Pull request checklist

* [x] My changes **do not** require documentation changes
  * [ ] Otherwise: Documentation issue linked to PR
* [x] My changes **should not** be added to the release notes for the next release
  * [ ] Otherwise: I've added my notes to `release_notes.md`
* [x] My changes **do not** need to be backported to a previous version
  * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [x] I have added all required tests (Unit tests, E2E tests)

PR #1905, opened 2023-09-13
Setup E2E test app for Net Framework

### LINKED ISSUE #965: Add E2E .NET Framework tests
Create E2E test app (probably [here](https://github.com/Azure/azure-functions-dotnet-worker/tree/main/test/E2ETests/E2EApps)) for .NET Framework to validate builds/PRs as we develop this worker.

surgupta-msft:

resolves #965 

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/setup-e2e-tests.ps1
+++ b/setup-e2e-tests.ps1
@@ -104,6 +104,8 @@ if (Test-Path $output)
   Remove-Item $output -Recurse -Force -ErrorAction Ignore
 }
 
+Write-Host "----- Executing tests for Dotnet version $DotnetVersion -----"
+
 $AdditionalPackArgs = @("-c", "Release", "-p:FunctionsRuntimeVersion=$FunctionsRuntimeVersion", "-p:DotnetVersion=$DotnetVersion")
 
 if ($SkipBuildOnPack -eq $true)
--- a/setup-e2e-tests.ps1
+++ b/setup-e2e-tests.ps1
@@ -1,6 +1,12 @@
 #Requires -Version 6
 
 param(
+    [Parameter(Mandatory=$true)]
+    [ValidateNotNullOrEmpty()]
+    [String]
+    [ValidateSet("net7", "netfx")]
+    $DotnetVersion,
+
     [Switch]
     $SkipStorageEmulator,
 
@@ -98,7 +104,7 @@ if (Test-Path $output)
   Remove-Item $output -Recurse -Force -ErrorAction Ignore
 }
 
-$AdditionalPackArgs = @("-c", "Release", "-p:FunctionsRuntimeVersion=$FunctionsRuntimeVersion")
+$AdditionalPackArgs = @("-c", "Release", "-p:FunctionsRuntimeVersion=$FunctionsRuntimeVersion", "-p:DotnetVersion=$DotnetVersion")
 
 if ($SkipBuildOnPack -eq $true)
 {
--- a/src/ci.yml
+++ b/src/ci.yml
@@ -75,13 +75,6 @@ jobs:
       projects: |
         test/**/*Tests.csproj
 
-  - task: DotNetCoreCLI@2
-    displayName: 'Build projects'
-    inputs:
-      command: 'build'
-      arguments: '-c Release -p:BuildNumber=$(buildNumber) -p:IsLocalBuild=False'
-      projects: 'DotNetWorker.sln'
-
   - pwsh: |
       $useIntegrationTestingCoreTools = $null
       if (-not ([bool]::TryParse($env:USECORETOOLSBUILDFROMINTEGRATIONTESTS, [ref] $useIntegrationTestingCoreTools)))
--- a/src/ci.yml
+++ b/src/ci.yml
@@ -62,13 +62,32 @@ jobs:
       {
         throw "UseCoreToolsBuildFromIntegrationTests can only be set to True or False. Current value is set to $env:USECORETOOLSBUILDFROMINTEGRATIONTESTS"
       }
-      ./setup-e2e-tests.ps1 -DotnetVersion $env:DOTNETVERSION `
+      ./setup-e2e-tests.ps1 -DotnetVersion 'net7' `
                             -UseCoreToolsBuildFromIntegrationTests:$useIntegrationTestingCoreTools `
                             -SkipBuildOnPack
-    displayName: "Setup E2E tests"
+    displayName: "Setup E2E tests - Net7"
 
   - task: DotNetCoreCLI@2
-    displayName: 'Run tests'
+    displayName: 'Run tests - Net7'
+    inputs:
+      command: 'test'
+      arguments: '--no-build -c Release'
+      projects: |
+        test/**/*Tests.csproj
+
+  - pwsh: |
+      $useIntegrationTestingCoreTools = $null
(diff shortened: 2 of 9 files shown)


test/E2ETests/E2EApps/E2EApp/E2EApp.csproj
```diff
@@ -1,26 +1,39 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-  <PropertyGroup>
+  <PropertyGroup Condition="$(DotnetVersion) != 'netfx'">
     <TargetFramework>net7.0</TargetFramework>
```
kshyju, line 3: Question: Can the `TargetFramework` value be a variable passed in from the caller of the test? We are officially supporting .net8 for the worker now. Wonder how easy it is to test on .net8 as well.

surgupta-msft: For this PR, it was easy to run Net7 because everything else was already upgraded to Net7 as part of this old commit - https://github.com/Azure/azure-functions-dotnet-worker/pull/1203. We can certainly pass any variable from caller of the test but I believe we need to upgrade all the dependencies to be compatible with Net8 first.

test/E2ETests/E2EApps/E2EApp/E2EApp.csproj
```diff
@@ -1,26 +1,39 @@
...
     <_FunctionsSkipCleanOutput>true</_FunctionsSkipCleanOutput>
     <AssemblyName>Microsoft.Azure.Functions.Worker.E2EApp</AssemblyName>
     <RootNamespace>Microsoft.Azure.Functions.Worker.E2EApp</RootNamespace>
     <Nullable>disable</Nullable>
   </PropertyGroup>
 
-  <ItemGroup>
+  <!-- Remove  any .cs files which are not expected to work with net framework -->  
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net48' ">
+      <Compile Remove="Blob\*.cs; Cosmos\*.cs; EventHubs\*.cs; Queue\*.cs; Table\*.cs"/>
```
jviau, line 22: Why are these being removed? I think we found this does compile when we are nut multi-targeting.

surgupta-msft: Yeah the build is working fine but I have noticed random failures in tests.

For example - 
- Samples changes that would be needed to make E2E build working with Net48 - [Link](https://github.com/Azure/azure-functions-dotnet-worker/compare/surgupta/netfx-e2etest-setup...surgupta/netfx-test?expand=1)
- Two pipeline runs on the same commitId and DotnetVersion as Net7 - [Run1](https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=152811&view=results), [Run2](https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=152816&view=results).
- Run1 has 1 test failure while Run2 has 2 test failures. (I have also seen upto 4 tests failures in other runs).

With the current approach tests are working fine consistently. I have updated the logic here to replace "Remove files" to "Include files" if that feels better.

```csharp
  <ItemGroup Condition=" '$(TargetFramework)' == 'net48' ">
      <Compile Remove="**\*.cs"/>
      <Compile Include="Timer\*.cs; Http\*.cs; *.cs"/>
  </ItemGroup>
```

I am open to suggestions here.

jviau: I see. I would stick with the `Remove` style you originally had. Do we have issues tracking the work to make those other tests function in net48?

src/ci.yml
```diff
@@ -62,19 +62,29 @@ jobs:
...
       arguments: '--no-build -c Release'
       projects: |
         test/**/*Tests.csproj
 
+  - task: DotNetCoreCLI@2
+    condition: eq(variables['DotnetVersion'], 'netfx')
+    displayName: 'Run tests for NetFramework'
+    inputs:
+      command: 'test'
+      arguments: '--no-build -c Release  --filter "FullyQualifiedName~TimerTrigger"' #only run http tests to avoid emulators on Linux
```
kshyju, line 84: Is the filter correct here? Did you mean "HttpTrigger" ?

setup-e2e-tests.ps1
```diff
@@ -98,7 +104,7 @@ if (Test-Path $output)
   Remove-Item $output -Recurse -Force -ErrorAction Ignore
 }
 
-$AdditionalPackArgs = @("-c", "Release", "-p:FunctionsRuntimeVersion=$FunctionsRuntimeVersion")
+$AdditionalPackArgs = @("-c", "Release", "-p:FunctionsRuntimeVersion=$FunctionsRuntimeVersion", "-p:DotnetVersion=$DotnetVersion")
```
kshyju, line 107: Can you remind me who sets this environment variable when we are triggering this pipeline for a PR build? We need to run it with both values rite? So does this gets called twice with different values?

src/ci.yml
surgupta-msft: Did a bit of modifications to the PR so that such checks are not needed. Removed this condition now.

setup-e2e-tests.ps1
surgupta-msft: Thank you for reminding me this question. @jviau do you know if the ask is to run tests with both net7 and net48 every time or is it either-or based on pipeline configuration? If not, I will double check with Fabio and update here.

test/E2ETests/E2EApps/E2EApp/E2EApp.csproj
surgupta-msft: Checked a few more times and realized that there are only 4 tests that are failing or showing flakiness. I have updated the PR to skip those 4 and created this [github issue](https://github.com/Azure/azure-functions-dotnet-worker/issues/1910) to track fix for them. This helped in getting rid of few checks I added to avoid those tests. Let me know if the updated PR looks better or should we consider switching back to the old state.

setup-e2e-tests.ps1
jviau: That depends on how our infra will handle it. _Ideally_, they both would run in the same pipeline together. However, I know we had issues with netfx tests before on the linux job, so we may need to address that.

surgupta-msft: Update ci.yml to run both everytime.

src/ci.yml
```diff
@@ -62,13 +62,32 @@ jobs:
...
+      {
+        throw "UseCoreToolsBuildFromIntegrationTests can only be set to True or False. Current value is set to $env:USECORETOOLSBUILDFROMINTEGRATIONTESTS"
+      }
+      ./setup-e2e-tests.ps1 -DotnetVersion 'netfx' `
+                            -UseCoreToolsBuildFromIntegrationTests:$useIntegrationTestingCoreTools `
+                            -SkipBuildOnPack
+    displayName: "Setup E2E tests - NetFramework"
+
+  - task: DotNetCoreCLI@2
+    displayName: 'Run tests - NetFramework'
```
liliankasem, line 90: just to confirm, we are not running these for linux right?

surgupta-msft: Skipped Netframework tests for linux

liliankasem (approved):

PR #1917, opened 2023-09-20
Analyzer to detect missing ASP.NET Core integration registration

### LINKED ISSUE #1902: Create analyzer to detect missing ASP.NET Core integration registration
The analyzer should detect missing registrations. Code fix to make the change.

surgupta-msft:

resolves #1902 

### Pull request checklist

  * [x] Otherwise: Added documentation in the PR

### Additional information

Additional PR information


docs/analyzer-rules/AZFW0014.md
```diff
@@ -0,0 +1,29 @@
...
+
+This rule is triggered when the ASP.NET function app is missing registration for the appropriate method.
+
+## Rule description
+
+The ASP.NET function app should have registration for `ConfigureFunctionsWebApplication` method. This rule will be violated when the app is missing this registration.
+
+## How to fix violations
+
+Add registration for `ConfigureFunctionsWebApplication` method in Function App. Refer to <ToDo: Add public Msft docs link>
```
surgupta-msft, line 19: Do we have any public docs for this available that I can add here?

kshyju: https://learn.microsoft.com/en-us/azure/azure-functions/dotnet-isolated-process-guide#aspnet-core-integration

sdk/Sdk.Analyzers/RegistrationExpectedInASPNetIntegration.cs
```diff
@@ -0,0 +1,41 @@
...
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze);
+            context.RegisterSymbolAction(AnalyzeMethod, SymbolKind.Namespace);
+        }
+
+        private static void AnalyzeMethod(SymbolAnalysisContext context)
+        {
+            bool isAspNetAssembly = context.Compilation.ReferencedAssemblyNames.Where(assembly => assembly.Name.Equals(AspNetExtensionAssemblyName)).Any();
```
kshyju, line 29: Replace `Where`  with `Any` (and remove the final Any call)

sdk/Sdk.Analyzers/RegistrationExpectedInASPNetIntegration.cs
```diff
@@ -0,0 +1,41 @@
...
+            SyntaxReference syntaxReference = context.Symbol.DeclaringSyntaxReferences.FirstOrDefault();
```
kshyju, line 30: We can return early/ext if AspNetExtensionAssemblyName reference is not present.

sdk/Sdk.Analyzers/RegistrationExpectedInASPNetIntegration.cs
```diff
@@ -0,0 +1,41 @@
...
+            string code = syntaxReference.SyntaxTree.GetText().ToString();
+
```
kshyju, line 32: I think we do not need to scan the entire source code from the app. So basically, exit if the current method we are analyzing is not the `main` method of that assembly.

test/Sdk.Analyzers.Tests/RegistrationExpectedInAspNetIntegrationTests.cs
```diff
@@ -0,0 +1,277 @@
...
+            string testCode = @"
+                using System;
+                using Microsoft.Extensions.DependencyInjection;
+                using Microsoft.Extensions.Hosting;
+
+                namespace SampleApp
+                {
+                    public class Program
+                    {
+                        public static void Main()
```
kshyju, line 138: Add another private or public method to this class in one of those tests.

surgupta-msft: done

sdk/Sdk.Analyzers/RegistrationExpectedInASPNetIntegration.cs
surgupta-msft: done

docs/analyzer-rules/AZFW0014.md
surgupta-msft: Thanks

sdk/Sdk.Analyzers/RegistrationExpectedInASPNetIntegration.cs
surgupta-msft: I was scanning completely because it's not necessary that there would be a `Main` method like the code sample [here]( https://github.com/Azure/azure-functions-dotnet-worker/blob/main/samples/AspNetIntegration/Program.cs). Is there any way to detect this implicit `Main`?

surgupta-msft: Updated code to check only Main method now

liliankasem (approved):

sdk/Sdk.Analyzers/Extensions/MethodSymbolExtensions.cs
```diff
@@ -43,5 +43,23 @@ public static bool IsFunction(this IMethodSymbol symbol, SymbolAnalysisContext a
 
             return attributes.Any(a => attributeType.IsAssignableFrom(a.AttributeClass, true));
         }
+
+        /// <summary>
+        /// Checks if a method symbol is a Main method. This also checks for implicit main in top-level statements
```
kshyju, line 48: 👍

sdk/Sdk.Analyzers/RegistrationExpectedInASPNetIntegration.cs
```diff
@@ -0,0 +1,48 @@
...
+            if (!isAspNetAssembly || !symbol.IsMainMethod())
+            {
+                return;
+            }
+
+            SyntaxReference syntaxReference = symbol.DeclaringSyntaxReferences.FirstOrDefault();
+            string code = syntaxReference.SyntaxTree.GetText().ToString();
+
+            if (!code.Contains(ExpectedRegistrationMethod))
```
kshyju, line 41: The string.Contains may fail if user has code like below.

```csharp
var host = new HostBuilder()
//.ConfigureFunctionsWebApplication()
 .Build();
```
Update one of the tests have this code.

kshyju: May be we should try something like this.

```csharp
SyntaxReference syntaxReference = symbol.DeclaringSyntaxReferences.FirstOrDefault();
var root = syntaxReference.SyntaxTree.GetRoot();

var methodCallExpressions = root.DescendantNodes().OfType<InvocationExpressionSyntax>();

bool methodInvocationPresent = methodCallExpressions.Any(invocation =>
    (invocation.Expression as MemberAccessExpressionSyntax)?.Name.Identifier.ValueText ==
    ExpectedRegistrationMethod);

if (methodInvocationPresent)
{
    return;
}
// Report diagnostic here

```

fabiocav (changes requested):
Haven't reviewed the implementation, but we need to make sure this analyzer is part of the ASP.NET Core integration package.

sdk/Sdk.Analyzers/RegistrationExpectedInASPNetIntegration.cs
surgupta-msft: Absolutely loved the suggestion here. Thank you!

test/Worker.Extensions.Tests/Worker.Extensions.Tests.csproj
```diff
@@ -19,6 +19,13 @@
             <PrivateAssets>all</PrivateAssets>
             <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
         </PackageReference>
+        <PackageReference Include="SmartAnalyzers.RoslynTestKit" Version="4.7.0" />
+        <PackageReference Include="coverlet.collector" Version="3.2.0">
+            <PrivateAssets>all</PrivateAssets>
+            <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
```
surgupta-msft, line 25: The changes here are taken from [Sdk.Analyzers.Tests.csproj](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/test/Sdk.Analyzers.Tests/Sdk.Analyzers.Tests.csproj) as we wanted to keep _extension analyzer tests_ outside of `Sdk.Analyzer.Tests`.

extensions/Worker.Extensions.Http.AspNetCore/src/Worker.Extensions.Http.AspNetCore.csproj
```diff
@@ -1,23 +1,54 @@
...
+        <Compile Include="Analyzer\*.cs" />
+    </ItemGroup>
+
+    <ItemGroup>
+        <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.3">
+            <PrivateAssets>all</PrivateAssets>
+            <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+        </PackageReference>
+        <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.11.0" />
+    </ItemGroup>
```
surgupta-msft, line 46: For reference, changes here are similar to [Sdk.Analyzers.csproj](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/sdk/Sdk.Analyzers/Sdk.Analyzers.csproj).

extensions/Worker.Extensions.Http.AspNetCore/src/Analyzer/RegistrationExpectedInASPNetIntegration.cs
```diff
@@ -0,0 +1,72 @@
...
+            context.EnableConcurrentExecution();
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze);
+            context.RegisterSymbolAction(AnalyzeMethod, SymbolKind.Method);
+        }
+
+        private static void AnalyzeMethod(SymbolAnalysisContext context)
+        {
+            var symbol = (IMethodSymbol)context.Symbol;
+
+            var isAspNetAssembly = context.Compilation.ReferencedAssemblyNames.Any(assembly => assembly.Name.Equals(AspNetExtensionAssemblyName));
```
kshyju, line 36: I wonder whether we need this code anymore as the analyzer is part of the assembly now.

test/Worker.Extensions.Tests/AspNetIntegration/RegistrationExpectedInAspNetIntegrationTests.cs
```diff
@@ -0,0 +1,341 @@
...
+                using Microsoft.Extensions.Hosting;
+
+                namespace SampleApp
+                {
+                    public class Program
+                    {
+                        public static void Main()
+                        {
+                            var host = new HostBuilder()
+                                .ConfigureFunctionsWorkerDefaults()
```
kshyju, line 189: What happens if someone calls `ConfigureFunctionsWorkerDefaults` followed by `ConfigureFunctionsWebApplication` ?

kshyju:
```csharp
var host = new HostBuilder()
    .ConfigureFunctionsWebApplication()
    .Build();
```

extensions/Worker.Extensions.Http.AspNetCore.Analyzers/Worker.Extensions.Http.AspNetCore.Analyzers.csproj
```diff
@@ -0,0 +1,36 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <VersionPrefix>1.0.0</VersionPrefix>
+        <OutputType>Library</OutputType>
+        <SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>
+        <IncludeBuildOutput>false</IncludeBuildOutput>
+        <TargetFramework>netstandard2.0</TargetFramework>
+        <PackageId>Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore.Analyzers</PackageId>
```
surgupta-msft, line 9: For reference, changes here are similar to [Sdk.Analyzers.csproj](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/sdk/Sdk.Analyzers/Sdk.Analyzers.csproj).

extensions/Worker.Extensions.Http.AspNetCore/src/Analyzer/RegistrationExpectedInASPNetIntegration.cs
surgupta-msft: yeah removed the check

test/Worker.Extensions.Tests/AspNetIntegration/RegistrationExpectedInAspNetIntegrationTests.cs
surgupta-msft: yeah, tried above scenario with an aspnet function app and it did not work. Didn't get proper exception also. Do we have an issue tracking to fix this behavior? For analyzer, I can add a new one for this scenario to keep the [error code](https://github.com/Azure/azure-functions-dotnet-worker/tree/main/docs/analyzer-rules) specific to one issue.

surgupta-msft: Update - had an offline sync on this. The investigation for this scenario will be tracked [here](https://github.com/Azure/azure-functions-dotnet-worker/issues/1947)

extensions/Worker.Extensions.Http.AspNetCore.Analyzers/RegistrationExpectedInASPNetIntegration.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore
+{
+    /// <summary>
+    /// Analyzer to verify whether expected registation is present for AspNetIntegration.
```
kshyju, line 13: nit: typo

kshyju (approved):
One question about organizing the test project.

test/Worker.Extensions.Tests/Worker.Extensions.Tests.csproj
```diff
@@ -19,6 +19,13 @@
             <PrivateAssets>all</PrivateAssets>
```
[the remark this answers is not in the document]
kshyju: Should we consider moving any new project/code in that structure already? @jviau 

  - extensions
       - Worker.Extensions.Http.AspNetCore
           - src
                   - Worker.Extensions.Http.AspNetCore
                   - Worker.Extensions.Http.AspNetCore.Analyzers.
           - tests
                   - Worker.Extensions.Http.AspNetCore.Analyzers.Tests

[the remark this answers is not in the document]
surgupta-msft: Not sure how early we are planning to shift to new restructuring of the folders. But meanwhile, I am thinking that it can look bit awkward to have only 1 category of tests inside `Extensions` folder and all remaining tests inside separate tests projects.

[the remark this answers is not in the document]
jviau: Yeah, the engineering work (when I get to it) is planned to restructure tests. But until then feel free to put tests wherever you feel they fit best. If you want you could add one similar to this: https://github.com/Azure/azure-functions-dotnet-worker/tree/main/test/Worker.Extensions.Rpc.Tests

kshyju:
also update release notes? Do you plan to do the version bump before releasing or with this PR itself?

test/Worker.Extensions.Tests/Worker.Extensions.Tests.csproj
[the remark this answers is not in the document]
surgupta-msft: Done

extensions/Worker.Extensions.Http.AspNetCore.Analyzers/tools/install.ps1
jviau, line 1: Are these files needed anymore? I thought this was redundant with SDK style projects. Do we support non-SDK projects?

surgupta-msft: Removed them. Thanks for catching.

extensions/Worker.Extensions.Http.AspNetCore.Analyzers/DiagnosticDescriptors.cs
```diff
@@ -0,0 +1,22 @@
...
+        public const string Usage = "Usage";
+
+        private static DiagnosticDescriptor Create(string id, string title, string messageFormat, string category, DiagnosticSeverity severity)
+        {
+            var helpLink = $"https://aka.ms/azfw-rules?ruleid={id}";
+            return new DiagnosticDescriptor(id, title, messageFormat, category, severity, isEnabledByDefault: true, helpLinkUri: helpLink);
+        }
+
+        public static DiagnosticDescriptor CorrectRegistrationExpectedInAspNetIntegration { get; }
+            = Create(id: "AZFW0014", title: "Missing expected registration in ASPNet Integration", messageFormat: "The registration for method '{0}' is expected for ASP.Net Integration.",
```
liliankasem, line 19:
```suggestion
            = Create(id: "AZFW0014", title: "Missing expected registration in ASP.Net Integration", messageFormat: "The registration for method '{0}' is expected for ASP.Net Integration.",
```

extensions/Worker.Extensions.Http.AspNetCore.Analyzers/RegistrationExpectedInASPNetIntegration.cs
```diff
@@ -0,0 +1,69 @@
...
+            }
+
+            var diagnostic = Diagnostic.Create(DiagnosticDescriptors.CorrectRegistrationExpectedInAspNetIntegration, symbol.Locations.First(), ExpectedRegistrationMethod);
+            context.ReportDiagnostic(diagnostic);
+        }
+
+        // Checks if a method symbol is a Main method. This also checks for implicit main in top-level statements
+        private static bool IsMainMethod(IMethodSymbol symbol)
+        {
+            var isMainMethod = symbol?.IsStatic == true && symbol.Name switch
```
liliankasem, line 58:
```suggestion
            return symbol?.IsStatic == true && symbol.Name switch
```

test/Worker.Extensions.Http.AspNetCore.Tests/Worker.Extensions.Http.AspNetCore.Tests.csproj
```diff
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
```
liliankasem, line 1: I think no real changes were made here?

liliankasem (approved):

extensions/Worker.Extensions.Http.AspNetCore.Analyzers/DiagnosticDescriptors.cs
surgupta-msft: Good catch. Updated at few other places also.

test/Worker.Extensions.Http.AspNetCore.Tests/Worker.Extensions.Http.AspNetCore.Tests.csproj
surgupta-msft: yeah, initially had few changes but no changes now. This is keep coming up here even after fixing a couple times.

jviau (commented):
I think we discussed having this analyzer only trigger if `ConfigureFunctionsWorkerDefaults` is present and `ConfigureFunctionsWebApplication` is not. Are we going with that approach or not?

I don't like the idea of having this analyzer trying to warn on the _absence_ of something on `IHostBuilder` without something _extra_ to indicate this is intended to be a function app and they are not configuring at as we expect.

surgupta-msft:
> I think we discussed having this analyzer only trigger if `ConfigureFunctionsWorkerDefaults` is present and `ConfigureFunctionsWebApplication` is not. Are we going with that approach or not?
> 
> I don't like the idea of having this analyzer trying to warn on the _absence_ of something on `IHostBuilder` without something _extra_ to indicate this is intended to be a function app and they are not configuring at as we expect.

(Providing a brief summary below about the things we have discussed to avoid any confusion)

My understanding is we changed _error_ to _warning_ because absence of `ConfigureFunctionsWebApplication()` may not mean that the app is configured incorrectly. The other thing we discussed which is tracked by a separate issue #1947 is the presence of both methods which can cause issues due to redundant calls. For that, we discussed that if we fix redundant calls issue then we may not need any analyzer for that case at all.

For this exact thing - _this analyzer only trigger if `ConfigureFunctionsWorkerDefaults` is present and `ConfigureFunctionsWebApplication` is not_. I will setup some time to close on this.

docs/analyzer-rules/AZFW0014.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Cause
+
+This rule is triggered when the ASP.NET function app is missing registration for the appropriate method.
+
+## Rule description
+
+The ASP.NET function app should have registration for `ConfigureFunctionsWebApplication` method. This rule will be violated when the app is missing this registration.
+
+## How to fix violations
```
jviau, line 17: Can we add a `When to suppress violations` section, saying it is okay to suppress this if `ConfigureFunctionsWebApplication` is called indirectly through a different method off the `IHostBuilder`.

test/extensions/Worker.Extensions.Http.AspNetCore.Tests/RegistrationExpectedInAspNetIntegrationTests.cs
```diff
@@ -0,0 +1,315 @@
...
+                            .WithSeverity(Microsoft.CodeAnalysis.DiagnosticSeverity.Warning)
+                            .WithSpan(13, 37, 13, 41)
+                            .WithArguments(ExpectedRegistrationMethod));
+
+            await test.RunAsync();
+        }
+
+
+        [Fact]
+        public async Task AspNetIntegration_CommentedRegistration_Diagnostics_Expected()
```
jviau, line 56:
```suggestion
        public async Task AspNetIntegration_CommentedRegistration_Diagnostics_NotExpected()
```

test/extensions/Worker.Extensions.Http.AspNetCore.Tests/RegistrationExpectedInAspNetIntegrationTests.cs
```diff
@@ -0,0 +1,315 @@
...
+                    }
+                }";
+
+            var test = new AnalyzerTest
+            {
+                ReferenceAssemblies = LoadRequiredDependencyAssemblies(),
+                TestCode = testCode
+            };
+
+            // test.ExpectedDiagnostics is an empty collection because ConfigureFunctionsWorkerDefaults() is not present
```
jviau, line 88: Can you asset the collection is empty?

docs/analyzer-rules/AZFW0014.md
surgupta-msft: done

test/extensions/Worker.Extensions.Http.AspNetCore.Tests/RegistrationExpectedInAspNetIntegrationTests.cs
surgupta-msft: The analyzer test infra internally does that. Meaning for this case, it is expecting Empty diagnostic collection. The test will fail if diagnostics are generated here (because it is expecting empty collection).

extensions/Worker.Extensions.Http.AspNetCore.Analyzers/RegistrationExpectedInASPNetIntegration.cs
```diff
@@ -0,0 +1,95 @@
...
+            var syntaxReference = symbol.DeclaringSyntaxReferences.FirstOrDefault();
+            var root = syntaxReference?.SyntaxTree.GetRoot();
+            var methodCallExpressions = root?.DescendantNodes().OfType<InvocationExpressionSyntax>();
+
+            var expectedMethodInvocationPresent = methodCallExpressions?.Any(invocation => (invocation.Expression as MemberAccessExpressionSyntax)?.Name.Identifier.ValueText == ExpectedRegistrationMethod);
+
+            var incorrectMethodCallExpressions = methodCallExpressions?.Where(invocation => (invocation.Expression as MemberAccessExpressionSyntax)?.Name.Identifier.ValueText == IncorrectRegistrationMethod);
+            var incorrectMethodInvocationPresent = incorrectMethodCallExpressions?.Any();
+
+            if (!(bool)incorrectMethodInvocationPresent)
```
kshyju, line 51: This casting expression will throw if `incorrectMethodInvocationPresent` has `null`. Consider making `incorrectMethodInvocationPresent` bool type instead of `bool?`.

extensions/Worker.Extensions.Http.AspNetCore.Analyzers/RegistrationExpectedInASPNetIntegration.cs
```diff
@@ -0,0 +1,95 @@
...
+            if (!IsMainMethod(symbol))
+            {
+                return;
+            }
+
+            var syntaxReference = symbol.DeclaringSyntaxReferences.FirstOrDefault();
+            var root = syntaxReference?.SyntaxTree.GetRoot();
+            var methodCallExpressions = root?.DescendantNodes().OfType<InvocationExpressionSyntax>();
+
+            var expectedMethodInvocationPresent = methodCallExpressions?.Any(invocation => (invocation.Expression as MemberAccessExpressionSyntax)?.Name.Identifier.ValueText == ExpectedRegistrationMethod);
```
kshyju, line 46: Move this closer to usage (after line 55).

extensions/Worker.Extensions.Http.AspNetCore.Analyzers/RegistrationExpectedInASPNetIntegration.cs
```diff
@@ -0,0 +1,95 @@
...
+        {
+            var symbol = (IMethodSymbol)context.Symbol;
+
+            if (!IsMainMethod(symbol))
+            {
+                return;
+            }
+
+            var syntaxReference = symbol.DeclaringSyntaxReferences.FirstOrDefault();
+            var root = syntaxReference?.SyntaxTree.GetRoot();
```
kshyju, line 43: I suggest doing a null check somewhere here and return, so that for the remaining code you can use non nullable types. Null checks on collections types (such as IEnumerable returned by the linq statement below) does not sounds right.

extensions/Worker.Extensions.Http.AspNetCore.Analyzers/RegistrationExpectedInASPNetIntegration.cs
```diff
@@ -0,0 +1,95 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore
+{
+    /// <summary>
+    /// Analyzer to verify whether expected registration is present for ASP.NET Core Integration.
+    /// </summary>
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class RegistrationExpectedInASPNetIntegration : DiagnosticAnalyzer
```
kshyju, line 17:
```suggestion
    public sealed class RegistrationExpectedInASPNetIntegration : DiagnosticAnalyzer
```

extensions/Worker.Extensions.Http.AspNetCore.Analyzers/Worker.Extensions.Http.AspNetCore.Analyzers.csproj
```diff
@@ -0,0 +1,34 @@
...
```
kshyju, line 9: Since we are not planning to publish a package from this project, Shall we consider setting `<IsPackable>false</IsPackable>` ?

Also, Should we remove the other properties which are used for a project which gets packed. Ex: PackageId, Description, VersionPrefix et..

extensions/Worker.Extensions.Http.AspNetCore.Analyzers/RegistrationExpectedInASPNetIntegration.cs
jviau: Because we have `incorrectMethodCallExpressions?.Any()`, it will have to be `bool?`. What I do in this case is null-coalesce for the null case:

Not sure what the behavior should be when `null`, but something like this (or `false` instead of `true`)
``` CSharp
if (!(incorrectMethodInvocationPresent ?? true))
```

fabiocav (approved):

extensions/Worker.Extensions.Http.AspNetCore.Analyzers/Worker.Extensions.Http.AspNetCore.Analyzers.csproj
surgupta-msft: I tried doing this but the analyzer is not working as expected. Some examples I found seem to be packing it, not sure if this is a requirement for Analyzer. Happy to debug this with you but don't think we should block on this.

PR #1924, opened 2023-09-22
Refactor CosmosDBConverter to remove use reflection

### LINKED ISSUE #1613: [SDK Binding] Refactor CosmosDbConverter to remove use of reflection
CosmosDBConverter issue capturing the work done for blobs as part of #1325

liliankasem:

resolves #1613

### Pull request checklist

### Additional information

This is the approach I could think of but if there is a better way I'm all for it!


liliankasem:
@fabiocav let me know if this is vastly different to what you had in mind - happy to make adjustments for improvement but figured it would be nice to get this out before I release cosmos 4.5.0

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
```diff
@@ -80,47 +80,43 @@ private CosmosDBInputAttribute GetBindingDataContent(ModelBindingData bindingDat
...
-            else
-            {
-                createPOCOMethod = GetType()
-                                    .GetMethod(nameof(CreatePOCOAsync), BindingFlags.Instance | BindingFlags.NonPublic)
-                                    .MakeGenericMethod(targetType);
-            }
-
+            var json = JsonSerializer.Serialize(data);
+            return JsonSerializer.Deserialize(json, targetType,
+                new JsonSerializerOptions() { PropertyNameCaseInsensitive = true });
```
kshyju, line 90: Use a cached/static version of `JsonSerializerOptions` instance instead of creating one in every Convert call.

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
```diff
@@ -80,47 +81,42 @@ private CosmosDBInputAttribute GetBindingDataContent(ModelBindingData bindingDat
             Type _ when targetType == typeof(CosmosClient) => CreateCosmosClient<CosmosClient>(cosmosAttribute),
             Type _ when targetType == typeof(Database) => CreateCosmosClient<Database>(cosmosAttribute),
             Type _ when targetType == typeof(Container) => CreateCosmosClient<Container>(cosmosAttribute),
-            _ => await CreateTargetObjectAsync(targetType, cosmosAttribute)
+            _ => ConvertDynamic(await CreateTargetObjectAsync(targetType, cosmosAttribute), targetType)
         };
 
-        private async Task<object> CreateTargetObjectAsync(Type targetType, CosmosDBInputAttribute cosmosAttribute)
+        private object ConvertDynamic(dynamic data, Type targetType)
```
jviau, line 87: Does this work? Since `CreateTargetObjectAsync` returns `Task<object>`

```suggestion
        private object ConvertDynamic(object data, Type targetType)
```

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
```diff
@@ -80,47 +81,42 @@ private CosmosDBInputAttribute GetBindingDataContent(ModelBindingData bindingDat
...
-        private async Task<object> CreatePOCOAsync<T>(Container container, CosmosDBInputAttribute cosmosAttribute)
+        private async Task<object> CreatePOCOAsync(Container container, CosmosDBInputAttribute cosmosAttribute)
         {
             if (String.IsNullOrEmpty(cosmosAttribute.Id) || String.IsNullOrEmpty(cosmosAttribute.PartitionKey))
             {
                 throw new InvalidOperationException("The 'Id' and 'PartitionKey' properties of a CosmosDB single-item input binding cannot be null or empty.");
             }
 
-            ItemResponse<T> item = await container.ReadItemAsync<T>(cosmosAttribute.Id, new PartitionKey(cosmosAttribute.PartitionKey));
+            ItemResponse<dynamic> item = await container.ReadItemAsync<dynamic>(cosmosAttribute.Id, new PartitionKey(cosmosAttribute.PartitionKey));
```
jviau, line 119: I think we can use `ReadItemStreamAsync`, pass back the `Stream` from `ResponseMessage.Content` and deserialize that via `JsonSerializer`. This should be more performant than using `dynamic`.

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
```diff
@@ -148,11 +144,11 @@ private async Task<object> CreatePOCOCollectionAsync<T>(Container container, Cos
             QueryRequestOptions queryRequestOptions = new();
             if (!String.IsNullOrEmpty(cosmosAttribute.PartitionKey))
             {
-                var partitionKey =  new PartitionKey(cosmosAttribute.PartitionKey);
+                var partitionKey = new PartitionKey(cosmosAttribute.PartitionKey);
                 queryRequestOptions = new() { PartitionKey = partitionKey };
             }
 
-            using (var iterator = container.GetItemQueryIterator<T>(queryDefinition: queryDefinition, requestOptions: queryRequestOptions))
+            using (var iterator = container.GetItemQueryIterator<dynamic>(queryDefinition: queryDefinition, requestOptions: queryRequestOptions))
```
jviau, line 151: Similarly here, can use `GetItemQueryStreamIterator`

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
liliankasem: I didn't even think to check for a Stream method - nice :D

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
liliankasem: Switched to stream by I'm still doing serialize to json string -> then deserialize to target type via the "ConvertObject" method. I don't really love this and was wondering if it would be better to deserialize every document to target type when building the list i.e.

```csharp
        private async Task<IList<object>> ExtractCosmosDocumentsAsync(FeedIterator iterator)
        {
            var documentList = new List<Stream>();
            while (iterator.HasMoreResults)
            {
                ResponseMessage response = await iterator.ReadNextAsync();
                var doc = return JsonSerializer.Deserialize(response.Content, targetType);
                documentList.Add(doc);
            }
            return documentList;
        }
``` 

But the serialize/deserialize approach felt less heavy?

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
```diff
@@ -80,57 +82,52 @@ private CosmosDBInputAttribute GetBindingDataContent(ModelBindingData bindingDat
             Type _ when targetType == typeof(CosmosClient) => CreateCosmosClient<CosmosClient>(cosmosAttribute),
             Type _ when targetType == typeof(Database) => CreateCosmosClient<Database>(cosmosAttribute),
             Type _ when targetType == typeof(Container) => CreateCosmosClient<Container>(cosmosAttribute),
-            _ => await CreateTargetObjectAsync(targetType, cosmosAttribute)
+            _ => ConvertObject(await CreateTargetObjectAsync(targetType, cosmosAttribute), targetType)
         };
 
-        private async Task<object> CreateTargetObjectAsync(Type targetType, CosmosDBInputAttribute cosmosAttribute)
+        private object ConvertObject(object data, Type targetType)
```
jviau, line 88: Moving to `Stream` will require using stream-specific JSON apis, we cannot treat it as object. But that is simple, can make this change:

``` CSharp
private async Task<object?> ConvertObjectAsync(Stream stream, Type targetType)
{
    using (stream) // I think we need to dispose stream when we are done with it.
    {
        return await JsonSerializer.DeserializeAsync(stream, targetType, _serializerOptions);
    }
}
```

liliankasem: We definitely do need to dispose, thanks for the reminder

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
jviau: I would look into using an `IAsyncEnumerable<Stream>` with the `yield` keyword. This would let you avoid have to realize the entire query right away into a `List`.

extensions/Worker.Extensions.Shared/Extensions/TypeExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// </summary>
+        public static bool TryGetCollectionElementType(this Type type, out Type? elementType)
+        {
+            if (!type.IsCollectionType())
+            {
+                elementType = null;
+                return false;
+            }
+
+            elementType = type.IsArray ? type.GetElementType() : type.GenericTypeArguments[0];
```
jviau, line 46: This will throw if I provide a type like so:

``` CSharp
public class MyList : List<string>
{
}
```

It will pass `IsCollectionType`, but `type.GenericTypeArguments[0]` will throw index out of bounds.

You will need to walk the type hierarchy until you get to either `IList<T>`, `ICollection<T>`, or `IEnumerable<T>` and then get the first generic argument.

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
```diff
@@ -124,10 +127,31 @@ private async Task<Stream> CreatePOCOAsync(Container container, CosmosDBInputAtt
                 throw new InvalidOperationException($"Unable to retrieve document with ID '{cosmosAttribute.Id}' and PartitionKey '{cosmosAttribute.PartitionKey}'");
             }
 
-            return item.Content;
+            return JsonSerializer.DeserializeAsync(item.Content, targetType, _serializerOptions)!;
+        }
+
+        private async Task<IList> CreateListAsync(Container container, CosmosDBInputAttribute cosmosAttribute, Type elementType, Type targetType)
+        {
+            var resultType = typeof(List<>).MakeGenericType(elementType);
```
jviau, line 135: Do we intend to support other concrete collection types? Such as `HashSet<T>`? If not the failure is going to be ambiguous - `TryGetCollectionElementType` will pass, but then the cast later on will fail. I suggest we do one of two things:

1. if we don't intend to support the other collection types, we should make that more explicit and fail if the collection type is not explicitly in the set of ones we support.
2. if we do intend to support them, then you would need to branch this logic:
    1. if `targetType` is a concrete type, we can activate it and dynamically call the `.Add` method.
    2. if `targetType` is not concrete, we will provide `List<T>` (I think that covers all collection interfaces we would support? Need to verify that).

jviau (changes requested):

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
liliankasem: I don't know if we place to support more collection types any time soon, we can go with option #2 just in case. Or just assume #1 for now and refactor for #2 when we get to that work item?

extensions/Worker.Extensions.Shared/Extensions/TypeExtensions.cs
liliankasem: Addresses in last commit - let me know what you think!

jviau (dismissed):

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
liliankasem: I think what I currently have implemented is #2; if that doesn't look right, I can change this to be option #1 and leave it there for now

extensions/Worker.Extensions.Shared/Extensions/TypeExtensions.cs
```diff
@@ -0,0 +1,82 @@
...
+            if (type.IsArray)
+            {
+                elementType = type.GetElementType();
+                return true;
+            }
+
+            // Traverse the inheritance hierarchy to find the first generic interface
+            while (type != typeof(object) || type is not null)
+            {
+                var interfaceType = type.GetInterfaces().FirstOrDefault(t => t.IsGenericType
```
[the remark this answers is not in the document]
liliankasem: You're right, it returns `typeof(int)` - what would be the best way to approach this specific scenario? Moving the `IsGeneric` check down and having a split branch i.e.

```csharp
var interfaceType = type.GetInterfaces().FirstOrDefault((t == typeof(IEnumerable) || t == typeof(ICollection) || t == typeof(IList)));

if (interfaceType is not null)
{
    if (interfaceType.IsGenericType)
    {
        // do a?
    }
    else
    {
        // do b?
    }
}
```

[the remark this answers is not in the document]
liliankasem: No that doesn't make sense either - I'm stumped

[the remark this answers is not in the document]
jviau: You can do something like: if generic, get generic type definition, check if that is `IEnumerable<>`, `ICollection<>` or `IList<>`. If so, then pull off the first generic argument.

https://learn.microsoft.com/en-us/dotnet/api/system.type.getgenerictypedefinition?view=net-7.0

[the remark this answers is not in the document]
liliankasem: Fixed! I realize now why I had a hard time with this, I didn't have "using System.Collections.Generic;" and the error message was not clear as to why `== typeof(IEnumerable<>)` was not okay 😔 Anywho, this case now correctly gets string instead of int

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
```diff
@@ -85,52 +88,104 @@ private CosmosDBInputAttribute GetBindingDataContent(ModelBindingData bindingDat
...
+            if (targetType.TryGetCollectionElementType(out Type? elementType))
+            {
+                if (elementType is null)
+                {
+                    throw new ArgumentNullException(nameof(elementType));
+                }
+
+                if (targetType.IsConcreteType())
+                {
+                    return await CreatePocoListConcreteAsync(container, cosmosAttribute, elementType, targetType);
```
jviau, line 107: Will this work for arrays? I think `IsConcereteType()` will return `true` for that

liliankasem: I will do some E2E testing, it's been a while since all the last couple changes 😞

liliankasem: Is it okay to do this `if (targetType.IsConcreteType() && !targetType.IsArray)`? And let arrays fall into the other branch? 

This is mostly working right now, the biggest issue is that all of the streams once deserialized have null properties :/ So trying to figure out why that might be happening

liliankasem: Ended up figuring this out and currently in deserialization hell. Turns out the content of "stream" is a json object with an array of documents so it's not so easy to just deserialize to a list/target type. E.g.:

```json
{{
  "_rid": "ltwyAJGNSgs=",
  "Documents": [
    {
      "id": "1",
      "description": "take out the rubbish",
      "_rid": "ltwyAJGNSgsBAAAAAAAAAA==",
      "_self": "dbs/ltwyAA==/colls/ltwyAJGNSgs=/docs/ltwyAJGNSgsBAAAAAAAAAA==/",
      "_etag": "\"0f00f247-0000-0700-0000-63bc7ae60000\"",
      "_attachments": "attachments/",
      "_ts": 1673296614
    }
  ],
  "_count": 1
}}
```

So will need to extract Documents from the stream first, then deserialize that to a list. And will also need to do AddRange and not Add (like I was doing before this refactor using `GetItemQueryIterator` i.e.

```cs
// pre refactor code
private async Task<IList<T>> ExtractCosmosDocumentsAsync<T>(FeedIterator<T> iterator)
        {
            var documentList = new List<T>();
            while (iterator.HasMoreResults)
            {
                FeedResponse<T> response = await iterator.ReadNextAsync();
                documentList.AddRange(response.Resource);
            }
            return documentList;
        }
```

jviau: Resolved this by providing a type to extract the `Document` property.

jviau:
@fabiocav this is ready for review now. I started the work of extracting common collection binding helpers that we can hopefully apply to other extensions in time.

extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs
```diff
@@ -125,70 +106,35 @@ private async Task<object> CreateTargetObjectAsync(Type targetType, CosmosDBInpu
...
+            if (string.IsNullOrEmpty(cosmosAttribute.Id) || string.IsNullOrEmpty(cosmosAttribute.PartitionKey))
             {
                 throw new InvalidOperationException("The 'Id' and 'PartitionKey' properties of a CosmosDB single-item input binding cannot be null or empty.");
             }
 
             ResponseMessage item = await container.ReadItemStreamAsync(cosmosAttribute.Id, new PartitionKey(cosmosAttribute.PartitionKey));
-
-            if (item is null || item.StatusCode is not System.Net.HttpStatusCode.OK || item.Content is null)
-            {
-                throw new InvalidOperationException($"Unable to retrieve document with ID '{cosmosAttribute.Id}' and PartitionKey '{cosmosAttribute.PartitionKey}'");
```
liliankasem: By removing this error, the unit tests now fail with a null ref exception instead:

`https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=158152&view=logs&j=1372d9e0-5cd3-5ef5-5e82-7ce7a218d320&t=89898c79-b51f-5488-1a35-13c9114bcb3d&l=218`

```
Expected: Unable to retrieve document with ID '1' a···
Actual:   Object reference not set to an instance o···
```

I think specifically removing the `item.Content is null` part of this check is important - I would suggest we bring back the null check and more intuitive error message.

liliankasem: Note: I think these are the only test failing right now, so should be good to go once this is addressed as they are all expecting `Unable to retrieve document with ID '{cosmosAttribute.Id}' and PartitionKey '{cosmosAttribute.PartitionKey}'`:

- ConvertAsync_POCO_ItemResponseNull_ThrowsException_ReturnsFailure
- ConvertAsync_ItemResponse_ResourceIsNull_ThrowsException_ReturnsFailed
- ConvertAsync_POCO_IdProvided_StatusNot200_ThrowsException_ReturnsFailure

jviau: I'll have to update the unit tests, as this call returning `null` is not how it will behave live.

jviau: - `ConvertAsync_POCO_ItemResponseNull_ThrowsException_ReturnsFailure` -> removed, we can trust the Cosmos SDK will not return null. If it does, something is really wrong
- `ConvertAsync_ItemResponse_ResourceIsNull_ThrowsException_ReturnsFailed` -> removed, either one of two things is happening: 1) the stream `.Content` is null, which we can trust won't happen - something is wrong with the Cosmos SDK here. Or 2) the stream is not null, but it deserializes to `null`. We should not verify this behavior, it may be a valid (custom converter that returns `null` for example).
- `ConvertAsync_POCO_IdProvided_StatusNot200_ThrowsException_ReturnsFailure` -> updated to be a `CosmosException`. This exception will be more flushed out in a live run, but it validates we at least throw in the non-success cases.

jviau (commented):
resetting my review.

test/Worker.Extensions.Shared.Tests/Reflection/ParameterBinderTests.cs
```diff
@@ -86,6 +98,19 @@ private static async IAsyncEnumerable<object> GetIntEnumerable(Type t)
             }
         }
 
+        private static async IAsyncEnumerable<object> GetPocoArray(Type t)
```
jviau, line 101: this method was not really needed - could use `GetIntEnumerable` still. But no harm leaving it in if you want.

jviau (approved):

PR #1933, opened 2023-09-26
Send env reload response with "Failure" as Status property value when app payload is not present

### LINKED ISSUE #1931: Handle specialization without app payload
> The instance (host) would still need to handle the specialization (assignment). The native worker should still fail that specialization, but the fallback should behave as expected in scenarios where we don't have a suitable payload for the native worker, which in this case, would be to index 0 functions and handle the empty payload as we've always done (after the fix to enable portal creation, that is)

kshyju:
Fixes #1931

Currently, If we get specialization request without a valid payload present in the `FunctionAppDirectory` , our code crashes when trying to load the assembly. This PR updates that code path so that it will send an instance of EnvironmentReloadResponse with "Failure" as the "Status" property value.

### DIFF
--- a/host/src/FunctionsNetHost/Grpc/IncomingGrpcMessageHandler.cs
+++ b/host/src/FunctionsNetHost/Grpc/IncomingGrpcMessageHandler.cs
@@ -50,13 +50,20 @@ namespace FunctionsNetHost.Grpc
                     Logger.LogTrace("Specialization request received.");
 
                     var envReloadRequest = msg.FunctionEnvironmentReloadRequest;
+                    var applicationExePath = PathUtils.GetApplicationExePath(envReloadRequest.FunctionAppDirectory);
+                    Logger.LogTrace($"application path {applicationExePath}");
+
+                    if (string.IsNullOrWhiteSpace(applicationExePath))
+                    {
+                        var ex = new InvalidOperationException($"Unable to find a valid function app payload at '{envReloadRequest.FunctionAppDirectory}'");
+                        responseMessage.FunctionEnvironmentReloadResponse = BuildFailedEnvironmentReloadResponse(ex);
+                        break;
+                    }
+
                     foreach (var kv in envReloadRequest.EnvironmentVariables)
                     {
                         EnvironmentUtils.SetValue(kv.Key, kv.Value);
                     }
-
-                    var applicationExePath = PathUtils.GetApplicationExePath(envReloadRequest.FunctionAppDirectory);
-                    Logger.LogTrace($"application path {applicationExePath}");
 
 #pragma warning disable CS4014
                     Task.Run(() =>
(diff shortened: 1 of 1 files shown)


kshyju:
@paulbatum @TsuyoshiUshio FYI

host/src/FunctionsNetHost/Grpc/IncomingGrpcMessageHandler.cs
```diff
@@ -50,14 +50,21 @@ private async Task Process(StreamingMessage msg)
                     Logger.LogTrace("Specialization request received.");
 
                     var envReloadRequest = msg.FunctionEnvironmentReloadRequest;
+                    var applicationExePath = PathUtils.GetApplicationExePath(envReloadRequest.FunctionAppDirectory);
+                    Logger.LogTrace($"application path {applicationExePath}");
+
+                    if (string.IsNullOrWhiteSpace(applicationExePath))
+                    {
+                        var ex = new InvalidOperationException($"Unable to find a valid function app payload at '{envReloadRequest.FunctionAppDirectory}'");
+                        responseMessage.FunctionEnvironmentReloadResponse = BuildFailedEnvironmentReloadResponse(ex);
```
brettsam, line 59: I'll admit I don't know how the host handles this. I assume it logs the exception and kills the host? Or does it retry?

kshyju: We log an error and set call the `SetException` method on TCS.
https://github.com/Azure/azure-functions-host/blob/c85df40731c1862d3b7b79fc0537044b5426c247/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L377C46-L377C46
The grpc channel to worker continue to stay open.

liliankasem (approved):

host/src/FunctionsNetHost/Grpc/IncomingGrpcMessageHandler.cs
brettsam: This error will show up in the customer's App Insights, right?

kshyju: No. This does not go to customer logs.

kshyju: > We log an error and set call the `SetException` method on TCS. https://github.com/Azure/azure-functions-host/blob/c85df40731c1862d3b7b79fc0537044b5426c247/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L377C46-L377C46 The grpc channel to worker continue to stay open.

Found out that host code is not gracefully handling this case. Let me follow up tomorrow.

kshyju: I will be making a change in the host which will check whether we have a valid payload before initiating the language worker specialization.

The change in this PR is still valid because without this change, if the worker config file does not exist in the payload, we used to crash the process. with this change, it will not crash, but sends an env reload response with "Faiulre" as status. Host will shutdown the application in that case.

PR #1937, opened 2023-09-29
Pass RestoreSources to inner-build restore task

### LINKED ISSUE #1863: WorkerExtensions.csproj does not use the repo's NuGet.config due to being in temp dir
It appears that WorkerExtensions.csproj is generated under the temp dir and is attempted to be restored at build time. Because it's in the temp dir, it will not pick up the repo's NuGet.config and so will try to restore from nuget.org by default. This causes problems if either there is a firewall to the public internet, or if there are packages in use which are not on nuget.org.

If this temp project much be generated, it should be under the `$(IntermediateOutputPath)` instead. This has added benefits of not placing files outside the repo entirely which makes a "git clean" unproductive and slowly fills the disk over time.

jviau:

resolves #1863

### Pull request checklist

### Additional information

This is a more explicit fix to the linked issue. It is already addressed by moving the `WorkerExtension.csproj` out of the temp directory and into intermediate output dir (so it would inherit nuget.config), but I prefer directly supplying the nuget sources to use.

### DIFF
--- a/sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
+++ b/sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
@@ -151,7 +151,13 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
 
   <!-- Invoke restore on WorkerExtension.csproj -->
   <Target Name="_WorkerExtensionsRestore">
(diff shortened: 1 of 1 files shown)


jviau:
The CI linked to this was corrupted, a manual run succeeded: https://dev.azure.com/azfunc/Azure%20Functions/_build/results?buildId=154877&view=results

liliankasem (approved):

PR #1945, opened 2023-10-04
Service Bus settlement support

### LINKED ISSUE #1538: Enhance Service Bus binding to enable calling service APIs
Issue to track enhancements to the Service Bus binding to enable use of SDK APIs from the .NET Worker binding targets.

JoshLove-msft:

resolves #1538

### Pull request checklist

### Additional information

Additional PR information


extensions/Worker.Extensions.ServiceBus/src/Properties/AssemblyInfo.cs
```diff
@@ -4,5 +4,5 @@
 using System.Runtime.CompilerServices;
 using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
 
-[assembly: ExtensionInformation("Microsoft.Azure.WebJobs.Extensions.ServiceBus", "5.12.0")]
+[assembly: ExtensionInformation("Microsoft.Azure.WebJobs.Extensions.ServiceBus", "5.13.0-alpha.20231004.1")]
```
JoshLove-msft, line 7: This will need to wait to merge until a non-alpha version is available.

extensions/Worker.Extensions.ServiceBus/src/Worker.Extensions.ServiceBus.csproj
```diff
@@ -15,17 +15,24 @@
   <Import Project="..\..\..\build\Extensions.props" />
 
   <ItemGroup>
-    <PackageReference Include="Azure.Messaging.ServiceBus" Version="7.14.0" />
+    <PackageReference Include="Azure.Messaging.ServiceBus" Version="7.16.1" />
     <PackageReference Include="Microsoft.Extensions.Azure" Version="1.6.3" />
+    <PackageReference Include="Google.Protobuf" Version="3.23.4" />
+    <PackageReference Include="Grpc.Tools" Version="2.56.2" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Rpc" Version="1.0.0" />
```
JoshLove-msft, line 22: The package ref can be removed.

JoshLove-msft:
Still need to add tests/samples

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActions.cs
```diff
@@ -0,0 +1,168 @@
...
+using System.Threading.Tasks;
+using Azure.Messaging.ServiceBus;
+using Google.Protobuf.WellKnownTypes;
+using Microsoft.Azure.Functions.Worker.Converters;
+using Microsoft.Azure.ServiceBus.Grpc;
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    [InputConverter(typeof(ServiceBusMessageActionsConverter))]
+    public class ServiceBusMessageActions
```
JoshLove-msft, line 13: Need a protected ctor for mocking.

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActions.cs
```diff
@@ -0,0 +1,168 @@
...
+        {
+            _settlement = settlement;
+        }
+
+        ///<inheritdoc cref="ServiceBusReceiver.CompleteMessageAsync(ServiceBusReceivedMessage, CancellationToken)"/>
+        public virtual async Task CompleteMessageAsync(
+            ServiceBusReceivedMessage message,
+            CancellationToken cancellationToken = default)
+        {
+            await _settlement.CompleteAsync(new() { Locktoken = message.LockToken}, cancellationToken: cancellationToken);
```
jviau, line 27:
```suggestion
            await _settlement.CompleteAsync(new() { Locktoken = message.LockToken}, cancellationToken: cancellationToken );
```

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActions.cs
```diff
@@ -0,0 +1,168 @@
...
+            string? deadLetterReason = default,
+            string? deadLetterErrorDescription = default,
+            CancellationToken cancellationToken = default)
+        {
+            var request = new DeadletterRequest()
+            {
+                Locktoken = message.LockToken,
+                PropertiesToModify = { TransformProperties(propertiesToModify) }
+            };
+            if (deadLetterReason != null)
```
jviau, line 57: is this check necessary? Is there any issue assigning regardless of null or not? Same with line 62..65

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActions.cs
jviau, line 27: Consider performing argument validation throughout this class to avoid null-refs

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActions.cs
```diff
@@ -0,0 +1,168 @@
...
+        {
+            var converted = new Dictionary<string, SettlementProperties>();
+            if (properties == null)
+            {
+                return converted;
+            }
+            // support all types listed here - https://learn.microsoft.com/en-us/dotnet/api/azure.messaging.servicebus.servicebusmessage.applicationproperties?view=azure-dotnet
+            foreach (var kvp in properties)
+            {
+                switch (kvp.Value)
```
jviau, line 93: Small organization suggestion: use new switch statements:

``` CSharp
SettlementProperties p = kvp.Value switch
{
    string s => new() { StringValue = s },
    bool b => new() { BoolValue = b },
   ....
};

converted.Add(kvp.Key, p);

```

extensions/Worker.Extensions.ServiceBus/src/Proto/settlement.proto
```diff
@@ -0,0 +1,66 @@
...
+
+// The defer message request containing the locktoken and properties to modify.
+message DeferRequest {
+  string locktoken = 1;
+  map<string, SettlementProperties> propertiesToModify = 2;
+}
+
+// The settlement property can be of any type listed below, which 
+// corresponds to the types specified in 
+// https://learn.microsoft.com/en-us/dotnet/api/azure.messaging.servicebus.servicebusmessage.applicationproperties?view=azure-dotnet#remarks
```
jsquire, line 51:
```suggestion
// https://learn.microsoft.com/dotnet/api/azure.messaging.servicebus.servicebusmessage.applicationproperties?view=azure-dotnet#remarks
```

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActionsConverter.cs
```diff
@@ -0,0 +1,35 @@
...
+            _settlement = settlement;
+        }
+
+        public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
+        {
+            try
+            {
+                return new ValueTask<ConversionResult>(ConversionResult.Success(new ServiceBusMessageActions(_settlement)));
+            }
+            catch (Exception exception)
```
jviau, line 29: Just a note to myself: is this expected for all input converters to add such trivial exception handling? If line 27 ever failed it would mean there is something seriously wrong. Is it better to let that exception bubble up?

extensions/Worker.Extensions.ServiceBus/src/Worker.Extensions.ServiceBus.csproj
```diff
@@ -15,17 +15,24 @@
   <Import Project="..\..\..\build\Extensions.props" />
 
   <ItemGroup>
-    <PackageReference Include="Azure.Messaging.ServiceBus" Version="7.14.0" />
+    <PackageReference Include="Azure.Messaging.ServiceBus" Version="7.16.1" />
     <PackageReference Include="Microsoft.Extensions.Azure" Version="1.6.3" />
+    <PackageReference Include="Google.Protobuf" Version="3.23.4" />
+    <PackageReference Include="Grpc.Tools" Version="2.56.2" />
```
jviau, line 21:
```suggestion
    <PackageReference Include="Grpc.Tools" Version="2.56.2" PrivateAssets="All" />
```

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActions.cs
JoshLove-msft: Yes, the generated code throws if you set to null.

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActions.cs
jsquire: I think this suggestion throws the space in the wrong spot.  It would be the curly brace right after `message.LockToken` rather than the closing paren, no?

extensions/Worker.Extensions.ServiceBus/src/Proto/settlement.proto
```diff
@@ -0,0 +1,66 @@
...
+message AbandonRequest {
+  string locktoken = 1;
+  map<string, SettlementProperties> propertiesToModify = 2;
+}
+
+// The deadletter message request containing the locktoken and properties to modify along with the reason/description.
+message DeadletterRequest {
+  string locktoken = 1;
+  map<string, SettlementProperties> propertiesToModify = 2;
+  optional string deadletterReason = 3;
```
jviau, line 39: recommend using `google.protobuf.StringValue` instead, then you can set to `null` in code.

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActions.cs
jviau: Ah left a comment on the proto file - should use `google.protobuf.StringValue` instead.

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActions.cs
jviau: Edited the suggestion.

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActionsConverter.cs
fabiocav: In this particular case, I would expect is to let this exception bubble up.

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActions.cs
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ServiceBusMessageActions"/> class for mocking use in testing.
+        /// </summary>
+        /// <remarks>
+        /// This constructor exists only to support mocking. When used, class state is not fully initialized, and
+        /// will not function correctly; virtual members are meant to be mocked.
+        ///</remarks>
+        protected ServiceBusMessageActions()
```
fabiocav, line 29: This is not currently being leveraged. Considered removing?

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActions.cs
```diff
@@ -0,0 +1,134 @@
...
+            await _settlement.DeferAsync(request, cancellationToken: cancellationToken);
+        }
+
+        private static Dictionary<string, SettlementProperties> TransformProperties(IDictionary<string, object>? properties)
+        {
+            var converted = new Dictionary<string, SettlementProperties>();
+            if (properties == null)
+            {
+                return converted;
+            }
```
fabiocav, line 94: Can we modify this method to return an `IDictionary<string, SettlementProperties>` and make the following modification?

``` csharp
 if (properties == null)
 {
     return ImmutableDictionary<string, SettlementProperties>.Empty;
 }

 var converted = new Dictionary<string, SettlementProperties>();
```

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActions.cs
```diff
@@ -0,0 +1,134 @@
...
+            // support all types listed here - https://learn.microsoft.com/en-us/dotnet/api/azure.messaging.servicebus.servicebusmessage.applicationproperties?view=azure-dotnet
+            foreach (var kvp in properties)
+            {
+                SettlementProperties settlementProperties = kvp.Value switch
+                {
+                    string stringValue => new SettlementProperties() { StringValue = stringValue },
```
fabiocav, line 100: Debating whether creating implicit custom converters would help make this a bit more readable and maintainable

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActions.cs
```diff
@@ -0,0 +1,134 @@
...
+                    Guid guidValue => new SettlementProperties() { StringValue = guidValue.ToString() },
+                    DateTimeOffset dateTimeOffsetValue => new SettlementProperties()
+                        { TimestampValue = Timestamp.FromDateTimeOffset(dateTimeOffsetValue) },
+                    // proto does not support DateTime, so use Timestamp from google.protobuf
+                    DateTime dateTimeValue => new SettlementProperties() { TimestampValue = Timestamp.FromDateTimeOffset(dateTimeValue) },
+                    // proto does not support Uri, so use string
+                    Uri uriValue => new SettlementProperties() { StringValue = uriValue.ToString() },
+                    // proto does not support TimeSpan, so use string
+                    TimeSpan timeSpanValue => new SettlementProperties() { StringValue = timeSpanValue.ToString() },
+                    _ => throw new NotSupportedException($"Unsupported property type {kvp.Value.GetType()}"),
```
fabiocav, line 126: Is there any additional information we can provide that would help the end user understand/fix the problem with this exception?

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActions.cs
JoshLove-msft: This is to support mocking.

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActions.cs
JoshLove-msft: I've changed the implementation to no longer use maps in the proto file. I will update this PR shortly.

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActionsConverter.cs
JoshLove-msft: My understanding was that catching and returning a failed result was the recommended pattern. This is what we do for the other SDK type converters. 
/cc @liliankasem

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActions.cs
JoshLove-msft: Added

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActions.cs
JoshLove-msft: No longer applies.

extensions/Worker.Extensions.ServiceBus/src/Proto/settlement.proto
```diff
@@ -30,38 +29,19 @@ message CompleteRequest {
...
 }
 
 // The deadletter message request containing the locktoken and properties to modify along with the reason/description.
 message DeadletterRequest {
   string locktoken = 1;
-  map<string, SettlementProperties> propertiesToModify = 2;
-  google.protobuf.StringValue deadletterReason= 3;
-  google.protobuf.StringValue deadletterErrorDescription = 4;
+  bytes propertiesToModify = 2;
+  string deadletterReason = 3;
```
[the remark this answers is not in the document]
JoshLove-msft: oops - copy + paste from the other repo.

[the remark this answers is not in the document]
JoshLove-msft: Updated mirror version here - https://github.com/Azure/azure-sdk-for-net/pull/39361
Once we move this extension to the azure-sdk-for-net repo we won't have to maintain two copies.

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActionsConverter.cs
liliankasem: Yeah I'm pretty sure every converter will fail with the exception:

e.g. cosmos: https://github.com/Azure/azure-functions-dotnet-worker/blob/7f4e1a1925cb11ec341d02da094dd25908a49e89/extensions/Worker.Extensions.CosmosDB/src/CosmosDBConverter.cs#L58-L61

We should keep the pattern the same for consistency and if this is a behaviour we want to change we should open an issue to make this change to all converters.

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActionsConverter.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Threading.Tasks;
+using Microsoft.Azure.Functions.Worker.Converters;
+using Microsoft.Azure.ServiceBus.Grpc;
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    /// <summary>
+    /// Converter to bind to <see cref="ServiceBusMessageActions" /> type parameter.
+    /// </summary>
+    internal class ServiceBusMessageActionsConverter : IInputConverter
```
liliankasem, line 14: I don't think this uses deferred binding and so does not need the `[SupportsDeferredBinding]` attribute, but just want to double check that's correct?

liliankasem:
I think this all looks good - would love to see some tests and samples added

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActionsConverter.cs
JoshLove-msft: Correct, this is just for binding to the client-like type ServiceBusMessageActions.

JoshLove-msft:
> I think this all looks good - would love to see some tests and samples added

Added!

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActions.cs
fabiocav: Sounds good. Unless I missed it, I don't think tests were in yesterday when asked.

fabiocav (approved):

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActions.cs
JoshLove-msft: Yeah it is actually a pattern we follow so that customers can mock the client types.

liliankasem (approved):

PR #1960, opened 2023-10-12
Sending EnableUserCodeException as a capability of native worker 

kshyju:
- Sending `EnableUserCodeException` as a capability of native worker. - this forces host to log it in a way it surfaces user's AI logs.
- Improving the exception message for env reload failure. - host will log it as it is received.
- Version bump to 1.0.2

### DIFF
--- a/host/src/FunctionsNetHost/Grpc/IncomingGrpcMessageHandler.cs
+++ b/host/src/FunctionsNetHost/Grpc/IncomingGrpcMessageHandler.cs
@@ -56,15 +56,17 @@ namespace FunctionsNetHost.Grpc
 
                     if (workerConfig?.Description is null)
                     {
-                        responseMessage.FunctionEnvironmentReloadResponse = BuildFailedEnvironmentReloadResponse(new FunctionAppPayloadNotFoundException());
+                        Logger.LogTrace($"Could not find a worker config in {envReloadRequest.FunctionAppDirectory}");
+                        responseMessage.FunctionEnvironmentReloadResponse = BuildFailedEnvironmentReloadResponse();
                         break;
                     }
 
                     // function app payload which uses an older version of Microsoft.Azure.Functions.Worker package does not support specialization.
                     if (!workerConfig.Description.CanUsePlaceholder)
                     {
-                        Logger.LogTrace("App payload uses an older version of worker package which does not support specialization.");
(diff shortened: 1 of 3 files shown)


host/src/FunctionsNetHost/Grpc/IncomingGrpcMessageHandler.cs
```diff
@@ -56,15 +56,17 @@ private async Task Process(StreamingMessage msg)
...
                         break;
                     }
 
                     // function app payload which uses an older version of Microsoft.Azure.Functions.Worker package does not support specialization.
                     if (!workerConfig.Description.CanUsePlaceholder)
                     {
-                        Logger.LogTrace("App payload uses an older version of worker package which does not support specialization.");
-                        responseMessage.FunctionEnvironmentReloadResponse = BuildFailedEnvironmentReloadResponse(new EnvironmentReloadNotSupportedException());
+                        Logger.LogTrace("App payload uses an older version of Microsoft.Azure.Functions.Worker SDK which does not support placeholder.");
+                        var e = new EnvironmentReloadNotSupportedException("This app is not using the latest version of Microsoft.Azure.Functions.Worker SDK and therefore does not leverage all performance optimizations. See https://aka.ms/azure-functions/dotnet/placeholders for more information.");
```
kshyju, line 68: Host change will follow which consumes this version - In that host version we will log what the worker sent as it is.

fabiocav (approved):

host/tools/build/Microsoft.Azure.Functions.DotnetIsolatedNativeHost.nuspec
```diff
@@ -4,7 +4,7 @@
     <id>Microsoft.Azure.Functions.DotNetIsolatedNativeHost</id>
     <title>Microsoft Azure Functions dotnet-isolated native host</title>
     <tags>dotnet-isolated azure-functions azure</tags>
-    <version>1.0.1</version>
+    <version>1.0.2</version>
```
brettsam, line 7: Had we released 1.0.1 already?

host/src/FunctionsNetHost/Grpc/IncomingGrpcMessageHandler.cs
brettsam, line 68: It may not be super-important for this scenario, but there won't be a StackTrace with this without throwing it, will there? You can throw/catch the exception to capture the StackTrace correctly.

PR #1968, opened 2023-10-19
Defaulting to source-generation

### LINKED ISSUE #1906: Defaulting of source metadata generator and worker indexing
Once the generator is otherwise ready, we need to cut a version of the SDK which makes the auto-registration on by default, removing the need for this line in project files:
```xml
<FunctionsAutoRegisterGeneratedMetadataProvider>True</FunctionsAutoRegisterGeneratedMetadataProvider>
```

That setting could still be explicitly set to "False" if needed.

Doing so would enable us to do the same thing for the worker indexing, I believe, and we could similarly remove the need for:
```xml
<FunctionsEnableWorkerIndexing>True</FunctionsEnableWorkerIndexing>
```

Again allowing it to be disabled if needed.

satvu:

resolves #1906 

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/sdk/Sdk.Generators/Constants.cs
+++ b/sdk/Sdk.Generators/Constants.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
         {
             internal const string MSBuildRootNamespace = "build_property.RootNamespace";
             internal const string GeneratedCodeNamespace = "build_property.FunctionsGeneratedCodeNamespace";
-            internal const string EnableSourceGen = "build_property.FunctionsEnableWorkerIndexing";
+            internal const string EnableSourceGen = "build_property.FunctionsEnableMetadataSourceGen";
             internal const string EnablePlaceholder = "build_property.FunctionsEnableExecutorSourceGen";
             internal const string AutoRegisterGeneratedFunctionsExecutor = "build_property.FunctionsAutoRegisterGeneratedFunctionsExecutor";
             internal const string AutoRegisterGeneratedMetadataProvider = "build_property.FunctionsAutoRegisterGeneratedMetadataProvider";
--- a/sdk/Sdk.Generators/Constants.cs
+++ b/sdk/Sdk.Generators/Constants.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
         {
             internal const string MSBuildRootNamespace = "build_property.RootNamespace";
             internal const string GeneratedCodeNamespace = "build_property.FunctionsGeneratedCodeNamespace";
-            internal const string EnableSourceGen = "build_property.FunctionsEnableMetadataSourceGen";
+            internal const string EnableSourceGen = "build_property.FunctionsEnableWorkerIndexing";
             internal const string EnablePlaceholder = "build_property.FunctionsEnableExecutorSourceGen";
             internal const string AutoRegisterGeneratedFunctionsExecutor = "build_property.FunctionsAutoRegisterGeneratedFunctionsExecutor";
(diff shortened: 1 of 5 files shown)


sdk/Sdk.Generators/Constants.cs
```diff
@@ -14,7 +14,7 @@ internal static class BuildProperties
         {
             internal const string MSBuildRootNamespace = "build_property.RootNamespace";
             internal const string GeneratedCodeNamespace = "build_property.FunctionsGeneratedCodeNamespace";
-            internal const string EnableSourceGen = "build_property.FunctionsEnableMetadataSourceGen";
+            internal const string EnableSourceGen = "build_property.FunctionsEnableWorkerIndexing";
```
satvu, line 17: Previously we had both `FunctionsEnableWorkerIndexing` and `FunctionsEnableMetadataSourceGen` since it is possible to use worker-indexing but still create and read from the `functions.json` file.  I think we can use just one prop as we only really want source-generated metadata if worker-indexing is enabled, and this first iteration reflects that.

satvu:
Still doing some local validation/testing as well, keeping in draft.

sdk/Sdk.Generators/Constants.cs
jviau: Did we already ship with this value? If so, we would need to fill the gap in behavior. We can rename it sure, but in the targets we should map the old property to the new one if the new one is not set yet.

kshyju: we only have [official docs](https://learn.microsoft.com/en-us/azure/azure-functions/dotnet-isolated-process-guide#performance-optimizations) for using `FunctionsEnableWorkerIndexing` prop.

jviau: Official docs or not, if we shipped it in a non-preview package we shouldn't break it.

sdk/release_notes.md
```diff
@@ -6,13 +6,12 @@
 
 ### Microsoft.Azure.Functions.Worker.Sdk <version> (meta package)
 
-- <entry>
+- Default to source-generated function metadata (#1968)
```
kshyju, line 10: Should we also mention something like "If you had `<FunctionsEnableWorkerIndexing>True</FunctionsEnableWorkerIndexing>` in your csproj, you can remove it after upgrading to this version"  in the release notes when we release this version of SDK ?

kshyju (approved):

sdk/release_notes.md
kshyju: Do you plan to include the version(`MinorProductVersion`) bump in a this PR itself?

satvu: @kshyju I will add that comment. I will include the version bump in the release prep PR for whatever package ends up having this change.

jviau (approved):
I see you reverted the changes I was concerned about in the earlier iteration (as they could be breaking). If you want to still make a change to use the new properties within the generator (but not necessary).

If you want to and avoid breaking changes, what I would do is:

1. Refactor targets to your desired state (with breaking changes).
2. Add back in the old properties and map them into the new ones as appropriate _only if the new one is not already set by the customer_.

satvu:
Blocked by https://github.com/Azure/azure-functions-host/issues/9536 - I don't believe we can release this without the release of this host change.

kshyju:
> Blocked by [Azure/azure-functions-host#9536](https://github.com/Azure/azure-functions-host/issues/9536) - I don't believe we can release this without the release of this host change.

Are we really blocked on that PR? I am positive the setting is enabled on stamp level for all stamps. @soninaren Can you confirm.

satvu:
Confirmed offline that this is not blocked - host PR for defaulting is not done, but all stamps have worker-indexing enabled so we can proceed.

PR #1984, opened 2023-10-23
Revert "Updating method used to retrieve a blob stream"

fabiocav:
Reverts Azure/azure-functions-dotnet-worker#1976


fabiocav:
Temporarily reverting this to inspect failures.

liliankasem (approved):

PR #1987, opened 2023-10-24
Added default encoding in case it is not specified by the caller

### LINKED ISSUE #1986: ReadAsStringAsync does not supply a default encoding
Issue:
Calling `Microsoft.Azure.Functions.Worker.Http.HttpRequestDataExtensions.ReadAsStringAsync` without specifying the encoding results in a `NullReferenceException`.

The documentation for `HttpRequestDataExtensions.ReadAsStringAsync` states that the `encoding` parameter defaults to UTF-8. 

Expected behaviour
A call such as this: `await req.ReadAsStringAsync()` works and UTF-8 as the default encoding.

endizhupani:

resolves #1986 

### Pull request checklist

### DIFF
--- a/src/DotNetWorker.Core/Http/HttpRequestDataExtensions.cs
+++ b/src/DotNetWorker.Core/Http/HttpRequestDataExtensions.cs
@@ -5,12 +5,11 @@ using System;
 using System.IO;
 using System.Net;
 using System.Text;
-using System.Threading.Tasks;
-using Microsoft.Extensions.Options;
-using Microsoft.Extensions.DependencyInjection;
-using Azure.Core.Serialization;
 using System.Threading;
-using System.Text.Json;
+using System.Threading.Tasks;
+using Azure.Core.Serialization;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.Azure.Functions.Worker.Http
 {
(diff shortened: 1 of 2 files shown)


endizhupani:

src/DotNetWorker.Core/Http/HttpRequestDataExtensions.cs
```diff
@@ -37,7 +36,7 @@ public static class HttpRequestDataExtensions
                 return null;
             }
 
-            using (var reader = new StreamReader(request.Body, bufferSize: 1024, detectEncodingFromByteOrderMarks: true, encoding: encoding, leaveOpen: true))
+            using (var reader = new StreamReader(request.Body, bufferSize: 1024, detectEncodingFromByteOrderMarks: true, encoding: encoding ?? Encoding.UTF8, leaveOpen: true))
```
kshyju, line 39: For non netfx case, we should not send the default value here, instead we should let the `StreamReader` constructor handle it (as it does today).

For netfx, let's override the value to Encoding.UTF8.

fabiocav: I'll mark this as resolved based on what we've discussed. I think we're fine with the behavior here as it would be consistent with the current implementation.

fabiocav (approved):

PR #1993, opened 2023-10-26
Adding "global::" namespace prefix to avoid compiler misintepretation for function classes with same name as its namespace

### LINKED ISSUE #1990: `FunctionsEnableExecutorSourceGen` introduces build errors if project, namespace, and function class share the same name
## Summary
If the name of a function class is the same with the namespace **AND** the project it resides in, enabling `FunctionsEnableExecutorSourceGen` will generate the following build error in generated source code:
> The type name 'Name' does not exist in the type 'Name'

## Minimal reproducible sample
In a project named `FunctionCodeGen`:
```cs
using System.Net;
using Microsoft.Azure.Functions.Worker;

namespace FunctionCodeGen
{
    public class FunctionCodeGen
    {
        [Function("FunctionCodeGen")]
        public HttpResponseData Run([HttpTrigger(AuthorizationLevel.Function, "get")] HttpRequestData req) => req.CreateResponse(HttpStatusCode.OK);
    }
}
```
```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <AzureFunctionsVersion>v4</AzureFunctionsVersion>
    <OutputType>Exe</OutputType>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.19.0" />
    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http" Version="3.1.0" />
    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.15.1" />
  </ItemGroup>
  <ItemGroup>
    <None Update="host.json">
      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
    </None>
    <None Update="local.settings.json">
      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
      <CopyToPublishDirectory>Never</CopyToPublishDirectory>
    </None>
  </ItemGroup>
  <ItemGroup>
    <Using Include="System.Threading.ExecutionContext" Alias="ExecutionContext" />
  </ItemGroup>
</Project>
```

## Analysis
The generated code causing error is:
```cs
namespace FunctionCodeGen
{
    // ...
    {
        // ...
        {
            // ...
            var i = _functionActivator.CreateInstance(instanceType, context) as FunctionCodeGen.FunctionCodeGen; // This line causes error
            // ...
        }
        // ...
    }
    // ...
}
```

The namespace for the generated class is determined by the project name. When project name and the namespace for the function class are the same, the generated class and the function class reside in the same namespace. If additionally the function class has the same name with the containing namespace, then the compiler interprets the generated `as FunctionCodeGen.FunctionCodeGen` as referencing a nested type called `FunctionCodeGen` within type `FunctionCodeGen`, which does not exist.

Renaming the project will change the namespace for the generated class, so the compiler will be able to tell that `as FunctionCodeGen.FunctionCodeGen` references a class called `FunctionCodeGen` in another namespace `FunctionCodeGen`, and the code compiles. Renaming either the namespace or the function class will change `as FunctionCodeGen.FunctionCodeGen` to something like `as FunctionCodeGenNamespace.FunctionCodeGenClass`, and the compiler will be able to tell the first token is a namespace and the second one is a class, avoiding the problem.

#### Mitigation

To mitigate the issue, add the below entry to your `.csproj` file.

```xml
<PropertyGroup>
  <FunctionsGeneratedCodeNamespace >YourProjectNamespace</FunctionsGeneratedCodeNamespace>
</PropertyGroup>
```

DL444:

resolves #1990 

### Pull request checklist

### Additional information

Turns out to less straightforward than I thought it would be. One reference to the name performs a string comparison with a public property, and `Type.GetType()` would fail with fully qualified name. So, I ended up keeping both types of names, and I'm not even sure if this is the right approach.


DL444:

test/Sdk.Generator.Tests/FunctionExecutor/FunctionExecutorGeneratorTests.cs
```diff
@@ -502,6 +502,77 @@ public async ValueTask ExecuteAsync(FunctionContext context)
...
+namespace TestProject
+{
+    public class TestProject
+    {
+        [Function(""FunctionA"")]
+        public HttpResponseData Foo([HttpTrigger(AuthorizationLevel.User, ""get"")] HttpRequestData r, FunctionContext c)
+        {
+            return r.CreateResponse(System.Net.HttpStatusCode.OK);
+        }
+    }
```
kshyju, line 525: Thank you! Can we also throw in a static class and static function to make sure that it also works?
```csharp
public static class Foo
{
    [Function("Foo")]
    public static HttpResponseData Run([HttpTrigger(AuthorizationLevel.Function, "get")] HttpRequestData req)
    {
        throw new NotImplementedException();
    }
}

````

kshyju (approved):

kshyju:
Don't merge yet.

fabiocav (approved):
Thank you for the contribution, @DL444

PR #1996, opened 2023-10-27
[ApplicationInsights] removing env var reads from hot path

### LINKED ISSUE #1827: 8x Performance degradation after updating Microsoft.Azure.Functions.Worker.Extensions.ApplicationInsights package
Our system has experienced a huge (8x) performance degradation after what seemed like a minor upgrade. See the details about the versions below.

image

We have already confirmed that reverting these packages solves the issue.

All functions are affected, but the simplest one is a HTTP trigger that reads several blobs from Azure Storage. It went from taking less than 15s to more than 90s (peaking at 112s). The red box contains all the requests with the latest version of the packages. Below is before the deployment and the first one is after reverting the changes.

image

We will try to isolate a specific package now, but this is just to make everyone else aware that this happened to us and to hopefully get some advice from the team.

I did notice that the https://www.nuget.org/packages/Microsoft.Azure.Functions.Worker.ApplicationInsights has been released just a few hours ago and is listed as "0 downloads". That's going to be my main suspect.

image

brettsam:
resolves #1827 

During the (long) investigation into the issue above we have found that in some cases environment variable reads can be very expensive in App Service. But we cannot fully cache the variables because during certain App Service events (slot swaps, for example), the environment variables can change out from underneath us. 

This change adds a background monitor to watch for the changes in these few environment variables, then triggers an Options change when it occurs. This keeps the expensive reads off the hot path.

### Pull request checklist

### DIFF
--- a/src/DotNetWorker.ApplicationInsights/Initializers/AppServiceEnvironmentVariableMonitor.cs
+++ b/src/DotNetWorker.ApplicationInsights/Initializers/AppServiceEnvironmentVariableMonitor.cs
@@ -1,0 +1,75 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Options;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Azure.Functions.Worker.ApplicationInsights.Initializers;
+
+internal class AppServiceEnvironmentVariableMonitor : BackgroundService, IOptionsChangeTokenSource<AppServiceOptions>
+{
+    private readonly TimeSpan _refreshInterval;
+
+    private IChangeToken _changeToken;
+    private CancellationTokenSource _cancellationTokenSource = new();
+
+    private readonly Dictionary<string, string?> _monitoredVariableCache = new(StringComparer.OrdinalIgnoreCase);
+
+    public AppServiceEnvironmentVariableMonitor() : this(TimeSpan.FromSeconds(5))
+    {
+    }
+
+    public AppServiceEnvironmentVariableMonitor(TimeSpan refreshInterval)
+    {
+        _refreshInterval = refreshInterval;
+        _changeToken = new CancellationChangeToken(_cancellationTokenSource.Token);
+    }
+
+    public string Name => string.Empty;
+
+    public IChangeToken GetChangeToken() => _changeToken;
+
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        while (!stoppingToken.IsCancellationRequested)
+        {
+            bool changeDetected = false;
+
+            foreach (string envVar in AppServiceOptionsInitializer.EnvironmentVariablesToMonitor)
+            {
+                string? currentVal = Environment.GetEnvironmentVariable(envVar);
+                _monitoredVariableCache.TryGetValue(envVar, out string? cachedVal);
+
+                if (!string.Equals(currentVal, cachedVal, StringComparison.Ordinal))
+                {
+                    changeDetected = true;
+                    _monitoredVariableCache[envVar] = currentVal;
+                }
+            }
+
+            if (changeDetected)
+            {
+                var oldTokenSource = Interlocked.Exchange(ref _cancellationTokenSource, new CancellationTokenSource());
+                Interlocked.Exchange(ref _changeToken, new CancellationChangeToken(_cancellationTokenSource.Token));
+
+                if (!oldTokenSource.IsCancellationRequested)
+                {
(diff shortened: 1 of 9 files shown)


src/DotNetWorker.ApplicationInsights/Initializers/AppServiceOptionsInitializer.cs
```diff
@@ -0,0 +1,37 @@
...
+    internal const string AzureWebsiteName = "WEBSITE_SITE_NAME";
+    internal const string AzureWebsiteSlotName = "WEBSITE_SLOT_NAME";
+    internal const string AzureWebsiteCloudRoleName = "WEBSITE_CLOUD_ROLENAME";
+    internal const string DefaultProductionSlotName = "production";
+
+    internal static string[] EnvironmentVariablesToMonitor = [AzureWebsiteName, AzureWebsiteSlotName, AzureWebsiteCloudRoleName];
+
+    public void Configure(AppServiceOptions options)
+    {
+        options.AzureWebsiteName = Environment.GetEnvironmentVariable(AzureWebsiteName);
```
kshyju, line 16: Should we consider using `IConfiguration` to read this value? Is `IConfiguration` available at this point?

src/DotNetWorker.ApplicationInsights/Initializers/AppServiceEnvironmentVariableMonitor.cs
```diff
@@ -0,0 +1,75 @@
...
+                if (!oldTokenSource.IsCancellationRequested)
+                {
+                    oldTokenSource.Cancel();
+                    oldTokenSource.Dispose();
+                }
+            }
+
+            try
+            {
+                await Task.Delay(_refreshInterval, stoppingToken);
```
kshyju, line 66: Is it possible that one of these env variables changes more than once (by platform) within the 5 second timespan? In that case, would we miss the second update?

src/DotNetWorker.ApplicationInsights/Initializers/AppServiceOptionsInitializer.cs
```diff
@@ -0,0 +1,37 @@
...
+using Microsoft.Extensions.Options;
+
+internal class AppServiceOptionsInitializer : IConfigureOptions<AppServiceOptions>
+{
+    internal const string AzureWebsiteName = "WEBSITE_SITE_NAME";
+    internal const string AzureWebsiteSlotName = "WEBSITE_SLOT_NAME";
+    internal const string AzureWebsiteCloudRoleName = "WEBSITE_CLOUD_ROLENAME";
+    internal const string DefaultProductionSlotName = "production";
+
+    internal static string[] EnvironmentVariablesToMonitor = [AzureWebsiteName, AzureWebsiteSlotName, AzureWebsiteCloudRoleName];
```
kshyju, line 12: Does this style of collection initializer works in net6/net7? Should we use `new [` ?

src/DotNetWorker.ApplicationInsights/Initializers/AppServiceOptionsInitializer.cs
brettsam: `IConfiguration` caches all the environment variables so unfortunately this doesn't really work as it'd never see a change. We *could* write a broader environment variable checker that'd invalidate the `IConfiguration` and have it re-populate if there's a change, but that's a much bigger issue and I don't think it's worth it. In this case, we only really care about a couple platform-level variables, so I elected to just check them directly. 

App Insights takes a similar approach here -- https://github.com/microsoft/ApplicationInsights-dotnet/blob/main/WEB/Src/WindowsServer/WindowsServer/Implementation/EnvironmentVariableMonitor.cs

src/DotNetWorker.ApplicationInsights/Initializers/AppServiceEnvironmentVariableMonitor.cs
brettsam: I *think* they'll only ever change at most once. I was tempted even to kill the timer once we detected the first change, but I can't confirm that that behavior will never change.

But if they change twice within 5 seconds, we'll see the last change (we'd miss that middle, but that's okay).

This does mean there's a small window where some telemetry would be wrong due to the env variable swap (up to 5 seconds until we notice the change), but that shouldn't be a huge issue. Heck, App Insights defaults to 60 seconds for checking this by default...

src/DotNetWorker.ApplicationInsights/Initializers/AppServiceOptionsInitializer.cs
brettsam: build failed -- so yeah :-)

kshyju (approved):

PR #2022, opened 2023-11-03
Change binding type name to camel case in source-gen

satvu:

Changes binding type name to camel case in source-gen as customers and partner services may check casing. Consistent with legacy behavior (#589).

### Pull request checklist

### DIFF
--- a/sdk/Sdk.Generators/Extensions/StringExtensions.cs
+++ b/sdk/Sdk.Generators/Extensions/StringExtensions.cs
@@ -38,7 +38,7 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
         {
             if (!string.IsNullOrEmpty(str))
             {
-                return Char.ToLowerInvariant(str.First()) + str.Substring(1);
+                return Char.ToLowerInvariant(str[0]) + str.Substring(1);
             }
             else
             {
--- a/sdk/Sdk.Generators/Extensions/StringExtensions.cs
+++ b/sdk/Sdk.Generators/Extensions/StringExtensions.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
 {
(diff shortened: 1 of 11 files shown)


sdk/Sdk.Generators/Extensions/StringExtensions.cs
```diff
@@ -29,5 +30,20 @@ public static string TrimStringsFromEnd(this string str, IReadOnlyList<string> s
...
         }
+
+        /// <summary>
+        /// Returns a copy of the string where the first character is in lower case.
+        /// </summary>
+        public static string ToLowerFirstCharacter(this string str)
+        {
+            if (!string.IsNullOrEmpty(str))
+            {
+                return Char.ToLowerInvariant(str.First()) + str.Substring(1);
```
jviau, line 41: `str[0]` - slightly less allocation than `First()`

sdk/Sdk.Generators/Extensions/StringExtensions.cs
```diff
@@ -29,5 +30,20 @@ public static string TrimStringsFromEnd(this string str, IReadOnlyList<string> s
 
             return result;
         }
+
+        /// <summary>
+        /// Returns a copy of the string where the first character is in lower case.
+        /// </summary>
+        public static string ToLowerFirstCharacter(this string str)
```
jviau, line 37: nit:
```suggestion
        public static string ToCamelCase(this string str)
```

jviau (approved):

kshyju (approved):

sdk/Sdk.Generators/Extensions/StringExtensions.cs
kshyju: I am not sure "ToCameCase" is appropriate name as we are not doing that. for example : for input "HTTPTRIGGER" , the ideal camelCase result should be "httpTrigger", but we would return "hTTPTRIGGER"

satvu: ^ This was also my reasoning.

jviau: Good points

PR #2064, opened 2023-11-14
Update ServiceBus dependencies

### LINKED ISSUE #2021: Introducing vulnerability: CVE-2023-36414 (Azure Identity SDK Remote Code Execution Vulnerability) while using ServiceBus binding
Our validation pipeline started to fail on the WhiteSource validation step with the High Severity alert: 
image

We followed the recommendation and installed the latest Azure.Identity package (v1.10.3). However, it didn't help.

We ran some investigation.
We created a bare-bone Azure Function project and added a single function there that is triggered by a service bus. 
[TestFunction.zip](https://github.com/Azure/azure-functions-dotnet-worker/files/13253236/TestFunction.zip)
We found there are 2 places in the build output path where Azure.Identity.dll exists.
They are _..\bin\Debug\net6.0_ folder and _..\bin\Debug\net6.0\.azurefunctions_
There is the latest version of Azure.Identity in the [..\bin\Debug\net6.0](_url_) folder. But in the _..\bin\Debug\net6.0\.azurefunctions_ there is the old one (Azure.Identity-1.7.0)

What else we found is the Azure.Identity version 1.7.0 is a default version of dependency for the latest version of Microsoft.Azure.Functions.Worker.Extensions.ServiceBus
image

When the project compiled with no latest version of Azure.Identity Installed both output folders contain the old version of it
image
image

However, after we Installed the latest Azure.Identity 1.10.3 (as part of Microsoft.Extensions.Azure)
image

 we see it in the _..\bin\Debug\net6.0_ folder but in the _..\bin\Debug\net6.0\.azurefunctions_ there is still old Azure.Identity:
image
image

PS. We also had this issue when we used the EventGrid binding but we switched to the EventGrid client call as a workaround.

jviau:

resolves #2021

### Pull request checklist

### Additional information

Updates dependencies to address CVE in `Azure.Identity`.

### DIFF
--- a/extensions/Worker.Extensions.ServiceBus/release_notes.md
+++ b/extensions/Worker.Extensions.ServiceBus/release_notes.md
@@ -4,8 +4,8 @@
 - My change description (#PR/#issue)
 -->
 
-### Microsoft.Azure.Functions.Worker.Extensions.ServiceBus 5.14.1
+### Microsoft.Azure.Functions.Worker.Extensions.ServiceBus 5.15.0
 
-- Fixed issue where deadlettering a message without specifying properties to modify could throw
-  an exception from out of proc extension.
-- Include underlying exception details in RpcException when a failure occurs.
+- Update `Microsoft.Azure.WebJobs.Extensions.ServiceBus` dependency to `5.13.4`
+- Update `Microsoft.Azure.Extensions` to `1.7.1`
+- Update `Azure.Messaging.ServiceBus` to `7.17.0`
--- a/extensions/Worker.Extensions.ServiceBus/release_notes.md
+++ b/extensions/Worker.Extensions.ServiceBus/release_notes.md
@@ -7,5 +7,5 @@
 ### Microsoft.Azure.Functions.Worker.Extensions.ServiceBus 5.15.0
 
 - Update `Microsoft.Azure.WebJobs.Extensions.ServiceBus` dependency to `5.13.4`
-- Update `Microsoft.Azure.Extensions` to `1.7.1`
+- Update `Azure.Identity` to `1.10.2`
 - Update `Azure.Messaging.ServiceBus` to `7.17.0`
--- a/extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActions.cs
+++ b/extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActions.cs
@@ -62,6 +62,7 @@ namespace Microsoft.Azure.Functions.Worker
         ///</remarks>
         protected ServiceBusMessageActions()
         {
+            _settlement = null!; // not expected to be used during mocking.
         }
 
         ///<inheritdoc cref="ServiceBusReceiver.CompleteMessageAsync(ServiceBusReceivedMessage, CancellationToken)"/>
--- a/extensions/Worker.Extensions.ServiceBus/src/Worker.Extensions.ServiceBus.csproj
+++ b/extensions/Worker.Extensions.ServiceBus/src/Worker.Extensions.ServiceBus.csproj
@@ -15,8 +15,9 @@
   <Import Project="..\..\..\build\Extensions.props" />
 
   <ItemGroup>
+    <PackageReference Include="Azure.Identity" Version="1.10.2" />
     <PackageReference Include="Azure.Messaging.ServiceBus" Version="7.17.0" />
-    <PackageReference Include="Microsoft.Extensions.Azure" Version="1.7.1" />
+    <PackageReference Include="Microsoft.Extensions.Azure" Version="1.6.3" />
     <PackageReference Include="Google.Protobuf" Version="3.23.4" />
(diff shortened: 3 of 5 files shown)


extensions/Worker.Extensions.ServiceBus/release_notes.md
```diff
@@ -4,8 +4,7 @@
...
 -->
 
-### Microsoft.Azure.Functions.Worker.Extensions.ServiceBus 5.14.1
+### Microsoft.Azure.Functions.Worker.Extensions.ServiceBus 5.15.0
 
-- Fixed issue where deadlettering a message without specifying properties to modify could throw
-  an exception from out of proc extension.
-- Include underlying exception details in RpcException when a failure occurs.
+- Update `Microsoft.Azure.WebJobs.Extensions.ServiceBus` dependency to `5.13.4`
+- Update `Microsoft.Azure.Extensions` to `1.7.1`
```
JoshLove-msft, line 10: Why do we mention this transitive dependency specifically?

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActions.cs
```diff
@@ -62,6 +62,7 @@ internal ServiceBusMessageActions(Settlement.SettlementClient settlement)
         ///</remarks>
         protected ServiceBusMessageActions()
         {
+            _settlement = null!; // not expected to be used during mocking.
```
JoshLove-msft, line 65: Why was this needed?

jviau: Nullability warning without it.

extensions/Worker.Extensions.ServiceBus/release_notes.md
JoshLove-msft: Nevermind, I see this is for the isolated worker dependency.

jviau: This is explicitly referenced in the csproj. I updated it as well as it is what brings in `Azure.Identity`.

JoshLove-msft (approved):

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActions.cs
JoshLove-msft: It might be better to new up an instance if possible in case it does get used in mocking scenarios.

extensions/Worker.Extensions.ServiceBus/src/Worker.Extensions.ServiceBus.csproj
```diff
@@ -16,7 +16,7 @@
 
   <ItemGroup>
     <PackageReference Include="Azure.Messaging.ServiceBus" Version="7.16.2" />
```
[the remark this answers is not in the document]
jviau: I updated this, but not entirely sold it is the right thing. Reving our dependencies increases chance of incompatible package dependencies. Say they have another dependency which has a stricter version range for this package, it is now a headache for the customer. And for what benefit? We didn't _need_ this upgrade. And if customers needed it, they can pin the newer version themselves.

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActions.cs
jviau: I don't have any connection to new this up with. So, the failure would just happen later and be more obscure.

extensions/Worker.Extensions.ServiceBus/src/Worker.Extensions.ServiceBus.csproj
```diff
@@ -15,8 +15,8 @@
   <Import Project="..\..\..\build\Extensions.props" />
 
   <ItemGroup>
-    <PackageReference Include="Azure.Messaging.ServiceBus" Version="7.16.2" />
-    <PackageReference Include="Microsoft.Extensions.Azure" Version="1.6.3" />
+    <PackageReference Include="Azure.Messaging.ServiceBus" Version="7.17.0" />
+    <PackageReference Include="Microsoft.Extensions.Azure" Version="1.7.1" />
```
jviau, line 19: This change will bring in `Azure.Identity/1.10.3`, which has the major version leap in the downstream dependency. If we don't want that, we can alternatively revert this and explicitly bring in `Azure.Identity/1.10.2`

kshyju: For host repo, we used to 1.10.2 to avoid introducing the breaking change (from 1.10.3)
https://github.com/Azure/azure-functions-host/pull/9646#issuecomment-1785587540

LenaVinogradov:
Any updates? we are waiting for the new package please.

liliankasem (changes requested):
As per @shyju's comment, we should explicitly bring in `Azure.Identity/1.10.2`

PR #2103, opened 2023-11-28
Generate correct ScriptFile value when generating metadata.

kshyju:
Fixing function metadata provider to correctly generate ScriptFile property value. Currently it always generate a value which ends with ".dll" and this causes NETFX function apps to fail during function load. This PR adds changes to determine whether the value should be ".dll" or ".exe" and use the appropriate value.

We only need to do this for the entry assembly. Dependent assemblies will be always .dll.

### Pull request checklist


satvu (approved):
LGTM - just need pipeline tests fixed.

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -40,7 +40,8 @@ public Parser(GeneratorExecutionContext context)
             /// Takes in candidate methods from the user compilation and parses them to return function metadata info as GeneratorFunctionMetadata.
             /// </summary>
             /// <param name="methods">List of candidate methods from the syntax receiver.</param>
-            public IReadOnlyList<GeneratorFunctionMetadata> GetFunctionMetadataInfo(List<IMethodSymbol> methods)
+            /// <param name="parsingContext">An instance of <see cref="FunctionsMetadataParsingContext"/>. Optional.</param>
```
liliankasem, line 43: nit: I think optional usually goes at the start of the description not the end

liliankasem (approved):

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
kshyju: Moved.

PR #2119, opened 2023-12-02
Avoid executing source generators outside of an Azure Functions project

### LINKED ISSUE #2114: Broken WorkerExtensionStartupCodeExecutor generation when FunctionsApp referenced by XUnit test project.
When referencing a function app by an XUnit test project (may happen in other ways as well) the namespace does not get generated correctly. The error shows up against the test project. This happens on the latest version of https://www.nuget.org/packages/Microsoft.Azure.Functions.Worker.Sdk/1.16.3

I would share my solution but its proprietary so I can't, rolling back to the immediate previous version (1.16.2) fixed the issue for me, should be easy enough to work out what caused this..

Below is what was generated.

[assembly: WorkerExtensionStartupCodeExecutorInfo(typeof(.WorkerExtensionStartupCodeExecutor))]

namespace 
{
    [global::System.ComponentModel.EditorBrowsableAttribute(global::System.ComponentModel.EditorBrowsableState.Never)]
    [global::System.Runtime.CompilerServices.CompilerGeneratedAttribute()]
    internal class WorkerExtensionStartupCodeExecutor : WorkerExtensionStartup
    {
        /// <summary>

kshyju:
resolves #2114

### Pull request checklist

Additional PR information

### DIFF
--- a/sdk/Sdk.Generators/Constants.cs
+++ b/sdk/Sdk.Generators/Constants.cs
@@ -5,7 +5,10 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
 {
     internal static class Constants
     {
-        internal const string Isolated = "isolated";
+        internal static class ExecutionModel
+        {
+            internal const string Isolated = "isolated";
+        }
 
         internal static class Languages
         {
--- a/sdk/Sdk.Generators/Constants.cs
+++ b/sdk/Sdk.Generators/Constants.cs
@@ -5,6 +5,8 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
 {
     internal static class Constants
     {
+        internal const string Isolated = "isolated";
+
         internal static class Languages
         {
             internal const string DotnetIsolated = "dotnet-isolated";
@@ -12,8 +14,8 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
 
         internal static class BuildProperties
         {
+            internal const string FunctionsExecutionModel = "build_property.FunctionsExecutionModel";
             internal const string MSBuildTargetFrameworkIdentifier = "build_property.TargetFrameworkIdentifier";
-            internal const string MSBuildRootNamespace = "build_property.RootNamespace";
             internal const string GeneratedCodeNamespace = "build_property.FunctionsGeneratedCodeNamespace";
             internal const string EnableSourceGen = "build_property.FunctionsEnableMetadataSourceGen";
             internal const string EnablePlaceholder = "build_property.FunctionsEnableExecutorSourceGen";
--- a/sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
+++ b/sdk/Sdk.Generators/ExtensionStartupRunnerGenerator.cs
@@ -62,6 +62,11 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
 
         public void Execute(GeneratorExecutionContext context)
         {
+            if (!context.IsRunningInAzureFunctionProject())
+            {
+                return;
+            }
+            
             var extensionStartupTypeNames = GetExtensionStartupTypes(context);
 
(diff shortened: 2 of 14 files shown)


jviau:
Reading the linked issue, the root cause is that the SDK package ref is transitively included in the xunit project, so it too includes all those targets. Disabling the source generators this way mitigates a piece of the problem, but it is larger than that. The better fix is to just not include that package transitively. This can be done via:

1. Ask customers to add `PrivateAssets="all"` to their SDK package ref (we can also update templates to include this)
2. Bigger change - we can look into using an actual SDK msbuild element. This would let us get our targets in and also avoid it being included transitively. However, there are other consequences of this we would need to explore (not to mention a complete shift in how customers reference our SDK).

So, I am not sold this is absolutely necessary because even with this I would still advise customers to perform option 1. This won't cause any harm including it, but it isn't right to say the root problem is solved with this change (there are still other SDK targets that _could_ cause issues being brought in).

fabiocav:
> Reading the linked issue, the root cause is that the SDK package ref is transitively included in the xunit project, so it too includes all those targets. Disabling the source generators this way mitigates a piece of the problem, but it is larger than that. The better fix is to just not include that package transitively. This can be done via:
> 
> 1. Ask customers to add `PrivateAssets="all"` to their SDK package ref (we can also update templates to include this)
> 2. Bigger change - we can look into using an actual SDK msbuild element. This would let us get our targets in and also avoid it being included transitively. However, there are other consequences of this we would need to explore (not to mention a complete shift in how customers reference our SDK).
> 
> So, I am not sold this is absolutely necessary because even with this I would still advise customers to perform option 1. This won't cause any harm including it, but it isn't right to say the root problem is solved with this change (there are still other SDK targets that _could_ cause issues being brought in).

While I agree with the above, the fact that this update introduces an issue that will lead to confusion and require a project change leads to a poor experience, so we should avoid that. Whether the approach is what's scoped here (which I agree, shouldn't cause problems) or something else, that's less important, but we need to ensure an SDK reference update within the same major won't break user's builds.

sdk/Sdk.Generators/Constants.cs
```diff
@@ -5,15 +5,20 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
...
+        }
+
         internal static class Languages
         {
             internal const string DotnetIsolated = "dotnet-isolated";
         }
 
         internal static class BuildProperties
         {
+            internal const string FunctionsExecutionModel = "build_property.FunctionsExecutionModel";
```
jviau, line 20: I don't think this will work in the end, as `FunctionsExecutionModel` is set in: https://github.com/Azure/azure-functions-dotnet-worker/blob/5ff83ce0fdf244536cea093686540b424dbfc1ea/sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.props#L14

and due to the transitive reference, this will also be part of the xunit project, so the condition will pass and generators still run.

We need to find something that is different between the xunit project (rather a non-functions project) and the functions project. But that is hard because bringing in the functions SDK at all _is_ our signal that this is a function project! We can brainstorm something else to use here, but also need to ensure it doesn't false disable the source generator unintetnionally.

May we can check if the project is a library vs executable?

jviau (changes requested):

sdk/Sdk.Generators/Constants.cs
kshyju: @jviau  Are you sure about this? My understanding is that the MSBuild properties are not propagated through project references, unless one explicitly do that via `<Import Project="path\to\your\propsfile.props" />` in the second project (in our case - the Test project)

My local validation also confirms this understanding. Are you still able to see the source generation happening inside the test projects when the artifacts from this PR is used? It is very much possible I missed something in my testing.

kshyju: image

jviau (approved):

PR #2148, opened 2023-12-11
Adding a Info level log entry after function app assembly load

kshyju:
Adding a log statement after worker assembly is successfully loaded during language worker specialization. We had a previous log statement, but that was a trace level log entry. Changed it to INFO level and including the process id and APP_POOL_ID environment variable value.

### Pull request checklist

### DIFF
--- a/host/src/FunctionsNetHost/Grpc/IncomingGrpcMessageHandler.cs
+++ b/host/src/FunctionsNetHost/Grpc/IncomingGrpcMessageHandler.cs
@@ -87,7 +87,13 @@ namespace FunctionsNetHost.Grpc
 
                     Logger.LogTrace($"Will wait for worker loaded signal.");
                     WorkerLoadStatusSignalManager.Instance.Signal.WaitOne();
-                    Logger.LogTrace($"Received worker loaded signal. Forwarding environment reload request to worker.");
+
+                    var logMessage = $"FunctionApp assembly loaded successfully. ProcessId:{Environment.ProcessId}";
+                    if (OperatingSystem.IsWindows())
+                    {
+                        logMessage += $", AppPoolId:{Environment.GetEnvironmentVariable(EnvironmentVariables.AppPoolId)}";
+                    }
+                    Logger.Log(logMessage);
 
                     await MessageChannel.Instance.SendInboundAsync(msg);
                     _specializationDone = true;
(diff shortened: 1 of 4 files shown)


host/src/FunctionsNetHost/Grpc/IncomingGrpcMessageHandler.cs
```diff
@@ -87,7 +87,13 @@ private async Task Process(StreamingMessage msg)
 
                     Logger.LogTrace($"Will wait for worker loaded signal.");
                     WorkerLoadStatusSignalManager.Instance.Signal.WaitOne();
-                    Logger.LogTrace($"Received worker loaded signal. Forwarding environment reload request to worker.");
+
+                    var logMessage = $"FunctionApp assembly loaded successfully. ProcessId:{Environment.ProcessId}";
+                    if (OperatingSystem.IsWindows())
```
kshyju, line 92: The OS specific implementation returns `true` or `false` (using conditional PREPROCESSOR DIRECTIVE SYMBOLS).

host/src/FunctionsNetHost/Grpc/IncomingGrpcMessageHandler.cs
```diff
@@ -87,7 +87,13 @@ private async Task Process(StreamingMessage msg)
...
+                    {
+                        logMessage += $", AppPoolId:{Environment.GetEnvironmentVariable(EnvironmentVariables.AppPoolId)}";
+                    }
+                    Logger.Log(logMessage);
```
safihamid, line 96: nit: everywhere else in this file uses Logger.LogTrace.
not sure if there is any difference or not.

safihamid (approved):
:shipit:

host/src/FunctionsNetHost/Grpc/IncomingGrpcMessageHandler.cs
kshyju: Entries with Logger.Trace will be logged only when `AZURE_FUNCTIONS_FUNCTIONSNETHOST_TRACE` env variable is set to `1`. Logger.Log will be alwys logged (more like LogInfo). There is some planned improvement to this logging code and we can fix the name to be more readable at that point.

satvu (approved):

Issue #1929, opened 2023-09-26
When using ASP.NET Core integration, ObjectDisposedException is thrown when the request is cancelled


xin9le:
# Summary

When using ASP.NET Core integration, if a request thrown to an HttpTrigger developed is cancelled by the client side, an `ObjectDisposedException` occurs. The stack trace is as follows. This does not occur when ASP.NET Core integration is not used.

## Stack trace
```
System.Private.CoreLib: Exception while executing function: Functions.Http_Sample. System.Private.CoreLib: Result: Failure
Exception: System.ObjectDisposedException: Request has finished and HttpContext disposed.
 Object name: 'HttpContext'.
    at Microsoft.AspNetCore.Http.DefaultHttpContext.ThrowContextDisposed()
    at Microsoft.AspNetCore.Http.DefaultHttpContext.get_Features()
    at Microsoft.AspNetCore.Routing.RoutingHttpContextExtensions.GetRouteData(HttpContext httpContext)
    at Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore.FunctionsHttpProxyingMiddleware.Invoke(FunctionContext context, FunctionExecutionDelegate next) in D:\a\_work\1\s\extensions\Worker.Extensions.Http.AspNetCore\src\FunctionsMiddleware\FunctionsHttpProxyingMiddleware.cs:line 52
    at Microsoft.Azure.Functions.Worker.FunctionsApplication.InvokeFunctionAsync(FunctionContext context) in D:\a\_work\1\s\src\DotNetWorker.Core\FunctionsApplication.cs:line 77
    at Microsoft.Azure.Functions.Worker.Handlers.InvocationHandler.InvokeAsync(InvocationRequest request) in D:\a\_work\1\s\src\DotNetWorker.Grpc\Handlers\InvocationHandler.cs:line 88
Stack:    at Microsoft.AspNetCore.Http.DefaultHttpContext.ThrowContextDisposed()
    at Microsoft.AspNetCore.Http.DefaultHttpContext.get_Features()
    at Microsoft.AspNetCore.Routing.RoutingHttpContextExtensions.GetRouteData(HttpContext httpContext)
[...]

PR #2153, opened 2023-12-14
[ASP.NET Core Integration] Check for cancellation before executing action

satvu:

partial mitigation for #1929, see #2180 for the long-term fix.

### Pull request checklist

Issue created for E2E testing of ASP.NET integration.

### DIFF
--- a/extensions/Worker.Extensions.Http.AspNetCore/src/FunctionsMiddleware/FunctionsHttpProxyingMiddleware.cs
+++ b/extensions/Worker.Extensions.Http.AspNetCore/src/FunctionsMiddleware/FunctionsHttpProxyingMiddleware.cs
@@ -53,7 +53,7 @@ namespace Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore
             {
                 ActionContext actionContext = new ActionContext(httpContext, httpContext.GetRouteData(), new ActionDescriptor());
 
-                if (!httpContext.RequestAborted.IsCancellationRequested)
+                if (!httpContext.RequestAborted.IsCancellationRequested && !context.CancellationToken.IsCancellationRequested)
                 {
                     await actionResult.ExecuteResultAsync(actionContext);
                 }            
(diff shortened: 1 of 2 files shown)


jviau:
@satvu I think this is covering a symptom of the bug, but not the root cause. This approach will avoid us making a call to something that was disposed due to cancellation, but the root issue is that we dispose this object _at all_ before the function execution is complete, regardless of cancellation state.

extensions/Worker.Extensions.Http.AspNetCore/src/FunctionsMiddleware/FunctionsHttpProxyingMiddleware.cs
```diff
@@ -53,7 +53,10 @@ public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next
             {
                 ActionContext actionContext = new ActionContext(httpContext, httpContext.GetRouteData(), new ActionDescriptor());
 
-                await actionResult.ExecuteResultAsync(actionContext);
+                if (!httpContext.RequestAborted.IsCancellationRequested && !context.CancellationToken.IsCancellationRequested)
```
jviau, line 56: Can we add a comment that this is a mitigation and the dispose exception is still technically possible?

jviau (approved):

liliankasem:
> @satvu I think this is covering a symptom of the bug, but not the root cause. This approach will avoid us making a call to something that was disposed due to cancellation, but the root issue is that we dispose this object _at all_ before the function execution is complete, regardless of cancellation state.
> 
> See my comment [here](https://github.com/Azure/azure-functions-dotnet-worker/issues/1929#issuecomment-1743890718)
> 
> Our cancellation token callbacks have introduced a race condition with function execution, which leads to ObjectDisposedException.

I wonder if this is related to another bug I found last month:

> Sometimes when we send a cancelled invocation with a cancellation request to the worker, the CT being used in the
[BindFunctionInputAsync](https://github.com/Azure/azure-functions-dotnet-worker/blob/2b51979d1c3a93c08c322f788138a1fd2b25f51e/src/DotNetWorker.Core/Context/Features/DefaultFunctionInputBindingFeature.cs#L32) method (DefaultFunctionInputBindingFeature.cs) is already cancelled, this throws a Task cancelled exception leading to a 500. I think this is a race of sorts as it happens every other time but not consistently. I can repro by hitting the a HTTP function endpoint several times in a row.

satvu:
Closing as we are prioritizing a long-term fix and skipping this mitigation.

PR #2174, opened 2023-12-30
Fix issue where IReadOnlyList<T> wasn't a valid binding type.

### LINKED ISSUE #2173: CosmosDB Extension throws exception when Input is of type IReadOnlyList.
I recently deployed an updated function leveraging the 4.5.0 version of the CosmosDB extension and am getting an exception when trying to bind CosmosDB inputs to IReadOnlyList.

I tracked it down to the [ParameterBinder](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/extensions/Worker.Extensions.Shared/Reflection/ParameterBinder.cs) and it looks like it's caused by this code:

`
        private static bool IsListInterface(Type type)
        {
            return type.IsGenericType &&
                (type.GetGenericTypeDefinition() == typeof(IList<>)
                || type.GetGenericTypeDefinition() == typeof(ICollection<>)
                || type.GetGenericTypeDefinition() == typeof(IEnumerable<>));
        }
`

The exception I get matches the one from line 62 of ParameterBinder.BindCollectionAsync so I'm confident that if I'm not on the right function I'm close.

I've downgraded the CosmosDB package I'm using to 4.4.2 and won't be able to upgrade until this is fixed.  Looking at the code hierarchy it may affect other extensions as they pick up the shared code as well.

savagemonitor:

resolves #2173 

### Pull request checklist

  * [ ] ~~Otherwise: Documentation issue linked to PR~~
  * [ ] ~~Otherwise: I've added my notes to `release_notes.md`~~
  * [ ] ~~Otherwise: Backport tracked by issue/PR #issue_or_pr~~

### Additional information

I leveraged Linq to give what I felt was a "cleaner" approach where the IsListInterface can succinctly look at the value of the Type as well as the value of all of the Interfaces of the type to see if it's valid.  I think this should cover every situation where the passed in interface type inherits from one of the valid types.

I know this isn't an optimal runtime solution since LINQ can be slower but since we're talking reflection I figured it was a good tradeoff.  If there are performance issues with this approach I'd love to learn more.

### DIFF
--- a/extensions/Worker.Extensions.Shared/Reflection/ParameterBinder.cs
+++ b/extensions/Worker.Extensions.Shared/Reflection/ParameterBinder.cs
@@ -16,9 +16,10 @@ namespace Microsoft.Azure.Functions.Worker.Extensions
     internal static class ParameterBinder
     {
         /// <summary>
-        /// An array of types that reasonably map back to a class that implements a generic list data structure. 
+        /// Read only property that contains all of the generic interfaces implemented by <see cref="System.Collections.Generic.List"/>.
         /// </summary>
-        private static Type[] validListInterfaceTypes = new Type[] {typeof(IList<>), typeof(ICollection<>), typeof(IEnumerable<>)};
+        /// <remarks>This property calls ToList at the end to force the resolution of the LINQ methods that leverage deferred execution.</remarks>
+        private static readonly IEnumerable<Type> validListInterfaceTypes = typeof(List<>).GetInterfaces().Where(t => t.IsGenericType).Select(t => t.GetGenericTypeDefinition()).ToList();
 
         private const BindingFlags DeclaredOnlyLookup = BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.Static | BindingFlags.DeclaredOnly;
 
@@ -97,11 +98,15 @@ namespace Microsoft.Azure.Functions.Worker.Extensions
             }
         }
 
+        /// <summary>
+        /// A method that determines if a Type is a generic interface of the <see cref="System.Collections.Generic.List"/> concrete class.
+        /// </summary>
+        /// <param name="type">A generic interface type to be tested against the types available on the generic <see cref="System.Collections.Generic.List"/> type.</param>
+        /// <returns>True if the type is a generic type and it's an interface of the generic <see cref="System.Collections.Generic.List"/> type otherwise false.</returns>
         private static bool IsListInterface(Type type)
         {
             return type.IsGenericType &&
-                (validListInterfaceTypes.Contains(type.GetGenericTypeDefinition()) ||
-                 type.GetInterfaces().Select(t => t.GetGenericTypeDefinition()).Intersect(validListInterfaceTypes).Any());
+                validListInterfaceTypes.Any(t => t == type.GetGenericTypeDefinition());
         }
 
         private static Action<object> GetAddMethod(object collection)
--- a/extensions/Worker.Extensions.Shared/Reflection/ParameterBinder.cs
+++ b/extensions/Worker.Extensions.Shared/Reflection/ParameterBinder.cs
@@ -96,7 +96,8 @@ namespace Microsoft.Azure.Functions.Worker.Extensions
             return type.IsGenericType &&
                 (type.GetGenericTypeDefinition() == typeof(IList<>)
                 || type.GetGenericTypeDefinition() == typeof(ICollection<>)
-                || type.GetGenericTypeDefinition() == typeof(IEnumerable<>));
+                || type.GetGenericTypeDefinition() == typeof(IEnumerable<>)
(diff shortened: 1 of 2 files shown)


savagemonitor:

extensions/Worker.Extensions.Shared/Reflection/ParameterBinder.cs
```diff
@@ -94,9 +100,8 @@ await foreach (object item in pageable)
         private static bool IsListInterface(Type type)
         {
             return type.IsGenericType &&
-                (type.GetGenericTypeDefinition() == typeof(IList<>)
-                || type.GetGenericTypeDefinition() == typeof(ICollection<>)
-                || type.GetGenericTypeDefinition() == typeof(IEnumerable<>));
+                (validListInterfaceTypes.Contains(type.GetGenericTypeDefinition()) ||
+                 type.GetInterfaces().Select(t => t.GetGenericTypeDefinition()).Intersect(validListInterfaceTypes).Any());
```
jviau, line 104: Checking all interfaces is too broad, there is no guarantee we can instantiate the declared interface. For example, I could add this type:

``` CSharp
public interface ICustomList<T> : IList<T> { }
```

This would pass the `IsListInterface` check, but we cannot actually satisfy that type with `List<T>` - or any other concrete type we know of.

I think the best option will be to explicitly add support for `IReadOnlyList<T>`, and also `IReadOnlyCollection<T>` as `List<T>` implements it as well.

jviau (changes requested):
Thank you for the contribution! I agree supporting `IReadOnlyList<T>` (and `IReadOnlyCollection<T>`) would be nice. Some comments on how to go about that.

extensions/Worker.Extensions.Shared/Reflection/ParameterBinder.cs
savagemonitor: I get what you're saying as I read the intent of the method differently than how you did.  After reading your comment and re-reading the code I get that the intent is to identify if the Type is a generic interface on List<T>.

I don't think a static list of interfaces is the right solution though because if .NET changes the set of interfaces that List<T> supports and someone tries to leverage one of the new interfaces then they'll be blocked until you rev this shared code.

I think a better fix would be to check the passed in Type against the generic interfaces of List<> to see if it's one of them.  Then as List<T>'s interfaces change you don't need to rev this shared code.  I'll see if I can code that up later tonight.

savagemonitor: My latest commit changes the IsListInterface to check against every generic interface on `List<T>` to see if the passed in type is one on `List<T>`.  I've also added a test that will verify that every generic interface will return a list as expected.

I ended up using a little more LINQ then I intended to though and I understand this has a performance hit.  I didn't spot any perf tests so I don't know the exact impact.  I expect it to be small but I'm sure that even small increases are a problem for something this widely used.

If you'd rather not take any performance hit I can go back to the static list though I'd keep the test I added to validate that all of the interfaces are supported.  That way if `List<T>` gets updated in the future it will be easy to figure out that more interfaces need to be added.

extensions/Worker.Extensions.Shared/Reflection/ParameterBinder.cs
```diff
@@ -14,6 +15,12 @@ namespace Microsoft.Azure.Functions.Worker.Extensions
     /// </summary>
     internal static class ParameterBinder
     {
+        /// <summary>
+        /// Read only property that contains all of the generic interfaces implemented by <see cref="System.Collections.Generic.List"/>.
+        /// </summary>
+        /// <remarks>This property calls ToList at the end to force the resolution of the LINQ methods that leverage deferred execution.</remarks>
+        private static readonly IEnumerable<Type> validListInterfaceTypes = typeof(List<>).GetInterfaces().Where(t => t.IsGenericType).Select(t => t.GetGenericTypeDefinition()).ToList();
```
jviau, line 22: for performance, can make this a `HashSet<Type>`:

``` CSharp
new HashSet<Type>(typeof(List<>).GetInterfaces().Where(t => t.IsGenericType).Select(t => t.GetGenericTypeDefinition()));
```

Can then use `.Contains` in the `IsListInterface` check
```

savagemonitor: I like it and will update my PR with it.

The thing it doesn't solve is that building the HashSet is always going to be a performance hit as the `List<>` type will be reflected on to get the types then the LINQ queries will execute.  I don't really see a way to make that aspect faster beyond replacing the LINQ code or going to a static list.

jviau: Being static that will happen only once, so not a big concern.

jviau (approved):

jviau:
Thank you for the contribution!

PR #2177, opened 2024-01-04
Locate the Entrypoint first by name and take the method with the FunctionAttribute

### LINKED ISSUE #2124: Unable to load Function, a function with the id already exists.
I have a queue trigger function that receives events from a queue storage, and the function execution sometimes fails with an error message:

```Result: Failure Exception: System.InvalidOperationException: Unable to load Function '[Function name]'. A function with the id '[Function id]' name already exists. at Microsoft.Azure.Functions.Worker.FunctionsApplication.LoadFunction(FunctionDefinition definition) in D:\a\_work\1\s\src\DotNetWorker.Core\FunctionsApplication.cs:line 62 at Microsoft.Azure.Functions.Worker.GrpcWorker.FunctionLoadRequestHandler(FunctionLoadRequest request, IFunctionsApplication application, IMethodInfoLocator methodInfoLocator) in D:\a\_work\1\s\src\DotNetWorker.Grpc\GrpcWorker.cs:line 269```

Basically I have a system topic that is subscribed to "BlobCreated" events in a blob container, which sends the events to the queue storage and those events are then handled by the queue trigger function.

After I upload items to the blob container (100 or less at a time), the function works fine for a while until this error starts occurring. What could possibly cause this error and what does it mean exactly?

I am using these package versions:
```
<PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Storage.Queues" Version="5.2.0" />
<PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.20.0" />
<PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.16.3" />
`````

host.json:
```
{
    "version": "2.0",
    "logging": {
        "applicationInsights": {
            "samplingSettings": {
                "isEnabled": true,
                "excludedTypes": "Request"
            }
        }
    },
    "extensions": {
        "queues": {
            "batchSize": 2,
            "newBatchThreshold": 1,
            "maxDequeueCount": 1,
            "messageEncoding": "base64"
        }
    }
}
```

Program.cs (with business logic removed):

```
using Microsoft.Extensions.Azure;
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Hosting;

var host = new HostBuilder()
    .ConfigureServices((services) =>
    {
		... services here
    })
    .Build();

host.Run();
```

Function looks like this:
```
[Function("HandleData")]
public async Task Run([QueueTrigger("data-queue")] BlobEvent input)
{
    ...handle data
}
```

There are no other functions in the Function App

### PR DESCRIPTION
**Pull request by DerGuru**
resolves #3909 

### Pull request checklist

* [X] My changes **do not** require documentation changes
* [X] I've added my notes to `release_notes.md`
* [X] My changes **do not** need to be backported to a previous version
* [X] I have added all required tests (Unit tests, E2E tests)

### Additional information
I made it more complex, than it strictly had to be, but in order to not break every unit test, I decided for this approach.
I also would have preferred, to use the type of FunctionAttribute, but since this is declared in a different assembly, which is not added as a dependency, I decided to use the FullName of the type instead.
 
Additional PR information
I changed as little as possible.

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -3,6 +3,7 @@
 <!-- Please add your release notes in the following format:
 - My change description (#PR/#issue)
 -->
+- Locate the Azure Function Entrypoint first by name and take the method with the FunctionAttribute (#2124)
 
 ### Microsoft.Azure.Functions.Worker (metapackage) 1.20.1
 
--- a/src/DotNetWorker.Core/Invocation/DefaultMethodInfoLocator.cs
+++ b/src/DotNetWorker.Core/Invocation/DefaultMethodInfoLocator.cs
@@ -33,7 +33,8 @@ namespace Microsoft.Azure.Functions.Worker.Invocation
             Type? functionType = assembly.GetType(typeName);
             IEnumerable<MethodInfo> possibleEntryPoints = functionType?.GetMethods().Where(x => x.Name.Equals(methodName)) ?? Enumerable.Empty<MethodInfo>();
 
-            MethodInfo? methodInfo =  possibleEntryPoints.FirstOrDefault(m => m.CustomAttributes.Any(ca => ca.AttributeType.FullName == "Microsoft.Azure.Functions.Worker.FunctionAttribute"))
+            MethodInfo? methodInfo =  
+                   possibleEntryPoints.FirstOrDefault(m => m.CustomAttributes.Any(ca => ca.AttributeType.FullName == "Microsoft.Azure.Functions.Worker.FunctionAttribute"))
                 ?? possibleEntryPoints.FirstOrDefault();
 
             if (methodInfo == null)
--- a/src/DotNetWorker.Core/Invocation/DefaultMethodInfoLocator.cs
+++ b/src/DotNetWorker.Core/Invocation/DefaultMethodInfoLocator.cs
@@ -1,4 +1,6 @@
 ﻿using System;
+using System.Collections;
+using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 #if NET5_0_OR_GREATER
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by DerGuru** [comment]

**Comment by DerGuru** [comment]
Sorry for the mess with with the commits ... this is my very first pull request.

**Comment by kshyju** [comment]
@DerGuru Thanks for your contribution. The PR description says this change fixes issue 3909 which does not exist.  What problem are we fixing here?

**Comment by DerGuru** [comment]
Hi,The issue was closed already, but I fell into the same pittrap.I had a class with an Azure Function doing one thing and needed it (a bit differently but almost the same) another time.Original became the base class, two derived classes using the previous class’s Original Function method with a different parameter. Done, I thought. But then the Function would not start.I googled the error message and found the closed issue. Nevertheless I wasn‘t the only one and so I had a look at the code and fixed it.I am not completely sure about the performance, but since there is a Regular Expression at work, the hit cannot be that hard. I could have made it a bit faster, but then every Unit Test I saw, that was using the DefaultMethodInfoLocator would have failed.I worked around it already in my code and if you decide, to keep the old behavior, I am fine with it.Kind Regards AndreasAm 04.01.2024 um 17:31 schrieb Shyju Krishnankutty ***@***.***>:﻿
@DerGuru Thanks for your contribution. The PR description says this change fixes issue 3909 which does not exist.  What problem are we fixing here?

**Comment by kshyju** [comment]
@DerGuru Closed issues are still available to view. Can you share the link please?
Apologies. I am still struggling to understand the use case you mentioned. It would be great if you can share the relevant code and the error message you received (if any)/ actual output. We would like to understand the issue before going further to absorb this change.

**Comment by DerGuru** [comment]
Hi, 

sorry for the delay. Here is the [link](https://github.com/Azure/azure-functions-host/issues/3909) to the pit I fell into as well.
All your questions will be answered in the opening post there. It was created for Function Apps v1.0 but still current behaviour.
I had a litte different constellation of classes though.

```csharp
public abstract class BaseClass
{
    protected abstract Task<IEnumerable<Entity>> GetEntites();
    protected async Task Execute(bool doIt) 
    {
        var entities = GetEntites();
        await foreach (var entity in entities)
        {
            if (doIt)
               await DoWith(entity);
            else
               await JustAddToReport(entity);
        }
    }
}

public class DerivedClass1 : BaseClass 
{
    protected override async Task<IEnumerable<Entity>> GetEntities()
    {
        // get and return entities
    }

    [Function("DoOne")]
    public async Task Execute([0 0 4 * * *] TimerTrigger trigger)
    {
        await base.Execute(true); 
    }
}


public class DerivedClass2 : BaseClass 
{
    protected override async Task<IEnumerable<Entity>> GetEntities()
    {
        // get and return entities
    }
    [Function("DoTwo")]
    public async Task Execute([0 0 5 * * *] TimerTrigger trigger)
    {
        await base.Execute(false); //just a report for now
    }
}
```

DerGuru:
Ahhh I just saw, that it was created for Azure-Functions-Host. But I guess, the code has moved since 2018.

kshyju:
@DerGuru I am going to close this PR based on the previous comments/discussion. But as I mentioned in the first comment, we absolutely welcome external contributions, when you have one. Thank you!

Issue #1159, opened 2022-11-08
Allow for multiple IFunctionMetadataProvider implementations


jviau:
# Overview

To enable functions being exposed from referenced assemblies - via the source generator or otherwise - we should expand the `IFunctionMetadataProvider` contract to allow for multiple implementations. Right now, it is assumed only one of them is registered to DI, limiting the ability to provide additional function metadata.

# Use Cases

## Functions from multiple projects

Customers may structure their application with functions split between multiple projects. These assemblies may have their own `IFunctionMetadataProvider` (from code-gen or otherwise) which needs to be registered in addition to the current projects generated provider. Supporting multiple providers will enable this use case.

## Functions from extensions

Some extensions may want to expose function metadata definitions in their own way. For example, to support class-based orchestrators and activities in Durable Functions, the extension could provide functions metadata which all point to the same implementation which shims to the target class to invoke.

# Concerns

The existing `DefaultFunctionMetadataProvider` only wants to be registered if there is no other provider: https://github.com/Azure/azure-functions-dotnet-worker/blob/9c3380d36af1ed79040707f62cdf2141410c53de/src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs#L55. However, it is not as simple as that. Really this is only wants to be registered if there is no provider effectively _replacing_ it - in this case the generated provider. This is because these two providers are aiming to provide metadata for the **same** set of functions. However, not all providers will want to replace that metadata provider. Some thought/design will need to go into solving the differentiation between overlapping providers and the conditional registration of the default provider.

danielmarbach:
We stumbled over this issue while looking into a problem we are facing. We are generating a function that gets triggered with a service bus trigger with a generator. Unfortunately, the order of source generator invocations is not deterministic and we are running after the function metadata provider source generator, which causes our function to not be recognized and added to the DirectFunctionExecutor internal dictionary. 

Is there are reason the whole metadata gathering is not more composable? For example, why not allow registering multiple providers that are then aggregated by the host automatically to populate the metadata dictionary?

danielmarbach:
@fabiocav @jviau I have added a spike draft PR for discussion https://github.com/Azure/azure-functions-dotnet-worker/pull/2183#pullrequestreview-1810109597.

danielmarbach:
Quick question: Is it possible to execute functions when only the metadata is know when the source gened function executor is in use?

jviau:
closed by #3145

PR #2183, opened 2024-01-08
Spike function metadata customization

danielmarbach:
This is an attempt to spike function metadata customization as outlined in https://github.com/Azure/azure-functions-dotnet-worker/issues/1159. This PR is deliberately not "done" in order to use the PR as a discussion point. I made inline comments where necessary.

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Emitter.cs
+++ b/sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Emitter.cs
@@ -66,6 +66,7 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
                                  {
                                      builder.ConfigureServices(s => 
                                      {
+                                         s.TryAddSingleton<IFunctionMetadataProviderAggregator, SkipDefaultFunctionMetadataProviderAggregator>();
                                          s.AddSingleton<IFunctionMetadataProvider, GeneratedFunctionMetadataProvider>();
                                      });
                                      return builder;
@@ -130,7 +131,7 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
 
                     builder.Append(BuildRetryOptions(function.Retry));
 
-                    builder.AppendLine($$"""                
+                    builder.AppendLine($$"""
                                     ScriptFile = "{{function.ScriptFile}}"
                     """);
                     builder.AppendLine($$"""            };""");
--- a/src/DotNetWorker.Core/FunctionMetadata/DefaultFunctionMetadataProviderAggregator.cs
+++ b/src/DotNetWorker.Core/FunctionMetadata/DefaultFunctionMetadataProviderAggregator.cs
@@ -1,0 +1,29 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Functions.Worker.Core.FunctionMetadata;
+
+public class DefaultFunctionMetadataProviderAggregator : IFunctionMetadataProviderAggregator
+{
+    private readonly IEnumerable<IFunctionMetadataProvider> _functionMetadataProviders;
+
+    public DefaultFunctionMetadataProviderAggregator(IEnumerable<IFunctionMetadataProvider> functionMetadataProviders)
+    {
+        _functionMetadataProviders = functionMetadataProviders;
+    }
+
+    public async ValueTask<ImmutableArray<IFunctionMetadata>> GetFunctionMetadataAsync(string scriptRoot)
+    {
+        var aggregatedMetadata = new List<IFunctionMetadata>();
+        foreach (var functionMetadataProvider in _functionMetadataProviders)
+        {
+            var functionMetadata = await functionMetadataProvider.GetFunctionMetadataAsync(scriptRoot).ConfigureAwait(false);
+            aggregatedMetadata.AddRange(functionMetadata);
+        }
+        return aggregatedMetadata.ToImmutableArray();
+    }
+}
--- a/src/DotNetWorker.Core/FunctionMetadata/DefaultMetadataProviderAttribute.cs
+++ b/src/DotNetWorker.Core/FunctionMetadata/DefaultMetadataProviderAttribute.cs
@@ -1,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.Functions.Worker.Core.FunctionMetadata;
+
+[AttributeUsage(AttributeTargets.Class)]
+public sealed class DefaultMetadataProviderAttribute : Attribute
+{
+}
--- a/src/DotNetWorker.Core/FunctionMetadata/IFunctionMetadataProviderAggregator.cs
+++ b/src/DotNetWorker.Core/FunctionMetadata/IFunctionMetadataProviderAggregator.cs
@@ -1,0 +1,19 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Immutable;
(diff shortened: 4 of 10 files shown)


src/DotNetWorker.Core/FunctionMetadata/DefaultFunctionMetadataProviderAggregator.cs
```diff
@@ -0,0 +1,29 @@
...
+    private readonly IEnumerable<IFunctionMetadataProvider> _functionMetadataProviders;
+
+    public DefaultFunctionMetadataProviderAggregator(IEnumerable<IFunctionMetadataProvider> functionMetadataProviders)
+    {
+        _functionMetadataProviders = functionMetadataProviders;
+    }
+
+    public async ValueTask<ImmutableArray<IFunctionMetadata>> GetFunctionMetadataAsync(string scriptRoot)
+    {
+        var aggregatedMetadata = new List<IFunctionMetadata>();
```
danielmarbach, line 21: Should the function id be used to deduplicate the metadata?

src/DotNetWorker.Core/FunctionMetadata/DefaultMetadataProviderAttribute.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.Functions.Worker.Core.FunctionMetadata;
+
+[AttributeUsage(AttributeTargets.Class)]
+public sealed class DefaultMetadataProviderAttribute : Attribute
```
danielmarbach, line 9: Currently using an attribute to mark the default internal one. I went back and forth on this one but at the end I ended up adding this attribute to the Core project. Maybe someone has a better idea?

src/DotNetWorker.Core/FunctionMetadata/IFunctionMetadataProviderAggregator.cs
```diff
@@ -0,0 +1,19 @@
...
+
+using System.Collections.Immutable;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Functions.Worker.Core.FunctionMetadata;
+
+/// <summary>
+/// Aggregates all the function metadata.
+/// </summary>
+public interface IFunctionMetadataProviderAggregator
```
danielmarbach, line 12: Not particularly excited of this name but it gets the discussion started

src/DotNetWorker.Core/FunctionMetadata/SkipDefaultFunctionMetadataProviderAggregator.cs
```diff
@@ -0,0 +1,33 @@
...
+
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Functions.Worker.Core.FunctionMetadata;
+
+public class SkipDefaultFunctionMetadataProviderAggregator : IFunctionMetadataProviderAggregator
```
danielmarbach, line 12: Made public so that the corresponding aggregator can be registered by the extensions

src/DotNetWorker.Core/FunctionMetadata/SkipDefaultFunctionMetadataProviderAggregator.cs
```diff
@@ -0,0 +1,33 @@
...
+    public SkipDefaultFunctionMetadataProviderAggregator(IEnumerable<IFunctionMetadataProvider> functionMetadataProviders)
+    {
+        _functionMetadataProviders = functionMetadataProviders
+            .Where(p => p.GetType().GetCustomAttribute<DefaultMetadataProviderAttribute>() == default)
+            .ToArray();
+    }
+
+    public async ValueTask<ImmutableArray<IFunctionMetadata>> GetFunctionMetadataAsync(string scriptRoot)
+    {
+        var aggregatedMetadata = new List<IFunctionMetadata>(_functionMetadataProviders.Length);
```
danielmarbach, line 25: Should the function id be used to deduplicate the metadata?

src/DotNetWorker.Grpc/FunctionMetadata/DefaultFunctionMetadataProvider.cs
```diff
@@ -12,6 +12,7 @@
 
 namespace Microsoft.Azure.Functions.Worker
 {
+    [DefaultMetadataProvider]
```
danielmarbach, line 15: Marked this one for now. Another option could be to use a the full type name instead?

src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs
```diff
@@ -44,8 +44,10 @@ public static IServiceCollection AddGrpc(this IServiceCollection services)
             services.AddSingleton<IUserMetricWriter>(p => p.GetRequiredService<GrpcFunctionsHostLogWriter>());
             services.AddSingleton<IWorkerDiagnostics, GrpcWorkerDiagnostics>();
 
+            // Default aggregator that collects metadata from all the metadata providers
+            services.TryAddSingleton<IFunctionMetadataProviderAggregator, DefaultFunctionMetadataProviderAggregator>();
```
danielmarbach, line 48: Allowing the aggregator to be overridden by the extensions

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Emitter.cs
```diff
@@ -66,6 +66,7 @@ public static IHostBuilder ConfigureGeneratedFunctionMetadataProvider(this IHost
                                  {
                                      builder.ConfigureServices(s => 
                                      {
+                                         s.TryAddSingleton<IFunctionMetadataProviderAggregator, SkipDefaultFunctionMetadataProviderAggregator>();
```
danielmarbach, line 69: Would have to see whether `AddSingleton` or `TryAddSingleton` is better. For now using the `TryAdd` to potentially allow full customization. But I wonder if that is even possible due to the non-deterministic nature of the auto configuration extensions call order?

kshyju:
@danielmarbach Is this still a valid PR to be reviewed?

danielmarbach:
@kshyju why shouldn't it be? :)

I got no feedback at all so far. Like I mentioned in the description the intent was to do a spike to explore options for https://github.com/Azure/azure-functions-dotnet-worker/issues/1159 and get some feedback and insights form team

kshyju:
cc @jviau

jviau:
@danielmarbach sorry for taking so long to get to this. I have updated the description of #1159 to reflect some internal discussions we have had. Overall the requirements for this feature have been simplified, but I also clarified some breaking change concerns I want to make sure we avoid. Do you want to re-review that issue to see if your PR is inline with the new description?

I think one main thing to call out is avoiding impacting the service cardinality of `IFunctionMetadataProvider`, as adding more implementations of this to the service container may be a behavior breaking change.

danielmarbach:
@jviau At the moment the requirements you mentioned in the linked issue could be fulfilled. Because the source generator can register the skip aggregator which then skips the primary one but executes all the auxiliary ones. 

The problem I see with saying the primary one can be replaced by another primary one it becomes a bit tricky to differentiate those cases. I'm also not sure whether you want to handle this with "should not be registered" (as you mentioned in the issue) but rather express this conceptually in the code so that you never have to rely on DI registration orders.

But if you want to go down that route then we might want to work with a magic string that represents the primary and then check whether there is another primary and switch to that one or throw if there are more than two primaries. But all this feels a bit awkward

> I think one main thing to call out is avoiding impacting the service cardinality of IFunctionMetadataProvider, as adding more implementations of this to the service container may be a behavior breaking change.

Yeah currently this PR suggests registering multiple metadata providers. Otherwise you would have to introduce different marker interfaces for the different metadata providers to make sure only ever one IFunctionMetadataProvider stays registered

danielmarbach:
Would it help to introduce IAuxiliaryFunctionMetadataProvider that is allowed to have a different cardinality and leave the current interface the single replaceable one?

jviau:
> Would it help to introduce IAuxiliaryFunctionMetadataProvider that is allowed to have a different cardinality and leave the current interface the single replaceable one?

This sounds like a reasonable approach.

PR #2190, opened 2024-01-09
Fix unit tests when ran locally

jviau:
Quick fix for tests that were failing locally due to not accounting for daylight savings differences on local machine.


satvu (approved):

jviau:

jviau:
@microsoft-github-policy-service rerun

PR #2210, opened 2024-01-12
Decouple FunctionsEnableMetadataSourceGen from FunctionsEnableWorkerIndexing to actually make it work properly when set

### LINKED ISSUE #2209: FunctionsEnableMetadataSourceGen doesn't disable the metadata generator
Potentially related issues

https://github.com/Azure/azure-functions-dotnet-worker/issues/2186
https://github.com/Azure/azure-functions-dotnet-worker/issues/1058

as suggested by @jviau in https://github.com/Azure/azure-functions-dotnet-worker/issues/2186#issuecomment-1885330533 it should be possible to control the individual source generators with

```xml
<FunctionsEnableExecutorSourceGen>false</FunctionsEnableExecutorSourceGen>
<FunctionsEnableMetadataSourceGen>false</FunctionsEnableMetadataSourceGen>
```

while `FunctionsEnableExecutorSourceGen` works setting `FunctionsEnableMetadataSourceGen` to `false` doesn't lead to disabling the metadata source generator. Only when `FunctionsEnableWorkerIndexing` is set to `false` the metadata source generator is actually disabled so in essence  `FunctionsEnableMetadataSourceGen` has no effect.

danielmarbach:

resolves #2209 

Unfortunately, I have no context of the project to be able to say this change of decoupling the properties is safe, but at least from the expectations I had from the description of the property and how the executor generation works, I would have expected to be able to disable the metadata source gen independently of the function indexing.

I did smoke test the package locally with the dev build, and it seems to work (see after GIF) and my function is still executing.

### Pull request checklist

### Additional information

This changes potentially requires a backport
Happy to add necessary release notes once I get some feedback from the team.

#### Before

Before

#### After
After

### DIFF
--- a/sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
+++ b/sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
@@ -37,9 +37,9 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
         <FunctionsEnablePlaceholder Condition="$(FunctionsEnablePlaceholder) == ''">false</FunctionsEnablePlaceholder>
         <FunctionsEnableWorkerIndexing Condition="$(FunctionsEnableWorkerIndexing) == '' Or $(FunctionsEnableWorkerIndexing)">true</FunctionsEnableWorkerIndexing>
         <FunctionsEnableMetadataSourceGen Condition="$(FunctionsEnableMetadataSourceGen) == '' Or $(FunctionsEnableMetadataSourceGen)">true</FunctionsEnableMetadataSourceGen>
-        <FunctionsAutoRegisterGeneratedMetadataProvider Condition="$(FunctionsAutoRegisterGeneratedMetadataProvider) == '' Or (FunctionsAutoRegisterGeneratedMetadataProvider)">true</FunctionsAutoRegisterGeneratedMetadataProvider>
+        <FunctionsAutoRegisterGeneratedMetadataProvider Condition="$(FunctionsAutoRegisterGeneratedMetadataProvider) == '' Or $(FunctionsAutoRegisterGeneratedMetadataProvider)">true</FunctionsAutoRegisterGeneratedMetadataProvider>
         <FunctionsEnableExecutorSourceGen Condition="$(FunctionsEnableExecutorSourceGen) == '' Or $(FunctionsEnableExecutorSourceGen)">true</FunctionsEnableExecutorSourceGen>
-        <FunctionsAutoRegisterGeneratedFunctionsExecutor Condition="$(FunctionsAutoRegisterGeneratedFunctionsExecutor) == '' Or (FunctionsAutoRegisterGeneratedFunctionsExecutor)">true</FunctionsAutoRegisterGeneratedFunctionsExecutor>
+        <FunctionsAutoRegisterGeneratedFunctionsExecutor Condition="$(FunctionsAutoRegisterGeneratedFunctionsExecutor) == '' Or $(FunctionsAutoRegisterGeneratedFunctionsExecutor)">true</FunctionsAutoRegisterGeneratedFunctionsExecutor>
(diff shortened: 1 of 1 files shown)


sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
```diff
@@ -36,11 +36,10 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
       
         <FunctionsEnablePlaceholder Condition="$(FunctionsEnablePlaceholder) == ''">false</FunctionsEnablePlaceholder>
         <FunctionsEnableWorkerIndexing Condition="$(FunctionsEnableWorkerIndexing) == '' Or $(FunctionsEnableWorkerIndexing)">true</FunctionsEnableWorkerIndexing>
-        <FunctionsEnableMetadataSourceGen>$(FunctionsEnableWorkerIndexing)</FunctionsEnableMetadataSourceGen>
-        <FunctionsAutoRegisterGeneratedMetadataProvider>$(FunctionsEnableWorkerIndexing)</FunctionsAutoRegisterGeneratedMetadataProvider>
+        <FunctionsEnableMetadataSourceGen Condition="$(FunctionsEnableMetadataSourceGen) == '' Or $(FunctionsEnableMetadataSourceGen)">true</FunctionsEnableMetadataSourceGen>
```
danielmarbach, line 39: I followed deliberately the same condition pattern already in this file for consistency but generally these conditions don't make a lot of sense. This doesn't protect anyone from setting a value that doesn't evaluate to a boolean. For example you could still set `foo` as a value which then makes the condition blow up. Wouldn't it be better actually have a condition like the following:

```xml
Condition="$(FunctionsEnableMetadataSourceGen) == '' Or $(FunctionsEnableMetadataSourceGen) != 'false'
```

this would then at least set the default value to `true` when it is empty or when an invalid value like `foo` is set.

danielmarbach:
@jviau you have been so kind to review my last contribution. Would you mind giving this one a go too?

satvu:
Hi @danielmarbach, thanks for being a contributor! We will review and come back to this PR. I believe these two properties can be decoupled as they were introduced separately when we first implemented worker indexing (still using `functions.metadata`) then replaced the `functions.metadata` file creation and read with the source-generated metadata. However, performance benefits are only gained from the combination of these two features, so the "worker-indexing only" path is not preferred.

These two props were coupled together with the intention of using `FunctionsEnableWorkerIndexing` as a catch-all to light up the parts that are needed to use source-generation (worker-indexing, source-generated metadata, source-generated executor, auto-registration of the generated metadata provider, etc). Additionally, some of the other properties this PR touches need to be coupled to work properly.

You've brought up a great point, and we will review to see if we can improve on this to give users the ability to control these features with more granularity without breaking themselves.

fabiocav (commented):
Thank you, @danielmarbach!

Indeed, decoupling the worker indexing behavior from the generated provider is how this is supposed to work, so this is a great change, but we do need to ensure at least a couple of things are done:

- The source generated provider (or any injected provider that doesn't make metadata information available to the metadata loader) will not work properly if worker indexing is disabled. Enabling the source generated provider without worker indexing will lead to issues, so it shouldn't be a valid state (should result in a build failure). To simplify the way to think about this, the only provider that will work with worder indexing disabled is the `DefaultFunctionMetadataProvider`, and it's not super realistic (or desirable) to expect others to support that mode.
- Having `FunctionsAutoRegisterGeneratedMetadataProvider` set to `true` when the generated metadata provider is disabled should also be treated as an invalid state.

kshyju:
@danielmarbach Are these changes still valid as of today? Is this good to be reviewed?

danielmarbach:
@kshyju I think so. I do not have the capacity nor the detailed knowledge to address the feedback of @fabiocav https://github.com/Azure/azure-functions-dotnet-worker/pull/2210#pullrequestreview-1846844982, unfortunately. So I leave it up to your team to decide how you want to proceed from here.

sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
kshyju: Is this not a breaking change? Anyone who has previously used `FunctionsEnableWorkerIndexing` to disable the metadata source gen + auto registration will be seeing a different behavior with this change.

PR #2217, opened 2024-01-16
[ApplicationInsights] Enable AAD Auth

RohitRanjanMS:
Enabling AAD Authentication in the ApplicationInsights worker package is achieved by configuring the APPLICATIONINSIGHTS_AUTHENTICATION_STRING. This configuration ensures a smooth Azure Active Directory Authentication experience across all ApplicationInsights pipelines.
- Host AI Integration
- Snapshot
- Profiler
- Java AI Agent (Released)
- Node AI Agent (Inprogress)
- Python AI Agent (Planned)
- Isolated AI package (with this proposed change)

I will update the doc after the release.
### Pull request checklist


src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
```diff
@@ -10,21 +10,34 @@
...
+            _options = options.Value;
+        }
+
         public void Initialize(TelemetryConfiguration configuration)
         {
             _telemetryClient = new TelemetryClient(configuration);
 
+            if (_options.TokenCredentialOptions != null)
+            {
+                configuration.SetAzureTokenCredential(_options.TokenCredentialOptions.CreateTokenCredential());
```
jviau, line 38: I don't think setting it here is correct. I can't remember exactly but I believe this may be a copy of `TelemetryConfiguration`.

Instead, I think you need to do the following as part of options configuration of `TelemetryConfiguration`

src/DotNetWorker.ApplicationInsights/Initializers/FunctionsApplicationInsightsOptionsInitializer.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace Microsoft.Azure.Functions.Worker.ApplicationInsights.Initializers
+{
+    internal class FunctionsApplicationInsightsOptionsInitializer : IConfigureOptions<FunctionsApplicationInsightsOptions>
+    {
+        private const string AppInsightsAuthenticationString = "APPLICATIONINSIGHTS_AUTHENTICATION_STRING";
+
+        public void Configure(FunctionsApplicationInsightsOptions options)
+        {
+            // Read AuthString from environment variable
+            if (!string.IsNullOrEmpty(Environment.GetEnvironmentVariable(AppInsightsAuthenticationString)))
```
jviau, line 13: Before reading environment, check if the same value is in `IConfiguration`, then fallback to reading environment directly.

jviau (dismissed):
Given that `TelemetryConfiguration` is already used in the IOptions pattern, we can directly configure the credential on it. No need to go through `FunctionsApplicationInsightsOptions`.

src/DotNetWorker.ApplicationInsights/FunctionsTelemetryModule.cs
RohitRanjanMS: Done.

src/DotNetWorker.ApplicationInsights/Initializers/FunctionsApplicationInsightsOptionsInitializer.cs
RohitRanjanMS: Done.

src/DotNetWorker.ApplicationInsights/DotNetWorker.ApplicationInsights.csproj
```diff
@@ -16,7 +16,8 @@
   <Import Project="..\..\build\Common.props" />
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.ApplicationInsights.PerfCounterCollector" Version="2.21.0" />
+    <PackageReference Include="Azure.Identity" Version="1.10.2" />
```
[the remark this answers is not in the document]
brettsam: Adding a new package should be okay for a minor version bump.

brettsam (approved):

andystaples (approved):

karpikpl:
Is there documentation that goes with this PR?
I see that portal now supports Identity for Insights, but it adds 2 connection string - one with AAD and one without (which defeats the purpose)
image

after removing the connection string with instrumentation key, nothing is collected, even though portal deployment correctly set IAM roles and `APPLICATIONINSIGHTS_AUTHENTICATION_STRING`

PR #2236, opened 2024-01-25
Fix & suppress warnings

jviau:
Fixes build warnings

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,8 @@
+<Project>
+
+  <PropertyGroup>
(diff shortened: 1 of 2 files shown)


Directory.Build.props
```diff
@@ -0,0 +1,8 @@
+<Project>
+
+  <PropertyGroup>
+    <!-- TODO: remove when updating away from .NET5.0 -->
```
fabiocav, line 4: 👍🏼

fabiocav (approved):

PR #2240, opened 2024-01-26
adding issue templates for worker repo

mattchenderson:
Adding issue templates, using the new forms feature. We could also just use the standard markdown template as well.

Draft for discussion purposes. I also think the strings need some work, and likely we will want to add other sections. But I think letting that be built up over time is fine. We currently have no template, so just having one big description field is good enough to get started.

### Pull request checklist

  * [] Otherwise: Documentation issue linked to PR

### DIFF
--- a/.github/ISSUE_TEMPLATE/01-bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/01-bug-report.yml
@@ -1,0 +1,18 @@
+name: Bug Report
+description: Create a report to help us improve
+labels: ["bug"]
+body: 
+- id: description
+  type: textarea
+  attributes:
+    label: Description
+    placeholder: Please provide a succinct description of the issue.
+  validations:
+    required: true
+- id: repro
+  type: textarea
+  attributes:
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/01-bug-report.yml
```diff
@@ -0,0 +1,18 @@
+name: Bug Report
+description: Create a report to help us improve
+labels: ["bug"]
+body: 
+- id: description
+  type: textarea
+  attributes:
+    label: Description
+    placeholder: Please provide a succinct description of the issue.
```
kshyju, line 9: How can we tell the user to include information such as :

1. Package versions (worker, worker.sdk , worker extensions if any)
2. Repro steps
3. Link to minimal repro app (if possible to provide)
4. If issue only happening on deployed environment, provide SKU details

fabiocav: I agree with all of the above. Let's get the changes in and iterate.

PR #2315, opened 2024-03-01
Improvements to NetHost logging & worker.config.

kshyju:
#### 1. Logging
Improvements to logging in FNH. Currently we query the env variable for every call to TraceLog method. Based on the usecases I have encountered so far, I do not see a reason for the env value to change after the placeholder started. So I think it makes sense to read that only once during startup.

Also adding some customization to not prepend the special log prefix we add to the messages. This is helpful for local debugging. Previous version did this only for Debug build and this PR is changing it to rely on an env variable.

#### 2. Worker.config

Updated worker config to include one more profile condition which evaluates the environment variable `INITIALIZED_FROM_PLACEHOLDER`. This will ensure that FunctionsNetHost will never be launched in non placeholder mode. During some recent investigations, we found that FNH is being launched in specialization time.  Java worker config [uses this condition](https://github.com/Azure/azure-functions-java-worker/blob/c92d38cdeb867f2f7e2ae9b2515c32c2ecfd1e41/worker.config.json#L34C49-L34C62).

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/host/src/FunctionsNetHost/Logger.cs
+++ b/host/src/FunctionsNetHost/Logger.cs
@@ -8,22 +8,13 @@ namespace FunctionsNetHost
     internal static class Logger
     {
         private static readonly string LogPrefix;
+        private static readonly bool IsTraceLogEnabled;
 
         static Logger()
         {
-#if !DEBUG
-            LogPrefix = "LanguageWorkerConsoleLog";
-#else
-            LogPrefix = "";
-#endif
-        }
-
-        internal static bool IsTraceLogEnabled
-        {
-            get
-            {
-                return string.Equals(EnvironmentUtils.GetValue(EnvironmentVariables.FunctionsNetHostTrace), "1");
-            }
+            IsTraceLogEnabled = string.Equals(EnvironmentUtils.GetValue(EnvironmentVariables.EnableTraceLogs), "1");
+            var disableLogPrefix = string.Equals(EnvironmentUtils.GetValue(EnvironmentVariables.DisableLogPrefix), "1");
+            LogPrefix = disableLogPrefix ? string.Empty : "LanguageWorkerConsoleLog";
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


satvu (approved):

host/tools/build/worker.config.json
```diff
@@ -2,24 +2,46 @@
...
-      "profileName": "DotnetIsolatedLinux",
+      "profileName": "DotnetIsolatedLinuxPlaceholder",
       "conditions": [
         {
           "conditionType": "hostProperty",
           "conditionName": "platform",
           "conditionExpression": "LINUX"
+        },
+        {
+          "conditionType": "environment",
```
fabiocav, line 17: 👍🏼

host/src/FunctionsNetHost/Logger.cs
```diff
@@ -8,22 +8,13 @@ namespace FunctionsNetHost
...
-#endif
-        }
-
-        internal static bool IsTraceLogEnabled
-        {
-            get
-            {
-                return string.Equals(EnvironmentUtils.GetValue(EnvironmentVariables.FunctionsNetHostTrace), "1");
-            }
+            IsTraceLogEnabled = string.Equals(EnvironmentUtils.GetValue(EnvironmentVariables.EnableTraceLogs), "1");
```
fabiocav, line 15: Had a chat with @kshyju. It would be helpful to re-evaluate this at specialization time so the site has the ability to override this behavior.

kshyju: @fabiocav Pushed an update which reloads the logger configuration on specialization.

fabiocav (approved):

host/src/FunctionsNetHost/Logger.cs
fabiocav: thank you!

Issue #2205, opened 2024-01-11
Azure V4 Function Multi-Response Output Binding failing when function async.


AndyWDev:
Hi Folks,

I just raised this as a bug, but not sure if I am also doing something incorrectly.

I am using VS2022 17.8.4. to develop an Azure v4 function running using the isolated worker process.  The application is using C# in .Net 8.0 running locally using the CosmosDB emulator.

The issue is that when I make the function async, the CosmosDB output fails with an error message:

> IFeatureCollection has been disposed

The error occurs after the function return statement but before the client receives the response. The HttpResponse portion is successfully returned to the client.

The function returns a multi-response class defined below which is used to write out the data to the local cosmosDB emulator and return a copy to the calling client in the http response.

```
public class MultiResponse {
  [CosmosDBOutput(databaseName: "mytestdb", containerName: "testContainer", partitionKey = "/myPart", Connection="myConnectionStr", CreateIfNotExists = true)]
public MyDoc Item {get; set;}
public HttpResponseData HttpResponse {get; set;}
}

public class MyDoc {
public String Id {get; set;}
public String Name {get; set;}
}
```

The function definition is defined as 

```
[Function("MyTestFunc")]
public async Task<MultiResponse> Run([HttpTrigger(AuthorizationLevel.Function, "get", "post", Route = null)] HttpRequestData req )
{
    MultiResponse resp = new MultiResponse
    {
        HttpResponse = req.CreateReponse(HttpStatusCode.OK),
        Item = new MyDoc() { Id = 1234, Name="Bob"}
    };

   resp.HttpResponse.Headers.Add("Content-Type", "Text/plain; charset=utf-8");
   await resp.HttpResponse.WriteStringAsync(JsonConvert.SerializeObject(resp.Item);  // copy for client.
   return resp;
}
```

When I change the Task<MultiResponse> to just MultiResponse in the function definition and make the contents of the function synchronous, it appears to work as expected and my CosmosDB instance contains the  record.

**Edit:** Update with further info.
I found that if I removed the async and task keywords from the Run() function making it synchronous I received the AllowSynchronousIO warning which is ok as that is what it is for.

I then tried with just removing the async keyword and leaving Task<MultiResponse> and after removing any await keywords (and replacing with Task<Response<...>> and Task.Result calls) where I had a call to read some data (not shown) I found that it works, I get both the HTTPResponse and CosmosDB gets updated.

[...]

PR #2322, opened 2024-03-05
[BUG FIX] HTTP Trigger Functions with Multiple Output Bindings

### LINKED ISSUE #2079: Using Multiple Output Bindings with ASP.NET Core Integration
I was trying to combine [multiple output bindings](https://learn.microsoft.com/en-us/azure/azure-functions/dotnet-isolated-process-guide#multiple-output-bindings) with [ASP.NET Core Integration](https://learn.microsoft.com/en-us/azure/azure-functions/dotnet-isolated-process-guide#aspnet-core-integration).

Here's the type I'm returning from my HTTP triggered function:

```cs
public class NewOrderOutput
{
    public IActionResult? HttpResponse { get; set;}
    [QueueOutput("orders")]
    public OrderMessage? OrderMessage { get; set; }
    [CosmosDBOutput("db", "orders", Connection = "CosmosDB", CreateIfNotExists = false)]
    public OrderDocument? OrderDocument { get; set; } 
}
```

And here's using it
```cs
[Function(nameof(NewOrder))]
public async Task<NewOrderOutput> NewOrder(
    [HttpTrigger(AuthorizationLevel.Anonymous, "POST")] HttpRequest req, 
    [FromBody] OrderWebhook order)
{

    var response = new ContentResult()
    {
        Content = $"New order received for {order.ProductName} x {order.Quantity}"
    };

    var message = new OrderMessage() /// ...
    var doc = new OrderDocument()  //...
    return new NewOrderOutput() { OrderMessage = message, HttpResponse = response, OrderDocument = doc  };
}
```

When I call the function, I get no body in the HTTP response. However, if I change this up to use `HttpRequestData` and `HttpResponseData` it all works fine.

Should this feature work with `IActionResult`, or am I doing something wrong?

satvu:

resolves #2079 

### Pull request checklist

### Additional information

Currently, function metadata generation in both [legacy](https://github.com/Azure/azure-functions-dotnet-worker/blob/60489bbe72c0beb6339fdc456e3fa9d90fcc3f4e/sdk/Sdk/FunctionMetadataGenerator.cs#L289) and [source-generation](https://github.com/Azure/azure-functions-dotnet-worker/blob/60489bbe72c0beb6339fdc456e3fa9d90fcc3f4e/sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs#L471) have special handling of `HttpResponseData` types in multiple output binding scenarios. Neither of these generators handle new types introduced in ASP.NET Core Integration (like `IActionResult`). 

This PR introduces the following features:

- Introduction of `HttpResponseOutputAttribute`
  - Using this attribute on the intended HTTP response object will signal to generators to correctly generate function metadata
  - Explicit labeling will make the addition of new types easier and more regression-proof
- Handling of new attribute in source generator
-  ASP.NET Core Integration package changes
  - Introduction of `TryGetHttpResponse<T>` 
  - Refactoring to handle http response types in multiple output binding scenarios (where http is not a `$return` binding and has to be searched for inside of the output bindings list instead).
- Samples update

Not included:

- Fix in legacy generator (will follow up in separate PR).


extensions/Worker.Extensions.Http/src/HttpResponseAttribute.cs
```diff
@@ -0,0 +1,20 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    /// <summary>
+    /// Attribute used to mark an HTTP Response on an HTTP Trigger function with multiple output bindings.
+    /// </summary>
+    public sealed class HttpResponseOutputAttribute : OutputBindingAttribute
```
satvu, line 11: Would appreciate name suggestions - I am avoiding `HttpResponseAttribute` since it can be shortened to `HttpResponse` (and VS suggests that), but the source-generator pulls the `HttpResponse` symbol for comparison instead of the attribute in the source-generator in that case.

test/Sdk.Generator.Tests/FunctionExecutor/DependentAssemblyTest.cs
```diff
@@ -18,10 +20,10 @@ public DependentAssemblyTest()
             {
                 var abstractionsExtension = Assembly.LoadFrom("Microsoft.Azure.Functions.Worker.Extensions.Abstractions.dll");
                 var httpExtension = Assembly.LoadFrom("Microsoft.Azure.Functions.Worker.Extensions.Http.dll");
-                var hostingExtension = Assembly.LoadFrom("Microsoft.Extensions.Hosting.dll");
-                var diExtension = Assembly.LoadFrom("Microsoft.Extensions.DependencyInjection.dll");
-                var hostingAbExtension = Assembly.LoadFrom("Microsoft.Extensions.Hosting.Abstractions.dll");
-                var diAbExtension = Assembly.LoadFrom("Microsoft.Extensions.DependencyInjection.Abstractions.dll");
+                var hostingExtension = typeof(HostBuilder).Assembly;
```
satvu, line 23: Test cleanup items added to avoid dependency mismatches and errors. Can pull out to separate PR to go in before this one if desired.

extensions/Worker.Extensions.Http/src/HttpResponseAttribute.cs
mattchenderson: Existing convention for blobs, etc. would just leave it as `HttpOutputAttribute`, right?

samples/AspNetIntegration/MultipleOutputBindings.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        public class MyOutputType
+        {
+            [HttpResponseOutput]
+            public IActionResult HttpResponse { get; set; }
+
+            [QueueOutput("functionstesting2", Connection = "AzureWebJobsStorage")]
+            public string Name { get; set; }
+        }
+    }
```
mattchenderson, line 38: Can we set this up as a doc snippet? We can swap the reference to /samples/Extensions/MultiOutput/MultiOutput.cs's "docsnippet_multiple_outputs" in favor of one here.

extensions/Worker.Extensions.Http/src/HttpResponseAttribute.cs
satvu: @mattchenderson I also like that but had a brief convo with @fabiocav where we were between response/output and wondered if response made more sense intuitively ...

mattchenderson: I'm fine with `HttpResponseOutputAttribute`

samples/AspNetIntegration/MultipleOutputBindings.cs
mattchenderson:
```suggestion
    //<docsnippet_aspnetcore_multiple_outputs>
    public class MultipleOutputBindings
    {
        private readonly ILogger<MultipleOutputBindings> _logger;

        public MultipleOutputBindings(ILogger<MultipleOutputBindings> logger)
        {
            _logger = logger;
        }

        [Function("MultipleOutputBindings")]
        public MyOutputType Run([HttpTrigger(AuthorizationLevel.Function, "post")] HttpRequest req)
        {
            _logger.LogInformation("C# HTTP trigger function processed a request.");
            var myObject = new MyOutputType
            {
                Result = new OkObjectResult("C# HTTP trigger function processed a request."),
                MessageText= "some output"
            };

            return myObject;
        }

        public class MyOutputType
        {
            [HttpResponseOutput]
            public IActionResult Result { get; set; }

            [QueueOutput("myQueue")]
            public string MessageText { get; set; }
        }
    }
    //</docsnippet_aspnetcore_multiple_outputs>
```

sdk/Sdk.Generators/Constants.cs
```diff
@@ -41,7 +41,8 @@ internal static class Types
             internal const string BindingPropertyNameAttribute = "Microsoft.Azure.Functions.Worker.Extensions.Abstractions.BindingPropertyNameAttribute";
             internal const string DefaultValue = "Microsoft.Azure.Functions.Worker.Extensions.Abstractions.DefaultValueAttribute";
 
-            internal const string HttpResponse = "Microsoft.Azure.Functions.Worker.Http.HttpResponseData";
+            internal const string HttpResponseData = "Microsoft.Azure.Functions.Worker.Http.HttpResponseData";
+            internal const string HttpResponseAttribute = "Microsoft.Azure.Functions.Worker.HttpResponseOutputAttribute";
```
mattchenderson, line 45: I definitely agree with moving away from `HttpResponse` as the name here. That felt like a potential hazard.

Just noting that we seem to be inconsistent about the conventions for these string member names and how they map back to the types. I would probably err on the side of more verbose names which are distinct from the types they represent. but if we want them all to match, we should make the second one here `HttpResponstOutputAttribute`.

mattchenderson (changes requested):
My main request is for changes to the sample so it can be leveraged in the docs.

extensions/Worker.Extensions.Http/src/HttpResponseAttribute.cs
kshyju, line 11: Shall we specify `AttributeUsage` ? This will make it clear that where/how it is supposed to be used.

    [AttributeUsage(AttributeTargets.Property)]

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/KnownFunctionMetadataTypes.cs
```diff
@@ -52,7 +54,9 @@ internal KnownFunctionMetadataTypes(Compilation compilation)
 
         public INamedTypeSymbol? DefaultValue { get => _defaultValue.Value; }
 
-        public INamedTypeSymbol? HttpResponse { get => _httpResponse.Value; }
+        public INamedTypeSymbol? HttpResponseData { get => _httpResponseData.Value; }
+
+        public INamedTypeSymbol? HttpResponseOutputAtribute { get => _httpResponseOutputAttribute.Value; }
```
kshyju, line 59: typo (Atribute)

kshyju: https://devblogs.microsoft.com/visualstudio/visual-studio-spell-checker-preview-now-available/

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -519,6 +520,21 @@ private bool TryGetReturnTypePropertyBindings(ITypeSymbol returnTypeSymbol, bool
                 return true;
             }
 
+            private bool HasHttpResponseattribute(IPropertySymbol prop)
```
kshyju, line 523:
```suggestion
            private bool HasHttpResponseAttribute(IPropertySymbol prop)
```

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -519,6 +520,21 @@ private bool TryGetReturnTypePropertyBindings(ITypeSymbol returnTypeSymbol, bool
...
+            {
+                var attributes = prop.GetAttributes();
+                if (attributes.Length == 1)
```
kshyju, line 526: What if customer code has some extra attributes on this property? We should not assume it has only 1.

test/Sdk.Generator.Tests/FunctionMetadataProviderGeneratorTests/IntegratedTriggersAndBindingsTests.cs
```diff
@@ -177,6 +189,115 @@ public static IHostBuilder ConfigureGeneratedFunctionMetadataProvider(this IHost
...
+                            throw new NotImplementedException();
+                        }
+                    }
+
+                    public class MyOutputType
+                    {
+                        [QueueOutput("functionstesting2", Connection = "AzureWebJobsStorage")]
+                        public string Name { get; set; }
+
+                        [HttpResponseOutput]
```
kshyju, line 221: Let's ensure a test is covering the multiple attribute decoration case as well.

test/Sdk.Generator.Tests/Sdk.Generator.Tests.csproj
```diff
@@ -23,8 +24,8 @@
     <PackageReference Include="Microsoft.Extensions.Hosting" Version="8.0.0" />
     <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="8.0.0" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.8.0" />
-    <PackageReference Include="xunit" Version="2.6.6" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.5.6">
+    <PackageReference Include="xunit" Version="2.7.0"/>
```
kshyju, line 27: With this change, do we have multiple versions of these packages in the solution ? Perhaps we should consolidate (separate PR may be good)

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -467,8 +467,9 @@ private bool TryGetReturnTypePropertyBindings(ITypeSymbol returnTypeSymbol, bool
                         return false;
                     }
 
-                    // Check if this attribute is an HttpResponseData type attribute
-                    if (prop is IPropertySymbol property && SymbolEqualityComparer.Default.Equals(property.Type, _knownFunctionMetadataTypes.HttpResponse))
```
dandenton, line 471: Previously it worked to have an output type of `HttpResponse`.  Would this require existing `HttpResponse` response properties to start using `HttpResponseOutputAtribute` making this a breaking change?

satvu: We're adding this new case as part of an "OR" statement, so previous scenarios should still work - for example, `HttpResponseData` would still not require the attribute. I don't believe the output type of `HttpResponse` in this scenario was not a documented approach from our end, but I'll make a note to test it.

dandenton: It wasn't documented but I was just in the process of process of migrating from in-proc function to isolated and ran into #2205 (exception `IFeatureCollection has been disposed.` when returning the `HttpResponseData`) as we typically use IActionResult response using `Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore` for http. I switched to use HttpResponse on the output object to fix the issue.  We're currently returning HttpResponse now which is working, at least until we update to this version.  Is there any way to provide an Analyzer or something that would warn on upgrade that this is necessary?

test/Sdk.Generator.Tests/FunctionMetadataProviderGeneratorTests/IntegratedTriggersAndBindingsTests.cs
satvu: https://github.com/Azure/azure-functions-dotnet-worker/blob/e0cd2c152acf07b7813af38fac44c20b92704d9b/test/Sdk.Generator.Tests/FunctionMetadataProviderGeneratorTests/DiagnosticResultTests.cs#L106 - these already exist, and refactored so that the properties are checked for multiple bindings before processing potential http attributes.

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
satvu: @dandenton Good callout - tracking this issue [here](https://github.com/Azure/azure-functions-dotnet-worker/issues/2352). 

CC: @fabiocav @mattchenderson

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -481,35 +481,47 @@ private bool TryGetReturnTypePropertyBindings(ITypeSymbol returnTypeSymbol, bool
                         foundHttpOutput = true;
                         bindingsList.Add(GetHttpReturnBinding(prop.Name));
                     }
-                    else if (prop.GetAttributes().Length > 0) // check if this property has any attributes
+
+                    if (prop.GetAttributes().Length > 0)
```
kshyju, line 485: Store result of `prop.GetAttributes()` to a variable and use that in line 485 and 487.

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -481,35 +481,47 @@ private bool TryGetReturnTypePropertyBindings(ITypeSymbol returnTypeSymbol, bool
...
+                            if (foundHttpOutput)
                             {
-                                // validate that there's only one binding attribute per property
-                                if (foundPropertyOutputAttr)
-                                {
-                                    _context.ReportDiagnostic(Diagnostic.Create(DiagnosticDescriptors.MultipleBindingsGroupedTogether, Location.None, new object[] { "Property", prop.Name.ToString() }));
-                                    bindingsList = null;
-                                    return false;
-                                }
+                                _context.ReportDiagnostic(Diagnostic.Create(DiagnosticDescriptors.MultipleHttpResponseTypes, Location.None, new object[] { returnTypeSymbol.Name }));
```
kshyju, line 501: The diagnostic error uses "HttpResponseData" in the message, which may not be accurate if someone is applying `HttpResult` attribute on more than one properties.

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -481,35 +481,47 @@ private bool TryGetReturnTypePropertyBindings(ITypeSymbol returnTypeSymbol, bool
...
-                    else if (prop.GetAttributes().Length > 0) // check if this property has any attributes
+
+                    if (prop.GetAttributes().Length > 0)
                     {
-                        var foundPropertyOutputAttr = false;
+                        var bindingAttributes = prop.GetAttributes().Where(p => p.AttributeClass!.IsOrDerivedFrom(_knownFunctionMetadataTypes.BindingAttribute));
+
+                        if (bindingAttributes.Count() > 1)
+                        {
+                            _context.ReportDiagnostic(Diagnostic.Create(DiagnosticDescriptors.MultipleBindingsGroupedTogether, Location.None, new object[] { "Property", prop.Name.ToString() }));
```
kshyju, line 491: Can you remind me what code can cause this to diagnostic to be triggered? (487 to 491) ? What attribute customer may write (which inherits from bindingattribute) ?

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -520,16 +532,13 @@ private bool TryGetReturnTypePropertyBindings(ITypeSymbol returnTypeSymbol, bool
                 return true;
             }
 
-            private bool HasHttpResponseattribute(IPropertySymbol prop)
+            private bool HasHttpResultAttribute(ISymbol prop)
```
kshyju, line 535: This method is still checking for the first attribute of the symbol. Would it work for something like below, which is valid code customer may write.

```csharp
[SuppressMessage("Microsoft.Naming", "Foo", Justification = "Bar")]
[HttpResult]
public string Api_Response { set; get; }
```

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
satvu: [Here](https://github.com/Azure/azure-functions-dotnet-worker/blob/c1b2182e4611609e9b517828e871369a714e96b2/test/FunctionMetadataGeneratorTests/FunctionMetadataGeneratorTests.cs#L1362) is an example of stacking output bindings attributes from legacy tests. The check here is in case a user stacks multiple binding attributes on a property in the return type (ex. trying to bind one prop as a queue output and a blob output).

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
fabiocav: I'm tempted to say this should be within the scope of the changes.

fabiocav: Or at the very least, let's get this going in parallel

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
satvu: Edited the error, and double checked that this doc was still general enough for the new case: https://github.com/Azure/azure-functions-dotnet-worker/blob/main/docs/analyzer-rules/AZFW0007.md

extensions/Worker.Extensions.Http.AspNetCore/src/FunctionContext/FunctionContextExtensions.cs
```diff
@@ -47,5 +47,32 @@ internal static bool TryGetRequest(this FunctionContext context, [NotNullWhen(tr
...
             return request is not null;
         }
+
+        /// <summary>
+        /// Tries to get an HTTP Response of type T by first checking the invocation result then the output bindings of a function.
+        /// </summary>
+        /// <param name="context">The <see cref="FunctionContext"/></param>
+        /// <param name="result">The HTTP response for the respective function.</param>
+        /// <returns></returns>
+        public static bool TryGetHttpResponse<T>(this FunctionContext context, [NotNullWhen(true)] out T? result)
```
kshyju, line 57: If we are exposing a new public API, let's make sure it is properly documented. (Returns element is empty here).

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -480,35 +481,49 @@ private bool TryGetReturnTypePropertyBindings(ITypeSymbol returnTypeSymbol, bool
...
+                            bindingsList.Add(bindingDict!);
 
-                                returnTypeHasOutputBindings = true;
-                                foundPropertyOutputAttr = true;
-                            }
+                            returnTypeHasOutputBindings = true;
                         }
+
                     }
+
```
kshyju, line 525: njit: empty line

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -467,8 +467,9 @@ private bool TryGetReturnTypePropertyBindings(ITypeSymbol returnTypeSymbol, bool
                         return false;
                     }
 
-                    // Check if this attribute is an HttpResponseData type attribute
-                    if (prop is IPropertySymbol property && SymbolEqualityComparer.Default.Equals(property.Type, _knownFunctionMetadataTypes.HttpResponse))
+                    // Check for HttpResponseData type for legacy apps
+                    if (prop is IPropertySymbol property &&
+                            (SymbolEqualityComparer.Default.Equals(property.Type, _knownFunctionMetadataTypes.HttpResponseData)))
```
fabiocav, line 472: nit: move `&&` to this line

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -480,35 +481,49 @@ private bool TryGetReturnTypePropertyBindings(ITypeSymbol returnTypeSymbol, bool
                         foundHttpOutput = true;
                         bindingsList.Add(GetHttpReturnBinding(prop.Name));
                     }
-                    else if (prop.GetAttributes().Length > 0) // check if this property has any attributes
+
+                    var propAttributes = prop.GetAttributes();
```
fabiocav, line 485: We shouldn't continue after this check. If the block above succeeds, we don't need keep going.

extensions/Worker.Extensions.Http/src/HttpResultAttribute.cs
```diff
@@ -0,0 +1,15 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    /// <summary>
+    /// Attribute used to mark an HTTP Response on an HTTP Trigger function with multiple output bindings.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Property)]
```
fabiocav, line 11: Change to:
``` csharp
[AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
```

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -480,35 +481,49 @@ private bool TryGetReturnTypePropertyBindings(ITypeSymbol returnTypeSymbol, bool
...
-                                {
-                                    bindingsList = null;
-                                    return false;
-                                }
+                            foundHttpOutput = true;
+                            bindingsList.Add(GetHttpReturnBinding(prop.Name));
+                        }
+                        else
+                        {
+                            if (!TryCreateBindingDict(bindingAttributes.FirstOrDefault(), prop.Name, prop.Locations.FirstOrDefault(), out IDictionary<string, object>? bindingDict))
```
fabiocav, line 513: Can we take the opportunity to rename this to `TryCreateBindingDictionary`?

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
fabiocav: A couple of them :)

fabiocav (changes requested):

sdk/Sdk.Generators/DiagnosticDescriptors.cs
```diff
@@ -45,7 +45,7 @@ private static DiagnosticDescriptor Create(string id, string title, string messa
         public static DiagnosticDescriptor MultipleHttpResponseTypes { get; }
                   = Create(id: "AZFW0007",
```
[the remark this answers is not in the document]
satvu: The documentation was already generic and had no mention of `HttpResponseData`

mattchenderson:
Just confirming - the sample changes have been dropped from this PR, correct? For docs, should I plan on an inline revision then?

satvu:
> Just confirming - the sample changes have been dropped from this PR, correct? For docs, should I plan on an inline revision then?

For now, yes - the sample with the docsnippet will be added after the SDK is released and we can update the sample app reference.

extensions/Worker.Extensions.Http.AspNetCore/src/FunctionsMiddleware/FunctionsHttpProxyingMiddleware.cs
```diff
@@ -49,26 +59,53 @@ public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next
...
+                    ActionContext actionContext = new ActionContext(httpContext, httpContext.GetRouteData(), new ActionDescriptor());
+                    await actionResult.ExecuteResultAsync(actionContext);
+                    break;
+
+                case IResult iResult:
+                    await iResult.ExecuteAsync(httpContext);
+                    break;
+
+                default:
+                    throw new InvalidOperationException($"Unexpected return type {httpResult?.GetType()} for HTTP function {context.FunctionDefinition.Name}.");
```
fabiocav, line 93: This will introduce an exception in cases where it didn't happen before (e.g., returning a POCO).

A void function is also supported, and the response will be automatically handled in those cases. This exception would prevent that.

This is highlighting potential test coverage issues.

extensions/Worker.Extensions.Http.AspNetCore/src/FunctionsMiddleware/FunctionsHttpProxyingMiddleware.cs
```diff
@@ -49,26 +59,53 @@ public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next
...
-                // processing is required.
-                invocationResult.Value = null;
+                HandleHttpResultInOutputBindings(context, httpContext);
             }
 
             // allows asp.net middleware to continue
             _coordinator.CompleteFunctionInvocation(invocationId);
         }
 
+        private static async void HandleHttpResult(object? httpResult, FunctionContext context, HttpContext httpContext)
```
fabiocav, line 75: Please avoid `async void`.

For more information, please see https://learn.microsoft.com/en-us/archive/msdn-magazine/2013/march/async-await-best-practices-in-asynchronous-programming#avoid-async-void

extensions/Worker.Extensions.Http.AspNetCore/src/FunctionsMiddleware/FunctionsHttpProxyingMiddleware.cs
```diff
@@ -49,26 +59,53 @@ public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next
...
+        {
+            switch (httpResult)
+            {
+                case HttpResponseData httpResponseData:
+                    context.GetInvocationResult().Value = null;
```
fabiocav, line 80: This is overriding the invocation result value in the output binding case as well, which wouldn't be expected.

extensions/Worker.Extensions.Http.AspNetCore/src/FunctionsMiddleware/FunctionsHttpProxyingMiddleware.cs
```diff
@@ -49,26 +59,53 @@ public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next
...
+            }
+        }
+
+        private static void HandleHttpResultInOutputBindings(FunctionContext context, HttpContext httpContext)
+        {
+            var httpOutputBinding = context.GetOutputBindings<object>().FirstOrDefault(a => string.Equals(a.BindingType, HttpBindingType, StringComparison.OrdinalIgnoreCase));
+
+            if (httpOutputBinding is null)
+            {
+                throw new InvalidOperationException($"No HTTP Response type could be found in the function {context.FunctionDefinition.Name}'s output bindings.");
```
fabiocav, line 103: Similar issue to the one with the new exception above, this shouldn't fail (it's not strictly required)

sdk/Sdk.Generators/Sdk.Generators.csproj
```diff
@@ -10,7 +10,7 @@
     <IncludeBuildOutput>false</IncludeBuildOutput>
     <EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>
     <MinorProductVersion>2</MinorProductVersion>
```
fabiocav, line 12: We should update the minor version for this.

sdk/Sdk/Sdk.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <MinorProductVersion>17</MinorProductVersion>
-    <PatchProductVersion>2</PatchProductVersion>
```
fabiocav, line 5: Minor bump

test/Sdk.Generator.Tests/FunctionMetadataProviderGeneratorTests/NestedTypesTest.cs
```diff
@@ -20,10 +22,10 @@ public NestedTypesTests()
                 // load all extensions used in tests (match extensions tested on E2E app? Or include ALL extensions?)
                 var abstractionsExtension = Assembly.LoadFrom("Microsoft.Azure.Functions.Worker.Extensions.Abstractions.dll");
                 var httpExtension = Assembly.LoadFrom("Microsoft.Azure.Functions.Worker.Extensions.Http.dll");
-                var hostingExtension = Assembly.LoadFrom("Microsoft.Extensions.Hosting.dll");
-                var diExtension = Assembly.LoadFrom("Microsoft.Extensions.DependencyInjection.dll");
-                var hostingAbExtension = Assembly.LoadFrom("Microsoft.Extensions.Hosting.Abstractions.dll");
-                var diAbExtension = Assembly.LoadFrom("Microsoft.Extensions.DependencyInjection.Abstractions.dll");
+                var hostingExtension = typeof(HostBuilder).Assembly;
```
fabiocav, line 25: Did you run into issues with all the tests being updated?

test/DotNetWorkerTests/AspNetCore/FunctionsHttpProxyingMiddlewareTests.cs
```diff
@@ -54,7 +54,12 @@ private static (FunctionContext FunctionContext, HttpContext HttpContext, Mock<I
                 { "test", new TestBindingMetadata("test", triggerType, BindingDirection.In ) }
             };
 
-            var functionDef = new TestFunctionDefinition(inputBindings: inputBindings);
+            var outputBindings = new Dictionary<string, BindingMetadata>()
+            {
+                { "outputBinding", new TestBindingMetadata("$return", "http", BindingDirection.Out ) }
+            };
+
+            var functionDef = new TestFunctionDefinition(inputBindings: inputBindings, outputBindings: outputBindings);
```
fabiocav, line 62: nit: `functionDef` -> `functionDefinition`

test/DotNetWorkerTests/AspNetCore/FunctionsHttpProxyingMiddlewareTests.cs
```diff
@@ -54,7 +54,12 @@ private static (FunctionContext FunctionContext, HttpContext HttpContext, Mock<I
                 { "test", new TestBindingMetadata("test", triggerType, BindingDirection.In ) }
             };
 
-            var functionDef = new TestFunctionDefinition(inputBindings: inputBindings);
+            var outputBindings = new Dictionary<string, BindingMetadata>()
```
fabiocav, line 57: Can you clarify what you're trying to validate with this? 

Having an HTTP output binding/result wouldn't be an expected state for non-HTTP triggers.

test/DotNetWorkerTests/AspNetCore/FunctionsHttpProxyingMiddlewareTests.cs
```diff
@@ -54,7 +54,12 @@ private static (FunctionContext FunctionContext, HttpContext HttpContext, Mock<I
...
+            {
+                { "outputBinding", new TestBindingMetadata("$return", "http", BindingDirection.Out ) }
+            };
+
+            var functionDef = new TestFunctionDefinition(inputBindings: inputBindings, outputBindings: outputBindings);
 
             var functionContext = new TestFunctionContext(functionDef, new TestFunctionInvocation(), CancellationToken.None)
```
fabiocav, line 64: Not seeing tests for the new behavior added to the proxying middleware. Can we please have those added?

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -479,36 +480,48 @@ private bool TryGetReturnTypePropertyBindings(ITypeSymbol returnTypeSymbol, bool
...
-                                bindingsList.Add(bindingDict!);
+                            bindingsList.Add(bindingDict!);
 
-                                returnTypeHasOutputBindings = true;
-                                foundPropertyOutputAttr = true;
-                            }
+                            returnTypeHasOutputBindings = true;
                         }
-                    }
+                    }
```
fabiocav, line 524: empty spaces

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -288,7 +288,7 @@ private bool TryGetParameterInputAndTriggerBindings(IMethodSymbol method, out bo
 
                             string bindingName = parameter.Name;
 
-                            if (!TryCreateBindingDict(attribute, bindingName, Location.None, out IDictionary<string, object>? bindingDict, supportsDeferredBinding))
+                            if (!TryCreateBindingDictionary(attribute, bindingName, Location.None, out IDictionary<string, object>? bindingDict, supportsDeferredBinding))
```
fabiocav, line 291: can we please rename `bindingDict` to `bindings` or `bindingDictionary`?

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -164,7 +164,7 @@ private bool TryGetMethodOutputBinding(IMethodSymbol method, out bool hasMethodO
 
                 if (outputBindingAttribute != null)
                 {
-                    if (!TryCreateBindingDict(outputBindingAttribute, Constants.FunctionMetadataBindingProps.ReturnBindingName, Location.None, out IDictionary<string, object>? bindingDict))
+                    if (!TryCreateBindingDictionary(outputBindingAttribute, Constants.FunctionMetadataBindingProps.ReturnBindingName, Location.None, out IDictionary<string, object>? bindingDict))
```
fabiocav, line 167: can we please rename `bindingDict` to `bindings` or `bindingDictionary`?

fabiocav (changes requested):

extensions/Worker.Extensions.Http.AspNetCore/src/FunctionsMiddleware/FunctionsHttpProxyingMiddleware.cs
```diff
@@ -49,26 +59,53 @@ public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next
...
 
             // allows asp.net middleware to continue
             _coordinator.CompleteFunctionInvocation(invocationId);
         }
 
+        private static async void HandleHttpResult(object? httpResult, FunctionContext context, HttpContext httpContext)
+        {
+            switch (httpResult)
+            {
+                case HttpResponseData httpResponseData:
```
fabiocav, line 79: We cannot handle this result based on the abstract type. This is only safe when working with `AspNetCoreHttpRequestData` (which we automatically inject when the integration is enabled), as that implementation is backed by the ASP.NET Core types.

fabiocav (changes requested):

fabiocav:
I've also pushed some changes to the middleware to help with some of the issues. Done a bit quickly, so the tests mentioned and a review, would be good.

test/Sdk.Generator.Tests/FunctionMetadataProviderGeneratorTests/NestedTypesTest.cs
satvu: Yeah - adding the framework reference failed everything.

sdk/release_notes.md
```diff
@@ -4,7 +4,11 @@
 - My change description (#PR/#issue)
 -->
 
-### Microsoft.Azure.Functions.Worker.Sdk 1.17.2 (meta package)
+### Microsoft.Azure.Functions.Worker.Sdk 1.17.3
```
fabiocav, line 7: Need to update this to match the updated version.

sdk/release_notes.md
```diff
@@ -4,7 +4,11 @@
...
 
-- Revert changes from #1946
+- Updating to use `Microsoft.NET.Sdk.Functions.Generators` 1.2.2 (#2247)
+
+### Microsoft.Azure.Functions.Worker.Sdk.Generators 1.2.2
```
fabiocav, line 11: Same here.

sdk/release_notes.md
```diff
@@ -8,3 +8,11 @@
...
 - Re-add SDK refactors that were reverted in #2313 (#2347)
 - Address assembly scanning regression (#2347)
+- Updating to use `Microsoft.NET.Sdk.Functions.Generators` 1.2.2 (#2247)
+
+### Microsoft.Azure.Functions.Worker.Sdk.Generators 1.3.0
+
+- Introduces handling for `HttpResultAttribute`.
+
+
+
```
kshyju, line 18: nit: multiple empty lines

test/Sdk.Generator.Tests/FunctionMetadataProviderGeneratorTests/IntegratedTriggersAndBindingsTests.cs
```diff
@@ -177,6 +189,147 @@ public static IHostBuilder ConfigureGeneratedFunctionMetadataProvider(this IHost
...
+                    public class MyOutputType
+                    {
+                        [QueueOutput("functionstesting2", Connection = "AzureWebJobsStorage")]
+                        public string Name { get; set; }
+
+                        [HttpResult]
+                        public IActionResult HttpResponse { get; set; }
+                    }
+
+                    public class MyOutputType2
```
kshyju, line 233: Is it valid to write something like below? (even with legacy generator approach / in-proc model)

```csharp
public class MyOutputTypeToMultipleDestinations
{
    [QueueOutput("functionstesting2", Connection = "AzureWebJobsStorage")]
    [ServiceBusOutput("my-queue", Connection = "ServiceBusConnection")]
    public string Name { get; set; }


    [HttpResult]
    public IActionResult HttpResponse { get; set; }
}
```
Does this cause a diagnostic error today? Do we have a test for that?

extensions/Worker.Extensions.Http.AspNetCore/src/FunctionsMiddleware/FunctionsHttpProxyingMiddleware.cs
```diff
@@ -47,40 +51,67 @@ public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next
...
-            if (invocationResult?.Value is IActionResult actionResult)
+            var responseHandled = await TryHandleHttpResult(context.GetInvocationResult().Value, context, httpContext, true)
+                || await TryHandleOutputBindingsHttpResult(context, httpContext);
+            
+            if (!responseHandled)
             {
-                ActionContext actionContext = new ActionContext(httpContext, httpContext.GetRouteData(), new ActionDescriptor());
-
-                await actionResult.ExecuteResultAsync(actionContext);
+                var logger = context.InstanceServices.GetRequiredService<ExtensionTrace>();
```
kshyju, line 59: Do we need to query this service every time or should we consider caching it in the middleware class (which is singleton) level? 

```csharp
private ExtensionTrace? _extensionTracelogger = null;

{
    ....
    ....
    _extensionTracelogger = _extensionTracelogger ?? context.InstanceServices.GetRequiredService<ExtensionTrace>();
    _extensionTracelogger.NoHttpResponseReturned(context.FunctionDefinition.Name, context.InvocationId);
}

```

sdk/release_notes.md
```diff
@@ -8,3 +8,11 @@
 
 - Re-add SDK refactors that were reverted in #2313 (#2347)
 - Address assembly scanning regression (#2347)
+- Updating to use `Microsoft.NET.Sdk.Functions.Generators` 1.2.2 (#2247)
+
+### Microsoft.Azure.Functions.Worker.Sdk.Generators 1.3.0
+
+- Introduces handling for `HttpResultAttribute`.
```
kshyju, line 15: Could we improve this to be a bit more specific please?

kshyju (approved):

test/Sdk.Generator.Tests/FunctionMetadataProviderGeneratorTests/IntegratedTriggersAndBindingsTests.cs
satvu: That is not valid and results in a diagnostic error - [here](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/docs/analyzer-rules/AZFW0005.md) is the doc for it, [here](https://github.com/Azure/azure-functions-dotnet-worker/blob/add69528ff7127b1db4cb8df89a2a3fff9bb4788/test/Sdk.Generator.Tests/FunctionMetadataProviderGeneratorTests/DiagnosticResultTests.cs#L54) is a test for it, and [here](https://github.com/Azure/azure-functions-dotnet-worker/blob/add69528ff7127b1db4cb8df89a2a3fff9bb4788/sdk/Sdk/FunctionMetadataGenerator.cs#L332) is where this same check/error is thrown in the legacy generator.

extensions/Worker.Extensions.Http.AspNetCore/src/FunctionsMiddleware/FunctionsHttpProxyingMiddleware.cs
fabiocav: This is a good suggestion. We don't **need** to retrieve the service every time, however, this happens in a failure path that is not as likely to occur and wanted to avoid the additional boilerplate required for the marginal difference given the scenario and the fragility that having that uninitialized class level variable introduce here.

This is also part of the reason why this isn't resolved as a constructor parameter by default (i.e., in most cases, it will never be resolved and there would be an additional hit), which would be the preferred approach if we want to have a class level variable.

sdk/release_notes.md
fabiocav: Indeed. We want to have something that describes the feature here.

test/DotNetWorkerTests/AspNetCore/FunctionsHttpProxyingMiddlewareTests.cs
```diff
@@ -47,18 +51,125 @@ public async Task Middleware_NoOpsOnNonHttpTriggers()
...
+            // In a simple HTTP trigger function, there is only one input (the trigger), and the HTTP response
+            // The HTTP response will be in the InvocationResult
+            var mockActionResult = GetMockActionResult();
+            var bindingFeatures = test.FunctionContext.Features.Get<IFunctionBindingsFeature>();
+            bindingFeatures.InvocationResult = mockActionResult.Object;
+
+            var funcMiddleware = new FunctionsHttpProxyingMiddleware(test.MockCoordinator.Object);
+            await funcMiddleware.Invoke(test.FunctionContext, mockDelegate.Object);
+
+            mockActionResult.Verify();
```
fabiocav, line 69: Not blocking, but it would be good to track improvements to the new tests to improve code reuse/reduce duplication.

fabiocav (approved):

Issue #2323, opened 2024-03-05
[Docs] Add docs for new HTTP output binding


satvu:
Docs update needed for changes in #2322 prior/alongside the official release of those changes.

mattchenderson:
This will need:

- Update to HTTP example in multi-output reference
- Update to the HTTP trigger reference
- Update to the HTTP output reference

fabiocav:
Flagging a dependency on releasing the work done in #2322

fabiocav:
Done.

PR #2343, opened 2024-03-11
ASP.NET Sample Multiple Output Bindings

satvu:

Follow up to #2322 - the sample requires an SDK bump once the package with the fix is released.

Docs issue #2323 

### Pull request checklist


liliankasem (approved):

samples/AspNetIntegration/AspNetIntegration.csproj
```diff
@@ -8,7 +8,7 @@
   </PropertyGroup>
     
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.17.0" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.17.3-preview2" />
```
kshyju, line 11: Do we usually use the preview version packages in the samples? I thought we didn't.

satvu: No, but the bits needed are unfortunately only in the preview package.

samples/AspNetIntegration/MultipleOutputBindings.cs
```diff
@@ -0,0 +1,38 @@
...
+        [Function("MultipleOutputBindings")]
+        public MyOutputType Run([HttpTrigger(AuthorizationLevel.Function, "post")] HttpRequest req)
+        {
+            _logger.LogInformation("C# HTTP trigger function processed a request.");
+            var myObject = new MyOutputType
+            {
+                Result = new OkObjectResult("C# HTTP trigger function processed a request."),
+                MessageText = "some output"
+            };
+            return myObject;
```
kshyju, line 25: nit q: should we inline the `myObject` variable so that we can reduce 1 line? Thinking that because this code snippet goes in the docs and I think it is good to have less number of lines.

samples/AspNetIntegration/MultipleOutputBindings.cs
```diff
@@ -0,0 +1,38 @@
...
+    //<docsnippet_aspnetcore_multiple_outputs>
+    public class MultipleOutputBindings
+    {
+        private readonly ILogger<MultipleOutputBindings> _logger;
+        public MultipleOutputBindings(ILogger<MultipleOutputBindings> logger)
+        {
+            _logger = logger;
+        }
+        [Function("MultipleOutputBindings")]
+        public MyOutputType Run([HttpTrigger(AuthorizationLevel.Function, "post")] HttpRequest req)
```
kshyju, line 17: Should we include both `get` and `post` ?

kshyju (approved):

samples/AspNetIntegration/MultipleOutputBindings.cs
satvu: discussed offline - original abstraction model samples also only have `post` and this is the most used case

PR #2344, opened 2024-03-11
[Source-Gen][Bug Fix] Array input in SignalR generated correctly

### LINKED ISSUE #2100: Isolated functions with .NET 8 - SignalR extension throws warning CS8785
After upgrading our application to .NET 8, I started receiving the following error:

```
CSC : warning CS8785: Generator 'FunctionMetadataProviderGenerator' failed to generate source. It will not contribute to the output and compilation errors may occur as a result. Exception was of type 'InvalidOperationException' with message 'TypedConstant is an array. Use Values property.'.
```

**Steps to reproduce:**

- using Visual Studio 17.8.1
- create a new Function app targeting .NET 8 Isolated
- add the SignalR extension + a simple Negotiate method 

```
[Function("negotiate")]
public string Negotiate([HttpTrigger(AuthorizationLevel.Function)] HttpRequestData req, [SignalRConnectionInfoInput(HubName = "TestFunctions", UserId = "{headers.user}",
            IdToken = "{headers.token}",
            ClaimTypeList = new []{"user_id"})] string connectionInfo
    )
    {
        return connectionInfo;
    }
```
Minimal repo:
https://github.com/Msirkovsky/signalR-net8-issue-CS8785

satvu:

Will go in after #2322 which has test structure refactoring that this PR depends on, resolves #2100

### Pull request checklist


sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -672,6 +672,24 @@ private bool TryGetAttributeProperties(AttributeData attributeData, Location? at
                 return true;
             }
 
+            private bool WriteArgumentToMetadata(TypedConstant? namedArgument)
```
kshyju, line 675: The method name suggests it is "writing" something. but this method do some checks and return a boolean value. Should we rename this method to make it more readable?

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -623,9 +623,9 @@ private bool TryGetAttributeProperties(AttributeData attributeData, Location? at
 
                 foreach (var namedArgument in attributeData.NamedArguments)
                 {
-                    if (namedArgument.Value.Value != null)
+                    if (WriteArgumentToMetadata(namedArgument.Value))
```
kshyju, line 626: What part was exactly crashing and why? The exception thrown from roslyn suggests "Use Values property." but I don't see we use the `Values` property in this PR.

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
kshyju, line 675: method can be static?

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
satvu: The use of `Values` is [elsewhere](https://github.com/Azure/azure-functions-dotnet-worker/blob/d45568b5ce9a84e4df509ddc132cf2a6b05e97dc/sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs#L735) outside of the PR - the problem was that we were null-checking `namedArgument.Value.Value` in this spot, which is invalid in the case that `namedArgument.Value` is an array type and caused a crash.

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -623,9 +623,9 @@ private bool TryGetAttributeProperties(AttributeData attributeData, Location? at
 
                 foreach (var namedArgument in attributeData.NamedArguments)
                 {
-                    if (namedArgument.Value.Value != null)
+                    if (IsNamedArgumentValid(namedArgument.Value))
                     {
-                        if (string.Equals(namedArgument.Key, Constants.FunctionMetadataBindingProps.IsBatchedKey) && !attrProperties.ContainsKey("cardinality"))
+                        if (string.Equals(namedArgument.Key, Constants.FunctionMetadataBindingProps.IsBatchedKey) && !attrProperties.ContainsKey("cardinality") && namedArgument.Value.Value != null)
```
liliankasem, line 628: nit: getting a bit long, can we break this down?
```suggestion
                        if (string.Equals(namedArgument.Key, Constants.FunctionMetadataBindingProps.IsBatchedKey)
                        && !attrProperties.ContainsKey("cardinality") && namedArgument.Value.Value != null)
```

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -672,6 +672,24 @@ private bool TryGetAttributeProperties(AttributeData attributeData, Location? at
                 return true;
             }
 
+            private static bool IsNamedArgumentValid(TypedConstant? namedArgument)
```
liliankasem, line 675: Should we name this to be specific to the array check? What makes this named argument valid?

We should also move this function to the extension class maybe?

https://github.com/Azure/azure-functions-dotnet-worker/blob/main/sdk/Sdk.Analyzers/Extensions/TypeSymbolExtensions.cs

liliankasem (approved):

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
satvu: Named arguments are valid if they have a non-null value or if they are an array type (both checks are in this method) - could move this to an extension class but it's specific to this scenario.

"Valid" in this case means "we can/should write this to the function metadata". For example, with this [input binding](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/extensions/Worker.Extensions.CosmosDB/src/CosmosDBInputAttribute.cs) which has nullable properties, even if you add a prop explicitly as a named argument with a null value in the attribute, it won't be written to the metadata. Validated that this is legacy generator behavior and emulating it here.

PR #1946, opened 2023-10-04
Refactor SDK targets to improve incremental build, publish issues, and nuget feed issues 

jviau:
### Issue describing the changes in this PR

resolves #1863, #1834

### Pull request checklist

* [x] My changes **do not** require documentation changes
  * [ ] Otherwise: Documentation issue linked to PR
* [ ] My changes **should not** be added to the release notes for the next release
  * [x] Otherwise: I've added my notes to `release_notes.md`
* [x] My changes **do not** need to be backported to a previous version
  * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [x] I have added all required tests (Unit tests, E2E tests)

### Additional information

This PR refactors our SDK targets to improve the inner-build stability and quality. The inner build is used to collect functions host extension payload to be deployed with the application. The changes are:

1. Pass nuget restore sources directly to inner build
2. `WorkerExtensions.csproj` is now generated and build in the intermediate output directory under a set path
3. Incremental build support improved (due to point 2)
4. No longer copy items to output or publish directory manually, instead we populate the appropriate item groups so that `CopyToOutputDirectory` and `CopyToPublishDirectory` targets from the base .NET SDK will perform the copies for us.

PR #2347, opened 2024-03-13
Re-add SDK refactor with referenced assembly scanning fixed.

jviau:

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

This re-adds the SDK overhaul that was reverted in #2313. The cause for the reversion has also been addressed. The root issue was that we were identifying what referenced assemblies to scan by looking at all dll's next to the final output assembly of the functions project. With the SDK refactoring this was no longer possible as our targets ran earlier (before being copied to final output location). Instead, we directly use the `ReferencePath` item group from MSBuild to identify all assemblies that will be copied to the final output directory.

### DIFF
--- a/build/Sdk.slnf
+++ b/build/Sdk.slnf
@@ -1,0 +1,10 @@
+{
+  "solution": {
+    "path": "..\\DotNetWorker.sln",
+    "projects": [
+      "sdk\\Sdk.Analyzers\\Sdk.Analyzers.csproj",
+      "sdk\\Sdk.Generators\\Sdk.Generators.csproj",
+      "sdk\\Sdk\\Sdk.csproj"
+    ]
+  }
+}
--- a/sdk/Sdk/FunctionMetadataGenerator.cs
+++ b/sdk/Sdk/FunctionMetadataGenerator.cs
@@ -7,6 +7,7 @@ using System.Diagnostics;
 using System.Dynamic;
 using System.IO;
 using System.Linq;
+using Microsoft.Build.Framework;
 using Mono.Cecil;
 using Mono.Collections.Generic;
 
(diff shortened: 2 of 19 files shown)


fabiocav:
@jviau it might be helpful to share a link for a diff against a branch with the state we had in 1.17.0 so we can focus on the delta. Is that something you can provide?

sdk/Sdk/FunctionMetadataGenerator.cs
jviau, line 1: This file contains the fix.

**Before**: The task would look at the final output directory (ie: `/bin/{config}/{tfm}`) and scan DLLs there. This does not work anymore because our targets run earlier in the build and no DLL's have been copied there yet.

**After**: The task already had the list of referenced assemblies, which basically _is_ the set of DLLs which will be copied to the final location (but in their original locations). The change now is to just use that list of assemblies. I added a minor optimization to skip any assembly coming from a framework reference. These are .NETcore and AspNetCore assemblies that are not ever copied to final output (as the runtime will have its own copy). We don't need to scan those because a framework would never include functions concepts inside it.

jviau:
> @jviau it might be helpful to share a link for a diff against a branch with the state we had in 1.17.0 so we can focus on the delta. Is that something you can provide?

https://github.com/Azure/azure-functions-dotnet-worker/compare/sdk-1.17.0..jviau/sdk

Here is a diff. It isn't very helpful though as it also includes every unrelated commit between the two as well. Do you know a way to get a cleaner diff?

In the meantime, I can guide the review a bit here:

1. I left a comment explaining the core change [here](https://github.com/Azure/azure-functions-dotnet-worker/pull/2347#discussion_r1538298107)
2. All changes to `/test` and `/samples` folder is to support a new e2e test I added to verify this change. The e2e test changes lets us avoid writing to the test csproj's to update the SDK version. Instead we pass in the version to use as part of the build. And we always build/pack with a version of `99.99.99` during e2e test, so we know what version to pass in.

build/Sdk.slnf
```diff
@@ -0,0 +1,10 @@
+{
```
kshyju, line 1: This change was not in the [original SDK refactor PR](https://github.com/Azure/azure-functions-dotnet-worker/pull/1946/files). Do we need this?

sdk/Sdk/FunctionMetadataGenerator.cs
```diff
@@ -42,44 +43,33 @@ public FunctionMetadataGenerator(Action<TraceLevel, string, string> log)
...
+            var targetAssemblies = new List<string>() { assemblyPath };
 
-            var targetAssemblies = new List<string>(Directory.GetFiles(sourcePath, "*.dll"));
+            // We don't need to scan assemblies that come from a framework (.NET, AspNetCore), they won't have functions types in them.
+            targetAssemblies.AddRange(referencePaths.Where(x => x.GetMetadata("FrameworkReferenceName") == "").Select(x => x.ItemSpec));
+            _logger.LogMessage($"Found {targetAssemblies.Count} assemblies to evaluate.");
 
-            if (!assemblyPath.EndsWith(".dll"))
+            var resolver = new DefaultAssemblyResolver();
+            foreach (string referencePath in referencePaths.Select(p => Path.GetDirectoryName(p.ItemSpec)).Distinct())
```
kshyju, line 55: `referencePaths` is being enumerated more than once. Should we `ToArray` once and use that result in both places?

sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
```diff
@@ -8,437 +8,215 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
```
kshyju, line 8: Just to confirm, Changes in this file is same as what was in 1946?

jviau: Correct, I did not change that file

build/Sdk.slnf
jviau: This is part of the refactor to our test steps. 

it is used in `TestUtility.cs`
``` CSharp
public static readonly string SdkBuildProj = Path.Combine(PathToRepoRoot, "build", "Sdk.slnf");
```

sdk/Sdk/FunctionMetadataGenerator.cs
jviau: Are you worried about the parameter `IEnumerable<ITaskItem> referencePaths` being enumerated twice? Or is one of the two projections I have being enumerated twice?

For the former, a caller may still supply a concrete list. In which case `ToArray` will be overhead we don't want. At build it will be a concrete list, so our only concern would be during a test. Which I don't want to change this API for.

For the later, the 2 projections of this parameter are unique and enumerated only once each. A concrete list will not benefit.

kshyju (approved):
Let's get the preview version out for E2E testing.

fabiocav (approved):

Issue #2180, opened 2024-01-06
[ASP.NET Core Integration] Implement timeouts for HTTP coordinator related tasks


satvu:
### Context
In order to coordinate the ASP.NET pipeline and the gRPC/Functions pipeline while also handling cancellations, this [callback](https://github.com/Azure/azure-functions-dotnet-worker/blob/16a995deb65b44873cd3a56b5e6d4d3c22d64b7d/extensions/Worker.Extensions.Http.AspNetCore/src/Coordinator/ContextReference.cs#L38) was introduced that cancels coordination task when a function invocation is cancelled (token on `FunctionContext` is cancelled).

As noted in [this discussion comment](https://github.com/Azure/azure-functions-dotnet-worker/issues/1929#issuecomment-1743890718) we want to move away from reacting to the cancellation token, but we still need to make sure in the case of a disconnect, the pipeline is not stalled by these coordination tasks. 

### Solution

This issue will track the design and implementation of timeouts in the coordination tasks to pave the way for removing this callback.

_Note: https://github.com/Azure/azure-functions-host/pull/9523 will have implications on design and understanding of e2e cancellation scenarios here_.

satvu:
Resolved by #2356

PR #2356, opened 2024-03-19
Improvements to context coordination handling and observability in ASP.NET Core integration.

fabiocav:
Improvements to context coordination/synchronization handling and observability:

- Failure to receive any of the expected context synchronization calls will now result in a TimeoutException thrown with the appropriate exception information. Previously this would block indefinitely and failures here were difficult to diagnose.
- Debug logs are now emitted in the context coordination calls, improving observability.

No issue tracking. This is in response to internally reported incidents.

### Pull request checklist

### DIFF
--- a/extensions/Worker.Extensions.Http.AspNetCore/src/Coordinator/DefaultHttpCoordinator.cs
+++ b/extensions/Worker.Extensions.Http.AspNetCore/src/Coordinator/DefaultHttpCoordinator.cs
@@ -5,16 +5,23 @@ using System;
 using System.Collections.Concurrent;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
+using Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore.Infrastructure;
 
 namespace Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore
 {
     internal class DefaultHttpCoordinator : IHttpCoordinator
     {
-        private readonly ConcurrentDictionary<string, ContextReference> _contextReferenceList;
+        private const int HttpContextTimeoutInSeconds = 5;
+        private const int FunctionContextTimeoutInSeconds = 5;
+        private const int FunctionStartTimeoutInSeconds = 15;
 
-        public DefaultHttpCoordinator()
+        private readonly ConcurrentDictionary<string, ContextReference> _contextReferenceList;
+        private readonly ExtensionTrace _logger;
+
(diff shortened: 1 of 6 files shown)


extensions/Worker.Extensions.Http.AspNetCore/src/Worker.Extensions.Http.AspNetCore.csproj
```diff
@@ -6,11 +6,12 @@
     <Description>ASP.NET Core extensions for .NET isolated functions</Description>
 
     <!--Version information-->
-    <VersionPrefix>1.2.1</VersionPrefix>
+    <VersionPrefix>1.3.0</VersionPrefix>
     <TargetFramework>net6.0</TargetFramework>
   </PropertyGroup>
 
   <Import Project="..\..\..\build\Extensions.props" />
+
```
fabiocav, line 14: Intentional line.

extensions/Worker.Extensions.Http.AspNetCore/src/Coordinator/DefaultHttpCoordinator.cs
```diff
@@ -5,24 +5,40 @@
...
 using Microsoft.AspNetCore.Http;
+using Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore.Infrastructure;
 
 namespace Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore
 {
     internal class DefaultHttpCoordinator : IHttpCoordinator
     {
+        private const int HttpContextTimeoutSeconds = 5;
+        private const int FunctionContextTimeoutSeconds = 5;
+        private const int FunctionRunTimeoutSeconds = 15;
```
satvu, line 16: Would it make more sense to name this `FunctionStartTimeoutSeconds` or something similar? My first though reading this was a timeout for how long the function runs, but it's actually used with `FunctionStartTask`.

satvu (approved):
I think we can also attach this issue to the PR: https://github.com/Azure/azure-functions-dotnet-worker/issues/2180.

extensions/Worker.Extensions.Http.AspNetCore/src/Coordinator/DefaultHttpCoordinator.cs
fabiocav: Yeah, I like the name. The method exposed by the coordinator is called `RunFunctionInvocationAsync`, so I ended up deriving from that, but what you're suggesting would make this clearer.

PR #2404, opened 2024-04-17
ServiceBusSessionMessageActions management APIs: worker changes

### LINKED ISSUE #2374: [ServiceBusSessionMessageActions] Implement session management APIs
### Description

Add support for `ServiceBusSessionMessageActions` with and initial support with the following APIs:

- [ ] GetSessionStateAsync
- [ ] SetSessionStateAsync
- [ ] ReleaseSessionAsync
- [ ] RenewSessionLockAsync
- [ ] SessionLockedUntil (property)

aishwaryabh:

resolves #2374 

This PR adds support `ServiceBusSessionMessageActions`. The corresponding changes in the WebJobs extension have already been deployed in the latest package. This PR consumes those changes by updating the proto and creating the appropriate converters.

### Pull request checklist

### Additional information

Additional PR information


extensions/Worker.Extensions.ServiceBus/src/Proto/settlement.proto
```diff
@@ -19,6 +20,18 @@ service Settlement {
 
   // Defers a message
   rpc Defer (DeferRequest) returns (google.protobuf.Empty) {}
+
+  // Get
+  rpc Get (GetRequest) returns (GetResponse) {}
```
jviau, line 25: This naming is ambiguous as to what it is getting, can we expand upon it?

```suggestion
  rpc GetSession (GetSessionRequest) returns (GetSessionResponse) {}
```

Recommend similar naming changes for the other APIs added in this PR

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActions.cs
```diff
@@ -0,0 +1,95 @@
...
+        private readonly Settlement.SettlementClient _settlement;
+
+        private readonly string _sessionId;
+
+        internal ServiceBusSessionMessageActions(Settlement.SettlementClient settlement, string sessionId, DateTimeOffset sessionLockedUntil)
+        {
+            _settlement = settlement;
+            _sessionId = sessionId ?? throw new ArgumentNullException(nameof(sessionId));
+            SessionLockedUntil = sessionLockedUntil;
+
```
jviau, line 23: nit: remove empty line
```suggestion
```

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActions.cs
```diff
@@ -0,0 +1,95 @@
...
+    [InputConverter(typeof(ServiceBusSessionMessageActionsConverter))]
+    public class ServiceBusSessionMessageActions
+    {
+        private readonly Settlement.SettlementClient _settlement;
+
+        private readonly string _sessionId;
+
+        internal ServiceBusSessionMessageActions(Settlement.SettlementClient settlement, string sessionId, DateTimeOffset sessionLockedUntil)
+        {
+            _settlement = settlement;
```
jviau, line 20: Verify this is not null as well?

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
```diff
@@ -0,0 +1,49 @@
...
+        public ServiceBusSessionMessageActionsConverter(Settlement.SettlementClient settlement)
+        {
+            _settlement = settlement;
+        }
+
+        public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
+        {
+            try
+            {
+                context.FunctionContext.BindingContext.BindingData.TryGetValue("SessionId", out object? sessionId);
```
jviau, line 32: Consider adding `if(!...) { throw some error }` around these `TryGetValue` calls.

The error should be descriptive that this expected a binding data (provide the name) and it was not present.

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
```diff
@@ -0,0 +1,49 @@
...
+
+                // Get the sessionLockedUntil property from the SessionActions binding data
+                context.FunctionContext.BindingContext.BindingData.TryGetValue("SessionActions", out object? sessionActions);
+                JsonDocument jsonDocument = JsonDocument.Parse(sessionActions.ToString());
+                jsonDocument.RootElement.TryGetProperty("SessionLockedUntil", out JsonElement sessionLockedUntil);
+
+                var sessionActionResult = new ServiceBusSessionMessageActions(_settlement, sessionId?.ToString(), sessionLockedUntil.GetDateTimeOffset());
```
jviau, line 39: If we believe `sessionId` can be `null` here, and we know it must be non-null (as `ServiceBusSessionMessageActions` asserts that), we should perform our own verification and throw with a descriptive message.

I prefer not to rely on `ArgumentException` for reporting bad behavior in production - they are more a development time guide. This is at least for cases where we believe the arg exception may be encountered.

jviau (changes requested):

aishwaryabh:
Going to work on this soon

extensions/Worker.Extensions.ServiceBus/src/Proto/settlement.proto
aishwaryabh: done!

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
aishwaryabh: done!

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
aishwaryabh: added the checks!

H-Finch-404:
Any ETA on when this will be ready & released? Nice to see this being finally worked on!

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActions.cs
```diff
@@ -0,0 +1,94 @@
...
+            CancellationToken cancellationToken = default)
+        {
+            var request = new RenewSessionLockRequest()
+            {
+                SessionId = _sessionId,
+            };
+
+            var result = await _settlement.RenewSessionLockAsync(request, cancellationToken: cancellationToken);
+            SessionLockedUntil = result.LockedUntil.ToDateTimeOffset();
+
```
jviau, line 91: nit: remove empty line

samples/Extensions/Extensions.csproj
```diff
@@ -23,6 +22,10 @@
     <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.17.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
   </ItemGroup>
+   <ItemGroup>
+     <!-- Sample uses project reference, but actual apps should use 'PackageReference' -->
+     <ProjectReference Include="../../extensions/Worker.Extensions.ServiceBus/src/Worker.Extensions.ServiceBus.csproj" />
```
jviau, line 27: Ah I see this project has a bunch of different samples in it.

@fabiocav , @liliankasem  - what do you two think of using p2p refs for this project? Package references makes it difficult to add new in-development features to the samples. Or is that intentional and this should only demonstrate already release features?

liliankasem: I believe this was intentional to only demo released features; FunctionApp is the sample project that uses project references for testing. I typically only update samples for extensions once they've released and in their own PR.

(I'm not opposed to switching this, just sharing what I'm aware of)

jviau: Given that explanation, @aishwaryabh do you have any issue saving the sample change till this is released?

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActions.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ServiceBusMessageActions"/> class for mocking use in testing.
+        /// </summary>
+        /// <remarks>
+        /// This constructor exists only to support mocking. When used, class state is not fully initialized, and
+        /// will not function correctly; virtual members are meant to be mocked.
+        ///</remarks>
+        protected ServiceBusSessionMessageActions()
+        {
+            _settlement = null!; // not expected to be used during mocking.
```
fabiocav, line 34: This will generate a warning with nullable checks enabled. 

Either make the field nullable or suppress for this case (preferable)

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActions.cs
```diff
@@ -0,0 +1,94 @@
...
+            CancellationToken cancellationToken = default)
+        {
+            var request = new GetSessionStateRequest()
+            {
+                SessionId = _sessionId,
+            };
+
+            GetSessionStateResponse result = await _settlement.GetSessionStateAsync(request, cancellationToken: cancellationToken);
+            byte[] byteArray = result.SessionState.ToByteArray();
+            BinaryData binaryData = BinaryData.FromBytes(byteArray);
```
fabiocav, line 50: We should be able to avoid copies/allocations here by creating the `BinaryData` instance over the `ReadonlyMemory<byte>` exposed by `ByteString`:

For example:
``` csharp
BinaryData binaryData = new BinaryData(result.SessionState.Memory);
```

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActions.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        ///<inheritdoc cref="ServiceBusReceiver.CompleteMessageAsync(ServiceBusReceivedMessage, CancellationToken)"/>
+        public virtual async Task SetSessionStateAsync(
+            BinaryData sessionState,
+            CancellationToken cancellationToken = default)
+        {
+            var request = new SetSessionStateRequest()
+            {
+                SessionId = _sessionId,
+                SessionState = ByteString.CopyFrom(sessionState.ToArray()),
```
fabiocav, line 62: Should be able to make similar optimizations here using the `BinaryData` bytes (`ToMemory`) and passing the Span.

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActions.cs
```diff
@@ -0,0 +1,94 @@
...
+using Microsoft.Azure.Functions.Worker.Converters;
+using Microsoft.Azure.ServiceBus.Grpc;
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    [InputConverter(typeof(ServiceBusSessionMessageActionsConverter))]
+    public class ServiceBusSessionMessageActions
+    {
+        private readonly Settlement.SettlementClient _settlement;
+
```
fabiocav, line 15: nit: remove blank line

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActions.cs
```diff
@@ -0,0 +1,94 @@
...
+using System.Threading.Tasks;
+using Azure.Messaging.ServiceBus;
+using Google.Protobuf;
+using Microsoft.Azure.Functions.Worker.Converters;
+using Microsoft.Azure.ServiceBus.Grpc;
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    [InputConverter(typeof(ServiceBusSessionMessageActionsConverter))]
+    public class ServiceBusSessionMessageActions
```
fabiocav, line 12: Missing public documentation for the type.

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActions.cs
```diff
@@ -0,0 +1,94 @@
+﻿using System;
```
fabiocav, line 1: Missing file header.

extensions/Worker.Extensions.ServiceBus/src/ServiceBusMessageActions.cs
```diff
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
```
fabiocav, line 1: This file is also missing a file header, let's take the opportunity to fix (will track making sure this is checked)

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Threading.Tasks;
+using Microsoft.Azure.Functions.Worker.Converters;
+using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
+using Microsoft.Azure.ServiceBus.Grpc;
+using System.Text.Json;
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    /// <summary>
+    /// Converter to bind to <see cref="ServiceBusSessionMessageActions" /> or <see cref="ServiceBusSessionMessageActions[]" /> type parameters.
```
fabiocav, line 14: Fix documentation syntax:
 `"ServiceBusSessionMessageActions[]"` -> `"ServiceBusSessionMessageActions{}"`

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
+        {
+            try
+            {
+                var foundSessionId = context.FunctionContext.BindingContext.BindingData.TryGetValue("SessionId", out object? sessionId);
+                if (!foundSessionId)
+                {
+                    throw new InvalidOperationException("Expecting SessionId within binding data and value was not present.");
```
fabiocav, line 35: Is the exception message here sufficient to help users understand what is wrong/how to fix the problem?

I believe the primary root cause for this would be attempting to bind to `ServiceBusSessionMessageActions`/`ServiceBusSessionMessageActions[]` when sessions are disabled, correct?

If that is the case, can we:

1. Improve the exception message to share that information. That would put the user facing the exception in a better position to fix the problem
2. Consider an analyzer. Wondering if we should consider an analyzer for this, so if a user is binding to a type that requires session, we detect whether the trigger is setting `IsSessionEnabled` to true. *this wouldn't have to be part of this PR, but let's open an issue*. Avoiding this at compilation time would be great.

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
```diff
@@ -0,0 +1,62 @@
...
+                }
+
+                // Get the sessionLockedUntil property from the SessionActions binding data
+                var foundSessionActions = context.FunctionContext.BindingContext.BindingData.TryGetValue("SessionActions", out object? sessionActions);
+                if (!foundSessionActions)
+                {
+                    throw new InvalidOperationException("Expecting SessionActions within binding data and value was not present.");
```
fabiocav, line 42: For the other error conditions here, is there any additional information we can provide? Fine if not (as this would still be used for troubleshooting), but if we have the ability to share more information about the causes, let's add that.

samples/Extensions/Extensions.csproj
fabiocav: @liliankasem is correct, but I think this is indeed causing unnecessary churn and may also lead to stale samples. No need to block, but let's discuss changes do this.

aishwaryabh: Sure no worries I can remove these samples for now then and create another PR once the worker packages have been released

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
aishwaryabh: I'm not sure if we can add any more info to these error messages. I think the primary root cause for these errors would be is isSessionEnabled is not set for the queue, but the user would hit the error in one of the earlier error messages before.

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActions.cs
aishwaryabh: Updated! Let me know if this looks ok now

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActions.cs
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ServiceBusMessageActions"/> class for mocking use in testing.
+        /// </summary>
+        /// <remarks>
+        /// This constructor exists only to support mocking. When used, class state is not fully initialized, and
+        /// will not function correctly; virtual members are meant to be mocked.
+        ///</remarks>
+        #pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
```
jviau, line 37: You can alternatively suppress this by adding:

``` csharp
_sessionId = null!;
```

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActions.cs
```diff
@@ -0,0 +1,99 @@
...
+        protected ServiceBusSessionMessageActions()
+        #pragma warning restore CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
+        {
+            _settlement = null!; // not expected to be used during mocking.
+        }
+
+        public virtual DateTimeOffset SessionLockedUntil { get; private set; }
```
jviau, line 44: If this is for mocking, does it need `protected set;`?

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActions.cs
```diff
@@ -0,0 +1,99 @@
...
+            CancellationToken cancellationToken = default)
+        {
+            var request = new GetSessionStateRequest()
+            {
+                SessionId = _sessionId,
+            };
+
+            GetSessionStateResponse result = await _settlement.GetSessionStateAsync(request, cancellationToken: cancellationToken);
+            BinaryData binaryData = new BinaryData(result.SessionState.Memory);
+            return await Task.FromResult(binaryData);
```
jviau, line 57:
```suggestion
            return binaryData;
```

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActions.cs
aishwaryabh: added it!

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
fabiocav: For the message, can we add something like: 

``` csharp
$"Sessions must be enabled when binding to {nameof(ServiceBusSessionMessageActions)}."
```

fabiocav (approved):
Approving with a small comment about the exception message.

jviau (approved):

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
aishwaryabh: Added it!

PR #2414, opened 2024-04-18
Check if _functionCompletionTask is already complete

### LINKED ISSUE #2413: 'functionCompletionTask' is being set when already complete in Extensions.Http.AspNetCore
### Description

There appears to be a bug in `Extensions.Http.AspNetCore` during function completion. When a Http invocation gets cancelled, an `InvalidOperationException` is thrown instead of expected `TaskCanceledException`.

This occurs because during cancellation we cancel the CancellationTokenSource in the invocation handler:

https://github.com/Azure/azure-functions-dotnet-worker/blob/ca7afd23f6a7f68059e00bc642c13ef9efd21dee/src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs#L147

This completes the `_functionCompletionTask`:

https://github.com/Azure/azure-functions-dotnet-worker/blob/ca7afd23f6a7f68059e00bc642c13ef9efd21dee/extensions/Worker.Extensions.Http.AspNetCore/src/Coordinator/ContextReference.cs#L35-L45

When `CompleteFunction` is called the `_functionCompletionTask` task is already complete, but we try to set the result here anyway:

https://github.com/Azure/azure-functions-dotnet-worker/blob/ca7afd23f6a7f68059e00bc642c13ef9efd21dee/extensions/Worker.Extensions.Http.AspNetCore/src/Coordinator/ContextReference.cs#L53-L60

Which leads to the `InvalidOperationException` exception.

`Exception: System.InvalidOperationException: An attempt was made to transition a task to a final state when it had already completed.`

### Repro

Repro can be found in this GH issue:

- https://github.com/Azure/azure-functions-dotnet-worker/issues/2376

### Fix

First we can implement a check at the start of `CompleteFunction` to see if the task is complete before trying to set a result i.e.

```csharp
if (_functionCompletionTask.Task.IsCompleted)
{
    return;
}
```

This still leads to a different exception `FunctionInvocationException` with a `TaskCanceledException` inner exception.

```logs
Executed 'Functions.BlobStream' (Failed, Id=c4cb3e5e-a039-4a44-a4d2-172ac6be0668, Duration=25692ms)
      Microsoft.Azure.WebJobs.Host.FunctionInvocationException: Exception while executing function: Functions.BlobStream
       ---> System.InvalidOperationException: Failed to proxy request with ForwarderError: ResponseBodyCanceled
       ---> System.Threading.Tasks.TaskCanceledException: The operation was canceled.
         at System.Net.Http.CancellationHelper.ThrowOperationCanceledException(Exception innerException, CancellationToken cancellationToken)
         at System.Net.Http.HttpConnection.ContentLengthReadStream.ReadAsync(Memory`1 buffer, CancellationToken cancellationToken)
         at Yarp.ReverseProxy.Forwarder.StreamCopier.CopyAsync(Stream input, Stream output, Int64 promisedContentLength, StreamCopierTelemetry telemetry, ActivityCancellationTokenSource activityToken, Boolean autoFlush, CancellationToken cancellation)
```

If there are other enhancements to make here, they should be considered.

liliankasem:

resolves #2413

### Pull request checklist

### DIFF
--- a/extensions/Worker.Extensions.Http.AspNetCore/src/Coordinator/ContextReference.cs
+++ b/extensions/Worker.Extensions.Http.AspNetCore/src/Coordinator/ContextReference.cs
@@ -52,6 +52,11 @@ namespace Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore
 
         internal void CompleteFunction()
         {
+            if (_functionCompletionTask.Task.IsCompleted)
+            {
+                return;
+            }
+
             if (_httpContextValueSource.Task.IsCompleted)
             {
                 if (_httpContextValueSource.Task.IsCanceled || _token.IsCancellationRequested)


satvu (approved):

jviau (approved):
Just want to confirm that thread safety is not needed for this method?

fabiocav:
> > Just want to confirm that thread safety is not needed for this method?
> 
> I don't believe so; each call to SetHttpContextAsync and SetFunctionContextAsync operates on its own instance of `ContextReference`

Though they're operating on the same reference instance, the operations for the cancellation handling and function completion could potentially lead to a race here. 

We could lock here, but switching to `TrySetCanceled` and `TrySetResult` here might be a better option.

liliankasem:
> > > Just want to confirm that thread safety is not needed for this method?
> > 
> > 
> > I don't believe so; each call to SetHttpContextAsync and SetFunctionContextAsync operates on its own instance of `ContextReference`
> 
> Though they're operating on the same reference instance, the operations for the cancellation handling and function completion could happen on different threads and potentially lead to a race here.
> 
> We could lock here, but switching to `TrySetCanceled` and `TrySetResult` here might be a better option.

Ah yup, that makes sense:

https://github.com/Azure/azure-functions-dotnet-worker/pull/2431

PR #2448, opened 2024-05-08
Fix SignalR extensions DI issue

Y-Sindo:
`AddHostedService` with implementation type as `IHostedService` is forbidden.

resolves https://github.com/MicrosoftDocs/azure-docs/issues/121723

### Pull request checklist

### DIFF
--- a/extensions/Worker.Extensions.SignalRService/src/SignalRServiceDependencyInjectionExtensions.cs
+++ b/extensions/Worker.Extensions.SignalRService/src/SignalRServiceDependencyInjectionExtensions.cs
@@ -34,18 +34,17 @@ namespace Microsoft.Extensions.DependencyInjection
             services.AddAzureClientsCore();
             services.TryAddSingleton<HubContextProvider>();
             services.TryAddSingleton<ServiceManagerOptionsSetup>();
-            if (TryGetClientType(hubType, out var clientType))
+            var genericType = TryGetClientType(hubType, out var clientType) ?
+                typeof(ServiceHubContextInitializer<,>).MakeGenericType(hubType, clientType) :
+                typeof(ServiceHubContextInitializer<>).MakeGenericType(hubType);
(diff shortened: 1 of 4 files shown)


vicancy (approved):

extensions/Worker.Extensions.SignalRService/src/SignalRServiceDependencyInjectionExtensions.cs
```diff
@@ -44,7 +44,7 @@ public static IServiceCollection AddServerlessHub(this IServiceCollection servic
             {
                 var genericDefinition = typeof(ServiceHubContextInitializer<>);
                 var genericType = genericDefinition.MakeGenericType(hubType);
-                return services.AddHostedService(sp => (IHostedService)ActivatorUtilities.CreateInstance(sp, genericType, configure));
+                return services.AddSingleton(sp => (IHostedService)ActivatorUtilities.CreateInstance(sp, genericType, configure));
```
jviau, line 47: I am not sure this will work - or at least be stable. I think there will be issues if this method is ever called twice. Service registration calls should be idempotent.

I think you need to either track and no-op if `hubType` is already registered, OR, redesign this so you can call `AddHostedService` (and know the concrete type of the hosted service, so it is registered with that impl type).

jviau (changes requested):

extensions/Worker.Extensions.SignalRService/src/SignalRServiceDependencyInjectionExtensions.cs
Y-Sindo: Update:
* Tracked if `hubType` is registered by trying to find a singleton `genericType`. 
* Added some tests to verify that it works.

`AddHostedService` is a convenient method of `TryAddEnumereable`, and it finally adds a singleton service to the service collection.

PR #2455, opened 2024-05-09
[WIP] Implement CosmosDBOptions

### LINKED ISSUE #2450: [Cosmos Extension] Implement missing configuration options
### Description

Expose CosmosClientOptions to allow users to configure the cosmos client.

liliankasem:

resolves #2450

### Pull request checklist

### DIFF
--- a/extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBBindingOptions.cs
+++ b/extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBBindingOptions.cs
@@ -9,6 +9,12 @@ using Microsoft.Azure.Functions.Worker.Extensions.CosmosDB;
 
 namespace Microsoft.Azure.Functions.Worker
 {
+    /// <summary>
+    /// Internal options for configuring the CosmosDB binding.
+    /// This class is used internally by the Azure Functions runtime to manage the CosmosDB connection and clients.
+    /// It is not intended to be used directly in user code.
+    /// Any public configuration options should be set on the <see cref="CosmosDBOptions"/> class, which is publicly accessible.
+    /// </summary>
     internal class CosmosDBBindingOptions
     {
         public string? ConnectionName  { get; set; }
@@ -20,6 +26,10 @@ namespace Microsoft.Azure.Functions.Worker
         public TokenCredential? Credential { get; set; }
 
         public CosmosSerializer? Serializer { get; set; }
+
+        public ConnectionMode? ConnectionMode { get; set; }
+
+        public string? UserAgentSuffix { get; set; }
 
         internal string BuildCacheKey(string connection, string region) => $"{connection}|{region}";
 
@@ -34,20 +44,7 @@ namespace Microsoft.Azure.Functions.Worker
 
             string cacheKey = BuildCacheKey(ConnectionName!, preferredLocations);
 
-            CosmosClientOptions cosmosClientOptions = new ()
-            {
-                ConnectionMode = ConnectionMode.Gateway
-            };
-
-            if (!string.IsNullOrEmpty(preferredLocations))
-            {
-                cosmosClientOptions.ApplicationPreferredRegions = Utilities.ParsePreferredLocations(preferredLocations);
-            }
-
-            if (Serializer is not null)
-            {
-                cosmosClientOptions.Serializer = Serializer;
-            }
+            var cosmosClientOptions = Utilities.BuildClientOptions(ConnectionMode, Serializer, preferredLocations, UserAgentSuffix);
 
             return ClientCache.GetOrAdd(cacheKey, (c) => CreateService(cosmosClientOptions));
         }
--- a/extensions/Worker.Extensions.CosmosDB/release_notes.md
+++ b/extensions/Worker.Extensions.CosmosDB/release_notes.md
@@ -4,8 +4,7 @@
 - My change description (#PR/#issue)
 -->
 
-### Microsoft.Azure.Functions.Worker.Extensions.CosmosDB 4.8.1
+### Microsoft.Azure.Functions.Worker.Extensions.CosmosDB 4.9.0
 
-- Updating `Microsoft.Azure.WebJobs.Extensions.CosmosDB` reference to 4.6.1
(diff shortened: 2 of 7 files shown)


extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBBindingOptions.cs
```diff
@@ -21,6 +21,10 @@ internal class CosmosDBBindingOptions
```
liliankasem, line 21: I'm setting the options like this:

```csharp
builder.Services.Configure<CosmosDBOptions>((cosmosOptions) =>
{
    cosmosOptions.ConnectionMode = ConnectionMode.Direct;
    cosmosOptions.UserAgentSuffix = "MyApp";
});
```

Seeing as null in logs:

> [2024-05-09T20:28:49.395Z] CosmosDBOptions
[2024-05-09T20:28:49.395Z] {
[2024-05-09T20:28:49.395Z]   "ConnectionMode": null
[2024-05-09T20:28:49.395Z] }

But the ConnectionMode is being set properly when I log it as part of the function code:

`_logger.LogInformation("Cosmos client options: {0} - {1}", client.ClientOptions.ConnectionMode, client.ClientOptions.ApplicationName);`

> Cosmos client options: Direct - MyApp

Any thoughts on why it's null when logged during startup?

jviau (commented):
Looks good. I recommend maybe adding some xmldoc to the `CosmosDBBindingOptions` to explain how it is internal, and any public options should be mapped onto `CosmosDBOptions`.

My primary question for this work though is: have we thought about exposing `CosmosClientOptions` directly? What would the implications of that be?

jviau (approved):

extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBBindingOptions.cs
liliankasem: I'm not logging it myself, it's just part of the host start up logs? I'm not actually sure what does this i.e.

```log
[2024-05-09T20:28:49.392Z] FunctionResultAggregatorOptions
[2024-05-09T20:28:49.393Z] {
[2024-05-09T20:28:49.393Z]   "BatchSize": 1000,
[2024-05-09T20:28:49.393Z]   "FlushTimeout": "00:00:30",
[2024-05-09T20:28:49.393Z]   "IsEnabled": true
[2024-05-09T20:28:49.393Z] }
[2024-05-09T20:28:49.395Z] SingletonOptions
[2024-05-09T20:28:49.395Z] {
[2024-05-09T20:28:49.395Z]   "LockPeriod": "00:00:15",
[2024-05-09T20:28:49.395Z]   "ListenerLockPeriod": "00:00:15",
[2024-05-09T20:28:49.395Z]   "LockAcquisitionTimeout": "10675199.02:48:05.4775807",
[2024-05-09T20:28:49.395Z]   "LockAcquisitionPollingInterval": "00:00:05",
[2024-05-09T20:28:49.395Z]   "ListenerLockRecoveryPollingInterval": "00:01:00"
[2024-05-09T20:28:49.395Z] }
[2024-05-09T20:28:49.395Z] CosmosDBOptions
[2024-05-09T20:28:49.395Z] {
[2024-05-09T20:28:49.395Z]   "ConnectionMode": null
[2024-05-09T20:28:49.395Z] }
[2024-05-09T20:28:49.395Z] ScaleOptions
[2024-05-09T20:28:49.395Z] {
[2024-05-09T20:28:49.396Z]   "ScaleMetricsMaxAge": "00:02:00",
[2024-05-09T20:28:49.396Z]   "ScaleMetricsSampleInterval": "00:00:10",
[2024-05-09T20:28:49.396Z]   "MetricsPurgeEnabled": true,
[2024-05-09T20:28:49.396Z]   "IsTargetScalingEnabled": true,
[2024-05-09T20:28:49.396Z]   "IsRuntimeScalingEnabled": false
[2024-05-09T20:28:49.396Z] }
[2024-05-09T20:28:49.398Z] Starting JobHost
```

liliankasem:
> Looks good. I recommend maybe adding some xmldoc to the `CosmosDBBindingOptions` to explain how it is internal, and any public options should be mapped onto `CosmosDBOptions`.
> 
> My primary question for this work though is: have we thought about exposing `CosmosClientOptions` directly? What would the implications of that be?

That's a good question: I think originally we were not exposing CosmosClientOptions directly because we were following the WebJobs way to keep things consistent - though I don't know how much that matters now.

I do wonder if it exposes us to more issues given we do some customization like `ParsePreferredLocations()` and using the WorkerSerializer as the CosmosSerializer. We should be able to customize the options setup so these concerns can be addressed. We also get preferred locations from the [attribute](https://github.com/Azure/azure-functions-dotnet-worker/blob/785be42daec346159269b1faaa4a61e402f97c94/extensions/Worker.Extensions.CosmosDB/src/CosmosDBInputAttribute.cs#L97) so that adds another layer to things. 

I think overall, it's doable - I can be convinced either way

jviau:
I may be useful to experiment with exposing `CosmosClientOptions` directly. Particularly seeing how it works with the customizations we apply.

- What customizations can we apply directly to `CosmosClientOptions` during options config?
- What customizations do we need to apply at binding time?
    - This would most likely result in us creating a copy of the client from options, applying only binding-time customizations to the copy.

extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBBindingOptions.cs
fabiocav: The host automatically logs the options that implement `IOptionsFormatter` at initialization to show how those options were setup *for the extension*. So anything set on the worker will not have any impact on the host/extension side.

fabiocav (changes requested):
Can we discuss the design a bit? 

Sent a message on this yesterday, but I would like to avoid exposing our own options and consider exposing `CosmosClientOptions` first before we take this route. If there are no concerns with that, it would be a better approach, but if we expose our own model here, it would be harder (confusing) to expose `CosmosClientOptions` later.

PR #2483, opened 2024-05-23
 Implement `CosmosDBOptions` to allow configuration of the CosmosDB service client via `CosmosClientOptions`

### LINKED ISSUE #2450: [Cosmos Extension] Implement missing configuration options

liliankasem:

resolves #2450

### Pull request checklist

  * [ ] Otherwise: Documentation issue linked to PR - TBD: need to check if document the default being Gateway
* [x] I have added all required tests (Unit tests, E2E tests) - TODO

### Additional information

Implement CosmosDBExtensionOptions for the cosmos extension that will expose `CosmosClientOptions` and allow users to configure the CosmosClient.

### DIFF
--- a/extensions/Worker.Extensions.CosmosDB/release_notes.md
+++ b/extensions/Worker.Extensions.CosmosDB/release_notes.md
@@ -4,8 +4,27 @@
 - My change description (#PR/#issue)
 -->
 
-### Microsoft.Azure.Functions.Worker.Extensions.CosmosDB 4.8.1
+### Microsoft.Azure.Functions.Worker.Extensions.CosmosDB 5.0.0
 
-- Updating `Microsoft.Azure.WebJobs.Extensions.CosmosDB` reference to 4.6.1
-- Updating `Microsoft.Extensions.Azure` reference to 1.7.3
-- Updating `Microsoft.Azure.Cosmos` reference to 3.39.1
+- Expose `CosmosClientOptions` to allow configuration of the CosmosDB client (#2483)
+
+#### Breaking Change
+
+- **Default `ConnectionMode` Change:** The default `ConnectionMode` for `CosmosClientOptions` has been changed from `Gateway` to `Direct`.
+  This change is due to `Direct` being the default value for `ConnectionMode` in the Cosmos SDK; now that we are exposing those options,
+  it is not possible for us to tell whether the user has explicitly set the connection mode to `Direct` or not, so we must default to the
+  SDK's default value
+
+##### How to Maintain Previous Behavior
+
+To continue using `Gateway` mode, configure the `CosmosClientOptions` in your `Program` class as shown below:
+
+```csharp
+.ConfigureFunctionsWorkerDefaults((builder) =>
+{
+    builder.Services.Configure<CosmosClientOptions>((options) =>
+    {
+        options.ConnectionMode = ConnectionMode.Gateway;
+    });
+})
+```
--- a/extensions/Worker.Extensions.CosmosDB/release_notes.md
+++ b/extensions/Worker.Extensions.CosmosDB/release_notes.md
@@ -6,7 +6,7 @@
 
 ### Microsoft.Azure.Functions.Worker.Extensions.CosmosDB 5.0.0
 
-- Implement `CosmosDBOptions` to allow configuration of the CosmosDB service client via `CosmosClientOptions` (#2483)
+- Implement `CosmosDBExtensionOptions` to allow configuration of the CosmosDB service client via `CosmosClientOptions` (#2483)
 
 #### Breaking Change
 
--- a/extensions/Worker.Extensions.CosmosDB/release_notes.md
+++ b/extensions/Worker.Extensions.CosmosDB/release_notes.md
@@ -6,7 +6,7 @@
 
 ### Microsoft.Azure.Functions.Worker.Extensions.CosmosDB 5.0.0
 
-- Expose `CosmosClientOptions` to allow configuration of the CosmosDB client (#2483)
+- Implement `CosmosDBOptions` to allow configuration of the CosmosDB service client via `CosmosClientOptions` (#2483)
 
 #### Breaking Change
 
@@ -22,9 +22,9 @@ To continue using `Gateway` mode, configure the `CosmosClientOptions` in your `P
 ```csharp
 .ConfigureFunctionsWorkerDefaults((builder) =>
 {
-    builder.Services.Configure<CosmosClientOptions>((options) =>
+    builder.ConfigureCosmosDBExtensionOptions((options) =>
     {
-        options.ConnectionMode = ConnectionMode.Gateway;
+        options.ClientOptions.ConnectionMode = ConnectionMode.Gateway;
     });
 })
 ```
--- a/extensions/Worker.Extensions.CosmosDB/release_notes.md
+++ b/extensions/Worker.Extensions.CosmosDB/release_notes.md
@@ -4,27 +4,19 @@
 - My change description (#PR/#issue)
 -->
 
-### Microsoft.Azure.Functions.Worker.Extensions.CosmosDB 5.0.0
+### Microsoft.Azure.Functions.Worker.Extensions.CosmosDB 4.9.0
 
 - Implement `CosmosDBExtensionOptions` to allow configuration of the CosmosDB service client via `CosmosClientOptions` (#2483)
 
-#### Breaking Change
-
-- **Default `ConnectionMode` Change:** The default `ConnectionMode` for `CosmosClientOptions` has been changed from `Gateway` to `Direct`.
-  This change is due to `Direct` being the default value for `ConnectionMode` in the Cosmos SDK; now that we are exposing those options,
-  it is not possible for us to tell whether the user has explicitly set the connection mode to `Direct` or not, so we must default to the
-  SDK's default value
-
-##### How to Maintain Previous Behavior
-
-To continue using `Gateway` mode, configure the `CosmosClientOptions` in your `Program` class as shown below:
+#### Example Usage
 
 ```csharp
 .ConfigureFunctionsWorkerDefaults((builder) =>
 {
     builder.ConfigureCosmosDBExtensionOptions((options) =>
     {
-        options.ClientOptions.ConnectionMode = ConnectionMode.Gateway;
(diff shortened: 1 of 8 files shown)


extensions/Worker.Extensions.CosmosDB/release_notes.md
```diff
@@ -4,8 +4,27 @@
...
+  SDK's default value
+
+##### How to Maintain Previous Behavior
+
+To continue using `Gateway` mode, configure the `CosmosClientOptions` in your `Program` class as shown below:
+
+```csharp
+.ConfigureFunctionsWorkerDefaults((builder) =>
+{
+    builder.Services.Configure<CosmosClientOptions>((options) =>
```
jviau, line 25: I want customers to be able to configure `CosmosClientOptions` directly yes, but I also want it to be clear this configuration is specific for the functions worker cosmos extension. Configuring the options directly is a bit ambiguous what it is for. And what if some other unrelated library also configures cosmos that way?

We should come up with some design where the configuration is exceedingly clear that this is for only the functions worker cosmos extension

liliankasem: That is a very valid point, let me think on this a little more.

extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBOptions.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.Cosmos;
+
+namespace Microsoft.Azure.Functions.Worker
+{
+    public class CosmosDBOptions
```
jviau, line 8: Can we name this to make it clear it is for the extension? I like `CosmosDBExtensionOptions`

extensions/Worker.Extensions.CosmosDB/src/Worker.Extensions.CosmosDB.csproj
```diff
@@ -6,7 +6,7 @@
     <Description>Azure Cosmos DB extensions for .NET isolated functions</Description>
 
     <!--Version information-->
-    <VersionPrefix>4.8.1</VersionPrefix>
+    <VersionPrefix>5.0.0</VersionPrefix>
```
jviau, line 9: Does this need a major version rev?

jviau: Can we set connection mode to `Gateway` in `CosmosDBExtensionOptions`?

``` CSharp
CosmosClientOptions ClientOptions { get; set; } = new CosmosClientOptions() { ConnectionMode = ConnectionMode.Gateway };
```

liliankasem: Yup, now with the new change of moving to CosmosDBExtensionOptions this is no longer a breaking change since we can now set a default

extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBBindingOptions.cs
```diff
@@ -32,31 +40,33 @@ internal virtual CosmosClient GetClient(string preferredLocations = "")
...
-                cosmosClientOptions.ApplicationPreferredRegions = Utilities.ParsePreferredLocations(preferredLocations);
+                CosmosExtensionOptions.ClientOptions.ApplicationPreferredRegions = Utilities.ParsePreferredLocations(preferredLocations);
             }
 
-            if (Serializer is not null)
+            // Do not override if the serializer is configured via CosmosClientOptions
+            if (Serializer is not null && CosmosExtensionOptions.ClientOptions.Serializer is null)
             {
-                cosmosClientOptions.Serializer = Serializer;
+                CosmosExtensionOptions.ClientOptions.Serializer = Serializer;
```
jviau, line 59: Can we configured this in the options pipeline instead?

extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBBindingOptions.cs
```diff
@@ -32,31 +40,33 @@ internal virtual CosmosClient GetClient(string preferredLocations = "")
...
+            }
+
+            string cacheKey = BuildCacheKey(ConnectionName!, preferredLocations);
 
-            if (!string.IsNullOrEmpty(preferredLocations))
+            // Do not override if preferred locations is configured via CosmosClientOptions
+            if (!string.IsNullOrEmpty(preferredLocations) && CosmosExtensionOptions.ClientOptions.ApplicationPreferredRegions is null)
             {
-                cosmosClientOptions.ApplicationPreferredRegions = Utilities.ParsePreferredLocations(preferredLocations);
+                CosmosExtensionOptions.ClientOptions.ApplicationPreferredRegions = Utilities.ParsePreferredLocations(preferredLocations);
```
jviau, line 53: Because preferred locations are supplied per function, we can't mutate the global options here. That will impact other bindings with potentially different preferred locations.

We will need to apply preferred locations without impacting other functions.

- Can we make a copy of the options object?
- Can we use named options? Using the function name as the key? Then we will have a copy of options per function. This is a bit bigger of a scope creep for this PR though.
- Some other way to avoid this mutation?

extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBBindingOptions.cs
liliankasem: Done!

extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBBindingOptions.cs
liliankasem: Great callout here, I went with the copy approach.

jviau (changes requested):

extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBBindingOptions.cs
```diff
@@ -50,23 +48,20 @@ internal virtual CosmosClient GetClient(string preferredLocations = "")
...
             {
-                CosmosExtensionOptions.ClientOptions.ApplicationPreferredRegions = Utilities.ParsePreferredLocations(preferredLocations);
-            }
-
-            // Do not override if the serializer is configured via CosmosClientOptions
-            if (Serializer is not null && CosmosExtensionOptions.ClientOptions.Serializer is null)
-            {
-                CosmosExtensionOptions.ClientOptions.Serializer = Serializer;
+                // Copy to avoid modifying the original options
+                var cosmosClientOptions = CosmosExtensionOptions.ClientOptions;
```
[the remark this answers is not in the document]
liliankasem: resolved! pls review when you have a sec

extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBExtensionOptions.cs
```diff
@@ -11,5 +11,10 @@ public class CosmosDBExtensionOptions
         /// Gets or sets the CosmosClientOptions.
         /// </summary>
         public CosmosClientOptions ClientOptions { get; set; } = new() { ConnectionMode = ConnectionMode.Gateway };
+
+        internal CosmosDBExtensionOptions Clone()
+        {
+            return (CosmosDBExtensionOptions)this.MemberwiseClone();
```
jviau, line 17: Memberwise clone is a shallow copy. We will still have the same instance of `ClientOptions`. I am only concerned about how we apply `ApplicationPreferredRegions` to the client options, so no need for a clone here. Instead clone `ClientOptions` directly before applying preferred regions.

extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBBindingOptions.cs
```diff
@@ -48,8 +48,8 @@ internal virtual CosmosClient GetClient(string preferredLocations = "")
             // Do not override if preferred locations is configured via CosmosClientOptions
             if (!string.IsNullOrEmpty(preferredLocations) && CosmosExtensionOptions.ClientOptions.ApplicationPreferredRegions is null)
             {
-                // Copy to avoid modifying the original options
-                var cosmosClientOptions = CosmosExtensionOptions.ClientOptions;
+                // Clone to avoid modifying the original options
+                var cosmosClientOptions = CosmosExtensionOptions.Clone().ClientOptions;
```
jviau, line 52: We don't need a clone of extension options, but rather client options:

``` CSharp
CosmosClientOptions cosmosClientOptions = (CosmosClientOptions) CosmosExtensionOptions.ClientOptions.MeberwiseClone();
cosmosClientOptions.ApplicationPreferredRegions = Utilities.ParsePreferredLocations(preferredLocations);
return ClientCache.GetOrAdd(cacheKey, (c) => CreateService(cosmosClientOptions));
```

jviau (changes requested):

extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBExtensionOptions.cs
liliankasem: Oh I understand what you're saying now, I went down a different rabbit hole 🕳️

extensions/Worker.Extensions.CosmosDB/src/Config/CosmosDBBindingOptions.cs
jviau: Ah okay. We will need to either: create a manual copy of `CosmosClientOptions`, or use reflection to invoke `MemberwiseClone`. Calling clone on the extension options will not work, as you will be getting the same instance of client options. This means we are still mutating the original instance when setting preferred locations, which is what we need to address.

liliankasem: @jviau  I went with reflection via `CloningExtensions.cs` - let me know what you think

jviau (approved):

PR #2504, opened 2024-06-03
Adding POCO converter for Timer extension

### LINKED ISSUE #2411: Implement converter for TimerInfo that does not use customer provided serializers
### Description

Today, if a customer provides their own serializer and set a value such as  `UnmappedMemberHandling = JsonUnmappedMemberHandling.Disallow`, it breaks our converter. To fix this, for known POCOs that we provide, we should not use custom serializers.

### Task

- [ ] Create a converter for TimerInfo that does not use customer provided serializers.

kshyju:
Fixes #2411 

Adding a new converter in timer extension for POCO serialization. 

### Pull request checklist


extensions/Worker.Extensions.Shared/Serialization/JsonSerializerProvider.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    // Note: If you need an instance with different settings, create a new instance and use that
+    // instead of modifying this as it may break other places where this has been used.
+
+    /// <summary>
+    /// Option with the below settings:
+    ///  1. Deserialization supports case-insensitive properties.
+    ///  2. Serialization produces camelCase properties.
+    /// </summary>
+    internal static readonly JsonSerializerOptions Options = new JsonSerializerOptions
```
jviau, line 15: The heart of this issue is that the timer converter wants to control its own serialization behavior. As such, I don't think it is worth having this in a shared area. I would prefer the `TimerPocoConverter` had this directly as a field.

extensions/Worker.Extensions.Timer/src/TimerExtensionStartup.cs
```diff
@@ -0,0 +1,33 @@
...
+        public override void Configure(IFunctionsWorkerApplicationBuilder applicationBuilder)
+        {
+            if (applicationBuilder == null)
+            {
+                throw new ArgumentNullException(nameof(applicationBuilder));
+            }
+
+            applicationBuilder.Services.Configure<WorkerOptions>((workerOption) =>
+            {
+                workerOption.InputConverters.RegisterAt<TimerPocoConverter>(0);
```
jviau, line 29: Can we instead specify the converter on the `TimerTriggerAttribute`? We can also have this be a nested class of that attribute so it is easy to identify that this converter applies to only that type.

``` CSHarp
[InputConverter(typeof(Converter))]
[ConverterFallbackBehavior(ConverterFallbackBehavior.Disallow)]
public sealed class TimerTriggerAttribute : InputBindingAttribute
{
    internal class Converter : IInputConverter
    {
    }
}
```

Nesting the class is just a suggestion - if you want to leave it separate that is fine. But we should use the attribute - avoids adding global converters.

kshyju: I considered that. But that would be a breaking change since `TimerTriggerAttribute` is a public type and if any customer has written code to check the attributes on this type and did anything based on that, it will break.

jviau: I wouldn't worry about that as a breaking change. Attributes lay in a grey area for breaking changes - the attribute has to be documented as a feature we expect the customer to interact with directly. This attribute is not advertised/documented like that. Any customer usage of it will fall under unsupported reflection, and not be considered a breaking change.

extensions/Worker.Extensions.Shared/Serialization/JsonSerializerProvider.cs
kshyju: Moved to timer extension project.

extensions/Worker.Extensions.Timer/src/TimerExtensionStartup.cs
kshyju: Thoughts on this @fabiocav @brettsam ?

extensions/Worker.Extensions.Timer/src/Converters/TimerPocoConverter.cs
```diff
@@ -0,0 +1,61 @@
...
+using System;
+using System.IO;
+using System.Text;
+using System.Text.Json;
+using System.Threading.Tasks;
+using Microsoft.Azure.Functions.Worker.Converters;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Timer.Converters
+{
+    internal sealed class TimerPocoConverter : IInputConverter
```
liliankasem, line 13: If the only type we convert to is TimerInfo, is this better named as `TimerInfoConverter`? What other POCOS will the timer trigger convert?

extensions/Worker.Extensions.Timer/src/TimerExtensionStartup.cs
kshyju: Discussed during standup. Team agrees that this can be considered a non-breaking change. I will go ahead and make the changes.

kshyju: Pushed an iteration reflecting this change.

extensions/Worker.Extensions.Timer/src/Converters/TimerPocoConverter.cs
kshyju: Renamed to `TimerInfoConverter`

jviau (approved):

PR #2546, opened 2024-06-20
Handle cancellation token if function is invoked before cancellation token

### LINKED ISSUE #2415: Unexpected TaskCancelledException caused by pre-cancelled CancellationToken
### Problem Statement

When we send a cancelled invocation to the worker (i.e. invoke a function and the cancellation
token is cancelled) the CancellationToken being used through the pipeline in the dotnet-isolated
worker (via FunctionContext) is already cancelled **before** we get to the customer function. This leads
to a `TaskCancelledException` being thrown and a 500 error. The cause of this error is this:

`await _semaphoreSlim.WaitAsync(WaitTimeInMilliSeconds, context.CancellationToken);`

[code](https://github.com/Azure/azure-functions-dotnet-worker/blob/f694329065d20acbe3eacf48506f56e029f4cc87/src/DotNetWorker.Core/Context/Features/DefaultFunctionInputBindingFeature.cs#L32)

This is a problem because the customer function is not able to handle the cancellation token
as it never gets invoked.

### Investigation

- [ ] What does ASP.NET do?
- [ ] Attempt to repro with cancelling cts right before function invocation to understand if this is a race or not

### Design

We need to discuss the right behaviour for dotnet-isolated functions in this scenario.

- (a) We can leave the current behaviour as is and allow the pipeline to use the
  cancellation token to handle cancellation. Here, the customer function would not get invoked.
  We will still need to discuss how, if at all, we handle the `TaskCancelledException` exception.
- (b) We can change the behaviour so that items in the pipeline are not using a cancellation
  token and instead we use `CancellationToken.None`. This would mean that we continue and invoke
  the customer function where they are then able to handle the cancellation token.
- (c) Are there any other ways we could handle this scenario?

aishwaryabh:

resolves #2415 

We had a design discussion about this issue and came to the conclusion that we will not use the cancellation token in the pipeline and instead invoke `CancellationToken.None`. That way the customer can still have their function be invoked and handle the cancellation token as they want within their code.

To test this out I did the following from guidance from Lilian: 

1. use this sample for the worker app: [sample](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/samples/FunctionApp/HttpTriggerWithCancellation/HttpTriggerWithCancellation.cs)

2. Updated the tsc [here ](https://github.com/Azure/azure-functions-host/blob/d1067c54b27a8bb491867d692c9f7154f50552b8/src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs#L202-L210) to cancel the token after 5 seconds.

4. Invoked the HttpTriggerWithCancellation function, set a breakpoint after the token was cancelled, resumed function execution, and saw that a `TaskCancelledException` was invoked when the pipeline's cancellation token was passed in. With my changes, I saw that I was able to modify the function to handle the cancellation token as I wish.

### Pull request checklist

### Additional information

Additional PR information


src/DotNetWorker.Core/Context/Features/DefaultFunctionInputBindingFeature.cs
```diff
@@ -29,7 +29,10 @@ public async ValueTask<FunctionInputBindingResult> BindFunctionInputAsync(Functi
         {
             ObjectDisposedThrowHelper.ThrowIf(_disposed, this);
 
-            await _semaphoreSlim.WaitAsync(WaitTimeInMilliSeconds, context.CancellationToken);
+            // Setting second parameter to CancellationToken.None to prevent a TaskCancelledException if the 
+            // pipeline cancels the code before the function is invoked. This way the customer code will be invoked
+            // and they can handle the cancellation token.
+            await _semaphoreSlim.WaitAsync(WaitTimeInMilliSeconds, CancellationToken.None);
```
kshyju, line 35: Is this the only place we need to handle it? Anywhere else in the invocation pipeline the FunctioContext.CancellationToken is being used?

jviau:
The build errors appear unrelated to your change, but I am unsure how they appeared in the first place.

src/DotNetWorker.Core/Context/Features/DefaultFunctionInputBindingFeature.cs
aishwaryabh: Yeah I believe so. The only other place this is handled is [here](https://github.com/Azure/azure-functions-dotnet-worker/blob/e8e4b161e608a53a7cad40c27447e2a395ac7344/src/DotNetWorker.Core/Converters/CancellationTokenConverter.cs#L15) in the `CancellationTokenConverter`, but we'd want to pass it in during this instance

kshyju:
Looks like none of the tests in `DefaultModelBindingFeatureTests` is using a valid cancellation token flow. Should we add/update test(s) to reflect that?

kshyju:
> The build errors appear unrelated to your change, but I am unsure how they appeared in the first place.

We had some wildcard pattern used for nuget packages and one of the package had a [patch version bump a week ago](https://www.nuget.org/packages/Microsoft.CodeAnalysis.CSharp.Analyzer.Testing.XUnit/1.1.2), which brought a new version of `NuGet.Common` package as a transitive dependency. Our tests files were pinned to a lower version of `NuGet.Common` package and that caused the error.

@aishwaryabh  Please rebase after #2549 is merged.

aishwaryabh:
Added a unit test to handle the cancellation token

test/DotNetWorkerTests/DefaultModelBindingFeatureTests.cs
```diff
@@ -61,6 +61,47 @@ public async void BindFunctionInputAsync_Populates_ParametersUsingConverters()
             Assert.Equal("0ab4800e-1308-4e9f-be5f-4372717e68eb", guid.ToString());
         }
 
+        [Fact]
+        public async void BindFunctionInputAsync_IgnoreCancellationToken()
```
kshyju, line 65: We could also consider ensuring the Cancellation token parameter is populated as expected.

```csharp
// Arrange
var parameters = new List<FunctionParameter>()
{
    new("myQueueItem",typeof(Book)),
    new ("myGuid", typeof(Guid)),
    new ("token", typeof(CancellationToken)),
};
...
// Assert
var cancellationToken = TestUtility.AssertIsTypeAndConvert<CancellationToken>(parameterValuesArray[2]);
Assert.True(cancellationToken.IsCancellationRequested);
```

kshyju (approved):

test/DotNetWorkerTests/DefaultModelBindingFeatureTests.cs
aishwaryabh: Updated the unit test!

PR #2615, opened 2024-07-18
Adding net9.0 TFM

### LINKED ISSUE #2556: Update Core worker package to target .NET 9

kshyju:
fixes #2556

- Adding `net9.0` TFM to DotnetWorker, DotnetWorker.Core and DotnetWorker.Grpc.
- Bumped the SDK package version ([SDK changes to support net9.0 was checked-in](https://github.com/Azure/azure-functions-dotnet-worker/commit/3cc50f3ca19080161c5884287643074bb76e1f4b), but version was not bumped. Version bump is needed to test the local app)
- Added a new Net9 function app to sample. (For local testing, update tools/devpack.ps to update this project, as needed)
- Updating the below package references to use 9x version of them for `net9.0` tfm.
   - `Microsoft.Extensions.Hosting`
   - `Microsoft.Extensions.Hosting.Abstractions`
   - `System.Collections.Immutable`
   - `System.Diagnostics.DiagnosticSource`

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,7 +1,8 @@
 {
   "sdk": {
-    "version": "8.0.100",
-    "rollForward": "latestFeature"
+    "version": "9.0.100-preview.6.24328.19",
+    "allowPrerelease": true,
+    "rollForward": "latestPatch"
   },
   "msbuild-sdks": {
     "Microsoft.Build.NoTargets": "3.7.0",
--- a/samples/Net9FunctionApp/HelloHttp.cs
+++ b/samples/Net9FunctionApp/HelloHttp.cs
@@ -1,0 +1,17 @@
+﻿using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Extensions.Logging;
+
+namespace Net9FunctionApp
+{
+    public class HelloHttp(ILogger<HelloHttp> logger)
+    {
+        [Function("HelloHttp")]
+        public IActionResult Run([HttpTrigger(AuthorizationLevel.Anonymous, "get", "post")] HttpRequest req)
+        {
+            logger.LogInformation("C# HTTP trigger function processed a request.");
+            return new OkObjectResult("Welcome to Azure Functions!");
+        }
+    }
+}
--- a/samples/Net9FunctionApp/Net9FunctionApp.csproj
+++ b/samples/Net9FunctionApp/Net9FunctionApp.csproj
@@ -1,0 +1,29 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net9.0</TargetFramework>
+    <AzureFunctionsVersion>v4</AzureFunctionsVersion>
+    <OutputType>Exe</OutputType>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+  <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http" Version="3.2.0" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore" Version="1.3.2" />
+    <PackageReference Include="Microsoft.ApplicationInsights.WorkerService" Version="2.22.0" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.ApplicationInsights" Version="1.2.0" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.17.0" />
+  </ItemGroup>
+  <ItemGroup>
+    <None Update="host.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Update="local.settings.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <CopyToPublishDirectory>Never</CopyToPublishDirectory>
+    </None>
+  </ItemGroup>
+  <ItemGroup>
+    <Using Include="System.Threading.ExecutionContext" Alias="ExecutionContext" />
+  </ItemGroup>
+</Project>
--- a/samples/Net9FunctionApp/Net9FunctionApp.csproj
+++ b/samples/Net9FunctionApp/Net9FunctionApp.csproj
@@ -11,7 +11,7 @@
     <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http" Version="3.2.0" />
     <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore" Version="1.3.2" />
     <PackageReference Include="Microsoft.ApplicationInsights.WorkerService" Version="2.22.0" />
-    <PackageReference Include="Microsoft.Azure.Functions.Worker.ApplicationInsights" Version="1.2.0" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.ApplicationInsights" Version="1.3.0" />
     <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.17.0" />
   </ItemGroup>
   <ItemGroup>
--- a/sdk/Sdk/Sdk.csproj
+++ b/sdk/Sdk/Sdk.csproj
@@ -1,8 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <MinorProductVersion>17</MinorProductVersion>
-    <PatchProductVersion>4</PatchProductVersion>
+    <MinorProductVersion>18</MinorProductVersion>
+    <PatchProductVersion>0</PatchProductVersion>
+    <VersionSuffix>-preview1</VersionSuffix>
     <TargetFrameworks>netstandard2.0;net472</TargetFrameworks>
     <PackageId>Microsoft.Azure.Functions.Worker.Sdk</PackageId>
     <Description>This package provides development time support for the Azure Functions .NET Worker.</Description>
--- a/src/DotNetWorker.Core/DotNetWorker.Core.csproj
+++ b/src/DotNetWorker.Core/DotNetWorker.Core.csproj
@@ -24,24 +24,19 @@
     <PackageReference Include="Azure.Core" Version="1.37.0" />
   </ItemGroup>
 
-  <Choose>
-    <When Condition="'$(TargetFramework)' == 'net9.0'">
-      <ItemGroup>
-        <PackageReference Include="Microsoft.Extensions.Hosting" Version="9.0.0-preview.*" />
-        <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="9.0.0-preview.*" />
-        <PackageReference Include="System.Collections.Immutable" Version="9.0.0-preview.*" />
-        <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="9.0.0-preview.*" />
-      </ItemGroup>
-    </When>
-    <Otherwise>
-      <ItemGroup>
-        <PackageReference Include="Microsoft.Extensions.Hosting" Version="5.0.0" />
-        <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="5.0.0" />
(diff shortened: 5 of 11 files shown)


global.json
```diff
@@ -1,7 +1,8 @@
 {
   "sdk": {
-    "version": "8.0.100",
-    "rollForward": "latestFeature"
+    "version": "9.0.100-preview.6.24328.19",
```
kshyju, line 3: The version value does not support wild cards. `rollForward` should help pick the latest.

src/DotNetWorker.Core/DotNetWorker.Core.csproj
```diff
@@ -22,12 +22,27 @@
...
-    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="7.0.0" />
   </ItemGroup>
 
+  <Choose>
+    <When Condition="'$(TargetFramework)' == 'net9.0'">
+      <ItemGroup>
+        <PackageReference Include="Microsoft.Extensions.Hosting" Version="9.0.0-preview.*" />
+        <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="9.0.0-preview.*" />
+        <PackageReference Include="System.Collections.Immutable" Version="9.0.0-preview.*" />
+        <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="9.0.0-preview.*" />
```
mattchenderson, line 33: I think this being `preview.*` makes sense, but we don't typically do wildcards here, right? Just tactics for preview?

src/DotNetWorker.Core/DotNetWorker.Core.csproj
```diff
@@ -22,12 +22,27 @@
...
+      </ItemGroup>
+    </When>
+    <Otherwise>
```
mattchenderson, line 36: Should we make this explicit for each TFM? If doing so, I believe we can also just apply the condition to the item group directly. There's actually an example of that right below the current change.

sdk/Sdk/Sdk.csproj
```diff
@@ -1,8 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <MinorProductVersion>17</MinorProductVersion>
-    <PatchProductVersion>4</PatchProductVersion>
+    <MinorProductVersion>18</MinorProductVersion>
+    <PatchProductVersion>0</PatchProductVersion>
+    <VersionSuffix>-preview1</VersionSuffix>
```
mattchenderson, line 6: Is an SDK update needed with this set of changes, yet? No TFM or dependency changes, right?

mattchenderson: Nevermind - I saw the note in the PR description. Could you please provide a link to that changeset, though?

jviau:
What is the release plan for our packages after this is checked in? Will this be merged to `main`? If so, do we plan on keeping the `preview` tag until .net9 GAs?

src/DotNetWorker.Core/DotNetWorker.Core.csproj
kshyju: Correct. We usually ping to a specific version. We can change to that later.

sdk/Sdk/Sdk.csproj
kshyju: Added in PR description.

kshyju:
> What is the release plan for our packages after this is checked in? Will this be merged to `main`? If so, do we plan on keeping the `preview` tag until .net9 GAs?

Yes. This will go to main.
We have not waited until the GA release in November for previous .NET versions. Once the preview package is released and we don't encounter any major issues, we can remove the preview label on our package.

jviau:
> we can remove the preview label on our package.

We should discuss if this is something we should do. This would mean our non-preview package would be depending on preview packages (and an entirely preview TFM). This is discouraged with nuget (there are warnings for that).

src/DotNetWorker.Core/DotNetWorker.Core.csproj
```diff
@@ -22,12 +22,27 @@
...
     <PackageReference Include="Azure.Core" Version="1.37.0" />
-    <PackageReference Include="Microsoft.Extensions.Hosting" Version="5.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="5.0.0" />
-    <PackageReference Include="System.Collections.Immutable" Version="5.0.0" />
-    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="7.0.0" />
   </ItemGroup>
 
+  <Choose>
+    <When Condition="'$(TargetFramework)' == 'net9.0'">
+      <ItemGroup>
```
fabiocav, line 29: You can simplify this by moving the condition to your ItemGroup (see line 46 below)

src/DotNetWorker.Core/DotNetWorker.Core.csproj
fabiocav: Similar feedback above. I'm assuming @kshyju was trying to avoid another evaluation for the fallback (block below), but I still think that will be simpler.

src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
```diff
@@ -27,10 +27,23 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
-    <PackageReference Include="Microsoft.Extensions.Hosting" Version="5.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="5.0.0" />
   </ItemGroup>
 
+  <Choose>
+    <When Condition="'$(TargetFramework)' == 'net9.0'">
```
fabiocav, line 33: If we make the change to move condition to the ItemGroup above, let's be consistent here.

fabiocav (commented):
Really minor nit comments. Changes

src/DotNetWorker.Core/DotNetWorker.Core.csproj
kshyju: Moved condition to ItemGroup element instead of Choose-When pattern

src/DotNetWorker.Core/DotNetWorker.Core.csproj
kshyju: Moved condition to ItemGroup element instead of Choose-When pattern

samples/Net9FunctionApp/HelloHttp.cs
```diff
@@ -0,0 +1,17 @@
...
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Extensions.Logging;
+
+namespace Net9FunctionApp
+{
+    public class HelloHttp(ILogger<HelloHttp> logger)
+    {
+        [Function("HelloHttp")]
+        public IActionResult Run([HttpTrigger(AuthorizationLevel.Anonymous, "get", "post")] HttpRequest req)
```
surgupta-msft, line 11: I know this is everywhere in the samples but as a best practice we should keep `AuthorizationLevel` as `Function` here to set the correct example for users. I remember mathewc told us for [sdk type bindings](https://github.com/Azure/azure-functions-dotnet-worker/blob/21ce5a7b653a26fe55c1e9d0bfa4a614426e1ac9/samples/Extensions/Blob/BlobInputBindingSamples.cs#L35) samples.

samples/Net9FunctionApp/HelloHttp.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Extensions.Logging;
+
+namespace Net9FunctionApp
```
surgupta-msft, line 6: nit - should we name it as `Net9Worker` to keep it consistent with other tfm [sample apps](https://github.com/Azure/azure-functions-dotnet-worker/tree/main/samples)?

samples/Net9FunctionApp/Net9FunctionApp.csproj
```diff
@@ -0,0 +1,29 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net9.0</TargetFramework>
+    <AzureFunctionsVersion>v4</AzureFunctionsVersion>
+    <OutputType>Exe</OutputType>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+  <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
```
surgupta-msft, line 10: Is aspnet core mandatory here?

samples/Net9FunctionApp/Net9FunctionApp.csproj
```diff
@@ -0,0 +1,29 @@
...
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http" Version="3.2.0" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore" Version="1.3.2" />
+    <PackageReference Include="Microsoft.ApplicationInsights.WorkerService" Version="2.22.0" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.ApplicationInsights" Version="1.2.0" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.17.0" />
```
surgupta-msft, line 15: Update the latest versions of worker.sdk and worker.ApplicationInsights

samples/Net9FunctionApp/Net9FunctionApp.csproj
```diff
@@ -0,0 +1,29 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net9.0</TargetFramework>
+    <AzureFunctionsVersion>v4</AzureFunctionsVersion>
```
fabiocav, line 4: We should be able to omit this and rely on the default resolution from the SDK

src/DotNetWorker.Core/DotNetWorker.Core.csproj
```diff
@@ -22,6 +22,16 @@
 
   <ItemGroup>
     <PackageReference Include="Azure.Core" Version="1.37.0" />
```
surgupta-msft, line 24: Update Azure.Core to the latest version

fabiocav (approved):

src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
```diff
@@ -2,15 +2,15 @@
 
   <PropertyGroup>
     <OutputType>Library</OutputType>
-    <TargetFrameworks>net5.0;net6.0;net7.0;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net5.0;net6.0;net7.0;net9.0;netstandard2.0</TargetFrameworks>
```
surgupta-msft, line 5: wondering why there was no .Net8 here?

samples/Net9FunctionApp/HelloHttp.cs
kshyju: TBH, I felt this name is more readable. In my opinion, we should update the other ones too.

samples/Net9FunctionApp/Net9FunctionApp.csproj
kshyju: Not mandatory, but we include that in the template by default to get the best perf.

src/DotNetWorker.Grpc/DotNetWorker.Grpc.csproj
kshyju: We chose to not do it at that time :) Will be fixing in the 2x version.

src/DotNetWorker.Core/DotNetWorker.Core.csproj
kshyju: This package version update is unrelated to the changes we are adding in this PR (net9 TFM support). We can upgrade it in the 2x branch.

samples/Net9FunctionApp/Net9FunctionApp.csproj
kshyju: Updated AI package. can't update worker.sdk until we publish a version which supports net9. 1.17.0 has a glitch which will allow us to use it for a net9 app.

samples/Net9FunctionApp/HelloHttp.cs
kshyju: Do you know why it is a best practice?

Issue #2641, opened 2024-08-01
dotnet publish /t:PublishContainer on linux platform does not publish some dlls to a docker image


whitebehemoth:
### Description

If an isolated function is published using `PublishContainer` target on linux, some DLL are missing in the image. 

- Microsoft.Azure.Functions.Worker.Extensions
- Microsoft.Azure.WebJobs.Extensions.FunctionMetadataLoader
- Microsoft.Azure.WebJobs.Host.Storage
- Microsoft.WindowsAzure.Storage
- Microsoft.WindowsAzure.Storage.dll

With `WorkerExtensions.csproj` included into a solution four of them appears in the image, but not `FunctionMetadataLoader`. This results `LanguageWorkerOptions.WorkerConfigs` be empty on container start and have "No job functions found" error on the function start.

Please note, there is no issue with publishing on windows platform. 

our workaround for our devops pipeline:
1. publish without `/t:PublishContainer` option, 
2. build the image using a simple Dockerfile to copy the published folder, 

### Steps to reproduce

Issue can be reproduced using sample func on windows with wsl2 Ubuntu 22.04, on dotnet sdk 8.0.303 

1. run `func init --worker-runtime dotnet-isolated --docker`
2. run on linux `dotnet publish --configuration Release  --os linux --arch x64 --property:ContainerRepository=simplefunc --property:ContainerImageTag=linux --target:PublishContainer`
3. run the same command with different tag on windows `dotnet publish --configuration Release  --os linux --arch x64 --property:ContainerRepository=simplefunc --property:ContainerImageTag=win --target:PublishContainer`
4. compare images `docker scout compare --to simplefunc:win simplefunc:linux`
5. check the result:  5 packages are missing (scroll down to see them listed)
![image](https://github.com/user-attachments/assets/1862032b-8603-44a8-a9cb-a7b3c39f1d1e)

mattchenderson:
I am unable to reproduce this using SDK 8.0.4 / .NET 9 preview 6. For my setup, I am using version 4.0.5907 of the Core Tools, and I ran the commands as presented in the repro steps. This scaffolded the project with Microsoft.Azure.Functions.Worker 1.21.0 and Microsoft.Azure.Functions.Worker.Sdk 1.17.0 as the main dependencies, which I will note are not latest.

[...]

PR #2671, opened 2024-08-23
Adding default settings to support SDK container builds

### LINKED ISSUE #2617: Support SDK container builds from worker
### Description

[SDK container builds](https://learn.microsoft.com/en-us/dotnet/core/docker/publish-as-container?pivots=dotnet-8-0) are a nice feature of .NET (starting from .NET 7) that lets you get rid of a Dockerfile. The worker should support this approach and provide productive defaults.

The Web SDK does this today: https://github.com/dotnet/sdk/blob/df20c62f59589aec18ae09ff01754e5d97efe59a/src/WebSdk/Publish/Targets/Microsoft.NET.Sdk.Publish.props#L14-L16

We should ensure the same occurs for isolated worker projects by including the same from worker SDK builds (when TFM >= net7.0).

Our props/target should also include the right default `ContainerBaseImage`, etc. See @baronfel's gist: https://gist.github.com/baronfel/e97f46933d05365ea93d7855af742e58

Note that I think it is generally preferred these days that `_TargetFrameworkVersionWithoutV` be replaced with `$(TargetFrameworkVersion.TrimStart('vV'))` or other MSBuild built-ins for versioning.

This work has general benefits but is also needed for .NET Aspire work.

mattchenderson:

resolves #2617 

### Pull request checklist

### Additional PR information

Because the tests associated with these changes require a docker daemon, I've moved them out of the default test set by setting a trait and applying a default filter which removes it. The relevant test can still be run by explicitly specifying the other runtime settings file included in the PR. Discussion may be needed around running this test within CI.

In addition to passes for the newly added tests, I have also manually tested the resulting output in both Windows and Linux, using .NET SDK 8.0.4 as per the discussion in #2641. The resulting images have no differences in included packages.

~~I would like to discuss the impact of these changes, as they would modify the default base image and some other settings for apps that are already using SDK container builds. Those would likely be unsuccessful today, so it is likely safe to do so. However, this still could be a change that warrants a major version revision.~~ Edit: This now targets a new major version. See comment below.

If someone is setting the properties explicitly, those would continue to be honored. I also defined a new property (`FunctionsContainerOmitDefaultEnvsVars`) which can be used to suppress the environment variables being added should someone be customizing the other properties or have some other reason to modify them. It was not clear to me how I could otherwise conditionally avoid overwriting values that someone else had supplied for these.

### DIFF
--- a/sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.Publish.targets
+++ b/sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.Publish.targets
@@ -21,4 +21,30 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
   <Import Project="$(MSBuildThisFileDirectory)Microsoft.Azure.Functions.Worker.Sdk.Publish.$(PublishProtocol).targets"  
           Condition="Exists('$(MSBuildThisFileDirectory)Microsoft.Azure.Functions.Worker.Sdk.Publish.$(PublishProtocol).targets')" />
 
+    <Target
+          Name="AssignFunctionsBaseImage"
+          BeforeTargets="ComputeContainerBaseImage"
+          DependsOnTargets="_FunctionsPreBuild" >
+        <PropertyGroup>
+            <_FunctionsRuntimeMajorVersion>$(_AzureFunctionsVersionStandardized.TrimStart('vV'))</_FunctionsRuntimeMajorVersion>
+            <ContainerBaseImage Condition="'$(ContainerBaseImage)' == ''">mcr.microsoft.com/azure-functions/dotnet-isolated:$(_FunctionsRuntimeMajorVersion)-dotnet-isolated$(TargetFrameworkVersion.TrimStart('vV'))</ContainerBaseImage>
+            <ContainerWorkingDirectory Condition="'$(ContainerWorkingDirectory)' == ''">/home/site/wwwroot</ContainerWorkingDirectory>
+            <!-- Functions base images only support amd64 runtimes -->
+            <ContainerRuntimeIdentifier Condition="'$(ContainerRuntimeIdentifier)' == ''">linux-x64</ContainerRuntimeIdentifier>
+        </PropertyGroup>
+        <ItemGroup>
+            <ContainerEnvironmentVariable
+                    Condition="@(ContainerEnvironmentVariable->AnyHaveMetadataValue('Identity', 'AzureWebJobsScriptRoot')) == false"
+                    Include="AzureWebJobsScriptRoot"
+                    Value="$(ContainerWorkingDirectory)" />
+            <ContainerEnvironmentVariable
+                    Condition="@(ContainerEnvironmentVariable->AnyHaveMetadataValue('Identity', 'AzureFunctionsJobHost__Logging__Console__IsEnabled')) == false"
+                    Include="AzureFunctionsJobHost__Logging__Console__IsEnabled"
+                    Value="true" />
+        </ItemGroup>
+        <ItemGroup>
+            <ContainerAppCommand Include="/opt/startup/start_nonappservice.sh" />
+        </ItemGroup>
+    </Target>
+
 </Project>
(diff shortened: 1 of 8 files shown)


mattchenderson:
Discussed briefly with @fabiocav, and we agree that this indeed makes sense as a 2.0 for the SDK. I will rebase and retarget the `feature/2.x` branch.

sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.Publish.targets
```diff
@@ -21,4 +21,28 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
   <Import Project="$(MSBuildThisFileDirectory)Microsoft.Azure.Functions.Worker.Sdk.Publish.$(PublishProtocol).targets"  
           Condition="Exists('$(MSBuildThisFileDirectory)Microsoft.Azure.Functions.Worker.Sdk.Publish.$(PublishProtocol).targets')" />
 
+    <Target
+          Name="AssignFunctionsBaseImage"
+          BeforeTargets="ComputeContainerBaseImage"
+          DependsOnTargets="_FunctionsPreBuild" >
+        <PropertyGroup>
+            <_FunctionsRuntimeMajorVersion>$(_AzureFunctionsVersionStandardized.Split('-')[0].TrimStart('vV'))</_FunctionsRuntimeMajorVersion>
```
mattchenderson, line 29: This line can drop the `.Split('-')[0]` once https://github.com/Azure/azure-functions-dotnet-worker/pull/2676 is ported into the `feature/2.x` branch. The change would not be required, but it would clean things up to do so.

sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.Publish.targets
```diff
@@ -21,4 +21,28 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
...
+            <ContainerBaseImage Condition="'$(ContainerBaseImage)' == ''">mcr.microsoft.com/azure-functions/dotnet-isolated:$(_FunctionsRuntimeMajorVersion)-dotnet-isolated$(TargetFrameworkVersion.TrimStart('vV'))</ContainerBaseImage>
+            <ContainerWorkingDirectory Condition="'$(ContainerWorkingDirectory)' == ''">/home/site/wwwroot</ContainerWorkingDirectory>
+            <!-- Functions base images only support amd64 runtimes -->
+            <ContainerRuntimeIdentifier Condition="'$(ContainerRuntimeIdentifier)' == ''">linux-x64</ContainerRuntimeIdentifier>
+        </PropertyGroup>
+        <ItemGroup Condition="'$(FunctionsContainerOmitDefaultEnvVars)' != 'true'">
```
baronfel, line 35: In the issue description you mention wanting to defensively set this value, i.e. only setting the value if `AzureWebJobScriptRoot` is unset by a user at this point. You can check if the Item is already present by checking for an Item with that Identity:

```xml
<ContainerEnvironmentVariable
                    Condition="@(ContainerEnvironmentVariable->AnyHaveMetadataValue('Identity', 'AzureWebJobsScriptRoot')) == false"
                    Include="AzureWebJobsScriptRoot"
                    Value="$(ContainerWorkingDirectory)" />
```

mattchenderson: I like this MUCH better. Added to newer iteration. Thanks!

sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.Publish.targets
mattchenderson: This is addressed in the latest iteration.

brettsam (approved):

fabiocav (approved):

mattchenderson:
Pushing one more revision to fix the merge conflict on the release notes. Will then merge.

PR #2606, opened 2024-07-16
Add warning for v3 out of support. Fix incorrect build message.

jviau:
### Issue describing the changes in this PR

resolves #issue_for_this_pr

### Pull request checklist

* [x] My changes **do not** require documentation changes
  * [ ] Otherwise: Documentation issue linked to PR
* [ ] My changes **should not** be added to the release notes for the next release
  * [x] Otherwise: I've added my notes to `release_notes.md`
* [x] My changes **do not** need to be backported to a previous version
  * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [x] I have added all required tests (Unit tests, E2E tests)

### Additional information

Fixes the function version printed when defaulting during build. This also adds a warning for functions v3 being out of support.

PR #2676, opened 2024-08-28
Allowing version validation to include qualities for testing purposes

mattchenderson:
New version validation introduced in https://github.com/Azure/azure-functions-dotnet-worker/pull/2606 is too strict. It prohibits testing with quality tags, which is a part of the prerelease validation process for components like the tooling feed. The validation check should be for the major version format only.

This change will need to make it into the `feature/2.x` branch as well.

This should also inform changes on https://github.com/Azure/azure-functions-dotnet-worker/pull/2671 to remove a similar change in a later property, though those aren't strictly necessary (the secondary action shouldn't hurt). I think we can just rebase that once this change is in the `feature/2.x` branch.

### Pull request checklist

### DIFF
--- a/sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
+++ b/sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
@@ -63,7 +63,8 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
   <!-- Validating some expectations for the function app early on. -->
   <Target Name="_FunctionsPreBuild" BeforeTargets="BeforeBuild">
     <PropertyGroup>
-      <_AzureFunctionsVersionStandardized>$(AzureFunctionsVersion.ToLowerInvariant())</_AzureFunctionsVersionStandardized>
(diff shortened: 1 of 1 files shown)


sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
```diff
@@ -63,7 +63,7 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
   <!-- Validating some expectations for the function app early on. -->
   <Target Name="_FunctionsPreBuild" BeforeTargets="BeforeBuild">
     <PropertyGroup>
-      <_AzureFunctionsVersionStandardized>$(AzureFunctionsVersion.ToLowerInvariant())</_AzureFunctionsVersionStandardized>
+      <_AzureFunctionsVersionStandardized>$(AzureFunctionsVersion.ToLowerInvariant().Split('-')[0])</_AzureFunctionsVersionStandardized>
```
fabiocav, line 66: Can you please add a comment with expected version formats, so the expectation is clear to others reading this logic?

fabiocav (approved):

PR #2688, opened 2024-09-06
Fuction App Flex Consumption publish support

anvillan:

resolves (TBD)

### In this PR:
 - Add new `UseBlobContainerDeploy` (`bool`) property to the `ZipDeploy` target/task to identify when to use the 
 "`<publishUrl>/api/publish`" endpoint to publish a Flex Consumption function app
 - When polling for the ZipDeploy pubish status, append the `status_text` message (if defined) along with the status itself. E.g.: `"Deployment status is Failed: App setting WEBSITE_RUN_FROM_PACKAGE is not supported..."`.

### Pull request checklist


nzthiago:
@patelchandni @fabiocav - this PR is from the Visual Studio team to add publish support for Flex Consumption for the .NET Isolated worker. Please review.

fabiocav (commented):
@anvillan can you please add a note covering those changes to sdk/release_notes.md?

Thanks!

sdk/Sdk/Tasks/ZipDeploy/ZipDeploymentStatus.cs
```diff
@@ -54,10 +53,16 @@ public async Task<DeployStatus> PollDeploymentStatusAsync(string deploymentUrl,
             {
```
patelchandni, line 53: For OneDeploy /publish endpoint on flex consumption, we have the following deployment [statuses](https://msazure.visualstudio.com/Antares/_git/Antares-Kudu-Legion?path=%2Fsrc%2FKudu.Legion.Core%2FContracts%2FDeployStatus.cs&_a=contents&version=GBmain): 

Cancelled = -1 => treat as terminating with CANCELLED status *
Pending = 0 
Building = 1 
Deploying = 2 
Failed = 3 => treat as terminating with FAILED status
Success = 4 => treat as terminating with SUCCESS status
Conflict = 5 => treat as terminating with FAILED status *
PartialSuccess = 6 => treat as terminating with SUCCESS status *

* these are the new statuses to look for onedeploy on flex.

sdk/Sdk/Tasks/ZipDeploy/ZipDeployTask.cs
```diff
@@ -89,6 +95,12 @@ internal async System.Threading.Tasks.Task<bool> ZipDeployAsync(string zipToPubl
                 return false;
             }
 
+            // publish endpoint differs when using a blob storage container
+            var publishUriPath = useBlobContainerDeploy ? "publish" : "zipdeploy";
```
patelchandni, line 99: Apart from updating the /publish endpoint, we also need to pass a required parameter for remote build and deployer parameter for better telemetry: 
`/publish?Deployer=VisualStudio&RemoteBuild=false`

patelchandni (commented):
Suggested few changes based on new API behavior. Rest looks good, thank you.

sdk/Sdk/Tasks/ZipDeploy/ZipDeployTask.cs
fabiocav: This task could be executed outside of VS. This could be set using the agent provided if needed.

@patelchandni how is the deployer information being used?

patelchandni: @fabiocav - Mainly `RemoteBuild=false` is a required parameter for new API. 
And `Deployer=VisualStudio` is an optional parameter for better deployment telemetry internally. Agent is a lot of times very cryptic whereas we have provided an option to give straight forward value for deployer. So, Az CLI, VS Code, DevOps, and GitHub Action are all now passing this parameter.

fabiocav: I'm OK with logic that resolves a friendly deployer based on the agent. Not proposing the use of the literal agent, though, if using for telemetry, using that information would not be a bad idea as it would give you the ability to track behaviors across different versions of the extension.

The key thing is that we shouldn't assume VS in this path as that wouldn't be accurate.

sdk/Sdk/Tasks/ZipDeploy/ZipDeploymentStatus.cs
nzthiago: @patelchandni - can you also confirm, while Flex is the first Functions hosting plan to support OneDeploy, we can also bring support for other hosting plans in the future right? It should not be tightly coupled with Flex?

anvillan: @patelchandni I added the rest of the `enum` values: `PartiallySuccess` and `Conflict` and modified the code to treat them as `SUCCESS` and `FAILED` respectively.

As we discussed, `Cancelled` will be treated as `Unknown` because...
 - They both have `-1` value 
 - The final result of `ZipDeploy` task is a _binary_ response: either it succeeds or not. So a 3rd response type (cancelled) cannot be added to the existing functionality.

sdk/Sdk/Tasks/ZipDeploy/ZipDeployTask.cs
anvillan: Done. 
Modified the URLs as:
- ZipDeploy : `<publishUrl>/api/zipdeploy?isAsync=true`
- OneDeploy (thru ZipDeploy): `<publishUrl>/api/publish?RemoteBuild=false`

anvillan:
@patelchandni 

I added UTs for `ZipDeploy` task and `ZipDeploymentStatus`. They are a copy of what it is already in `azure_functions_vs_build_sdk` repo. Here: https://github.com/anvillan/azure-functions-vs-build-sdk/tree/main/test/Microsoft.NET.Sdk.Functions.MSBuild.Tests

I also added the new cases for the new _variations_ after the changes introduced in this PR.

sdk/Sdk/Tasks/ZipDeploy/ZipDeploymentStatus.cs
patelchandni: Thank you @anvillan

@nzthiago - Onedeploy API works quite different for other SKUs and takes different parameters. So yes, we can bring support for onedeploy for other SKUs, but the current changes are not sufficient for that. Also, some backend changes will be needed to support onedeploy on other function app SKUs.

That was one of the main reasons why I suggested to have a separate OneDeploy task so that we can put changes for all SKUs in one place in future. So if we do decide to support onedeploy for other hosting plans, then we should definitely have a separate task for onedeploy. And move flex code path to the same task.

sdk/Sdk/Tasks/ZipDeploy/ZipDeployTask.cs
patelchandni: Thank you @anvillan. To confirm, this OneDeploy option is only support for function apps on Flex consumption, right?

patelchandni (approved):
Looks good. Thank you.

anvillan:
@fabiocav, could you please take a look and help me to complete all the checks so this PR merges?

I can't see any actionable items for me to do.

CC: @nzthiago, @patelchandni

anvillan:
This change was merged with https://github.com/Azure/azure-functions-dotnet-worker/pull/2712. 

Closing this PR.

PR #2697, opened 2024-09-10
add missing dispose for process instances

SimonCropp:


src/DotNetWorker.Core/StartupHook.cs
```diff
@@ -35,7 +35,8 @@ public static void Initialize()
 #if NET5_0_OR_GREATER
         int processId = Environment.ProcessId;
 #else
-        int processId = Process.GetCurrentProcess().Id;
+        using var process = Process.GetCurrentProcess();
```
KondzioSSJ4, line 38: Maybe better would be making a exception method to allow get Process and then dispose it instantly
because this version means that this `process` variable will be disposed on the end of this method and there is no reason for that

Also...
If there is other and easier way to getting processId (from `Environment` as somebody did few lines higher)
then maybe  if would be better to group this code together and use in every location where you updated?

kshyju, line 38: Does this code compile? "Using declarations" is available from C# 8 onwards. For a .net framework function app that is not the case.

SimonCropp: c# version  has nothing to do with target framework. eg u can use c# 12 to target .net 472

fabiocav: +1 to this feedback.

We have the ability to abstract the symbol conditions away from the consumer, simplifying the callers so they can all use a single method (regardless of TFM) that will to the right (and most efficient) thing.

fabiocav (changes requested):

PR #2699, opened 2024-09-10
Updating ASP.NET Core extension version

fabiocav:
Updating extension version for release.

### Pull request checklist

### DIFF
--- a/extensions/Worker.Extensions.Http.AspNetCore/src/Worker.Extensions.Http.AspNetCore.csproj
+++ b/extensions/Worker.Extensions.Http.AspNetCore/src/Worker.Extensions.Http.AspNetCore.csproj
@@ -6,7 +6,8 @@
     <Description>ASP.NET Core extensions for .NET isolated functions</Description>
 
     <!--Version information-->
-    <VersionPrefix>1.3.3</VersionPrefix>
+    <VersionPrefix>2.0.0</VersionPrefix>
+    <VersionSuffix>-preview1</VersionSuffix>
     <TargetFramework>net6.0</TargetFramework>
   </PropertyGroup>
 


mattchenderson (approved):
Do we also need to update release notes here? Or was that already in the previous PR?

extensions/Worker.Extensions.Http.AspNetCore/src/Worker.Extensions.Http.AspNetCore.csproj
```diff
@@ -6,7 +6,8 @@
     <Description>ASP.NET Core extensions for .NET isolated functions</Description>
 
     <!--Version information-->
-    <VersionPrefix>1.3.3</VersionPrefix>
+    <VersionPrefix>2.0.0</VersionPrefix>
+    <VersionSuffix>-preview1</VersionSuffix>
```
kshyju, line 10: Unrelated to this change. But should we pin to specific versions other package references below (Worker.Extensions.Abstraction etc.,) ?

kshyju (approved):
Echoing the same question Matthew raised about TargetFramework value update.

extensions/Worker.Extensions.Http.AspNetCore/src/Worker.Extensions.Http.AspNetCore.csproj
```diff
@@ -6,7 +6,8 @@
...
     <TargetFramework>net6.0</TargetFramework>
```
mattchenderson, line 11: Should the TFM also be updated to reflect use of the newer types?

fabiocav: I don't want to combine that change with this, but that will likely change for 2.0. Let's discuss whether we want that change in place before the first drop.

extensions/Worker.Extensions.Http.AspNetCore/src/Worker.Extensions.Http.AspNetCore.csproj
fabiocav: This shouldn't be required, but I would like to follow up on that separately. Let's get the proper version in place and follow up.

fabiocav:
About @mattchenderson's comment:

> Do we also need to update release notes here? Or was that already in the previous PR?

Discussed offline., It was not in the original PR with the changes, but will be applied when the release is tagged.

Issue #876, opened 2022-05-09
Event metadata can get out-of-sync when using this sample code for event hubs trigger


wesselkranenborg:
In this repository there are some samples on how to fetch metadata from messages: https://github.com/Azure/azure-functions-dotnet-worker/blob/main/samples/Extensions/EventHubs/EventHubsTriggerMetadata.cs.

The code referring too is this:

```csharp
 [Function("EventHubsTriggerMetadata-Parameters")]
        public static void UsingParameters([EventHubTrigger("src-parameters", Connection = "EventHubConnectionAppSetting")] string[] messages,
            DateTime[] enqueuedTimeUtcArray,
            long[] sequenceNumberArray,
            string[] offsetArray,
            Dictionary<string, JsonElement>[] propertiesArray,
            Dictionary<string, JsonElement>[] systemPropertiesArray)
        {
            // You can directly access binding data via parameters.
            for (int i = 0; i < messages.Length; i++)
            {
                string message = messages[i];
                DateTime enqueuedTimeUtc = enqueuedTimeUtcArray[i];
                long sequenceNumber = sequenceNumberArray[i];
                string offset = offsetArray[i];

                // Note: The values in these dictionaries are sent to the worker as JSON. By default, System.Text.Json will not automatically infer primitive values
                //       if you attempt to deserialize to 'object'. See https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to?pivots=dotnet-5-0#deserialization-of-object-properties       
                //       
                //       If you want to use Dictionary<string, object> and have the values automatically inferred, you can use the sample JsonConverter specified
                //       here: https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-converters-how-to?pivots=dotnet-5-0#deserialize-inferred-types-to-object-properties
                //
                //       See the Configuration sample in this repo for details on how to add this custom converter to the JsonSerializerOptions, or for 
                //       details on how to use Newtonsoft.Json, which does automatically infer primitive values.
                Dictionary<string, JsonElement> properties = propertiesArray[i];
                Dictionary<string, JsonElement> systemProperties = systemPropertiesArray[i];
            }
        }
```

[...]

PR #2701, opened 2024-09-10
Set IncludeEmptyEntriesInMessagePayload capability to true by default

### LINKED ISSUE #1096: Always send IncludeEmptyEntriesInPayload capability
We added `IncludeEmptyEntriesInPayload` property to WorkerOptions in #1091. We can evaluate whether this flag can be removed in **2x** and have the dotnet worker always send this capability entry.

liliankasem:

resolves #1096

### Pull request checklist

### Additional information

This makes a change so that the worker will by default send the `IncludeEmptyEntriesInPayload` capability to the host. This means that empty entries will always be included in the message payload. This was initially created as a flag to avoid breaking users but there is no reason not to make this the permanent behaviour now as it's what would normally be expected.

Original bug: https://github.com/Azure/azure-functions-dotnet-worker/issues/876

### DIFF
--- a/src/DotNetWorker.Grpc/GrpcWorker.cs
+++ b/src/DotNetWorker.Grpc/GrpcWorker.cs
@@ -320,12 +320,12 @@ namespace Microsoft.Azure.Functions.Worker
             }
 
             // Add required capabilities; these cannot be modified and will override anything from WorkerOptions
-            capabilities["RpcHttpBodyOnly"] = bool.TrueString;
-            capabilities["RawHttpBodyBytes"] = bool.TrueString;
-            capabilities["RpcHttpTriggerMetadataRemoved"] = bool.TrueString;
-            capabilities["UseNullableValueDictionaryForHttp"] = bool.TrueString;
-            capabilities["TypedDataCollection"] = bool.TrueString;
-            capabilities["WorkerStatus"] = bool.TrueString;
+            capabilities[WorkerCapabilities.RpcHttpBodyOnly] = bool.TrueString;
+            capabilities[WorkerCapabilities.RawHttpBodyBytes] = bool.TrueString;
+            capabilities[WorkerCapabilities.RpcHttpTriggerMetadataRemoved] = bool.TrueString;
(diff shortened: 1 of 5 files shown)


src/DotNetWorker.Grpc/GrpcWorker.cs
```diff
@@ -320,12 +320,13 @@ private static IDictionary<string, string> GetWorkerCapabilities(WorkerOptions w
...
-            capabilities["UseNullableValueDictionaryForHttp"] = bool.TrueString;
-            capabilities["TypedDataCollection"] = bool.TrueString;
-            capabilities["WorkerStatus"] = bool.TrueString;
+            capabilities[WorkerCapabilities.RpcHttpBodyOnly] = bool.TrueString;
+            capabilities[WorkerCapabilities.RawHttpBodyBytes] = bool.TrueString;
+            capabilities[WorkerCapabilities.RpcHttpTriggerMetadataRemoved] = bool.TrueString;
+            capabilities[WorkerCapabilities.UseNullableValueDictionaryForHttp] = bool.TrueString;
+            capabilities[WorkerCapabilities.TypedDataCollection] = bool.TrueString;
+            capabilities[WorkerCapabilities.WorkerStatus] = bool.TrueString;
+            capabilities[WorkerCapabilities.IncludeEmptyEntriesInMessagePayload] = bool.TrueString;
```
kshyju, line 329: Instead of sending this always as enabled, should we just set the default value to "enabled". This allows users to use the previous behavior if they choose to do so by overriding the worker option property in their bootstrapping code.

liliankasem: I must have misunderstood the task, I thought we were trying to get rid of the capability as an option. I can bring it back and set it as default behaviour instead!

liliankasem: I am assuming it's similar for the enable user code change?

kshyju (approved):
Let's also capture this change in the release notes.

PR #2702, opened 2024-09-10
Set EnableUserCodeException capability to true by default

### LINKED ISSUE #1082: Set the value of EnableUserCodeException worker capability to true
- Set the value of `EnableUserCodeException` worker capability to True as default. 
- Evaluate whether `EnableUserCodeException` property shall be removed from [WorkerOptions](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/src/DotNetWorker.Core/Hosting/WorkerOptions.cs). (this was added in #939)

liliankasem:

resolves #1082

### Pull request checklist

### Additional information

This makes a change so that the worker will by default send the EnableUserCodeException capability to the host. This means that we will unwrap user-thrown exceptions that occur during an invocation and surface them as the original type. This was initially created as a flag to avoid breaking users who were expecting an RpcException for all their invocation exceptions, but there is no reason not to make this the permanent behaviour now as it would be expected that we correctly surface user exceptions in AppInsights instead of wrapping them. Marking as obsolete to potentially remove in future major versions.

Original issue: https://github.com/Azure/azure-functions-host/issues/4296

### Open question about this change:

Here is an example in the difference in behaviour between 'EnableUserCodeException' being on/off:

#### EnableUserCodeException: false

- Type = RpcException
- Message = 
`"System.AggregateException: One or more errors occurred. (A task was canceled.)\n ---> System.Threading.Tasks.TaskCanceledException: A task was canceled.\n   --- End of inner exception stack trace ---\n   at Moq.Behaviors.ThrowException.Execute(Invocation invocation) in /_/src/Moq/Behaviors/ThrowException.cs:line 22\n   at Moq.MethodCall.ExecuteCore(Invocation invocation) in /_/src/Moq/MethodCall.cs:line 109\n   at Moq.Setup.Execute(Invocation invocation) in /_/src/Moq/Setup.cs:line 84\n   at Moq.FindAndExecuteMatchingSetup.Handle(Invocation invocation, Mock mock) in /_/src/Moq/Interception/InterceptionAspects.cs:line 95\n   at Moq.Mock.Moq.IInterceptor.Intercept(Invocation invocation) in /_/src/Moq/Interception/Mock.cs:line 17\n   at Moq.CastleProxyFactory.Interceptor.Intercept(IInvocation underlying) in /_/src/Moq/Interception/CastleProxyFactory.cs:line 107\n   at Castle.DynamicProxy.AbstractInvocation.Proceed()\n   at Castle.Proxies.IFunctionsApplicationProxy.InvokeFunctionAsync(FunctionContext context)\n   at Microsoft.Azure.Functions.Worker.Handlers.InvocationHandler.InvokeAsync(InvocationRequest request) in /Users/likasem/source/functions/azure-functions-dotnet-worker/src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs:line 88"`
- StackTrace = same as message

#### EnableUserCodeException: true

- Type = AggregateException
- Message = "One or more errors occurred. (A task was canceled.)"
- StackTrace = 
`"   at Moq.Behaviors.ThrowException.Execute(Invocation invocation) in /_/src/Moq/Behaviors/ThrowException.cs:line 22\n   at Moq.MethodCall.ExecuteCore(Invocation invocation) in /_/src/Moq/MethodCall.cs:line 109\n   at Moq.Setup.Execute(Invocation invocation) in /_/src/Moq/Setup.cs:line 84\n   at Moq.FindAndExecuteMatchingSetup.Handle(Invocation invocation, Mock mock) in /_/src/Moq/Interception/InterceptionAspects.cs:line 95\n   at Moq.Mock.Moq.IInterceptor.Intercept(Invocation invocation) in /_/src/Moq/Interception/Mock.cs:line 17\n   at Moq.CastleProxyFactory.Interceptor.Intercept(IInvocation underlying) in /_/src/Moq/Interception/CastleProxyFactory.cs:line 107\n   at Castle.DynamicProxy.AbstractInvocation.Proceed()\n   at Castle.Proxies.IFunctionsApplicationProxy.InvokeFunctionAsync(FunctionContext context)\n   at Microsoft.Azure.Functions.Worker.Handlers.InvocationHandler.InvokeAsync(InvocationRequest request) in /Users/likasem/source/functions/azure-functions-dotnet-worker/src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs:line 88"`

When is `EnableUserCodeException=true`, the behaviour is more in line with what we would expect to see (instead of an all encapsulating RpcException). However, when it is an aggregate exception, we lose the inner exception (i.e. `TaskCanceledException`). To include an inner exception, we will need to update our [protobuf](https://github.com/Azure/azure-functions-language-worker-protobuf/blob/2a3fb29179e7e2dab6b0a8e7550e717f12295f8d/src/proto/FunctionRpc.proto#L618), and make host changes too.

⭐️ Is this still an okay change to make permanent? I think it is as it is already the current behaviour for most customers. We can always create a second work item to implement the InnerException changes across this repo+proto+host.


kshyju:
Is it our code which wraps the exception in the `AggregateException`?  I feel like users find it an undesired behavior. Logging their exception as it is might be more desirable.

liliankasem:
> Is it our code which wraps the exception in the `AggregateException`? I feel like users find it an undesired behavior. Logging their exception as it is might be more desirable.

Yeah I believe it is; not every exception is an aggregate exception - with task cancellation I noticed it was wrapped in aggregate, but I think other exceptions were not. I'll run some local E2E tests to validate this - it's been a minute :)

liliankasem:
> > Is it our code which wraps the exception in the `AggregateException`? I feel like users find it an undesired behavior. Logging their exception as it is might be more desirable.
> 
> Yeah I believe it is; not every exception is an aggregate exception - with task cancellation I noticed it was wrapped in aggregate, but I think other exceptions were not. I'll run some local E2E tests to validate this - it's been a minute :)

@kshyju - as expected, customer exceptions from their code show up properly, not wrapped in an aggregate:

Screenshot 2024-09-12 at 12 23 40 PM

I can't remember how I got the aggregate exception I mention in the PR desc (this is from a long time ago), I think it's when our code, during an invocation, throws an exception that we're doing some wrapping. Will try and repro a scenario like that but let me know if you have any ideas!

liliankasem:
Q: Do we want this to be permanently fixed in this way, or do we want to retain the ability for customers to disable this capability if they want? I don't see why they would want to keep their exceptions wrapped as an RpcException, this is technically a bug fix. But folks might have some dependency on "RpcException" that they don't want to change? 🤷🏻‍♀️

test/DotNetWorkerTests/Handlers/InvocationHandlerTests.cs
```diff
@@ -155,32 +157,6 @@ public async Task InvokeAsync_UserCodeThrowsException_OptionEnabled()
             Assert.True(response.Result.Exception.IsUserException);
         }
 
-        /// <summary>
-        /// Test keeping user code exception wrapped as RpcException. 
-        /// </summary>
-        [Fact]
-        public async Task InvokeAsync_UserCodeThrowsException_OptionDisabled()
```
mattchenderson, line 162: We could preserve this test if we keep the option (see review summary comment)

mattchenderson (commented):
> Q: Do we want this to be permanently fixed in this way, or do we want to retain the ability for customers to disable this capability if they want? I don't see why they would want to keep their exceptions wrapped as an RpcException, this is technically a bug fix. But folks might have some dependency on "RpcException" that they don't want to change? 🤷🏻‍♀️

I'm wondering if we should allow users to opt-into disablement, though we still get this default behavior change. Ideally, nobody uses that opt-in, but we at least defend that scenario for now in case there is a dependency. I don't think preserving that is necessarily bad, though I would in general like to reduce options if we can.

Along those lines, I wonder if we should keep the existing setting, so that the break is behavior and not API for those already using `EnableUserCodeException`. I don't like Booleans defaulting to true, but I wonder if that would provide the least churn. Those wanting to preserve the `RpcException` behavior would just be required to set `options.EnableUserCodeException = false;` And anybody with existing code calling `options.EnableUserCodeException = true;` has some redundant code that they could remove, but it's one less thing for that user to worry about when upgrading.

Were we to do that, I'd be in favor of a warning-level log at startup when user code exceptions are disabled.

fabiocav (commented):
> Q: Do we want this to be permanently fixed in this way, or do we want to retain the ability for customers to disable this capability if they want? I don't see why they would want to keep their exceptions wrapped as an RpcException, this is technically a bug fix. But folks might have some dependency on "RpcException" that they don't want to change?

Making this configurable if needed would be OK as it would be additive change, but I'm also inclined to say we should keep the option, flipping the default, so we can allow for users to opt-out and avoid a code change for those who have applied this configuration.

If we're keeping the property, let's obsolete this (https://learn.microsoft.com/en-us/dotnet/api/system.obsoleteattribute?view=net-8.0) so we can remove in a future version).

src/DotNetWorker.Core/Hosting/WorkerOptions.cs
```diff
@@ -32,13 +32,16 @@ public class WorkerOptions
...
         };
 
         /// <summary>
-        /// Gets or sets the flag for opting in to unwrapping user-code-thrown
-        /// exceptions when they are surfaced to the Host.
+        /// Gets or sets a value indicating whether exceptions thrown by user code should be unwrapped
+        /// and surfaced to the Host as their original exception type, instead of being wrapped in an RpcException.
+        /// The default value is <see langword="true"/>.
         /// </summary>
+        [Obsolete("This is now the default behaviour. This property may be unavailable in future releases.", false)]
```
fabiocav, line 45: nit: `behaviour` -> `behavior` (default en-US)

fabiocav (approved):

src/DotNetWorker.Core/Hosting/WorkerOptions.cs
liliankasem:
```suggestion
        [Obsolete("This is now the default behavior. This property may be unavailable in future releases.", false)]
```

iansall:
I don't have anything specific in my code to do with RpcExceptions or AggregateExceptions but have needed to set EnableUserCodeException to false otherwise I see issues like I have described in this issue:
https://github.com/Azure/azure-functions-durable-extension/issues/3024

tweediejd:
Where should we expect to see this change (user exception now bubble up instead of an RCP exception)?

Are we talking about RCP exceptions from within `orchestrations`?

We have upgraded all applicable `Azure.Functions.Worker.xxx` packages, but have yet to see any difference.

PR #2705, opened 2024-09-11
Set ValidateScopes to true when in dev environment

### LINKED ISSUE #606: Dependency Injection does not enable ValidateScopes for development environment
Hi,

When using dependency injection, one can register a singleton that depends on a scoped service.

```csharp
public class MyScopedClass { }
public class MySingletonClass { MySingletonClass (MyScopedClass scoped) { } }
<....>
services.AddSingleton<MySingletonClass>();
services.AddScoped<MyScopedClass>();
```

From AspNetCore, one assumes that this will throw an exception when used. [From the docs](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/web-host?view=aspnetcore-5.0#scope-validation):
> CreateDefaultBuilder sets ServiceProviderOptions.ValidateScopes to true if the app's environment is Development.

However, this is not the case in the isolated functions.

One can work around it by manually do this:
```csharp
.UseDefaultServiceProvider((context, options) =>
{
  options.ValidateScopes = context.HostingEnvironment.IsDevelopment();
})
``` 

Can this be enabled by default to minimize surprises?

liliankasem:

resolves #606

### Pull request checklist

### Additional information

I'm not sure if this is the right way to do this? Would using the factory be a better approach?

- https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.hosting.hostbuilder.useserviceproviderfactory?view=net-8.0


src/DotNetWorker/Hosting/WorkerHostBuilderExtensions.cs
```diff
@@ -171,6 +171,10 @@ public static IHostBuilder ConfigureFunctionsWorkerDefaults(this IHostBuilder bu
 
                     // Add default middleware
                     appBuilder.UseDefaultWorkerMiddleware();
+                })
+               .UseDefaultServiceProvider((context, options) =>
+                {
+                    options.ValidateScopes = context.HostingEnvironment.IsDevelopment();
```
kshyju, line 177: This looks like the correct change to me. What is the factory approach you mentioned in the PR description? How does that code change will look like?

liliankasem: I think the factory approach would look something like this:

```csharp
.UseServiceProviderFactory(new AutofacServiceProviderFactory())
    .ConfigureContainer<ContainerBuilder>(builder =>
    {
        new DefaultServiceProvider(yourOptionsHere);
    });
```

But that doesn't feel right to me, a little convoluted? And I don't know the perk of doing it this way either

mattchenderson (commented):
I was trying to wrap my head around this one, and I'm a bit confused based on the documented behavior. It seems like this is the default that .NET applies already.

> The default is the value of [IsDevelopment(IHostingEnvironment)](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.hosting.hostingenvironmentextensions.isdevelopment?view=net-8.0#microsoft-extensions-hosting-hostingenvironmentextensions-isdevelopment(microsoft-extensions-hosting-ihostingenvironment)).

Per https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.serviceprovideroptions.validatescopes?view=net-8.0#property-value

Is something else asserted to be overwriting that before we get to this point?

kshyju (approved):
Let's capture this change in the release notes.

PR #2749, opened 2024-10-06
Updating FunctionExecutorGenerator to avoid long if/else chains.

### LINKED ISSUE #2748: Function executor source generator uses a pattern that may lead to a compiler error
### Description

The function executor source generator (https://github.com/Azure/azure-functions-dotnet-worker/tree/main/sdk/Sdk.Generators/FunctionExecutor) uses a pattern that may lead to long `if`/`else` chains (this is dependent on the number of functions defined), and may end up triggering a compiler limitation, leading to error `CS8078: An expression is too long or complex to compile.`

This was originally reported at https://github.com/dotnet/roslyn/issues/72393 , and although the compiler team is addressing the limitation, the source generator issue should be fixed as we also have the opportunity to optimize the code it currently emits. 

### Steps to reproduce

Create a function app with a large number of functions (I don't have the exact repro number, but some of the examples reviewed are in the hundreds, with a recent support request for an app with 1200 functions).

The compilation will result in error `CS8078: An expression is too long or complex to compile.`

fabiocav:

resolves #2748 

This change updates the `FunctionExecutorGenerator` to avoid using long `if`/`else` chains that may lead to a compiler error.

The generation changes the emitted code from:
``` csharp
// ...
if (string.Equals(context.FunctionDefinition.EntryPoint, "FunctionName1", StringComparison.Ordinal))
{
 // generated body
}
else if (string.Equals(context.FunctionDefinition.EntryPoint, "FunctionName2", StringComparison.Ordinal))
{
   // generated body
}
```

to:
``` csharp
// ...
if (string.Equals(context.FunctionDefinition.EntryPoint, "FunctionName1", StringComparison.Ordinal))
{
 // generated body
    return;
}
if (string.Equals(context.FunctionDefinition.EntryPoint, "FunctionName2", StringComparison.Ordinal))
{
   // generated body
    return;
}
```

### Pull request checklist

### DIFF
--- a/sdk/Sdk.Generators/FunctionExecutor/FunctionExecutorGenerator.Emitter.cs
+++ b/sdk/Sdk.Generators/FunctionExecutor/FunctionExecutorGenerator.Emitter.cs
@@ -143,26 +143,24 @@ namespace Microsoft.Azure.Functions.Worker.Sdk.Generators
             {
                 var sb = new StringBuilder();
                 sb.Append(
-                   $$"""
+                   $"""
                 var inputBindingFeature = context.Features.Get<global::Microsoft.Azure.Functions.Worker.Context.Features.IFunctionInputBindingFeature>();
                             var inputBindingResult = await inputBindingFeature.BindFunctionInputAsync(context);
                             var inputArguments = inputBindingResult.Values;
-                {{(anyDefaultExecutor ? $"            _defaultExecutor = new Lazy<global::Microsoft.Azure.Functions.Worker.Invocation.IFunctionExecutor>(() => CreateDefaultExecutorInstance(context));{Environment.NewLine}" : string.Empty)}}
+                {(anyDefaultExecutor ? $"            _defaultExecutor = new Lazy<global::Microsoft.Azure.Functions.Worker.Invocation.IFunctionExecutor>(() => CreateDefaultExecutorInstance(context));{Environment.NewLine}" : string.Empty)}
                 """);
-
-                bool first = true;
 
                 foreach (ExecutableFunction function in functions)
                 {
                     var fast = function.Visibility == FunctionMethodVisibility.Public;
                     sb.Append($$"""
 
-                        {{(first ? string.Empty : "else ")}}if (string.Equals(context.FunctionDefinition.EntryPoint, "{{function.EntryPoint}}", StringComparison.Ordinal))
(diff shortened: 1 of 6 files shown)


sdk/Sdk.Generators/FunctionExecutor/FunctionExecutorGenerator.Emitter.cs
```diff
@@ -143,79 +143,86 @@ private static string GetMethodBody(IEnumerable<ExecutableFunction> functions, b
...
+                        end:
+                        return;
+            """);
                 return sb.ToString();
             }
 
             private static string EmitFastPath(ExecutableFunction function)
             {
                 var sb = new StringBuilder();
-                int functionParamCounter = 0;
```
fabiocav, line 174: Made changes to this method to improve readability and move declarations closer to use.

sdk/Sdk.Generators/FunctionExecutor/FunctionExecutorGenerator.Emitter.cs
```diff
@@ -143,79 +143,86 @@ private static string GetMethodBody(IEnumerable<ExecutableFunction> functions, b
...
             """);
-                    first = false;
                 }
 
+                sb.Append(
+            """
+
+                        end:
+                        return;
+            """);
```
fabiocav, line 172: These are the core changes to address the issue.

sdk/Sdk.Generators/FunctionExecutor/FunctionExecutorGenerator.Emitter.cs
```diff
@@ -143,79 +143,86 @@ private static string GetMethodBody(IEnumerable<ExecutableFunction> functions, b
...
                         }
+
             """);
-                    first = false;
                 }
 
+                sb.Append(
+            """
+
+                        end:
```
jviau, line 170: if the end section only includes a `return;` statement - can we just inline that return statement inside each `if` statement?

fabiocav: 🤦🏼  Yes, yes... it absolutely can. I was originally planning on additional changes here but backed those out and ended up keeping the same pattern. Fixed. There will be some additional changes to optimize this soon.

jviau (approved):

PR #2753, opened 2024-10-07
Remove unused FunctionsToolingSuffix

jviau:

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

Removes the properties `_ToolingSuffix` and `FunctionsToolingSuffix` from the SDK files. Their only usage was for an error, but beyond that there is no usage. This PR removes it, but we should discuss how valuable that error is to have. If we see value in this error, we should refactor to make it easier to add future versions.


liliankasem (approved):

mattchenderson:
I believe Visual Studio takes a dependency on these properties for some behaviors. We should not remove them without confirming that.

PR #2765, opened 2024-10-09
Adding a null check before initiating the internal Activity

### LINKED ISSUE #2733: Tracing is now very broken in Azure Functions
### Description

I'm not entirely sure what's changed, but I suspect this may have been a host change, but I'll list out what I'm saying based on multiple customer issues we've had.

There is now an empty Activity present in the execution path. This means that whenever a user adds tracing, this new Activity is the parent, however, it's not something that you can listen to since it does not have a source name.

This leads to the new ActivitySource named `Microsoft.Azure.Functions.Worker`, this does take over the Blank Activity when you listen to the source, however, it uses the TraceContext from the FunctionContext as it's parent, which doesn't use the inbound `traceparent` from the HttpRequest headers as it's parent. This means that your span has a parent which you're unable to access.

image
That is what a trace looks like right now with the OOTB setup.

```csharp
        services.AddOpenTelemetry()
            .ConfigureResource(resource => resource.AddService("practical-otel-azure-functions"))
            .UseFunctionsWorkerDefaults()
            .WithTracing(tracingBuilder =>
            {
                tracingBuilder
                    .SetSampler<AlwaysOnSampler>()
                    .AddSource("Microsoft.Azure.Functions.Worker")
                    .AddHttpClientInstrumentation()
                    .AddAspNetCoreInstrumentation()
                    .AddOtlpExporter();
            });
```

Aspire is rendering it in a semi-ok way, that's compensating for the fact that the data itself on those spans isn't right (missing parents, and missing root spans).

I think the big change is that, previously, there was not TraceContext on the function context, and therefore there was no active Activity. Since the upgrade, this has changed and now, there is an Activity which has a remote parent that there is no way to get access to (there's supposed to be a way to access it in Flex Consumption).

To be clear, this is a breaking change. Adding an Activity that is not trackable, and overriding the inbound `traceparent` causes functions to be untraceable due to the context propagation issues.

The AspNetCore span completely baffles me, since it's got a different parent to the FunctionsWorker span, and also somehow is outside of the hierarchy.

I'd love to know what caused this, and whether there is a plan to fix functions tracing for users who aren't upgrading to Flex Consumption.

### Steps to reproduce

Create a new HTTP Trigger function
Add OpenTelemetry
Send the data to somewhere you can see the raw data.
Run it locally
Hit the endpoint

Look at the telemetry and you'll see that it doesn't follow a linear hierarchy, and there is no span that doesn't have a parent (aka a root span)

RohitRanjanMS:
Adding a null check before starting the internal Activity in the `InvokeFunctionAsync` method. This will help in scenarios where customers want to consume the request telemetry emitted by AspNetCoreInstrumentation. Adding this null check ensures that if a request telemetry already exists (generated by AspNetCoreInstrumentation), the internal Activity won’t be started.

resolves #2733 

### Pull request checklist

### DIFF
--- a/src/DotNetWorker.Core/FunctionsApplication.cs
+++ b/src/DotNetWorker.Core/FunctionsApplication.cs
@@ -67,18 +67,23 @@ namespace Microsoft.Azure.Functions.Worker
 
         public async Task InvokeFunctionAsync(FunctionContext context)
         {
-            // This will act as an internal activity that represents remote Host activity. This cannot be tracked as this is not associate to an ActivitySource. 
-            using Activity activity = new Activity(nameof(InvokeFunctionAsync));
-            activity.Start();
+            Activity? activity = null;
 
-            if (ActivityContext.TryParse(context.TraceContext.TraceParent, context.TraceContext.TraceState, true, out ActivityContext activityContext))
+            if (Activity.Current is null)
             {
-                activity.SetId(context.TraceContext.TraceParent);
-                activity.SetSpanId(activityContext.SpanId.ToString());
-                activity.SetTraceId(activityContext.TraceId.ToString());
-                activity.SetRootId(activityContext.TraceId.ToString());
-                activity.ActivityTraceFlags = activityContext.TraceFlags;
-                activity.TraceStateString = activityContext.TraceState;
+                // This will act as an internal activity that represents remote Host activity. This cannot be tracked as this is not associate to an ActivitySource. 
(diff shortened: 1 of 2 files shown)


jviau (approved):
LGTM.

Something I realize we are missing with our approach of new'ing up the activity manually: we are missing out on a perf optimization to _not_ do this at all if the user is not listening to activities. How we will do this I am unsure, but it is a behavior with the `ActivitySource` API: if no one is listening to the source, it no-ops on activity creation.

PR #2769, opened 2024-10-10
Adding FunctionsApplicationBuilder

### LINKED ISSUE #2745: Refactor host builder APIs in Worker SDKs
Capturing initial context:
 - Layering: Move the APIs to Worker packages
 - Possible API changes (move away from FunctionsApplicationBuilder?)

brettsam:

resolves #2745 

Note: this is part 1 of decomposing PR #2747 

Adding support for `IHostApplicationBuilder` patterns in the Worker.

### Pull request checklist


src/DotNetWorker/Hosting/BootstrapHostBuilder.cs
```diff
@@ -0,0 +1,170 @@
...
+            }
+
+            if (_defaultsComplete)
+            {
+                // After defaults are run, we want to run these immediately so that they are observable by the imperative code
+                configureDelegate(Context, _builder.Services);
+            }
+            else
+            {
+                _configureServicesActions.Add(configureDelegate ?? throw new ArgumentNullException(nameof(configureDelegate)));
```
liliankasem, line 96: Do we need to have `throw new ArgumentNullException(nameof(configureDelegate)` again when we already check if this is null at the start of the method? It should not be possible for it to be null at this point, no?

src/DotNetWorker/Hosting/BootstrapHostBuilder.cs
```diff
@@ -0,0 +1,170 @@
...
+        public IHost Build()
+        {
+            // Functions configuration will never call this.
+            throw new InvalidOperationException();
+        }
+
+        public IHostBuilder ConfigureContainer<TContainerBuilder>(Action<HostBuilderContext, TContainerBuilder> configureDelegate)
+        {
+            // Functions configuration will never call this.
+            throw new InvalidOperationException();
```
liliankasem, line 111: nit: Do we want to add specific messages to these exceptions in case they are ever called somehow?

src/DotNetWorker/Hosting/BootstrapHostBuilder.cs
```diff
@@ -0,0 +1,170 @@
...
+
+        internal void ApplyServiceProviderFactory()
+        {
+            if (_serviceProviderFactory is null)
+            {
+                return;
+            }
+
+            _builder.ConfigureContainer(_serviceProviderFactory);
+        }
```
liliankasem, line 155: Should we check if ` _builder` is null as well?

src/DotNetWorker/Hosting/BootstrapHostBuilder.cs
```diff
@@ -0,0 +1,170 @@
+﻿using System;
+using System.Collections.Generic;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace Microsoft.Azure.Functions.Worker.Hosting
+{
+    internal class BootstrapHostBuilder : IHostBuilder
```
liliankasem, line 9: Can you add a summary comment/doc here?

src/DotNetWorker/Hosting/BootstrapHostBuilder.cs
brettsam: Nope! Copy/paste error. I've removed it.

src/DotNetWorker/Hosting/BootstrapHostBuilder.cs
brettsam: I don't think that's necessary. These are all only accessible internally so debugging them should be fairly easy with stack trace. And if we do hit them, it'd likely be during dev work.

src/DotNetWorker/Hosting/BootstrapHostBuilder.cs
brettsam: Yup. Also included a link to where this came from originally (the aspnetcore repo).

src/DotNetWorker/Hosting/BootstrapHostBuilder.cs
brettsam: Done.

src/DotNetWorker/Hosting/FunctionsApplicationBuilder.cs
```diff
@@ -0,0 +1,134 @@
...
+        Environment = bootstrapHostBuilder.Context.HostingEnvironment;
+
+        //Host = new ConfigureHostBuilder(bootstrapHostBuilder.Context, Configuration, Services);
+        //WebHost = new ConfigureWebHostBuilder(webHostContext, Configuration, Services);
+
+        return functionsWorkerApplicationBuilder;
+    }
+
+    /// <inheritdoc/>
+    public IFunctionsWorkerApplicationBuilder Use(Func<FunctionExecutionDelegate, FunctionExecutionDelegate> middleware)
```
mattchenderson, line 125: Is an equivalent needed for the other variations like `UseWhen`, etc.? Or will those call through here somehow?

src/DotNetWorker/Hosting/FunctionsApplicationBuilder.cs
```diff
@@ -0,0 +1,134 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Diagnostics.Metrics;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Extensions.Hosting;
+
+/// <summary>
+/// A builder for Azure Functions worker applications.
+/// </summary>
+public class FunctionsApplicationBuilder : IHostApplicationBuilder, IFunctionsWorkerApplicationBuilder
```
mattchenderson, line 16: Am I understanding correctly that one could use the config lambda parameter in 'ConfigureFunctionsWorkerDefaults` or the web variant, to basically get the inner builder here? Previously that was the only way, but now promoting this to this level, I guess we still have that way back into it too? Probably ultimately harmless, but maybe a touch confusing?

src/DotNetWorker/Hosting/FunctionsApplicationBuilder.cs
brettsam: `UseWhen()` is an extension method off of `IFunctionsWorkerApplicationBuilder`, so it works with this too.

src/DotNetWorker/Hosting/FunctionsApplication.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Microsoft.Extensions.Hosting
```
fabiocav, line 1: Asking to make sure it's an intentional choice: do we want this to be in `Microsoft.Extensions.Hosting`?

src/DotNetWorker/Hosting/FunctionsApplicationBuilder.cs
brettsam: There's no way to call `ConfigureFunctionsWorkerDefaults()` off of this. That hangs off of the `IHostBuilder`, which we're not exposing publicly here.

You could do something like:

```csharp
FunctionsApplicationBuilder funcBuilder = FunctionsApplication.CreateBuilder(args);
funcBuilder.Use<CustomMiddleware1>();

var innerBuilder = funcBuilder.Services.AddFunctionsWorkerDefaults();
innerBuilder.Use<CustomMiddleware2>();
```
Why would you? I dunno... :-)

The types are exposed in a variety of ways for composability, but ultimately it shouldn't matter how you call them.

src/DotNetWorker/Hosting/FunctionsApplicationBuilder.cs
```diff
@@ -0,0 +1,134 @@
...
+        _functionsWorkerApplicationBuilder = InitializeHosting(_bootstrapHostBuilder);
+    }
+
+    /// <inheritdoc/>
+    public IDictionary<object, object> Properties => ((IHostApplicationBuilder)_hostApplicationBuilder).Properties;
+
+    IConfigurationManager IHostApplicationBuilder.Configuration => Configuration;
+
+    /// <inheritdoc/>
+    public ConfigurationManager Configuration => _hostApplicationBuilder.Configuration;
```
fabiocav, line 53: `/// <inheritdoc/>` will not apply here since you have an explicit implementation of `IHostApplicationBuilder` right?

src/DotNetWorker/Hosting/FunctionsApplicationBuilder.cs
```diff
@@ -0,0 +1,134 @@
...
+    public IHostEnvironment Environment { get; private set; }
+
+    /// <inheritdoc/>
+    public ILoggingBuilder Logging => _hostApplicationBuilder.Logging;
+
+    /// <inheritdoc/>
+    public IMetricsBuilder Metrics => _hostApplicationBuilder.Metrics;
+
+    /// <inheritdoc />
+    public IServiceCollection Services => _hostApplicationBuilder.Services;
```
fabiocav, line 65: You may also have an issue with `/// <inheritdoc />` here as both interfaces expose this property.

src/DotNetWorker/Hosting/WorkerHostBuilderExtensions.cs
```diff
@@ -169,12 +169,19 @@ public static IHostBuilder ConfigureFunctionsWorkerDefaults(this IHostBuilder bu
...
-                })
-               .UseDefaultServiceProvider((context, options) =>
-                {
-                    options.ValidateScopes = context.HostingEnvironment.IsDevelopment();
+                    static bool ShouldSkipDefaultWorkerMiddleware(IDictionary<object, object> props)
+                    {
+                        return props is not null &&
+                            props.TryGetValue(FunctionsApplicationBuilder.SkipDefaultWorkerMiddlewareKey, out var skipObj) &&
+                            skipObj is bool skip &&
+                            skip;
```
fabiocav, line 177: nit: should be able to update the pattern here to:
``` csharp
skipObj is true
```

Replacing lines 176 and 177.

fabiocav (commented):
Mostly nit comments. Nothing blocking.

src/DotNetWorker/Hosting/BootstrapHostBuilder.cs
```diff
@@ -0,0 +1,180 @@
...
+
+        public BootstrapHostBuilder(HostApplicationBuilder builder)
+        {
+            if (builder is null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            _builder = builder;
+
```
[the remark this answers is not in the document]
fabiocav: Same throughout the rest of this file

src/DotNetWorker/Hosting/FunctionsApplicationBuilder.cs
brettsam: No, b/c you can't do this as it's not an `IHostBuilder`. It's an `IHostApplicationBuilder`.

```charp
builder.ConfigureFunctionsWebApplication(same_as_builder =>
{
    same_as_builder.UseMiddleware<ExceptionHandlingMiddleware>();
});
```

Maybe the confusing part is that I am going to be adding a new `ConfigureFunctionsWebApplication()` (which does not expose the callback) that is an extension off of `IHostApplicationBuilder`?: https://github.com/Azure/azure-functions-dotnet-worker/blob/brettsam/relayer/extensions/Worker.Extensions.Http.AspNetCore/src/FunctionsApplicationBuilderAspNetCoreExtensions.cs

src/DotNetWorker/Hosting/FunctionsApplicationBuilder.cs
mattchenderson: right, and it does indeed seem like it calls through here, so we still get the check for if it's built.  👍

src/DotNetWorker/Hosting/BootstrapHostBuilder.cs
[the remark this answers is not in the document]
brettsam: yeah we can't use that... i will swap to the other format though.

src/DotNetWorker/Hosting/FunctionsApplicationBuilder.cs
mattchenderson: Ah, that explains it. Cool - not exposing the callback removes the potential for this confusion.

src/DotNetWorker/Hosting/FunctionsApplication.cs
brettsam: Ahh good point. AspNet uses `Microsoft.AspNetCore.Builder` as their namespace -- we should probably mirror that.

src/DotNetWorker/Hosting/FunctionsApplicationBuilder.cs
brettsam: Good catch.

src/DotNetWorker/Hosting/FunctionsApplicationBuilder.cs
brettsam: that one seems to work...

liliankasem (approved):

fabiocav (approved):

PR #2772, opened 2024-10-11
Set extension RPC max message length

jviau:

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

Configures max message send and receive length for the gRPC `CallInvoker` in the `Microsoft.Azure.Functions.Worker.Extensions.Rpc` package.

### DIFF
--- a/extensions/Worker.Extensions.Rpc/src/GrpcHttpClientBuilderExtensions.cs
+++ b/extensions/Worker.Extensions.Rpc/src/GrpcHttpClientBuilderExtensions.cs
@@ -31,7 +31,18 @@ namespace Microsoft.Azure.Functions.Worker
 
             ValidateGrpcClient(builder);
             builder.Services.AddOptions<GrpcClientFactoryOptions>(builder.Name)
-                .Configure<IConfiguration>((options, config) => options.Address = config.GetFunctionsHostGrpcUri());
+                .Configure<IConfiguration>((options, config) =>
+                {
+                    options.Address = config.GetFunctionsHostGrpcUri();
+                    if (config.GetFunctionsHostMaxMessageLength() is int length)
+                    {
+                        options.ChannelOptionsActions.Add(o =>
+                        {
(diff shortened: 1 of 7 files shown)


extensions/Worker.Extensions.Rpc/src/GrpcHttpClientBuilderExtensions.cs
```diff
@@ -31,7 +31,18 @@ public static IHttpClientBuilder ConfigureForFunctionsHostGrpc(this IHttpClientB
...
-                .Configure<IConfiguration>((options, config) => options.Address = config.GetFunctionsHostGrpcUri());
+                .Configure<IConfiguration>((options, config) =>
+                {
+                    options.Address = config.GetFunctionsHostGrpcUri();
+                    if (config.GetFunctionsHostMaxMessageLength() is int length)
+                    {
+                        options.ChannelOptionsActions.Add(o =>
+                        {
+                            o.MaxReceiveMessageSize = length;
+                            o.MaxSendMessageSize = length;
```
fabiocav, line 42: Today, this defaults to `null`, and there will be no message limits enforced, correct?

The host will pass `grpcMaxMessageLength` by default, so this change will start setting a limit here, defaulting to `int.MaxValue`, so  ~2GB, which is large, but change in behavior to be aware of here.

jviau: The default is 4MB, not null. This will increase the limit.

liliankasem (approved):

extensions/Worker.Extensions.Rpc/src/GrpcHttpClientBuilderExtensions.cs
fabiocav: @jviau can you double check this? A resolved instance of `GrpcChannelOptions` will have the `MaxSendMessageSize` set to `null` with the OOB setup.

PR #2790, opened 2024-10-17
initial fix of duplicate registrations if AddFunctionsWorkerCore called twice

brettsam:
There will be more to improve here, but first crack at making AddFunctionsWorkerCore work if called multiple times.

### Pull request checklist

### DIFF
--- a/src/DotNetWorker.Core/Hosting/ServiceCollectionExtensions.cs
+++ b/src/DotNetWorker.Core/Hosting/ServiceCollectionExtensions.cs
@@ -27,6 +27,8 @@ namespace Microsoft.Extensions.DependencyInjection
     /// </summary>
     public static class ServiceCollectionExtensions
     {
+        private static bool _defaultConvertersAdded = false;
+
         /// <summary>
         /// Adds the core set of services for the Azure Functions worker.
         /// </summary>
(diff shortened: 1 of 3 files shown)


src/DotNetWorker.Core/Hosting/ServiceCollectionExtensions.cs
```diff
@@ -123,6 +125,13 @@ public static IFunctionsWorkerApplicationBuilder AddFunctionsWorkerCore(this ISe
         /// </summary>
         internal static IServiceCollection AddDefaultInputConvertersToWorkerOptions(this IServiceCollection services)
         {
+            if (_defaultConvertersAdded)
```
fabiocav, line 128: Should we consider adding a marker to the service collection so there isn't a static state tracking this? In some scenarios, that scope might be too broad.

fabiocav (commented):
Since the change to unblock duplicate calls was made against `main` , we should target that branch for this change as well.

src/DotNetWorker.Core/Hosting/ServiceCollectionExtensions.cs
brettsam: changed the approach here to use an `IConfigureOptions` class which lets us use `TryAddEnumerable`

src/DotNetWorker.Core/Hosting/ServiceCollectionExtensions.cs
```diff
@@ -81,7 +81,7 @@ public static IFunctionsWorkerApplicationBuilder AddFunctionsWorkerCore(this ISe
                     }
                 });
 
-            services.AddSingleton<ILoggerProvider, WorkerLoggerProvider>();
+            services.TryAddEnumerable(ServiceDescriptor.Singleton<ILoggerProvider, WorkerLoggerProvider>());
```
mattchenderson, line 84: Non-blocking question just for my own understanding, but why is this preferable to using `TryAddSingleton`?

brettsam: Depends on what the registered service is being used for. For `ILoggerProvider`, there will be many registered so we don't want to use `TryAddSingleton()` as it'd just flat-out skip this if anyone else had registered an `ILoggerProvider`.

Instead, `TryAddEnumerable()` checks if this specific implementation type has already been registered for this service type. Only then will it skip. This allows other types to be registered alongside it, but will prevent this specific duplication.

fabiocav (approved):

Issue #2873, opened 2024-11-17
ServiceBusSessionMessageActions not working with IsBatched=true


Meorir:
### Description

When trying to bind
ServiceBusSessionMessageActions sessionActions

while having IsBatched = true

causes this error:
`Result: Function 'ServiceBusHandler', Invocation id '5dae4abd-65bf-434d-97f9-ad28900a6c2d': An exception was thrown by the invocation.
Exception: Microsoft.Azure.Functions.Worker.FunctionInputConverterException: Error converting 1 input parameters for Function 'ServiceBusHandler': Cannot convert input parameter 'sessionActions' to type 'Microsoft.Azure.Functions.Worker.ServiceBusSessionMessageActions' from type 'System.String'. Error:System.InvalidOperationException: Expecting SessionId within binding data and value was not present. Sessions must be enabled when binding to ServiceBusSessionMessageActions.`

I assume batch processing in session enabled subscription (or queue) should be locked to a single session per invocation due to session locks.

[Function("ServiceBusHandler")]
	public async Task ServiceBusHandler(
			[ServiceBusTrigger("%ServiceBusTopicName%", 
			"%ServiceBusSubscriptionName%", 
			Connection = "ServiceBusConnectionString", 
			IsSessionsEnabled = true, 
			AutoCompleteMessages = false,
			IsBatched = true)]
			ServiceBusReceivedMessage[] messages,
			ServiceBusMessageActions messageActions,
			ServiceBusSessionMessageActions sessionActions
		)

When using with single message everything works fine. 
(except SetSessionStateAsync not accepting null as parameter, and not allowing to clear session state, but that is another bug)

### Steps to reproduce

Create a trigger for session-enabled ASB subscription:

[Function("ServiceBusHandler")]
	public async Task ServiceBusHandler(
			[ServiceBusTrigger("%ServiceBusTopicName%", 
			"%ServiceBusSubscriptionName%", 
			Connection = "ServiceBusConnectionString", 
			IsSessionsEnabled = true, 
			AutoCompleteMessages = false,
			IsBatched = true)]
			ServiceBusReceivedMessage[] messages,
			ServiceBusMessageActions messageActions,
			ServiceBusSessionMessageActions sessionActions
		)

After invocation the mapping throws an error

jviau:
@Meorir please see the issue I duplicated this to - you are experiencing the same problem, source gen is not running for your project.

jviau:
[...]

PR #2889, opened 2024-12-06
ServiceBusSessionMessageActions not working with IsBatched=true for ServiceBusReceivedMessage

aishwaryabh:

resolves [#2873](https://github.com/Azure/azure-functions-dotnet-worker/issues/2873)

When we have `isBatched` set to true and we have multiple `ServiceBusReceivedMessage` instances bound, the property that we receive from the binding data for the `ServiceBusSessionMesasgeActions` type is called `SessionIdArray` instead of `SessionId`. This PR accounts for that use case. I also removed the need for supporting deferred binding for `ServiceBusSessionMessageActions[]` as the SDK does not support this use case in the binding contract in `ServiceBusTriggerBindingStrategy`.

### Pull request checklist

### Additional information

Additional PR information


extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
```diff
@@ -49,7 +54,17 @@ public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
                     throw new InvalidOperationException("Expecting SessionLockedUntil within binding data of session actions and value was not present.");
                 }
 
-                var sessionActionResult = new ServiceBusSessionMessageActions(_settlement, sessionId!.ToString(), sessionLockedUntil.GetDateTimeOffset());
+                // Logic for if isBatched is true, then sessionIdArray will be used to get the sessionId.
+                var sessionIdRepeatedFieldArray = sessionIdArray as RepeatedField<string>;
+
+                if (foundSessionIdArray && (sessionIdRepeatedFieldArray == null || sessionIdRepeatedFieldArray.Count != 1))
```
liliankasem, line 60:
```suggestion
                if (foundSessionIdArray && (sessionIdRepeatedFieldArray is null || sessionIdRepeatedFieldArray.Count == 0))
```

liliankasem (approved):

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
fabiocav: @liliankasem did you mean to approve? This feedback indicates there's a bug with the current logic.

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
```diff
@@ -7,6 +7,11 @@
 using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
 using Microsoft.Azure.ServiceBus.Grpc;
 using System.Text.Json;
+using Microsoft.Azure.Functions.Worker.Core;
+using Microsoft.Azure.Functions.Worker.Extensions.ServiceBus;
+using Microsoft.Azure.Functions.Worker.Extensions;
+using System.Text;
+using Google.Protobuf.Collections;
```
fabiocav, line 14: Ideally, we shouldn't take a dependency on the protobuf types here.

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
liliankasem: Meant to just submit, updated my review

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
```diff
@@ -49,7 +54,17 @@ public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
                     throw new InvalidOperationException("Expecting SessionLockedUntil within binding data of session actions and value was not present.");
                 }
 
-                var sessionActionResult = new ServiceBusSessionMessageActions(_settlement, sessionId!.ToString(), sessionLockedUntil.GetDateTimeOffset());
+                // Logic for if isBatched is true, then sessionIdArray will be used to get the sessionId.
+                var sessionIdRepeatedFieldArray = sessionIdArray as RepeatedField<string>;
```
jviau, line 58: `RepeatedField<T>` implements `IList<T>`, so I recommend casting to that as it will be less dependent on precise payload (and more testable!)

```suggestion
                var sessionIdRepeatedFieldArray = sessionIdArray as IList<string>;
```

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
```diff
@@ -49,7 +54,17 @@ public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
...
+
+                if (foundSessionIdArray && (sessionIdRepeatedFieldArray == null || sessionIdRepeatedFieldArray.Count != 1))
+                {
+                     throw new InvalidOperationException($"Expecting batched SessionId within binding data and value was not present. Sessions must be enabled when binding to {nameof(ServiceBusSessionMessageActions)}.");
+                }
+
+                // If sessionIdRepeatedFieldArray has a value, it will only have one value within the array.
```
jviau, line 65: I don't think this is accurate. It looks like [ServiceBus ext](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Triggers/ServiceBusTriggerBindingStrategy.cs#L159) will send a value per message in the batch. In which case I think we need to support binding to an array.

jviau (changes requested):

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
liliankasem: did not let me change it without a new comment so this is that ☠️

liliankasem (changes requested):

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
aishwaryabh: Hmm I had asked Josh about this before and he said that

>  I think we technically allowing binding to an array of sessionIds but it should only ever have one value.

Let me confirm with him

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
aishwaryabh: changed to address feedback

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
jviau: I'll need some clarification for sure. I am confused because `sessionIdArray` will definitely have multiple values. It doesn't seem right for this binding to only consider the first element in that array.

aishwaryabh: Ok so I talked to Josh and he said that:
> Messages and sessions are different concepts. There could be multiple messages but each message would belong to the same session.

So even though we can technically have multiple session ids[ here](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Triggers/ServiceBusTriggerBindingStrategy.cs#L176) in the `ServiceBusTriggerBindingStrategy`, we are guaranteed that the value will always be the same. There is no scenario in which we would have different session IDs for the multiple messages.

JoshLove-msft: This is really an implementation detail that could theoretically change. Is there a reason that this is important? I would expect the behavior on the isolated worker to match the webjobs extension where you can still bind to an array of session IDs (although it will always have the same value in the current implementation).

JoshLove-msft: Looked at the existing implementation that takes a single sessionID in the internal ctor for ServiceBusSessionMessageActions. I think grabbing the first element of the array is fine here given that this is also an internal implementation detail that will need to stay in sync with the webjobs extensions implementation. We should just document in a comment that when using IsBatched=true, all messages will belong to the same session so we can just grab the first element.

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
```diff
@@ -30,7 +30,8 @@ public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
             try
             {
                 var foundSessionId = context.FunctionContext.BindingContext.BindingData.TryGetValue("SessionId", out object? sessionId);
-                if (!foundSessionId)
+                var foundSessionIdArray = context.FunctionContext.BindingContext.BindingData.TryGetValue("SessionIdArray", out object? sessionIdArray);
```
JoshLove-msft, line 33: If we find SessionId, do we need to still look up SessionIdArray?

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
aishwaryabh: I updated the comment!

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
JoshLove-msft: Also, do we need to maintain these two stack bool variables? Couldn't we just resolve down to the sessionId here?

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
```diff
@@ -49,7 +50,19 @@ public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
...
-                var sessionActionResult = new ServiceBusSessionMessageActions(_settlement, sessionId!.ToString(), sessionLockedUntil.GetDateTimeOffset());
+                // Logic for if isBatched is true, then sessionIdArray will be used to get the sessionId.
+                var sessionIdRepeatedFieldArray = sessionIdArray as IList<string>;
+
+                if (foundSessionIdArray && (sessionIdRepeatedFieldArray is null || sessionIdRepeatedFieldArray.Count == 0))
+                {
+                     throw new InvalidOperationException($"Expecting batched SessionId within binding data and value was not present. Sessions must be enabled when binding to {nameof(ServiceBusSessionMessageActions)}.");
+                }
+
+                // If sessionIdRepeatedFieldArray has a value (isBatched = true), we can just parse the first sessionId from the array, as all the values are guranteed to be the same.
```
JoshLove-msft, line 61:
```suggestion
                // If sessionIdRepeatedFieldArray has a value (isBatched = true), we can just parse the first sessionId from the array, as all the values are guaranteed to be the same.
```

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
```diff
@@ -49,7 +50,19 @@ public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
...
+                // This is becuase there can be multiple messages but each message would belong to the same session.
```
JoshLove-msft, line 62:
```suggestion
                // This is because there can be multiple messages, but each message would still belong to the same session.
```

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
jviau: Thank you for the clarification @aishwaryabh and @JoshLove-msft. This sounds good to me. Documenting that all values in `sessionIdArray` will always be identical will be helpful. Maybe we can even stop using that property altogether and send only `sessionId` from the WebJobs extension?

I also want to concur that if we go with this implementation which asserts only ever 1 session ID per batch, it will be a breaking change to have multiple session IDs in the future. So we need to be confident that we have no need for that to change, and if it is needed it will be a challenge to change it.

JoshLove-msft: > Maybe we can even stop using that property altogether and send only `sessionId` from the WebJobs extension?

I think this would be a breaking change, wouldn't it? Customers may be binding to sessionId as an array.

> I also want to concur that if we go with this implementation which asserts only ever 1 session ID per batch, it will be a breaking change to have multiple session IDs in the future.

I don't think it will happen given the existing public surface area and the implementation details of the Service Bus library, but if we did make this change, we would likely need to EditorBrowsableNever some properties.

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
aishwaryabh: makes sense! I created a method and just return the sessionId

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
```diff
@@ -80,5 +55,32 @@ public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
                 return new ValueTask<ConversionResult>(ConversionResult.Failed(exception));
             }
         }
+
+        private object? ParseSessionIdFromBindingData(ConverterContext context)
```
JoshLove-msft, line 59: Should this return a string?

aishwaryabh: yes updated!

test/Worker.Extensions.Tests/ServiceBus/ServiceBusSessionMessageConverterTests.cs
```diff
@@ -0,0 +1,133 @@
...
+
+            public override IInvocationFeatures Features => throw new NotImplementedException();
+        }
+
+        [Fact]
+        public async Task ConvertAsync_ReturnsSuccess()
+        {
+            var data = "{\"SessionLockedUntil\":\"2024-12-05T21:10:36.1193094+00:00\"}";
+            
+            var jsonObject = new
```
JoshLove-msft, line 58: Where is this used?

aishwaryabh: ah I forgot to remove this. done!

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
jviau: > Customers may be binding to sessionId as an array.

That was never correctly implemented, so won't be removing any functionality from the customer.

JoshLove-msft: Yes, they can do this today so we wouldn't want to break that scenario.

jviau (approved):

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
```diff
@@ -58,5 +54,32 @@ public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
...
+                }
+            }
+
+            if (sessionId == null)
+            {
+                throw new InvalidOperationException(
+                    $"Expecting SessionId or SessionIdArray within binding data and value was not present. Sessions must be enabled when binding to {nameof(ServiceBusSessionMessageActions)}.");
+            }
+
+            return sessionId.ToString();
```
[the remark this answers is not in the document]
JoshLove-msft: Yes, it is a string, and agreed on the preference for a cast.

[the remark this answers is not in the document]
aishwaryabh: added cast!

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
```diff
@@ -58,5 +54,32 @@ public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
                 return new ValueTask<ConversionResult>(ConversionResult.Failed(exception));
             }
         }
+
+        private string ParseSessionIdFromBindingData(ConverterContext context)
+        {
+            // Try to resolve sessionId directly
+            var bindingData = context.FunctionContext.BindingContext.BindingData;
+            bindingData.TryGetValue("SessionId", out object? sessionId);
```
liliankasem, line 62: Can we make `SessionId` and `SessionIdArray` constants?

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
```diff
@@ -58,5 +54,32 @@ public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
...
+
+            // If sessionId is not found and sessionIdRepeatedFieldArray has a value (isBatched = true), we can just parse the first sessionId from the array, as all the values are guaranteed to be the same.
+            // This is because there can be multiple messages but each message would belong to the same session.
+            // Note if web jobs extensions ever adds support for multiple sessions in a single batch, this logic will need to be updated.
+            if (sessionId == null && bindingData.TryGetValue("SessionIdArray", out object? sessionIdArray))
```
liliankasem, line 67: nit: you're using `is` on line 70 so keeping things consistent
```suggestion
            if (sessionId is null && bindingData.TryGetValue("SessionIdArray", out object? sessionIdArray))
```

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
```diff
@@ -58,5 +54,32 @@ public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
...
+            {
+                var sessionIdRepeatedArray = sessionIdArray as IList<string>;
+                if (sessionIdRepeatedArray is not null && sessionIdRepeatedArray.Count > 0)
+                {
+                    sessionId = sessionIdRepeatedArray[0]; // Use the first sessionId in the array
+                }
+            }
+
+            if (sessionId == null)
```
liliankasem, line 76: Same here

test/Worker.Extensions.Tests/ServiceBus/ServiceBusSessionMessageConverterTests.cs
```diff
@@ -0,0 +1,121 @@
...
+        public async Task ConvertAsync_Batch_ForReceivedMessage_ReturnsSuccess()
+        {
+            var data = "{\"SessionLockedUntil\":\"2024-12-05T21:10:36.1193094+00:00\"}";
+
+            IList<string> repeatedField = new List<string> { "test" };
+
+            var bindingDataDictionary = new Dictionary<string, object>
+            {
+                { "SessionIdArray", repeatedField },
+                { "SessionActions", JsonSerializer.Serialize(new
```
liliankasem, line 86: I don't actually know how these sessions work in SB, but for a batch scenario like this, can we have multiple actions? Id like for us to test batch properly with more than just `SessionIdArray` if that's a thing?

test/Worker.Extensions.Tests/ServiceBus/ServiceBusSessionMessageConverterTests.cs
```diff
@@ -0,0 +1,121 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
liliankasem, line 1: Can you also add E2E tests? There is a service bus [emulator](https://learn.microsoft.com/en-us/azure/service-bus-messaging/overview-emulator) now (I think we were waiting for this to exist).

You can create a service bus function here: https://github.com/Azure/azure-functions-dotnet-worker/tree/main/test/E2ETests/E2EApps/E2EApp

Then write code to test it here: https://github.com/Azure/azure-functions-dotnet-worker/tree/main/test/E2ETests/E2ETests

And finally, this script will need to update to setup the emulator: https://github.com/Azure/azure-functions-dotnet-worker/blob/main/setup-e2e-tests.ps1

This is a big ask so I am okay with this being setup in a new PR

jviau: > This is a big ask so I am okay with this being setup in a new PR

I think we should have a separate issue for adding E2E tests here, it is definitely a large work item.

test/Worker.Extensions.Tests/ServiceBus/ServiceBusSessionMessageConverterTests.cs
jviau: There are two action types: `ServiceBusMessageActions` and `ServiceBusSessionMessageActions`.

In a batch scenario the former is an array - one for each incoming message. In the later it is only a single element. Session + Batch will be for a bath of messages _all from the same session_ - so a single session actions works. It is an implementation oddity that the binding data is an array of session IDs, as it will just be an array of the same repeated session ID.

aishwaryabh: Yeah we can only test with `SessionIdArray` for the batched scenario since that's how the binding data is formatted today

test/Worker.Extensions.Tests/ServiceBus/ServiceBusSessionMessageConverterTests.cs
aishwaryabh: I can open an issue for this to follow up on!

aishwaryabh: https://github.com/Azure/azure-functions-dotnet-worker/issues/2904

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActionsConverter.cs
aishwaryabh: done! made the other strings consts as well

liliankasem (approved):

PR #2894, opened 2024-12-10
Update ContextReference & DefaultFromBodyConversionFeature to remove invocation cancellation token usage

### LINKED ISSUE #2417: ObjectDisposedException: Request has finished and HttpContext disposed. in Extensions.Http.AspNetCore
### Description

When using the Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore version 1.2.1 we sometimes receive an ObjectDisposedException resulting in a System.InvalidOperationException: An attempt was made to transition a task to a final state when it had already completed.

Here is the stack trace of the ObjectDisposedException

> Result: Failure Exception: System.ObjectDisposedException: Request has finished and HttpContext disposed. Object name: 'HttpContext'. at Microsoft.AspNetCore.Http.DefaultHttpContext.ThrowContextDisposed() at Microsoft.AspNetCore.Routing.RoutingHttpContextExtensions.GetRouteData(HttpContext httpContext) at Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore.FunctionsHttpProxyingMiddleware.Invoke(FunctionContext context, FunctionExecutionDelegate next) in D:\a_work\1\s\extensions\Worker.Extensions.Http.AspNetCore\src\FunctionsMiddleware\FunctionsHttpProxyingMiddleware.cs:line 54 at Microsoft.Azure.Functions.Worker.FunctionsApplication.InvokeFunctionAsync(FunctionContext context) in D:\a_work\1\s\src\DotNetWorker.Core\FunctionsApplication.cs:line 77 at Microsoft.Azure.Functions.Worker.Handlers.InvocationHandler.InvokeAsync(InvocationRequest request) in D:\a_work\1\s\src\DotNetWorker.Grpc\Handlers\InvocationHandler.cs:line 88 Stack: at Microsoft.AspNetCore.Http.DefaultHttpContext.ThrowContextDisposed() at Microsoft.AspNetCore.Routing.RoutingHttpContextExtensions.GetRouteData(HttpContext httpContext) at Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore.FunctionsHttpProxyingMiddleware.Invoke(FunctionContext context, FunctionExecutionDelegate next) in D:\a_work\1\s\extensions\Worker.Extensions.Http.AspNetCore\src\FunctionsMiddleware\FunctionsHttpProxyingMiddleware.cs:line 54 at Microsoft.Azure.Functions.Worker.FunctionsApplication.InvokeFunctionAsync(FunctionContext context) in D:\a_work\1\s\src\DotNetWorker.Core\FunctionsApplication.cs:line 77 at Microsoft.Azure.Functions.Worker.Handlers.InvocationHandler.InvokeAsync(InvocationRequest request) in D:\a_work\1\s\src\DotNetWorker.Grpc\Handlers\InvocationHandler.cs:line 88

This file hase already been changed, so line numbers are wrong. https://github.com/Azure/azure-functions-dotnet-worker/pull/2322/files/6149e16e951188e97bf7fcff6918b45ae233bde4#diff-dc49550c24c24c88566ae78776b0400d415dfc8d1a5d623571903671a953e715

The GetRouteData call in now on line 72 instead of 54.

We don't have a reproducible sample as this exception did only happens seemingly random. 

As a quick fix we removed AspNetCore package from our function.

### Steps to reproduce

Only saw it on a deployed linux azure function with a single instance.

liliankasem:

resolves #2417

### Pull request checklist

  * [x] Otherwise: Backport tracked by issue/PR #issue_or_pr

### Additional information

1. The `ContextReference` to remove and reference or use of the invocation cancellation token
2. The `DefaultFromBodyConversionFeature` to remove and reference or use of the invocation cancellation token

### Issue

Currently, ContextReference registers the invocation’s cancellation token to multiple internal tasks and value sources. One of these, _functionCompletionTask, is responsible for disposing of the HttpContext. When the invocation is canceled, this results in the HttpContext being disposed prematurely, leading to an ObjectDisposedException when the function attempts to complete the HTTP response.

We have reason to believe that canceling _functionCompletionTask propagates cancellation through the ASP.NET Core pipeline, causing the HttpContext to be disposed earlier than expected. Since HttpContext is typically disposed of naturally when the request completes, this explicit disposal is unnecessary and introduces unwanted side effects.

### Proposed Change

This PR removes the cancellation token registration from ContextReference, ensuring that:

- Function executions are not forcefully canceled when the invocation is aborted.
- HttpContext remains available until the invocation naturally completes.
- Cancellation behavior aligns with ASP.NET Core, where execution continues unless explicitly handled by user code.

By following the design principle that framework code should propagate cancellation but never react to it, this change ensures that handling cancellations remains the responsibility of customer code while preventing unintended HttpContext disposal.

### Impact

With this change, function executions will continue running even if the client disconnects, which is the default behavior in ASP.NET Core. Users who want to cancel function execution explicitly can still do so by checking the CancellationToken within their function logic.

#### _functionStartTask and _functionCompletionTask

These tasks will no longer be canceled automatically when the cancellation token is triggered. This means that if a client disconnects or the invocation is canceled, these tasks will not be marked as canceled, and they will continue to track the function’s start and completion until the function completes. I believe this is behaviour we want.

#### _httpContextValueSource and _functionContextValueSource

Removing the cancellation token registration will prevent these value sources from being canceled automatically when the cancellation token is triggered. This means that HttpContext and FunctionContext will not be disposed of prematurely, allowing them to remain available until the function invocation completes.  I believe this is behaviour we want.

#### This change is beneficial because it treats canceled invocations just like regular invocations, without any special handling. This gives customers the flexibility to manage cancellation in a way that best fits their needs.


sdk/Sdk.Generators/FunctionExecutor/FunctionExecutorGenerator.Parser.cs
```diff
@@ -28,7 +28,7 @@ internal ICollection<ExecutableFunction> GetFunctions(IEnumerable<IMethodSymbol>
 
                 foreach (IMethodSymbol method in methods.Where(m=>m.DeclaredAccessibility == Accessibility.Public))
                 {
-                    _context.CancellationToken.ThrowIfCancellationRequested();
+                    _context.CancellationToken.ThrowIfCancellationRequested();  // Why is this using the invocations CancellationToken?
```
jviau, line 31: This is the source generator cancellation token, not functions invocation (ie: cancel source generation itself, not cancellation consumption in the generated code). This usage is good.

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Emitter.cs
```diff
@@ -22,7 +22,7 @@ internal sealed class Emitter
 
             public static string Emit(GeneratorExecutionContext context, IReadOnlyList<GeneratorFunctionMetadata> funcMetadata, bool includeAutoRegistrationCode)
             {
-                string functionMetadataInfo = AddFunctionMetadataInfo(funcMetadata, context.CancellationToken);
+                string functionMetadataInfo = AddFunctionMetadataInfo(funcMetadata, context.CancellationToken);  // Why is this using the invocations CancellationToken?
```
jviau, line 25: Same here - source gen cancellation, not functions invocation.

sdk/Sdk.Generators/FunctionMetadataProviderGenerator/FunctionMetadataProviderGenerator.Parser.cs
```diff
@@ -34,7 +34,7 @@ public Parser(GeneratorExecutionContext context)
 
             private Compilation Compilation => _context.Compilation;
 
-            private CancellationToken CancellationToken => _context.CancellationToken;
+            private CancellationToken CancellationToken => _context.CancellationToken; // Why is this using the invocations CancellationToken?
```
jviau, line 37: Same here - source gen cancellation, not functions invocation.

sdk/Sdk.Analyzers/BindingTypeCodeRefactoringProvider.cs
```diff
@@ -21,7 +21,7 @@ public sealed class BindingTypeCodeRefactoringProvider : CodeRefactoringProvider
     {
         public override async Task ComputeRefactoringsAsync(CodeRefactoringContext context)
         {
-            SyntaxNode root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+            SyntaxNode root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);  // Why is this using the invocations CancellationToken?
```
jviau, line 24: This is cancellation of running analyzers. This usage is good.

extensions/Worker.Extensions.Http/src/DefaultFromBodyConversionFeature.cs
```diff
@@ -56,19 +56,19 @@ internal class DefaultFromBodyConversionFeature : IFromBodyConversionFeature
             }
             else if (targetType == typeof(byte[]))
             {
-                result = ReadBytes(requestData, context.CancellationToken);
+                result = ReadBytes(requestData, context.CancellationToken);  // Why is this using the invocations CancellationToken?
```
jviau, line 59: This usage is questionable. We shouldn't be short circuiting a customer's function invocation during binding when it is cancelled. Glancing over [MVC's input formatting implementation](https://github.com/dotnet/aspnetcore/blob/3e08b597894891f05556dc0bee23da53cb1b3e11/src/Mvc/Mvc.Core/src/Formatters/SystemTextJsonInputFormatter.cs#L66), I don't see them using CancellationToken even when it is available in APIs that support it. They _could_ pass the CT (available in `InputFormatterContext`), but they don't.

I think this reinforces that we shouldn't be using CT here either.

extensions/Worker.Extensions.Http.AspNetCore/src/Coordinator/ContextReference.cs
```diff
@@ -59,7 +62,7 @@ internal void CompleteFunction()
 
             if (_httpContextValueSource.Task.IsCompleted)
             {
-                if (_httpContextValueSource.Task.IsCanceled || _token.IsCancellationRequested)
+                if (_httpContextValueSource.Task.IsCanceled || _invocationToken.IsCancellationRequested || _token.IsCancellationRequested)
```
jviau, line 65: No piece of code will ever cancel `_token`, so it could be removed. My other comment also touches on how `CancellationToken` may not exactly be the right type.

extensions/Worker.Extensions.Http.AspNetCore/src/Coordinator/ContextReference.cs
```diff
@@ -32,10 +33,12 @@ public ContextReference(string invocationId)
 
         public TaskCompletionSource<FunctionContext> FunctionContextValueSource { get => _functionContextValueSource; set => _functionContextValueSource = value; }
 
-        internal void SetCancellationToken(CancellationToken token)
+        internal void SetCancellationToken(CancellationToken contextToken, CancellationToken invocationToken)
```
jviau, line 36: Wonder if we can just use a `Task` instead of `CancellationToken` for tracking the invocation, considering we only care about completion.

sdk/Sdk.Generators/FunctionExecutor/FunctionExecutorGenerator.Parser.cs
liliankasem: I should have double checked the context of my blind search xD

extensions/Worker.Extensions.Http/src/DefaultFromBodyConversionFeature.cs
liliankasem: Yup I was wondering the same thing, if we're going by the principal that we only propagate, then it doesn't make sense for us to act here, especially with this being the only scenario where we do use the CT to cancel an action we are taking (inconsistent)

extensions/Worker.Extensions.Http.AspNetCore/src/Coordinator/ContextReference.cs
```diff
@@ -76,10 +64,10 @@ internal void CompleteFunction()
 
         public void Dispose()
         {
-            if (_tokenRegistration != default)
-            {
-                _tokenRegistration.Dispose();
-            }
+            _functionStartTask?.TrySetCanceled();
```
fabiocav, line 67: We should be able to remove all of this.

extensions/Worker.Extensions.Http.AspNetCore/src/Coordinator/ContextReference.cs
```diff
@@ -59,9 +46,10 @@ internal void CompleteFunction()
 
             if (_httpContextValueSource.Task.IsCompleted)
             {
-                if (_httpContextValueSource.Task.IsCanceled || _token.IsCancellationRequested)
+                if (_httpContextValueSource.Task.IsCanceled || _invocationCancellationToken.IsCancellationRequested)
```
fabiocav, line 49: We can remove `_httpContextValueSource.Task.IsCanceled`, as we won't be setting this to canceled anywhere.

extensions/Worker.Extensions.Http.AspNetCore/src/Coordinator/DefaultHttpCoordinator.cs
```diff
@@ -43,8 +44,7 @@ public async Task<FunctionContext> SetHttpContextAsync(string invocationId, Http
 
         public async Task<HttpContext> SetFunctionContextAsync(string invocationId, FunctionContext context)
         {
-            var contextRef = _contextReferenceList.GetOrAdd(invocationId, static id => new ContextReference(id));
-            contextRef.SetCancellationToken(context.CancellationToken);
+            var contextRef = _contextReferenceList.GetOrAdd(invocationId, id => new ContextReference(id));
```
jviau, line 47: Why was the `static` modifier removed?

extensions/Worker.Extensions.Http/src/DefaultFromBodyConversionFeature.cs
```diff
@@ -56,19 +56,19 @@ internal class DefaultFromBodyConversionFeature : IFromBodyConversionFeature
...
+                Memory<byte> bytes = ReadBytes(requestData);
                 result = bytes;
             }
             else if (HasJsonContentType(requestData))
             {
                 ObjectSerializer serializer = requestData.FunctionContext.InstanceServices.GetService<IOptions<WorkerOptions>>()?.Value?.Serializer
                  ?? throw new InvalidOperationException("A serializer is not configured for the worker.");
 
-                result = serializer.Deserialize(requestData.Body, targetType, context.CancellationToken);
+                result = serializer.Deserialize(requestData.Body, targetType, CancellationToken.None);
```
jviau, line 71: Do you want to change to `DeserializeAsync` in this PR, or keep that change separate?

liliankasem: I'd like to keep them separate PRs

extensions/Worker.Extensions.Http.AspNetCore/src/Coordinator/DefaultHttpCoordinator.cs
liliankasem: Leftovers from a different change that I forgot to undo, thanks for flagging!

jviau (approved):
Approving. If you have any formal explanation for why removing that token registration in `ContextReference` is a good change that would be helpful.

I know I suggested that change 😆 but it is still a significant behavior change, wondering what the impact of it really is.

liliankasem:
> Approving. If you have any formal explanation for why removing that token registration in `ContextReference` is a good change that would be helpful.
> 
> I know I suggested that change 😆 but it is still a significant behavior change, wondering what the impact of it really is.

I may have something in the design doc! I'll fill in any blanks and can update the PR desc with more info

test/E2ETests/E2ETests/AspNetCore/CancellationEndToEndTests.cs
```diff
@@ -0,0 +1,67 @@
...
+            try
+            {
+                var task = HttpHelpers.InvokeHttpTrigger(functionName, cancellationToken: cts.Token);
+
+                await Task.Delay(3000);
+                cts.Cancel();
+
+                var response = await task;
+            }
+            catch (Exception)
```
brettsam, line 43: Does this test always expect an exception? If so... use `Assert.Throws()` instead (or store the exception and validate it outside the `catch`) so that if it happens to *not* throw, the test will fail. As it is, if an exception is not thrown, the test will pass and we wouldn't notice.

test/E2ETests/E2ETests/AspNetCore/CancellationEndToEndTests.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            _fixture = fixture;
+            _fixture.TestLogs.UseTestLogger(testOutputHelper);
+        }
+
+        [IgnoreOnNetFxTestRunTheory]
+        [InlineData("HttpWithCancellationTokenNotUsed", "Work completed.", "Succeeded")]
+        [InlineData("HttpWithCancellationTokenIgnored", "TaskCanceledException: A task was canceled", "Failed")]
+        [InlineData("HttpWithCancellationTokenHandled", "Request was cancelled.", "Succeeded")]
+        public async Task HttpTriggerFunctions_WithCancellationToken_BehaveAsExpected(string functionName, string expectedMessage, string invocationResult)
```
brettsam, line 30: Were you able to run this with your changes backed out? If so, which scenario behaved differently?

test/E2ETests/E2ETests/AspNetCore/CancellationEndToEndTests.cs
```diff
@@ -0,0 +1,67 @@
...
+                cts.Cancel();
+
+                var response = await task;
+            }
+            catch (Exception)
+            {
+                IEnumerable<string> logs = null;
+                await TestUtility.RetryAsync(() =>
+                {
+                    logs = _fixture.TestLogs.CoreToolsLogs.Where(p => p.Contains($"Executed 'Functions.{functionName}'"));
```
brettsam, line 48: is this b/c the logs don't necessarily show up immediately?

test/E2ETests/E2ETests/AspNetCore/CancellationEndToEndTests.cs
liliankasem: Ah that is a great point! Yes, we're always expecting an exception here because we are cancelling the CTS when we invoke the function to mimic RequestAborted -> `await httpClient.SendAsync(request, cancellationToken);` - this will always throw. But it's not really behaviour we need to "test" as this is coming from our helper and not out function/worker code. But agree that we need to assert it throws to not get a false positive

test/E2ETests/E2ETests/AspNetCore/CancellationEndToEndTests.cs
liliankasem: Yeah, the function has delays and can take up to 5 seconds to run (giving us plenty of time to cancel). I don't want to do a Task.Delay because that is flakey, instead we will just wait for the func invocation to complete.

test/E2ETests/E2ETests/AspNetCore/CancellationEndToEndTests.cs
liliankasem: I have not, but I think it's a good idea to try that out!

tools/devpack.ps1
```diff
@@ -38,42 +47,44 @@ if (!(Test-Path $localPack))
   New-Item -Path $localPack -ItemType directory | Out-Null
 }
 Write-Host
-Write-Host "---Updating project with local SDK pack---"
+Write-Host "---Updating projects with local SDK pack---"
 Write-Host "Packing Core .NET Worker projects to $localPack"
-& "dotnet" "pack" $sdkProject "-p:PackageOutputPath=$localPack" "-nologo" "-p:BuildNumber=$buildNumber" $AdditionalPackArgs
+& "dotnet" "pack" $sdkProject "-p:PackageOutputPath=$localPack" "-nologo" "-p:Version=2.0.1" "-p:VersionSuffix=$buildNumber" $AdditionalPackArgs
```
brettsam, line 52: is this version meant to always be hardcoded?

test/E2ETests/E2EApps/E2EAspNetCoreApp/NuGet.Config
```diff
@@ -0,0 +1,7 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
+    <add key="AzureFunctionsTempStaging" value="https://azfunc.pkgs.visualstudio.com/e6a70c92-4128-439f-8012-382fe78d6396/_packaging/AzureFunctionsTempStaging/nuget/v3/index.json" />
+  </packageSources>
+</configuration>
```
brettsam, line 7: do we need this nuget config? it should inherit from parent directories... (unless this is adding something new)

liliankasem: The "setup-e2e-tests" script adds a local nuget repo with the latest built worker/sdk so I believe this file is needed.

tools/devpack.ps1
liliankasem: No, I don't intend for this to be permenant. It's just a funky state where aspnetcore bits are all on 2.0.0. Then the test script builds 2.0.0-local20250125 which is considered a "downgrade" from 2.0.0 so the build and pipeline fail. I should be releasing 2.1.0 after this PR and my plan is to switch back to what we had before `"-p:BuildNumber=$buildNumber"` which should continue to work going forward. I just didn't feel like spending time on version # parsing was worth the time for something like this.

brettsam (approved):

PR #2912, opened 2025-01-04
Added support in ServiceBusSessionMessageActions for null in GetSessionState and SetSessionState

### LINKED ISSUE #2548: Add support to clear Service Bus session state
### Description

Service Bus session state support was added in #2102, but clearing the state is not yet supported.

As mentioned there as well, the Azure Service Bus SDK and the previous functions in-process SDK both support using `SetSessionStateAsync(null)` (or equivalent). The key is using `null` to clear the state, which is not possible with the current isolated SDK.

Can this be added to the isolated SDK?
There's no full support for session state management until clearing is possible (there will be an ever increasing amount of dangling session states). Workarounds to try setting the state via the Azure Service Bus client SDK models itself is also not possible, due to the state locking mechanism inside the client.

Meorir:
Added support for ServiceBusSessionMessageActions:
* SetSessionState(null)
* and returning null from GetSessionState

resolves #2548 

### Pull request checklist

### DIFF
--- a/extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActions.cs
+++ b/extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActions.cs
@@ -43,7 +43,7 @@ namespace Microsoft.Azure.Functions.Worker
         public virtual DateTimeOffset SessionLockedUntil { get; protected set; }
 
         ///<inheritdoc cref="ServiceBusReceiver.CompleteMessageAsync(ServiceBusReceivedMessage, CancellationToken)"/>
-        public virtual async Task<BinaryData> GetSessionStateAsync(
+        public virtual async Task<BinaryData?> GetSessionStateAsync(
             CancellationToken cancellationToken = default)
         {
             var request = new GetSessionStateRequest()
@@ -52,19 +52,25 @@ namespace Microsoft.Azure.Functions.Worker
             };
 
             GetSessionStateResponse result = await _settlement.GetSessionStateAsync(request, cancellationToken: cancellationToken);
-            BinaryData binaryData = new BinaryData(result.SessionState.Memory);
-            return binaryData;
+
+            if (result.SessionState is null || result.SessionState.IsEmpty)
+            {
+                return null;
+            }
+
+            return new BinaryData(result.SessionState.Memory);
         }
 
         ///<inheritdoc cref="ServiceBusReceiver.CompleteMessageAsync(ServiceBusReceivedMessage, CancellationToken)"/>
         public virtual async Task SetSessionStateAsync(
-            BinaryData sessionState,
+            BinaryData? sessionState,
             CancellationToken cancellationToken = default)
+
         {
             var request = new SetSessionStateRequest()
             {
                 SessionId = _sessionId,
-                SessionState = ByteString.CopyFrom(sessionState.ToMemory().Span),
+                SessionState = sessionState is null ? ByteString.Empty : ByteString.CopyFrom(sessionState.ToMemory().Span),
             };
 
             await _settlement.SetSessionStateAsync(request, cancellationToken: cancellationToken);
--- a/extensions/Worker.Extensions.ServiceBus/release_notes.md
+++ b/extensions/Worker.Extensions.ServiceBus/release_notes.md
@@ -4,6 +4,6 @@
 - My change description (#PR/#issue)
 -->
 
-### Microsoft.Azure.Functions.Worker.Extensions.ServiceBus <version>
+### Microsoft.Azure.Functions.Worker.Extensions.ServiceBus 5.24.0
 
-- <entry>
+- Added 'null' support in SetSessionState and GetSessionState methods (#2548)
--- a/test/extensions/Worker.Extensions.Tests/ServiceBus/ServiceBusSessionMessageActionsTests.cs
+++ b/test/extensions/Worker.Extensions.Tests/ServiceBus/ServiceBusSessionMessageActionsTests.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
@@ -10,6 +10,7 @@ using Google.Protobuf;
 using Google.Protobuf.WellKnownTypes;
 using Grpc.Core;
 using Microsoft.Azure.ServiceBus.Grpc;
+using Moq;
 using Xunit;
 
 namespace Microsoft.Azure.Functions.Worker.Extensions.Tests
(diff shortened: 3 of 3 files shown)


Meorir:

test/Worker.Extensions.Tests/ServiceBus/ServiceBusSessionMessageActionsTests.cs
```diff
@@ -49,6 +49,24 @@ public async Task CanRenewSessionLock()
             await messageActions.RenewSessionLockAsync();
         }
 
+        [Fact]
+        public async Task CanSetAndGetSessionStateWithNull()
+        {
+
```
jviau, line 55: nit: remove empty newline

test/Worker.Extensions.Tests/ServiceBus/ServiceBusSessionMessageActionsTests.cs
```diff
@@ -49,6 +49,24 @@ public async Task CanRenewSessionLock()
...
+            string testString = "TestString";
+            var client = new SessionMockSettlementClient("test", ByteString.CopyFromUtf8(testString));
```
jviau, line 57: Is it possible to use a Moq mock instead of manual mock here?

test/Worker.Extensions.Tests/ServiceBus/ServiceBusSessionMessageActionsTests.cs
```diff
@@ -49,6 +49,24 @@ public async Task CanRenewSessionLock()
...
+            var message = ServiceBusModelFactory.ServiceBusReceivedMessage(lockTokenGuid: Guid.NewGuid(), sessionId: "test");
+            var messageActions = new ServiceBusSessionMessageActions(client, message.SessionId, message.LockedUntil);
+
+            var afterSetState = await messageActions.GetSessionStateAsync();
+            Assert.Equal(testString, afterSetState.ToString());
+
+            await messageActions.SetSessionStateAsync(null);
+            var afterSetNullState = await messageActions.GetSessionStateAsync();
+            Assert.Null(afterSetNullState);
+
```
jviau, line 67: nit: remove empty newline.

test/Worker.Extensions.Tests/ServiceBus/ServiceBusSessionMessageActionsTests.cs
```diff
@@ -86,5 +104,46 @@ public override AsyncUnaryCall<RenewSessionLockResponse> RenewSessionLockAsync(R
                 return new AsyncUnaryCall<RenewSessionLockResponse>(Task.FromResult(response), Task.FromResult(new Metadata()), () => Status.DefaultSuccess, () => new Metadata(), () => { });
             }
         }
+
+
```
jviau, line 108: nit: remove multiple empty newlines, keeping only 1

test/Worker.Extensions.Tests/ServiceBus/ServiceBusSessionMessageActionsTests.cs
Meorir: Ok

test/Worker.Extensions.Tests/ServiceBus/ServiceBusSessionMessageActionsTests.cs
```diff
@@ -49,6 +50,55 @@ public async Task CanRenewSessionLock()
...
+            var messageActions = new ServiceBusSessionMessageActions(settlement: mockClient.Object,sessionId: message.SessionId,sessionLockedUntil: message.LockedUntil);
+
+            var initialState = await messageActions.GetSessionStateAsync();
+            Assert.Equal(initialTestString, initialState.ToString());
+
+            await messageActions.SetSessionStateAsync(null);
+            var afterSetNullState = await messageActions.GetSessionStateAsync();
+            Assert.Null(afterSetNullState);
+        }
+        private static AsyncUnaryCall<T> CreateAsyncUnaryCall<T>(T response) where T : class
```
jviau, line 97: nit, newline after `}`
```suggestion

        private static AsyncUnaryCall<T> CreateAsyncUnaryCall<T>(T response) where T : class
```

test/Worker.Extensions.Tests/ServiceBus/ServiceBusSessionMessageActionsTests.cs
```diff
@@ -49,6 +50,55 @@ public async Task CanRenewSessionLock()
...
+                    It.IsAny<CancellationToken>())
+                )
+                .Returns<SetSessionStateRequest, Metadata, DateTime?, CancellationToken>((request, _, _, _) =>
+                {
+                    currentSessionState = request.SessionState;
+                    return CreateAsyncUnaryCall(new Empty());
+                });
+
+            var message = ServiceBusModelFactory.ServiceBusReceivedMessage(lockTokenGuid: Guid.NewGuid(), sessionId: sessionId);
+            var messageActions = new ServiceBusSessionMessageActions(settlement: mockClient.Object,sessionId: message.SessionId,sessionLockedUntil: message.LockedUntil);
```
jviau, line 88:
```suggestion
            var messageActions = new ServiceBusSessionMessageActions(settlement: mockClient.Object, sessionId: message.SessionId, sessionLockedUntil: message.LockedUntil);
```

test/Worker.Extensions.Tests/ServiceBus/ServiceBusSessionMessageActionsTests.cs
```diff
@@ -49,6 +50,55 @@ public async Task CanRenewSessionLock()
             await messageActions.RenewSessionLockAsync();
         }
 
+        [Fact]
+        public async Task CanSetAndGetSessionStateWithNull()
```
jviau, line 54: Can you split this test into two different tests:

1. Verifying `null` is returned from `GetSessionStateAsync` when `result.SessionState == null`.
1. Verifying ability to pass `null` to `SetSessionStateAsync`

You shouldn't need any other tests/calls to the client / mock for your set of changes.

test/Worker.Extensions.Tests/ServiceBus/ServiceBusSessionMessageActionsTests.cs
```diff
@@ -49,6 +50,55 @@ public async Task CanRenewSessionLock()
...
+            var mockClient = new Mock<Settlement.SettlementClient>();
+            
+            mockClient
+                .Setup(x => x.GetSessionStateAsync(
+                    It.Is<GetSessionStateRequest>(r => r.SessionId == sessionId),
+                    It.IsAny<Metadata>(),
+                    It.IsAny<DateTime?>(),
+                    It.IsAny<CancellationToken>())
+                )
+                .Returns(() => CreateAsyncUnaryCall(new GetSessionStateResponse
```
jviau, line 69: A lambda expression should be necessary, you can simplify as:

``` csharp
mockClient.Setup(m => m.GetSessionStateAsync(<setup>)).ReturnsAsync(<BinaryData object>);
```

test/Worker.Extensions.Tests/ServiceBus/ServiceBusSessionMessageActionsTests.cs
```diff
@@ -49,6 +50,55 @@ public async Task CanRenewSessionLock()
...
+            mockClient
+                .Setup(x => x.SetSessionStateAsync(
+                    It.Is<SetSessionStateRequest>(r => r.SessionId == sessionId),
+                    It.IsAny<Metadata>(),
+                    It.IsAny<DateTime?>(),
+                    It.IsAny<CancellationToken>())
+                )
+                .Returns<SetSessionStateRequest, Metadata, DateTime?, CancellationToken>((request, _, _, _) =>
+                {
+                    currentSessionState = request.SessionState;
```
jviau, line 83: For the test verifying the return on `SetSessionStateAsync`, you shouldn't need to capture this parameter in a lambda. Instead, you use a mock verify call to assert how it was called.

``` csharp
mockClient.Verify(m => m.SetSessionStateAsync(<It.* statements to verify precisely how it should be called>), Times.Once());
```

jviau (changes requested):

extensions/Worker.Extensions.ServiceBus/src/ServiceBusSessionMessageActions.cs
```diff
@@ -43,7 +43,7 @@ protected ServiceBusSessionMessageActions()
         public virtual DateTimeOffset SessionLockedUntil { get; protected set; }
 
         ///<inheritdoc cref="ServiceBusReceiver.CompleteMessageAsync(ServiceBusReceivedMessage, CancellationToken)"/>
-        public virtual async Task<BinaryData> GetSessionStateAsync(
+        public virtual async Task<BinaryData?> GetSessionStateAsync(
```
JoshLove-msft, line 46: Wouldn't this be considered a breaking change?

Meorir: For now, i belive the only way to check for null session state is to 
`if(binaryDataFromBytes.ToMemory().Length == 0)`
`if(binaryDataFromBytes.ToArray().Length == 0)`

If breaking changes are not to be introduced, maybe it would be better to return BinaryData.Empty rather then BinaryData.FromBytes(memory) in case of empty. Below are my findings.

```
ReadOnlyMemory<byte> memory = System.Array.Empty<byte>();

BinaryData binaryDataFromBytes = BinaryData.FromBytes(memory);
BinaryData binaryDataFromDataEmpty = BinaryData.Empty;

if (binaryDataFromBytes == BinaryData.Empty) //False
{
	Console.WriteLine("binaryDataFromBytes == BinaryData.Empty");
}

if (binaryDataFromDataEmpty == BinaryData.Empty) //True
{
	Console.WriteLine("binaryDataFromDataEmpty == BinaryData.Empty");
}

if (binaryDataFromBytes.ToMemory().Length == 0) //True
{
	Console.WriteLine("binaryDataFromBytes == ToMemory().Length == 0");
}


if (binaryDataFromDataEmpty.ToMemory().Length == 0) //True
{
	Console.WriteLine("binaryDataFromDataEmpty == ToMemory().Length == 0");
}

if (binaryDataFromBytes.ToArray().Length == 0) //True
{
	Console.WriteLine("binaryDataFromBytes == ToArray().Length == 0");
}


if (binaryDataFromDataEmpty.ToArray().Length == 0) //True
{
	Console.WriteLine("binaryDataFromDataEmpty == ToArray().Length == 0");
}

```

The documentation (for receiver, not ServiceBusSessionMessageActions) states:
https://learn.microsoft.com/en-us/azure/service-bus-messaging/message-sessions
> The methods for managing session state, SetState and GetState, can be found on the session receiver object. A session that had previously no session state returns a null reference for GetState. The previously set session state can be cleared by passing null to the SetState method on the receiver.

jviau: Since `BinaryData` is a `class`, I not sure if adding `?` is a breaking change. I think it is only compiler metadata impacted, not runtime impact.

Meorir: In-proc returns null for empty SessionState

test/Worker.Extensions.Tests/ServiceBus/ServiceBusSessionMessageActionsTests.cs
Meorir: It seems Moq has some problems with AsyncUnaryCall and does not treat it as awaitable.

Meorir:
Rebased commit on current main. Modified tests.

PR #2940, opened 2025-01-28
Bump Worker.Extension.Http for release

liliankasem:


kshyju (approved):
Does this version of extension have any dependency on specific versions of core package?

liliankasem:
> Does this version of extension have any dependency on specific versions of core package?

It should not. All changes were made in the extension only

PR #2947, opened 2025-01-31
Update EventGrid & SendGrid extension dependencies and prepare for release

liliankasem:
Update EventGrid & SendGrid extension dependencies and prepare for release

### DIFF
--- a/extensions/Worker.Extensions.EventGrid/release_notes.md
+++ b/extensions/Worker.Extensions.EventGrid/release_notes.md
@@ -4,6 +4,7 @@
 - My change description (#PR/#issue)
 -->
 
-### Microsoft.Azure.Functions.Worker.Extensions.EventGrid <version>
+### Microsoft.Azure.Functions.Worker.Extensions.EventGrid 4.5.0
 
-- <entry>
+- Update dependency "Azure.Messaging.EventGrid" to v4.29.0
+- Update dependency "Microsoft.Azure.Functions.Worker.Core" to v1.20.0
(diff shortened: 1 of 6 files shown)


jviau (approved):

satvu (approved):

briandunnington:
this doesnt appear to update the dependency on Microsoft.Azure.WebJobs.Extensions.EventGrid (so still targeting 3.4.2) which depends on an old version of Azure.Core which has component governance issues. since it uses the extension attributes to load the package, end-users cannot override the package version to manually force using a newer version.

liliankasem:
> this doesnt appear to update the dependency on Microsoft.Azure.WebJobs.Extensions.EventGrid (so still targeting 3.4.2) which depends on an old version of Azure.Core which has component governance issues. since it uses the extension attributes to load the package, end-users cannot override the package version to manually force using a newer version.

It's actually targetting 3.4.3:

`<WebJobsExtension Include="Microsoft.Azure.WebJobs.Extensions.EventGrid" Version="3.4.3" />`

But yes, we are aware :) We are waiting on the EventGrid team to release a new WebJobs extension that updates `Microsoft.Extensions.Azure` (this is the package bringing in the old Azure.Core version) before we release a new Worker Extension.

briandunnington:
thanks @liliankasem (i missed that it had already updated the WebJobs reference to 3.4.3)

looking forward to getting that old Azure.Core reference gone so all of our automated compliance tools stop yelling at me =)

liliankasem:
> so all of our automated compliance tools stop yelling at me =)

Same :)

PR #3020, opened 2025-04-07
Better error message when Azure Functions Core Tools is missing in `dotnet run` use case.

### LINKED ISSUE #2976: `dotnet run` should facilitate acquisition of the Core Tools
### Description

If the Core Tools is not installed, `dotnet run` will fail with an exception similar to the following:

> 	Unhandled exception: System.ComponentModel.Win32Exception (2): An error occurred trying to start process 'func' with working directory 'bin\Debug\net9.0\'. The system cannot find the file specified.
>	   at System.Diagnostics.Process.StartWithCreateProcess(ProcessStartInfo startInfo)
>	   at Microsoft.DotNet.Cli.Utils.Command.Execute(Action`1 processStarted)
>	   at Microsoft.DotNet.Tools.Run.RunCommand.Execute()
>	   at System.CommandLine.Invocation.InvocationPipeline.Invoke(ParseResult parseResult)
>	   at System.CommandLine.ParseResult.Invoke()
>   at Microsoft.DotNet.Cli.Program.ProcessArgs(String[] args, TimeSpan startupTime, ITelemetry telemetryClient)

This is not clearly actionable for the user. When we are in this situation, we should provide clear instructions on how to acquire the Core Tools.

kshyju:
resolves #2976

Currently, when `func` is not available, a wrapped Win32/Native exception is [thrown](https://github.com/dotnet/runtime/blob/831d23e56149cd59c40fc00c7feb7c5334bd19c4/src/libraries/System.Diagnostics.Process/src/System/Diagnostics/Process.Windows.cs#L612). This PR updates the logic to first check if `func` exists on the machine. If not, it sets `RunCommand` and `RunArguments` to display a user-friendly error message.

### Pull request checklist

### DIFF
--- a/sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
+++ b/sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
@@ -124,7 +124,7 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
   </Target>
 
   <!-- This target is used to set up the run arguments for the function app, supporting `dotnet run` -->
-  <Target Name="_FunctionsComputeRunArguments" BeforeTargets="ComputeRunArguments" DependsOnTargets="_CheckForFuncInstallation" Condition="'$(FuncExists)' == 'true'">
+  <Target Name="_FunctionsComputeRunArguments" BeforeTargets="ComputeRunArguments" DependsOnTargets="_FunctionsCheckForCoreTools" Condition="'$(FuncExists)' == 'true'">
     <PropertyGroup>
       <RunCommand>func</RunCommand>
       <RunArguments>host start $(RunArguments)</RunArguments>
@@ -132,12 +132,8 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
     </PropertyGroup>
   </Target>
 
-  <Target Name="_CheckForFuncInstallation">
-    <PropertyGroup>
-      <PathCheckCommand Condition="'$(OS)' == 'Windows_NT'">where func</PathCheckCommand>
-      <PathCheckCommand Condition="'$(OS)' != 'Windows_NT'">which func</PathCheckCommand>
-    </PropertyGroup>
-    <Exec Command="$(PathCheckCommand)" IgnoreExitCode="true">
+  <Target Name="_FunctionsCheckForCoreTools">
+    <Exec Command="func --version" IgnoreExitCode="true">
       <Output TaskParameter="ExitCode" PropertyName="FuncExitCode" />
     </Exec>
     <PropertyGroup>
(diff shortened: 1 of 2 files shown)


sdk/Sdk/Targets/Microsoft.Azure.Functions.Worker.Sdk.targets
```diff
@@ -124,14 +124,37 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
...
+    </Exec>
+    <PropertyGroup>
+      <FuncExists Condition="'$(FuncExitCode)' == '0'">true</FuncExists>
+    </PropertyGroup>
+  </Target>
   
+  <Target Name="_FunctionsComputeRunArgumentsWhenFuncNotExists" BeforeTargets="ComputeRunArguments" DependsOnTargets="_CheckForFuncInstallation" Condition="'$(FuncExists)' != 'true'">
+    <PropertyGroup>
+      <ErrorMessage>Azure Functions Core Tools is missing. Please install it and try again. Please visit https://aka.ms/azfunc-coretools-not-installed for more details.</ErrorMessage>
+      <RunCommand Condition="'$(OS)' == 'Windows_NT'">cmd.exe</RunCommand>
```
jviau, line 151: Can we hook into msbuild targets and just use `Error` task here instead of passing to `RunCommand`?

Also, I wonder if we can use `cmd` and `sh` to invoke `func` installed via npm or the like?

kshyju: We want to show this error when user executes the `dotnet run` command. So we need to rely on the "ComputeRunArguments" target before hook. The runtime code tries to execute whatever is passed via `RunCommand` and `RunArguments`. Using `<Error>` also works but produces a failure with the targets file info etc. See below.

image

With the RunCommand approach, it's much nicer.

image

I felt the second one is better. But open to suggestions. @mattchenderson Thoughts?

Let me check the npm install use case.

jviau: I like the first one personally - it catches the attention and makes it clear we never got to the point of even running the application.

kshyju: It appears that `dotnet run` does not work today when core tools is installed via npm.

image

jviau: @kshyju yes there was a customer reported issue where we found that problem. The problem is that npm doesn't install a func executable - it installs func scripts, like func.cmd, func.ps1, func.sh. These are not supported by dotnet run by default. I think we can work around this by using `cmd` or `bash` (OS dependent) as the process, with `func` as an argument.

I did a similar thing for our azurite fixture: https://github.com/Azure/azure-functions-host/blob/dev/test/WebJobs.Script.Tests.Integration/Fixtures/AzuriteFixture.cs#L152

jviau (approved):

PR #3056, opened 2025-05-15
Update HttpResultAttribute analyzer to include case where return type is Task<T>

satvu:

Discovered this missing case from [this comment](https://github.com/Azure/azure-functions-dotnet-worker/issues/2682#issuecomment-27205736890) where the customer was on a package version that should have notified them to use the `HttpResult` attribute.

### Pull request checklist

### Additional information

task-analyzer-works

Code fix also works as expected.

### DIFF
--- a/extensions/Worker.Extensions.Http.AspNetCore.Analyzers/src/CodeFixForHttpResultAttributeExpected.cs
+++ b/extensions/Worker.Extensions.Http.AspNetCore.Analyzers/src/CodeFixForHttpResultAttributeExpected.cs
@@ -66,14 +66,9 @@ namespace Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore
                     .FirstAncestorOrSelf<TypeSyntax>();
                 var typeSymbol = semanticModel.GetSymbolInfo(typeNode).Symbol;
 
-                // Unwrap Task<T> if present
-                var taskType = semanticModel.Compilation.GetTypeByMetadataName("System.Threading.Tasks.Task`1");
-                if (typeSymbol is INamedTypeSymbol namedTypeSymbol &&
-                    namedTypeSymbol.ConstructedFrom != null &&
-                    SymbolEqualityComparer.Default.Equals(namedTypeSymbol.ConstructedFrom, taskType) &&
-                    namedTypeSymbol.TypeArguments.Length == 1)
+                if (SymbolUtils.TryUnwrapTaskOfT(typeSymbol, semanticModel, out var innerSymbol))
                 {
-                    typeSymbol = namedTypeSymbol.TypeArguments[0];
+                    typeSymbol = innerSymbol;
                 }
 
                 var typeDeclarationSyntaxReference = typeSymbol.DeclaringSyntaxReferences.FirstOrDefault();
--- a/extensions/Worker.Extensions.Http.AspNetCore.Analyzers/src/CodeFixForHttpResultAttributeExpected.cs
+++ b/extensions/Worker.Extensions.Http.AspNetCore.Analyzers/src/CodeFixForHttpResultAttributeExpected.cs
@@ -64,8 +64,18 @@ namespace Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore
 
                 var typeNode = root.FindNode(this._diagnostic.Location.SourceSpan)
                     .FirstAncestorOrSelf<TypeSyntax>();
-
                 var typeSymbol = semanticModel.GetSymbolInfo(typeNode).Symbol;
+
+                // Unwrap Task<T> if present
+                var taskType = semanticModel.Compilation.GetTypeByMetadataName("System.Threading.Tasks.Task`1");
+                if (typeSymbol is INamedTypeSymbol namedTypeSymbol &&
+                    namedTypeSymbol.ConstructedFrom != null &&
+                    SymbolEqualityComparer.Default.Equals(namedTypeSymbol.ConstructedFrom, taskType) &&
+                    namedTypeSymbol.TypeArguments.Length == 1)
+                {
+                    typeSymbol = namedTypeSymbol.TypeArguments[0];
+                }
+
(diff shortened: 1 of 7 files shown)


extensions/Worker.Extensions.Http.AspNetCore.Analyzers/src/CodeFixForHttpResultAttributeExpected.cs
```diff
@@ -64,8 +64,18 @@ protected override async Task<Document> GetChangedDocumentAsync(CancellationToke
 
                 var typeNode = root.FindNode(this._diagnostic.Location.SourceSpan)
                     .FirstAncestorOrSelf<TypeSyntax>();
-
                 var typeSymbol = semanticModel.GetSymbolInfo(typeNode).Symbol;
+
+                // Unwrap Task<T> if present
+                var taskType = semanticModel.Compilation.GetTypeByMetadataName("System.Threading.Tasks.Task`1");
+                if (typeSymbol is INamedTypeSymbol namedTypeSymbol &&
```
liliankasem, line 71: Can this be a helper method in our one of our `extensions/`?

extensions/Worker.Extensions.Http.AspNetCore.Analyzers/src/HttpResultAttributeExpectedAnalyzer.cs
```diff
@@ -61,6 +61,16 @@ private static void AnalyzeMethod(SyntaxNodeAnalysisContext context)
             var returnType = methodDeclaration.ReturnType;
             var returnTypeSymbol = semanticModel.GetTypeInfo(returnType).Type;
 
+            // Unwrap Task<T> if present
+            var taskType = semanticModel.Compilation.GetTypeByMetadataName("System.Threading.Tasks.Task`1");
```
liliankasem, line 65: reusable code

extensions/Worker.Extensions.Http.AspNetCore.Analyzers/src/CodeFixForHttpResultAttributeExpected.cs
```diff
@@ -64,8 +64,18 @@ protected override async Task<Document> GetChangedDocumentAsync(CancellationToke
...
+                    namedTypeSymbol.ConstructedFrom != null &&
```
liliankasem, line 72: nit
```suggestion
                    namedTypeSymbol.ConstructedFrom is not null &&
```

test/extensions/Worker.Extensions.Http.AspNetCore.Tests/HttpResultAttributeExpectedTests.cs
```diff
@@ -347,6 +347,119 @@ public class MyOutputType
             await test.RunAsync();
         }
 
+        [Fact]
+        public async Task HttpResultAttribute_WhenReturnTypeIsWrappedInTask_Expected()
+        {
```
surgupta-msft, line 352: Can we add some tests which ensure that we don't suggest code fix when its not needed? For example, what if there is another method in this class (which is not a trigger) but has Task<poco> as return type, something like below -

```csharp
Task<poco> fun() { }
```

extensions/Worker.Extensions.Http.AspNetCore.Analyzers/src/HttpResultAttributeExpectedAnalyzer.cs
```diff
@@ -61,6 +61,11 @@ private static void AnalyzeMethod(SyntaxNodeAnalysisContext context)
             var returnType = methodDeclaration.ReturnType;
             var returnTypeSymbol = semanticModel.GetTypeInfo(returnType).Type;
 
+           if (SymbolUtils.TryUnwrapTaskOfT(returnTypeSymbol, semanticModel, out var innerSymbol))
```
watfordsuzy, line 64:
```suggestion
            if (SymbolUtils.TryUnwrapTaskOfT(returnTypeSymbol, semanticModel, out var innerSymbol))
```

surgupta-msft (approved):

extensions/Worker.Extensions.Http.AspNetCore.Analyzers/src/CodeFixForHttpResultAttributeExpected.cs
```diff
@@ -64,8 +64,13 @@ protected override async Task<Document> GetChangedDocumentAsync(CancellationToke
...
                     .FirstAncestorOrSelf<TypeSyntax>();
-
                 var typeSymbol = semanticModel.GetSymbolInfo(typeNode).Symbol;
+
+                if (SymbolUtils.TryUnwrapTaskOfT(typeSymbol, semanticModel, out var innerSymbol))
+                {
+                    typeSymbol = innerSymbol;
+                }
+
                 var typeDeclarationSyntaxReference = typeSymbol.DeclaringSyntaxReferences.FirstOrDefault();
```
surgupta-msft, line 74: typeSymbol can be null here, right? as it is returned by method TryUnwrap(). In that case, we will need a null check here

extensions/Worker.Extensions.Http.AspNetCore.Analyzers/src/HttpResultAttributeExpectedAnalyzer.cs
```diff
@@ -61,6 +61,11 @@ private static void AnalyzeMethod(SyntaxNodeAnalysisContext context)
...
+            {
+                returnTypeSymbol = innerSymbol;
```
surgupta-msft, line 66: same thing here for returnTypeSymbol. Need a null check.

extensions/Worker.Extensions.Http.AspNetCore.Analyzers/src/CodeFixForHttpResultAttributeExpected.cs
satvu: innerSymbol is not null if `TryUnwrapTaskOfT` returns true so it would not be `null` at this line.

surgupta-msft (approved):

PR #3121, opened 2025-07-24
Default empty, nullable input bindings to null instead of empty string

liliankasem:

resolves https://github.com/Azure/azure-functions-mcp-extension/issues/70

### Pull request checklist

### Additional information

When an argument is not required, it may be omitted from the tool call request. This should lead to a null input binding value. However, it seems like strings, for example, will get the empty string.

For the draft, I have a couple of suggestions on how to address this issue (see code comments).

### DIFF
--- a/src/DotNetWorker.Core/Context/Features/DefaultFunctionInputBindingFeature.cs
+++ b/src/DotNetWorker.Core/Context/Features/DefaultFunctionInputBindingFeature.cs
@@ -55,7 +55,17 @@ namespace Microsoft.Azure.Functions.Worker.Context.Features
 
                     TryGetBindingSource(parameter.Name, functionBindings, out object? source);
 
+                    // Option 1: Check if source is an empty string and parameter is reference or nullable type
+                    if (source is string str && string.IsNullOrEmpty(str))
+                    {
+                        // OR we can set `parameterValues[i] = null`. However, we already have checks for HasDefaultValue & IsReferenceOrNullableType
+                        // in the `(bindingResult.Status == ConversionStatus.Unhandled)` case below which will hanlde what we need it to.
+                        // So it depends on if we want to short circuit or not.
+                        source = null;
+                    }
+
                     ConversionResult bindingResult = await ConvertAsync(context, parameter, _converterContextFactory, inputConversionFeature, source);
+
 
                     if (bindingResult.Status == ConversionStatus.Succeeded)
                     {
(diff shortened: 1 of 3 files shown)


src/DotNetWorker.Core/Context/Features/DefaultFunctionInputBindingFeature.cs
```diff
@@ -53,6 +53,13 @@ public async ValueTask<FunctionInputBindingResult> BindFunctionInputAsync(Functi
                 {
                     FunctionParameter parameter = context.FunctionDefinition.Parameters[i];
 
+                    // Option 1: If the parameter does not have a default value and is nullable type, set it to null.
+                    if (!parameter.HasDefaultValue && parameter.IsReferenceOrNullableType)
```
liliankasem, line 57: Actually on second thought, not having a default value does not necessarily mean that its an empty/null item. So maybe we could check if source is null or empty here then check for these parameters? But that might not be any difference to just checking for string empty in our converters

src/DotNetWorker.Core/Converters/TypeConverter.cs
```diff
@@ -10,6 +10,7 @@ internal class TypeConverter : IInputConverter
     {
         public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
         {
+            // Option 1: If the target type is a string, check for null or empty on the source here
```
liliankasem, line 13: I feel like this only solves the issue for strings and not all types/cases where we may want a null value

liliankasem: Actually, it looks like source is always an empty string regarless of the binding type

src/DotNetWorker.Core/Context/Features/DefaultFunctionInputBindingFeature.cs
```diff
@@ -55,8 +55,18 @@ public async ValueTask<FunctionInputBindingResult> BindFunctionInputAsync(Functi
 
                     TryGetBindingSource(parameter.Name, functionBindings, out object? source);
 
+                    // Option 1: Check if source is an empty string and parameter is reference or nullable type
+                    if (source is string str && string.IsNullOrEmpty(str))
```
liliankasem, line 59: I think this is my preferred approach. Regardless of the binding type, source is always a string that needs to be convertered into a Type. So I think this is a safe change

PR #3123, opened 2025-07-25
Adding net10 TFM to worker packages

### LINKED ISSUE #3114: NET 10 Worker package updates
### Description

- Version: 2.50.0-preview1
- Fork a new branch.
- Don't drop any existing TFMs. Add .NET10 to worker packages.
- Bump dependencies to 10x (for all TFMs)
-  Aspnet extension - Add .NET10 TFM.
- ApplicationInsights extension - no action now

kshyju:
Adding net10 TFM to worker packages and related changes. 

Fixes #3114

### Pull request checklist

The AspNetCore extension has a package reference to Microsoft.Azure.Functions.Worker  with pinned version. So I plan to have that as a follow up PR once this version of the worker package is out.

### DIFF
--- a/eng/ci/host/official-release.yml
+++ b/eng/ci/host/official-release.yml
@@ -94,13 +94,13 @@ extends:
             artifactName: $(artifact_name)
             pipeline: build
 
-          steps:
-          - task: 1ES.PublishNuget@1
-            displayName: Publish packages
-            inputs:
-              packagesToPush: $(packages_pattern)
-              packageParentPath: $(drop_path)
-              publishVstsFeed: $(nuget_feed)
-              nuGetFeedType: internal
-              allowPackageConflicts: true
-              publishPackageMetadata: true
+        steps:
+        - task: 1ES.PublishNuget@1
+          displayName: Publish packages
+          inputs:
+            packagesToPush: $(packages_pattern)
+            packageParentPath: $(drop_path)
+            publishVstsFeed: $(nuget_feed)
+            nuGetFeedType: internal
+            allowPackageConflicts: true
+            publishPackageMetadata: true
--- a/src/DotNetWorker.ApplicationInsights/DotNetWorker.ApplicationInsights.csproj
+++ b/src/DotNetWorker.ApplicationInsights/DotNetWorker.ApplicationInsights.csproj
@@ -6,8 +6,9 @@
     <AssemblyName>Microsoft.Azure.Functions.Worker.ApplicationInsights</AssemblyName>
     <RootNamespace>Microsoft.Azure.Functions.Worker.ApplicationInsights</RootNamespace>
     <MajorProductVersion>2</MajorProductVersion>
-    <MinorProductVersion>0</MinorProductVersion>
+    <MinorProductVersion>50</MinorProductVersion>
     <PatchProductVersion>0</PatchProductVersion>
+    <VersionSuffix>-preview1</VersionSuffix>
     <PackageReadmeFile>README.md</PackageReadmeFile>    
     <BeforePack>$(BeforePack);GetReleaseNotes</BeforePack>
   </PropertyGroup>
@@ -17,7 +18,7 @@
   <ItemGroup>
     <PackageReference Include="Azure.Identity" Version="1.12.0" />
     <PackageReference Include="Microsoft.ApplicationInsights.PerfCounterCollector" Version="2.22.0" />
-    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="8.0.0" />
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="10.0.0-preview.6.25358.103" /> 
   </ItemGroup>
 
   <ItemGroup>
--- a/src/DotNetWorker.Core/DotNetWorker.Core.csproj
+++ b/src/DotNetWorker.Core/DotNetWorker.Core.csproj
@@ -2,15 +2,16 @@
 
   <PropertyGroup>
     <OutputType>Library</OutputType>
-    <TargetFrameworks>net6.0;net8.0;net9.0;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net6.0;net8.0;net9.0;net10.0;netstandard2.0</TargetFrameworks>
     <PackageId>Microsoft.Azure.Functions.Worker.Core</PackageId>
     <Description>This library provides the core functionality to build an Azure Functions .NET Worker, adding support for the isolated, out-of-process execution model.</Description>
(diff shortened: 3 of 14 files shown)


test/Worker.Extensions.Tests/Blob/StreamTests.cs
```diff
@@ -246,7 +246,7 @@ public async Task ConvertAsync_StreamCollection_WithFilePath_Throws_ReturnsFaile
             // Assert
             Assert.Equal(ConversionStatus.Failed, conversionResult.Status);
             Assert.IsType<NotSupportedException>(conversionResult.Error);
-            Assert.Contains("Deserialization of types without a parameterless constructor, a singular parameterized constructor, or a parameterized constructor annotated with 'JsonConstructorAttribute' is not supported.", conversionResult.Error.Message);
+            Assert.Contains("Deserialization of interface or abstract types is not supported.", conversionResult.Error.Message);
```
kshyju, line 249: The exception message thrown by STJ `JsonSerializer.Deserialize` in this scenario has changed between net8.0 and net9.0+. Minimal console app repro to show here: https://gist.github.com/kshyju/04bd2c8d2891bd8df8b2e0d711276984#file-stjserializationerrornet8vsnet9-cs

eng/ci/host/official-release.yml
```diff
@@ -94,13 +94,13 @@ extends:
...
-          - task: 1ES.PublishNuget@1
-            displayName: Publish packages
-            inputs:
-              packagesToPush: $(packages_pattern)
-              packageParentPath: $(drop_path)
-              publishVstsFeed: $(nuget_feed)
-              nuGetFeedType: internal
-              allowPackageConflicts: true
-              publishPackageMetadata: true
+        steps:
```
kshyju, line 97: This change is unrelated to the net10 TFM changes. But this is from another commit (PR includes 2 commits) which was added to dev after we forked the feature branch,

src/DotNetWorker.Core/DotNetWorker.Core.csproj
```diff
@@ -22,19 +23,12 @@
 
   <ItemGroup>
     <PackageReference Include="Azure.Core" Version="1.44.1" />
+    <PackageReference Include="Microsoft.Extensions.Hosting" Version="10.0.0-preview.6.25358.103" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
```
jviau, line 29: We might want to remove this item group. We at least would want these to be `10.x` also now. Removing may accomplish that directly.

kshyju: Switched to 10.x. Don't think we can remove that as our code relies on APIs from those and `netstandard2.0` is one of our TFM.

src/DotNetWorker.Core/DotNetWorker.Core.csproj
```diff
@@ -22,19 +23,12 @@
 
   <ItemGroup>
     <PackageReference Include="Azure.Core" Version="1.44.1" />
```
Frulfump, line 25: 1.47.1 updates System.Memory.Data to 8.0.1 from 6.0.1 (And System.ClientModel to 1.5.1) https://www.nuget.org/packages/Azure.Core/1.47.1

Edit: Also removes alot of unnecessary packages that we would get for .NET 8+ since the currently used version has a .NET 6 target with those and the version I suggested adds a .NET 8 target with those removed.

src/DotNetWorker.Core/DotNetWorker.Core.csproj
```diff
@@ -2,15 +2,16 @@
 
   <PropertyGroup>
     <OutputType>Library</OutputType>
-    <TargetFrameworks>net6.0;net8.0;net9.0;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net6.0;net8.0;net9.0;net10.0;netstandard2.0</TargetFrameworks>
```
Frulfump, line 5: When will net6.0 support be dropped? It's out of support since November 2024 but maybe you would do that in a new major instead?

jviau: Yes, we will only remove net6 in a major version rev. We do not want existing net6 customers, even if it is out of support, to fall back to the netstandard2.0 TFM. The netstandard2.0 TFM has some significant differences in its transitive references, and we wouldn't want net6 projects to shift to that.

src/DotNetWorker.Core/DotNetWorker.Core.csproj
jviau: @kshyju -- maybe we should pin `System.Memory.Data` to 10.x?

Frulfump: I think it's best if the major version matches the TFM to avoid downgrade issues and potential incompatabilities (testing matrix for packages don't usually include earlier TFMs I think)? (Currently only previews exist for .NET 10 and I assume you want this merged before .NET 10 GAs in November it's fine to use 8.x/9.x or if you are fine with rc(still considered preview by NuGet) packages those should release in September and October. I'd rather not pin the version unless there's a very compelling reason to do so.

fabiocav: As discussed, let's not pin transitive dependencies. We should, however, bump anything to latest for direct dependencies, but that doesn't have to happen in the scope of this PR.

src/DotNetWorker.ApplicationInsights/DotNetWorker.ApplicationInsights.csproj
```diff
@@ -17,7 +17,11 @@
   <ItemGroup>
     <PackageReference Include="Azure.Identity" Version="1.12.0" />
     <PackageReference Include="Microsoft.ApplicationInsights.PerfCounterCollector" Version="2.22.0" />
-    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="8.0.0" />
+    <!-- 
+      DotNetWorker.Core depends on Microsoft.Extensions.Hosting, which in turn depends on Microsoft.Bcl.AsyncInterfaces.
+      Uncommenting the line below would result in a "package downgrade" warning due to conflicting versions of Microsoft.Bcl.AsyncInterfaces (10.0.0 vs 8.0.0).
+    -->
+    <!-- <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="8.0.0" /> -->
```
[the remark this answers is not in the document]
fabiocav: Bump this up as well. We don't want to drop an explicit dependency.

PR #3145, opened 2025-08-08
Implementing function metadata transform support

liliankasem:

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

Rough draft of feature to allow augmentation of the FunctionMetadata. 

Works E2E with: https://github.com/Azure/azure-functions-mcp-extension/pull/88


src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs
```diff
@@ -48,7 +48,11 @@ public static IServiceCollection AddGrpc(this IServiceCollection services)
             services.TryAddSingleton<IWorkerDiagnostics, GrpcWorkerDiagnostics>();
 
             // FunctionMetadataProvider for worker driven function-indexing
-            services.TryAddSingleton<IFunctionMetadataProvider, DefaultFunctionMetadataProvider>();
+            // services.TryAddSingleton<IFunctionMetadataProvider, DefaultFunctionMetadataProvider>();
+            services.TryAddKeyedSingleton<IFunctionMetadataProvider>("core", new DefaultFunctionMetadataProvider());
```
liliankasem: Could also implement a new interface here instead of using Keyed services

jviau:
Interesting approach. We should discuss this a bit more. I have an existing draft PR (#2834) and issue (#1159) that this approach would also address.

src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs
jviau: I am guessing the goal of this is to negotiate a conflict with the source generated `IFunctionMetadataProvider`. An option would be to change it to generate your new `IFunctionMetadataTransformer` contract instead. That is the approach I took in my draft PR (I think)

src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs
```diff
@@ -12,6 +12,7 @@
 using Microsoft.Azure.Functions.Worker.Logging;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Logging;
```
aishwaryabh: nit: not needed

aishwaryabh (approved):

src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs
fabiocav: Leftover from the initial set of changes. Removed.

mattchenderson (approved):
PR should be retitled / commit message for squash-and-merge should be adjusted so that history does not include a reference to a type not being included.

mattchenderson (approved):

jviau (approved):

jviau:
Need to make sure to update AspNetCore extension to consume `IFunctionMetadataManager` and not the provider.

alanlima:
Hi @liliankasem 

That’s a really interesting approach and implementation.

I’m currently working on a similar requirement where I need to implement a custom metadata provider to support disabling functions in the staging slot. I’ve successfully migrated to the new Metadata Transformer, which has been great.

However, I’m curious about one aspect: why is the Transform method synchronous (sync) when the `GetFunctionMetadataAsync` method in the metadata provider is asynchronous (async)?

In our use case, we disable functions using feature flags via `IFeatureManager`, aiming to replicate the behaviour of the `FeatureGateAttribute` used in ASP.NET applications. Since IFeatureManager relies on async methods, applying this requirement forces us into a sync over async scenario — which is less than ideal.

Would love to hear your thoughts on this design decision and whether there’s any guidance or workaround you’d recommend.

PR #3161, opened 2025-09-04
Update Forward Limit

### LINKED ISSUE #3111: "ForwardLimit" not configured, causes loss of client IP when multiple proxies are involved
### Description

In the following code:
https://github.com/Azure/azure-functions-dotnet-worker/blob/main/extensions/Worker.Extensions.Http.AspNetCore/src/FunctionsHostBuilderExtensions.cs#L70-L75

The `ForwardedHeadersOptions.ForwardLimit` property is not explicitly set. According to the default behavior of `ForwardedHeadersMiddleware`, this means `ForwardLimit` defaults to `1`, which results in only the last IP in the `X-Forwarded-For` header being used.

This causes the actual client IP to be lost when there are more than one proxy involved (e.g., CloudFront + Azure Front Door). Instead of capturing the first IP (which is typically the originating client), the current configuration captures only the most recent proxy's IP.

### Steps to reproduce

1. Deploy an Azure Function using the .NET Isolated Worker model with CloudFront or similar reverse proxy in front.
2. Ensure that multiple IPs are included in the `X-Forwarded-For` header.
3. Observe that only the last IP is considered as `RemoteIpAddress`.

satvu:

resolves #3111

### Pull request checklist

9/4 - In draft. Once a limit is decided upon, will add tests and follow up for potential docs updates on the limits defined here.

### DIFF
--- a/extensions/Worker.Extensions.Http.AspNetCore/src/FunctionsHostBuilderExtensions.cs
+++ b/extensions/Worker.Extensions.Http.AspNetCore/src/FunctionsHostBuilderExtensions.cs
@@ -73,7 +73,7 @@ namespace Microsoft.Extensions.Hosting
                     // are sent by the host and will be processed by the worker.
                     options.ForwardedHeaders = ForwardedHeaders.All;
 
-                    options.ForwardedHeadersOptions.ForwardLimit = null;
+                    options.ForwardLimit = null;
                 });
             });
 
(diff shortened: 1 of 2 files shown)


extensions/Worker.Extensions.Http.AspNetCore/src/FunctionsHostBuilderExtensions.cs
```diff
@@ -72,6 +72,8 @@ internal static IHostBuilder ConfigureAspNetCoreIntegration(this IHostBuilder bu
                     // By default, the X-Forwarded-For, X-Forwarded-Host, and X-Forwarded-Proto headers
                     // are sent by the host and will be processed by the worker.
                     options.ForwardedHeaders = ForwardedHeaders.All;
+
+                    options.ForwardLimit = null;
```
satvu, line 76: `null` removes any limits -- this is a placeholder for now. Do we want an explicitly chosen value for this limit?

satvu:
Closing as this should be customer configurable, and we should not change the defaults from the conservative/cautious default.

PR #3178, opened 2025-09-24
Bump ServiceBus extension dependencies

### LINKED ISSUE #3175: BUG] Function Tools cannot be used to debug locally an app against the Azure Service Bus emulator
# Issue Transfer

This issue has been transferred from the Azure SDK for .NET repository, [#52774](https://github.com/Azure/azure-sdk-for-net/issues/52774).

## Azure SDK Notes

The latest stable versions of both `Azure.Messaging.ServiceBus` and `Microsoft.Azure.WebJobs.Extensions.ServiceBus` have support for the emulator.   Currently ,the latest version of `Microsoft.Azure.Functions.Worker.Extensions.ServiceBus` has a dependency on v7.18.1 of the Service Bus client library, which is more than a year old, is not compatible with the emulator, and is unsupported.   The dependency needs to be updated to the current stable `Azure.Messaging.ServiceBus` and should be kept current going forward.

### Description

I extended my function app with a new function that uses the Service Bus trigger and output binding and I installed the [Azure Service Bus Emulator](https://learn.microsoft.com/en-us/azure/service-bus-messaging/overview-emulator) using the [installer](https://github.com/Azure/azure-service-bus-emulator-installer) available on GitHub. However, when running `func start` command, the function app runtime failed to connect the emulated Service Bus with the following error.

```bash
[2025-09-11T13:31:32.745Z] Message processing error (Action=Receive, EntityPath=input, Endpoint=bongo-servicebus-test.localhost.localstack.cloud)
[2025-09-11T13:31:32.745Z] System.Private.CoreLib: Connection refused ErrorCode: ConnectionRefused (ServiceCommunicationProblem). For troubleshooting information, see https://aka.ms/azsdk/net/servicebus/exceptions/troubleshoot. System.Private.CoreLib: Connection refused.
```

I created a Python app that successfully connects to Azure Service Bus emulator and is able to send and receive messages. So the Service Bus emulator works as expected. So I used Visual Studio Code to debug the worker process launched by the `func start` command, and I found the command that was failing. I created a repro in C# and it failed against the Azure Service Bus emulator, as expected.

I made a not-so-quick search over the internet and I didn't find any tip to solve the issue. AI assistants didn't provide any useful suggestion. So I looked the issues in the [Azure Service Bus Emulator Installer](https://github.com/Azure/azure-service-bus-emulator-installer) repo and I found this [issue] (https://github.com/Azure/azure-service-bus-emulator-installer/issues/104): `Azure.Messaging.ServiceBus` versions older than `7.18.2` (at this time, the latest` 7.20.1`) do not support the Service Bus emulator. This issue was updated yesterday. After bumping up the `Azure.Messaging.ServiceBus` assembly version in the `.csproj` project file, my C# sample worked.

However, the worker process started when you run the `func start` command does not uses the latest version of the `Azure.Messaging.ServiceBus` assembly and it's not using the version you specify in your Azure Functions project. In order to verify this assumption, I proceeded as follows:

- I created a launch profile to attach to a dotnet process:
```json
  "version": "0.2.0",
  "configurations": [
    {
      "name": ".NET Core Attach (Azure Functions)",
      "type": "coreclr",
      "request": "attach",
      "processId": "${command:pickProcess}",
      "justMyCode": false
    }
  ]
}
```
- I started my function with `func start --verbose --dotnet-isolated-debug`. This option pauses the .NET Worker process until a debugger is attached.
- I looked the assemblies loaded in the DEBUG VIEW in Visual Studio Code.
- I installed [https://www.nuget.org/packages/ilspycmd/](ilspycmd) tool
- I run `ilspycmd /home/paolo/azure/func/dotnet_reg/bin/output/.azurefunctions/Azure.Messaging.ServiceBus.dll | grep AssemblyVersion`. The command returned `[assembly: AssemblyVersion("7.17.1.0")]` which is lower than `7.18.2`. This means that the current version of the Azure Function Tools uses a version of the `Azure.Messaging.ServiceBus` assembly that does not support connecting to the [Service Bus emulator by Microsoft](https://learn.microsoft.com/en-us/azure/service-bus-messaging/overview-emulator). 
In order to fix the issue, please update the `Azure.Messaging.ServiceBus` assembly referenced by this project and any other assembly referenced by this project, such as `azure-function-host` and `azure-webjobs-sdk`, to the latest version.

The `Azure.Messaging.ServiceBus` is not directly referenced by projects in the https://github.com/Azure/azure-functions-core-tools/ repo, but it's referenced indirectly because the Azure Function Tools projects reference assemblies that in their turn the `Azure.Messaging.ServiceBus` assembly. Hence, the final effect is the one I described in this issue. These assemblies like are stored in different repos, such as:

- [Azure Functions Host](https://github.com/azure/azure-functions-host)
- [Azure WebJobs SDK](https://github.com/Azure/azure-webjobs-sdk)
- [Azure SDK for .NET](https://github.com/Azure/azure-sdk-for-net)

### Steps to reproduce

Just try to run `func start` to debug locally a function app that uses the Service Bus trigger and/our output binding against the [Azure Service Bus Emulator](https://learn.microsoft.com/en-us/azure/service-bus-messaging/overview-emulator).

### Expected behavior

The [Microsoft.Azure.WebJobs.Extensions.ServiceBus](https://github.com/Azure/azure-sdk-for-net/tree/main/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus) supports the latest version of `Azure.Messaging.ServiceBus` assembly. This is not enough. All the assemblies used by the function tools, directly or indirectly, need to do the same.

### Actual behavior

It does used the latest version.

### Reproduction Steps

Just try to run func start to debug locally a function app that uses the Service Bus trigger and/our output binding against the [Azure Service Bus Emulator](https://learn.microsoft.com/en-us/azure/service-bus-messaging/overview-emulator).

liliankasem:

resolves #3175

### Pull request checklist

### Additional information

Update dependencies in the  ServiceBus extension to latest minor versions.

### DIFF
--- a/extensions/Worker.Extensions.ServiceBus/src/Worker.Extensions.ServiceBus.csproj
+++ b/extensions/Worker.Extensions.ServiceBus/src/Worker.Extensions.ServiceBus.csproj
@@ -6,7 +6,7 @@
     <Description>Azure Service Bus extensions for .NET isolated functions</Description>
 
     <!--Version information-->
-    <VersionPrefix>5.23.0</VersionPrefix>
+    <VersionPrefix>5.24.0</VersionPrefix>
 
     <!--Temporarily opting out of documentation. Pending documentation-->
     <GenerateDocumentationFile>false</GenerateDocumentationFile>
@@ -15,11 +15,11 @@
   <Import Project="..\..\..\build\Extensions.props" />
 
   <ItemGroup>
-    <PackageReference Include="Azure.Identity" Version="1.13.1" />
-    <PackageReference Include="Azure.Messaging.ServiceBus" Version="7.18.1" />
-    <PackageReference Include="Microsoft.Extensions.Azure" Version="1.12.0" />
-    <PackageReference Include="Google.Protobuf" Version="3.27.1" />
-    <PackageReference Include="Grpc.Tools" Version="2.64.0" PrivateAssets="All" />
+    <PackageReference Include="Azure.Identity" Version="1.16.0" />
+    <PackageReference Include="Azure.Messaging.ServiceBus" Version="7.20.1" />
+    <PackageReference Include="Microsoft.Extensions.Azure" Version="1.13.0" />
+    <PackageReference Include="Google.Protobuf" Version="3.32.1" />
+    <PackageReference Include="Grpc.Tools" Version="2.72.0" PrivateAssets="All" />
   </ItemGroup>
 
   <ItemGroup>
--- a/extensions/Worker.Extensions.ServiceBus/release_notes.md
+++ b/extensions/Worker.Extensions.ServiceBus/release_notes.md
@@ -4,6 +4,11 @@
 - My change description (#PR/#issue)
 -->
 
-### Microsoft.Azure.Functions.Worker.Extensions.ServiceBus <version>
+### Microsoft.Azure.Functions.Worker.Extensions.ServiceBus 5.24.0
 
-- <entry>
+- Update minor dependencies
+    - Azure.Identity to 1.16.0
+    - Azure.Messaging.ServiceBus to 7.20.1
+    - Microsoft.Extensions.Azure to 1.13.0
+    - Google.Protobuf to 3.32.1
+    - Grpc.Tools to 2.72.0


extensions/Worker.Extensions.ServiceBus/src/Worker.Extensions.ServiceBus.csproj
jviau, line 39: Want to block this PR for now as it doesn't actually address #3175. The fix for that will be to wait for a new ServiceBus WebJobs to be published and to update our reference to it.

jviau (changes requested):

PR #3187, opened 2025-10-06
[v1.x] Update core tools download URI

liliankasem:
Update to the latest approach for downloading core tools, matching what we have in main


liliankasem:
@microsoft-github-policy-service rerun

PR #3195, opened 2025-10-13
Worker instrumentation

### LINKED ISSUE #3198: It should be possible to completely disable tracing on the host
See  https://github.com/Azure/azure-functions-dotnet-worker/issues/2733 for the context.

I'd like to be able to disable tracing on function host and trace everything on the worker to reduce number of spans I might not care about.

What I tried:
-  removing `telemetryMode` from `host.json` and then no traces are exported, but the parent span is still generated. 
- disabling ASP.NET Core activity auto-creation (https://github.com/dotnet/aspnetcore/issues/18955) - the same

my config looks like this:

```json
{
  "version": "2.0",
  "logging": {
    "LogLevel": {
      "Microsoft.AspNetCore.Hosting.Diagnostics": "None"
    },
    "EventLog": {
      "LogLevel": {
        "Microsoft.AspNetCore.Hosting.Diagnostics": "None"
      }
    }
  }
}
```

## What I see 

I get the following  telemetry for it with https://github.com/lmolkova/testfuncotel/tree/aspnetcoreint 

Image

I.e. two activities - from worker ASP.NET Core and custom one in the function. 
Both of these activities have a parent (but different parents) which came from host. 

Also related to https://github.com/Azure/azure-functions-dotnet-worker/issues/2875

## What I expect

When I don't provide `telemetryMode`, don't have `APPLICATIONINSIGHTS_CONNECTION_STRING` and also explicitly disable ASP.NET Core activity creation in absence of a listener, the host should not create any activities.

### LINKED ISSUE #3199: Versioning for Span and Metrics specifications
Several span and metrics specifications are currently in the experimental phase. To ensure a smooth transition, we should implement versioning that allows customers to choose their preferred version and upgrade at their own pace, avoiding breaking changes.

Attributes are added to both the host and worker processes, and the semantic conventions must remain consistent across these processes. Customers should have the ability to specify the version they wish to use, enabling a gradual transition to newer versions without disruption.

Key considerations:

- Impact on cold start performance.
- Ideally, a single setting/configuration should influence both the worker and host processes.

RohitRanjanMS:
- Updated worker instrumentation to improve telemetry handling.
- Added support for OpenTelemetry schema version backward compatibility, allowing users to adopt newer schema versions at their own pace.
- Removed the previous “internal” activity that was created solely for context propagation but could not be tracked.
- Introduced a new internal activity that is now published and provides proper context propagation.

This is how it looks now -

Existing
Image

Proposed
Image

resolves #3198
resolves #3199 

### Pull request checklist

### DIFF
--- a/test/DotNetWorker.OpenTelemetry.Tests/EndToEndTests.cs
+++ b/test/DotNetWorker.OpenTelemetry.Tests/EndToEndTests.cs
@@ -87,7 +87,7 @@ public class EndToEndTests
 
         if (ActivityContext.TryParse(context.TraceContext.TraceParent, context.TraceContext.TraceState, true, out ActivityContext activityContext))
         {
-            Assert.Equal("Invoke", activity.OperationName);
+            Assert.Equal("function TestName", activity.OperationName);
             Assert.Equal(activity.TraceId, activityContext.TraceId);
             Assert.Equal(activity.TraceStateString, activityContext.TraceState);
             Assert.Equal(ActivityKind.Internal, activity.Kind);
@@ -115,7 +115,6 @@ public class EndToEndTests
             Assert.Equal(activity.TraceStateString, activityContext.TraceState);
             Assert.Equal(ActivityKind.Server, activity.Kind);
             Assert.Contains(activity.Tags, t => t.Key == TraceConstants.OTelAttributes_1_17_0.InvocationId && t.Value == context.InvocationId);
-            Assert.Contains(activity.Tags, t => t.Key == TraceConstants.InternalKeys.AzFuncLiveLogsSessionId && t.Value == context.TraceContext.Attributes[TraceConstants.InternalKeys.AzFuncLiveLogsSessionId]);
         }
         else
         {
@@ -141,16 +140,12 @@ public class EndToEndTests
     [Fact]
     public async Task ContextPropagation_EmptyVersion()
     {
-        try
-        {
-            using var host = InitializeHost(string.Empty);
-            var context = CreateContext(host);
-            await _application.InvokeFunctionAsync(context);
-        }
-        catch (Exception ex)
-        {
-            Assert.IsType<ArgumentException>(ex);
-        }
+        using var host = InitializeHost(string.Empty);
+        var context = CreateContext(host);
+        await _application.InvokeFunctionAsync(context);
+
+        var activity = OtelFunctionDefinition.LastActivity;
+        Assert.Null(activity);
     }
 
     [Fact]
@@ -236,7 +231,7 @@ public class EndToEndTests
         public async Task TestFunction(FunctionContext context)
         {
             LastActivity = Activity.Current;
-            Activity.Current.AddTag("CustomKey", "CustomValue");
+            Activity.Current?.AddTag("CustomKey", "CustomValue");
 
             var response = new HttpResponseMessage(HttpStatusCode.OK);
             var handler = new MockHttpMessageHandler(response);
--- a/test/DotNetWorkerTests/FunctionsApplicationTests.cs
+++ b/test/DotNetWorkerTests/FunctionsApplicationTests.cs
@@ -104,7 +104,7 @@ namespace Microsoft.Azure.Functions.Worker.Tests
         private static void AssertActivity(Activity activity, FunctionContext context)
         {
             Assert.Equal("Invoke", activity.DisplayName);
-            Assert.Equal(5, activity.Tags.Count());
+            Assert.Equal(2, activity.Tags.Count());
             Assert.Equal("https://opentelemetry.io/schemas/1.17.0", activity.Tags.Single(k => k.Key == "az.schema_url").Value);
             Assert.Equal(context.InvocationId, activity.Tags.Single(k => k.Key == "faas.execution").Value);
         }


src/DotNetWorker.Core/Context/TraceContext.cs
```diff
@@ -17,5 +19,10 @@ public abstract class TraceContext
         /// Gets the state data.
         /// </summary>
         public abstract string TraceState { get; }
+
+        /// <summary>
+        /// Gets the attributes associated with the trace.
+        /// </summary>
+        public abstract IReadOnlyDictionary<string, string> Attributes { get; }
```
jviau: Adding a new abstract member is a breaking change. You can either:

1. Make it `virtual` and either throw a NotImplementedException by default or be an empty dictionary.

src/DotNetWorker.Core/Diagnostics/Telemetry/TelemetryProviderBase.cs
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        // Version-specific tags
+        foreach (var kv in GetVersionSpecificAttributes(context))
+        {
+            yield return kv;
+        }
+    }
+
+    protected abstract IEnumerable<KeyValuePair<string, object>> GetVersionSpecificAttributes(FunctionContext context);
```
jviau: This could also be achieved by having the dervied versions re-implement `GetTelemetryAttributes` and then append their own.

``` CSharp
// TelemetryProviderV1_17_0.cs
public IEnumerable<KeyValuePair<string, object>> GetTelemetryAttributes(FunctionContext context)
{
    yield return new(TraceConstants.AttributeAzSchemaUrl, TraceConstants.OpenTelemetrySchemaMap[SchemaVersion]);
    yield return new(TraceConstants.FunctionInvocationIdKey, context.InvocationId);
    yield return new(TraceConstants.FunctionNameKey, context.FunctionDefinition.Name);
    yield return new(TraceConstants.AttributeFaasExecution, context.InvocationId);

    foreach (var kv in base.GetTelemetryAttributes(context))
    {
        yield return kv;
    }
}
```

src/DotNetWorker.Core/Diagnostics/Telemetry/TelemetryProviderV1_17_0.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    protected override OpenTelemetrySchemaVersion SchemaVersion
+        => OpenTelemetrySchemaVersion.V1_17_0;
+
+    protected override ActivityKind Kind
+        => ActivityKind.Server;
+
+    protected override IEnumerable<KeyValuePair<string, object>> GetVersionSpecificAttributes(FunctionContext context)
+    {
+        yield return new(TraceConstants.AttributeAzSchemaUrl, TraceConstants.OpenTelemetrySchemaMap[SchemaVersion]);
```
jviau: consider pulling this value to a static readonly.

src/DotNetWorker.Core/Diagnostics/TraceConstants.cs
```diff
@@ -7,8 +7,15 @@ namespace Microsoft.Azure.Functions.Worker.Diagnostics
 {
     internal class TraceConstants
     {
-        public const string FunctionsActivitySource = "Microsoft.Azure.Functions.Worker";
-        public const string FunctionsActivitySourceVersion = "1.0.0.0";
+        public static IReadOnlyDictionary<OpenTelemetrySchemaVersion, string> OpenTelemetrySchemaMap =
```
jviau: consider nested classes to organize this class a bit.

``` CSharp
internal static class TraceConstants
{
    public static class SchemaVersions
    {
        public const string V1_17_0 = "https://opentelemetry.io/schemas/1.17.0";
        public const string V1_37_0 = "https://opentelemetry.io/schemas/1.37.0";
    } 
}
```

src/DotNetWorker.Core/Hosting/ServiceCollectionExtensions.cs
```diff
@@ -126,6 +123,46 @@ internal static IServiceCollection AddDefaultInputConvertersToWorkerOptions(this
...
+
+                // If OTEL is on and a schema-capability is present, parse it (or throw)
+                if (otelFlag && options.Capabilities.TryGetValue(TraceConstants.WorkerOTelSchemaVersion, out var sv))
+                {
+                    version = ParseSchemaVersion(sv);
+                }
+
+                return version switch
+                {
+                    OpenTelemetrySchemaVersion.V1_37_0 => new TelemetryProviderV1_37_0(),
```
jviau: Consider moving this factory to `TelemetryProvider` itself:

``` CSharp
public abstract class TelemetryProvider
{
    // can even add the parser
    public static TelemetryProvider Create(string schema)
    {
        ...
    }

    public static TelemetryProvider Create(OpenTelemetrySchemaVersion schema)
    {
        ...
    }
}
```

src/DotNetWorker.Core/FunctionsApplication.cs
```diff
@@ -67,29 +68,8 @@ public void LoadFunction(FunctionDefinition definition)
...
-                    activity.ActivityTraceFlags = activityContext.TraceFlags;
-                    activity.TraceStateString = activityContext.TraceState;
-                }
-            }
-
-            var scope = new FunctionInvocationScope(context.FunctionDefinition.Name, context.InvocationId);
-
-            using var logScope = _logger.BeginScope(scope);
-            using Activity? invokeActivity = _functionActivitySourceFactory.StartInvoke(context);
+            using var logScope = _logger.BeginScope(_functionTelemetryProvider.GetTelemetryAttributes(context).ToList());
```
jviau: This is a significant change to log scope. Also what happens if OTel is not enabled here?

src/DotNetWorker.OpenTelemetry/OpenTelemetryConstants.cs
```diff
@@ -12,5 +12,11 @@ internal class OpenTelemetryConstants
         internal const string RegionNameEnvVar = "REGION_NAME";
         internal const string ResourceGroupEnvVar = "WEBSITE_RESOURCE_GROUP";
         internal const string OwnerNameEnvVar = "WEBSITE_OWNER_NAME";
+        internal const string WorkerSchemaVersion = "1.37.0";
```
jviau: This directly ties to OTel schema version to the package. This will prevent us from servicing old schema versions in new releases. Is that acceptable?

src/DotNetWorker.Core/Context/TraceContext.cs
```diff
@@ -23,6 +23,6 @@ public abstract class TraceContext
         /// <summary>
         /// Gets the attributes associated with the trace.
         /// </summary>
-        public abstract IReadOnlyDictionary<string, string> Attributes { get; }
+        public virtual IReadOnlyDictionary<string, string> Attributes => new Dictionary<string, string>();
```
jviau: consider using a common empty dictionary to avoid any allocations when this is not overridden.

Could even use `ImmutableDictionary<string, string>.Empty`

src/DotNetWorker.Core/Diagnostics/Telemetry/TelemetryProvider.cs
```diff
@@ -0,0 +1,86 @@
...
+
+internal abstract class TelemetryProvider : IFunctionTelemetryProvider
+{
+    private static readonly ActivitySource _source
+        = new(TraceConstants.ActivityAttributes.Name, TraceConstants.ActivityAttributes.Version);
+
+    protected abstract OpenTelemetrySchemaVersion SchemaVersion { get; }
+
+    protected abstract ActivityKind Kind { get; }
+
```
jviau: nit extra newlines

src/DotNetWorker.Core/Diagnostics/Telemetry/TelemetryProvider.cs
```diff
@@ -0,0 +1,86 @@
...
+        return Create(version);
+    }
+
+    public static TelemetryProvider Create(OpenTelemetrySchemaVersion version)
+    {
+        return version switch
+        {
+            OpenTelemetrySchemaVersion.V1_17_0 => new TelemetryProviderV1_17_0(),
+            OpenTelemetrySchemaVersion.V1_37_0 => new TelemetryProviderV1_37_0(),
+            _ => throw new InvalidOperationException($"Unsupported OpenTelemetry schema version: {version}")
```
jviau: Consider `ArgumentOutOfRangeException` or `ArgumentException`

src/DotNetWorker.Core/FunctionsApplication.cs
RohitRanjanMS: I’ve kept all existing attributes intact and added new ones according to v17.0 without any breaking changes.
If OpenTelemetry/AppInsights isn’t enabled, the activity will be null, but you’ll always have the scope.
The default specification version is v17.0

src/DotNetWorker.OpenTelemetry/OpenTelemetryConstants.cs
RohitRanjanMS: I’ll add support for new versions as they’re released. Version 1.37 will remain our baseline, and we’ll bump it to a newer base in the next major release.

src/DotNetWorker.Core/Hosting/ServiceCollectionExtensions.cs
```diff
@@ -149,7 +169,7 @@ private static void RunExtensionStartupCode(IFunctionsWorkerApplicationBuilder b
             var startupCodeExecutorInstance =
                 Activator.CreateInstance(startupCodeExecutorInfoAttr.StartupCodeExecutorType) as WorkerExtensionStartup;
             startupCodeExecutorInstance!.Configure(builder);
-        }
+        }
```
jviau: nit remove extra whitespace
```suggestion
        }
```

src/DotNetWorker.Core/Diagnostics/Telemetry/TelemetryProviderV1_37_0.cs
```diff
@@ -0,0 +1,36 @@
...
+
+internal sealed class TelemetryProviderV1_37_0 : TelemetryProvider
+{
+    private static readonly KeyValuePair<string, object> SchemaUrlAttribute =
+        new(TraceConstants.OTelAttributes_1_37_0.SchemaUrl, TraceConstants.OTelAttributes_1_37_0.SchemaVersion);
+    protected override OpenTelemetrySchemaVersion SchemaVersion
+        => OpenTelemetrySchemaVersion.V1_37_0;
+
+    protected override ActivityKind Kind
+        => ActivityKind.Internal;
```
jviau, line 17: Are we assuming the func host emits the `Server` span?

src/DotNetWorker.Core/FunctionsApplication.cs
jviau: Won't this new scope approach be adding OTel specific attributes even when customer is using legacy app insights SDK?

I also wonder if we need all of those in the log scope specifically? Or can they be part of the OTel resource?

src/DotNetWorker.OpenTelemetry/ResourceSemanticConventions.cs
```diff
@@ -14,7 +14,7 @@ internal static class ResourceSemanticConventions
         internal const string CloudProvider = "cloud.provider";
         internal const string CloudPlatform = "cloud.platform";
         internal const string CloudRegion = "cloud.region";
-        internal const string CloudResourceId = "cloud.resource.id";
+        internal const string CloudResourceId = "cloud.resource_id";
```
jviau, line 17: Is this not a schema breaking change?

cjaliaga: Good question. Wasn't always the cloud resource id defined as `cloud.resource_id`? Tried to see if there was a resource change from `cloud.resource.id` to `cloud.resource_id` but didn't find that in the schema specs.

RohitRanjanMS: This package is still in preview.
It was initially cloud.resource.id but was inconsistent with other attributes (which use underscore). I think this changed to cloud.resource_id some time back.

src/DotNetWorker.Core/Diagnostics/Telemetry/TelemetryProviderV1_37_0.cs
RohitRanjanMS: yes, for all the triggers.

src/DotNetWorker.Core/FunctionsApplication.cs
RohitRanjanMS: I've updated the interface to define separate methods for retrieving Scope and Tag attributes, keeping the design straightforward since each set contains distinct attributes.

For customers using legacy Application Insights, telemetry output remains consistent—there are no changes to the emitted attributes.

RohitRanjanMS: The only attribute that could potentially be moved to the resource level is HostInstanceId. However, I don't think it's available during initialization, it can't be captured by the resource detector. Currently, HostInstanceId is supplied as part of the invocation payload.

src/DotNetWorker.OpenTelemetry/ConfigureFunctionsOpenTelemetry.cs
```diff
@@ -18,13 +18,20 @@ public static IOpenTelemetryBuilder UseFunctionsWorkerDefaults(this IOpenTelemet
 
             builder.Services
                 // Tells the host to no longer emit telemetry on behalf of the worker.
-                .Configure<WorkerOptions>(workerOptions => workerOptions.Capabilities["WorkerOpenTelemetryEnabled"] = bool.TrueString);
+                .Configure<WorkerOptions>(workerOptions => workerOptions.Capabilities[OpenTelemetryConstants.WorkerOTelEnabled] = bool.TrueString)
```
scrocquesel: Why are you making a difference between otel and appinsight (with `WorkerApplicationInsightsLoggingEnabled`) to stop the host from emiting telemetry on behalf of the worker ? Which ever worker telemetry solution is chosen should not impact how the host works.
The host is configured with `telemetryMode` in `host.json`. If both the host and the worker MUST be configured with the same telemetry sdk, then it shouldn't be necessary to configure the host or it should be documented. IMHO, a user must have the choice until the openTelemetry implementation is fully working in the host, which is not.

Currently, having `telemetryMode=openTelemetry` in the host is unusable due to a lot of traces and logs that were historicaly filtered.
But then, if we let the host in telemetry mode appInsight, and move the worker to openTelemetry, if we do not set the capabilities `WorkerApplicationInsightsLoggingEnabled` our self in the worker, then we have duplicated traces.

jviau (approved):

src/DotNetWorker.OpenTelemetry/ConfigureFunctionsOpenTelemetry.cs
scrocquesel, line 21: Why are you making a difference between otel and appinsight (with `WorkerApplicationInsightsLoggingEnabled`) to stop the host from emiting telemetry on behalf of the worker ? Which ever worker telemetry solution is chosen should not impact how the host works.
The host is configured with `telemetryMode` in `host.json`. If both the host and the worker MUST be configured with the same telemetry sdk, then it shouldn't be necessary to configure the host or it should be documented. IMHO, a user must have the choice until the openTelemetry implementation is fully working in the host, which is not.

Currently, having `telemetryMode=openTelemetry` in the host is unusable due to a lot of traces and logs that were historicaly filtered.
But then, if we let the host in telemetry mode appInsight, and move the worker to openTelemetry, if we do not set the capabilities `WorkerApplicationInsightsLoggingEnabled` our self in the worker, then we have duplicated traces.

RohitRanjanMS: Hi @scrocquesel , thanks for reviewing the PR. The change is intentional—legacy Application Insights integration is in sustainment mode, and even the Azure Monitor team advises against using it. I’ve made efforts to reduce noise and filter out irrelevant telemetry. Core tools may take time to reflect this, but if you run it in Azure, you’ll notice the difference. This has a direct impact on monitoring costs, which I’m mindful of. We’re committed to improving the OpenTelemetry experience—please let me know if you spot any noisy or unnecessary telemetry, and I’ll look into it.

scrocquesel: > Hi @scrocquesel , thanks for reviewing the PR. The change is intentional—legacy Application Insights integration is in sustainment mode, and even the Azure Monitor team advises against using it. I’ve made efforts to reduce noise and filter out irrelevant telemetry. Core tools may take time to reflect this, but if you run it in Azure, you’ll notice the difference. This has a direct impact on monitoring costs, which I’m mindful of. We’re committed to improving the OpenTelemetry experience—please let me know if you spot any noisy or unnecessary telemetry, and I’ll look into it.

From the worker, with this PR, it is good. But from the host in default mode (appinsight), we have the issue of duplicate traces. We can move to open telemetry but then, there is lot of dependencies traces from azure client SDK, there is also requests for warm-up and some undocumented reccurent admin/status request traces to name a few. I prefer to stay with appinsight mode for now and just set the aforementioned capability to avoid duplicates trace and log while the worker is using otel. IMHO, this is actually the best experience.
I think I will move to otel for the host when we will be able to filter traces per trigger.

RohitRanjanMS: I have removed some of these - https://github.com/Azure/azure-functions-host/pull/11429/files#diff-c3d0241de259b8202b42b93c7e43f7e12e195a5332c31ef7bc4eabd7f2365dd2

RohitRanjanMS: If you notice any issues with the latest release, please feel free to create an issue—I'd be happy to take a look.

PR #3200, opened 2025-10-16
Enable IFunctionExecutor in InvocationFeatures

### LINKED ISSUE #1666: Allow for extensions to wrap function invocation delegate
# Background

Durable functions for dotnet isolated has a need for wrapping all `[OrchestrationTrigger]` function invocations with additional logic. Today we perform this wrapping via [middleware](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/Worker.Extensions.DurableTask/DurableTaskFunctionsMiddleware.cs). However, it was discovered this is extremely fragile for durables case because we need to tightly control the precise async behavior in our logic. This has led to [non-determinism issues](https://github.com/microsoft/durabletask-dotnet/issues/153) in durable, which arise from the customer introducing _any custom middleware_ between the Durable middleware and the function invocation. This makes orchestrations fragile and is a less than ideal experience for customers to put it lightly.

# Proposed Solution

Provide an API where an extension can wrap a function invocation's delegate with a custom delegate. This would most make sense to be doable as part of the middleware. So durable could retain its middleware, but instead of wrapping the rest of the pipeline it performs _some_ action on the `FunctionContext` (maybe a new `IInvocationFeatures` from `.Features`)?

# Alternative Solutions

## Controlling middleware order

If durable could control its middleware order, it could set its middleware to be the last. However, this is fragile as any other extension could come along and also declare its middleware to be last, and then we are back to square one with this issue.

## Replacing function definition invocation 

This would rely on introducing an extensibility model to function definition metadata generation and is most likely overkill for this scenario.

jviau:

resolves #1666

### Pull request checklist

### Additional information

This PR allows for an `IFunctionExecutor` implementation to be supplied on a per-invocation bases via invocation features. That feature will **not** be set by default. When it is not present, the existing behavior remains of using the DI registered `IFunctionExecutor`. This allows for extensions to selective override the execution behavior of specific invocations.

### DIFF
--- a/test/DotNetWorkerTests/FunctionMetadata/DefaultFunctionMetadataManagerTests.cs
+++ b/test/DotNetWorkerTests/FunctionMetadata/DefaultFunctionMetadataManagerTests.cs
@@ -7,6 +7,8 @@ using Microsoft.Extensions.Logging;
 using Moq;
 using Xunit;
 
+#nullable enable
+
 namespace Microsoft.Azure.Functions.Worker.Tests.FunctionMetadata
 {
     public class DefaultFunctionMetadataManagerTests
@@ -70,7 +72,7 @@ namespace Microsoft.Azure.Functions.Worker.Tests.FunctionMetadata
 
             var manager = new DefaultFunctionMetadataManager(
                 mockProvider.Object,
-                new[] { mockTransformer.Object },
+                [mockTransformer.Object],
                 mockLogger.Object);
 
             var ex = await Assert.ThrowsAsync<InvalidOperationException>(() => manager.GetFunctionMetadataAsync("test"));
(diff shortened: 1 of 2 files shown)


src/DotNetWorker.Core/Pipeline/FunctionExecutionMiddleware.cs
```diff
@@ -1,22 +1,24 @@
...
-        public FunctionExecutionMiddleware(IFunctionExecutor functionExecutor)
-        {
-            _functionExecutor = functionExecutor;
-        }
+        private readonly IFunctionExecutor _functionExecutor = functionExecutor
+            ?? throw new ArgumentNullException(nameof(functionExecutor));
 
         public Task Invoke(FunctionContext context)
         {
+            if (context.Features.Get<IFunctionExecutor>() is { } executor)
```
kshyju, line 17: So, extensions authors (or even function app developers) have to set their `IFunctionExecutor` implementation in a middleware which gets executed before `FunctionExecutionMiddleware` per invocation? Or is there another way these can be registered (like during app bootstrapping? )

jviau: Yes, it is set during middleware. FuctionExecutionMiddleware should always be the _last_ middleware in the pipeline, so any customer middleware will always have a chance to set this.

We will not support some other way to register this during app bootstraping - middleware will be the only way. Which I am happy with - this is the most flexible for customers and honestly not much work for them.

kshyju (approved):

src/DotNetWorker.Core/Pipeline/FunctionExecutionMiddleware.cs
kshyju: So, they have to handle everything needed for execution inside their `IFunctionExecutor` implementation (such as input binding etc..).

PR #3295, opened 2026-01-15
[MSBUILD SDK] Add dev.ps1 script for local-dev of Azure.Functions.Sdk

jviau:
Quick PR to improve local dev loop. Adds a ps1 script for a quick way to bootstrap the SDK for local dev. Also supports `dotnet watch` on the resolver csproj.

### DIFF
--- a/src/Azure.Functions.Sdk/dev.ps1
+++ b/src/Azure.Functions.Sdk/dev.ps1
@@ -23,7 +23,7 @@ function Find-Root
 $root = Find-Root
 $resolver = Join-Path -Path $root -ChildPath "test/Azure.Functions.Sdk.Resolver/Azure.Functions.Sdk.Resolver.csproj"
 
-$command = "dotnet publish $resolver"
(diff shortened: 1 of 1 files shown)


src/Azure.Functions.Sdk/dev.ps1
```diff
@@ -0,0 +1,49 @@
...
+$config = "-c debug"
+if ($Release)
+{
+    $config = "-c release"
+}
+
+$command += " $config"
+
+# Use IEX to split '$config' into separate switch and arg.
+$publishDir = Invoke-Expression "dotnet build $resolver $config -getproperty:PublishDir"
```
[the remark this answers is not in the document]
jviau: The value will be the same between the two.

RohitRanjanMS: I think this won't work if $root contains spaces.

jviau: Enclosed in quotes.

PR #3303, opened 2026-01-22
Propagate tags from worker to host

satvu:

Related to https://github.com/Azure/azure-functions-host/issues/11574 and the [PR](https://github.com/Azure/azure-functions-host/pull/11575/changes) resolving it.

Depends on https://github.com/Azure/azure-functions-language-worker-protobuf/pull/98. 

### Pull request checklist

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -6,12 +6,13 @@
 
 ### Microsoft.Azure.Functions.Worker (metapackage) <version>
 
+- Update `Microsoft.Azure.Functions.Worker.Core` to <version>
 - Update `Microsoft.Azure.Functions.Worker.Grpc` to <version>
 
 ### Microsoft.Azure.Functions.Worker.Core <version>
 
-- <entry>
+- Add support for propagating trace context tags from worker to host (#3303)
 
 ### Microsoft.Azure.Functions.Worker.Grpc <version>
 
-- Update protobuf version to <version> and add support for propagating tags from the worker to the functions host (#3303).
+- Update protobuf version to v1.12.0-protofile and add support for propagating tags from the worker to the functions host (#3303).


src/DotNetWorker.Core/Diagnostics/TraceConstants.cs
```diff
@@ -40,8 +42,22 @@ public static class OTelAttributes_1_37_0
...
+            OTelAttributes_1_37_0.InvocationId,
+            OTelAttributes_1_37_0.FunctionName,
+            OTelAttributes_1_37_0.Instance,
+            OTelAttributes_1_37_0.SchemaUrl,
+        };
+    }
+
     public static class InternalKeys
     {
+        public const string FunctionContextItemsKey = "TagsForFunctionActivity";
```
kshyju: Should we use a product prefix? ex: `AzureFunctions_` ?

kshyju: How about `ActivityTags` instead of `TagsForFunctionActivity`?

src/DotNetWorker.Core/Diagnostics/TraceConstants.cs
```diff
@@ -40,8 +42,22 @@ public static class OTelAttributes_1_37_0
         public const string SchemaVersion = "https://opentelemetry.io/schemas/1.37.0";
     }
 
+    public static class KnownAttributes
+    {
+        public static IReadOnlyList<string> All { get; } = new List<string>
```
kshyju: Change this to a HashSet so that the `Contains` call in `AddTraceContextTags` method becomes O(1) lookup instead O(N).

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
```diff
@@ -163,5 +170,29 @@ public bool TryCancel(string invocationId)
...
+                Attributes = { }
+            };
+
+            var invocationTags = context.Items.TryGetValue(TraceConstants.InternalKeys.FunctionContextItemsKey, out var tagsObj)
+                ? tagsObj as System.Collections.Generic.IDictionary<string, string>
+                : null;
+
+            if (invocationTags is not null)
+            {
+                foreach (var tag in invocationTags.Where(tag => !TraceConstants.KnownAttributes.All.Contains(tag.Key)))
```
kshyju: You can avoid the Linq overhead (from the `Where` method) by switching to a simple loop. Minor perf gain.

```
var known = TraceConstants.KnownAttributes.All;

foreach (var (key, value) in tags)
{
    if (!known.Contains(key))
    {
        traceContext.Attributes[key] = value;
    }
}
```

src/DotNetWorker.Core/Diagnostics/TraceConstants.cs
kshyju: Use `FrozenSet`, that is even better for perf than hashset. (.NET8 onwards only)

src/DotNetWorker.Core/FunctionsApplication.cs
```diff
@@ -74,6 +75,16 @@ public async Task InvokeFunctionAsync(FunctionContext context)
...
             {
                 await _functionExecutionDelegate(context);
+
+                var tags = invokeActivity?.Tags
+                    .GroupBy(kv => kv.Key)
+                    .ToDictionary(g => g.Key, g => g.Last().Value);
+
+                if (tags is not null)
+                {
+                    context.Items ??= new ConcurrentDictionary<object, object>();
```
satvu: Added this which resolved failing tests - during manual e2e testing, this wasn't needed because DefaultFunctionContext is initialized with Items set to `Dictionary<string, string>`. https://github.com/Azure/azure-functions-dotnet-worker/blob/63aa3583be998f2e3d0a974cf916bb0a4899f3d4/src/DotNetWorker.Core/Context/DefaultFunctionContext.cs#L36

In tests, the mock setup was using interfaces of the factory classes instead, so `context.Items` was `null` and this threw an error. This line also uses a `ConcurrentDictionary` while the default context class uses just `Dictionary` ... the items field is mostly used for middleware and the ASP.NET integration flow. The latter makes me slightly concerned about thread-safe access.

jviau: AspNetCore's HttpContext.Items appears to just be a regular dictionary under the hood. I don't think there is an expectation of thread safety for this dictionary, because it serves a _single_ invocation. We don't expect invocations to be.

With that said, I don't think this place should be initializing this dictionary -- it is not its responsibility to do that. Please update the tests to set it up correctly instead.

src/DotNetWorker.Core/FunctionsApplication.cs
```diff
@@ -74,6 +75,16 @@ public async Task InvokeFunctionAsync(FunctionContext context)
             try
             {
                 await _functionExecutionDelegate(context);
+
+                var tags = invokeActivity?.Tags
```
jviau: small nit: LINQ is elegant and all, but its performance can be 💩 . I often just use a `foreach` loop.

But this also begs the question: do we even need to transform to a dictionary? Can we just grab the Tags and store that in `Items` directly? Worth an investigation to see if the tags can outlive the activity (ie: still accessible post activity disposal).

release_notes.md
```diff
@@ -6,12 +6,12 @@
 
 ### Microsoft.Azure.Functions.Worker (metapackage) <version>
 
-- <entry>
+- Update `Microsoft.Azure.Functions.Worker.Grpc` to <version>
```
satvu, line 10: These changes can't be merged/released until the host changes are out. Putting in placeholders for now for the release notes.

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
```diff
@@ -163,5 +169,30 @@ public bool TryCancel(string invocationId)
...
         }
+
+        private void AddTraceContextTags(InvocationResponse response, FunctionContext context)
+        {
+            if (context.Items is null)
+            {
+                return;
+            }
+
+            var tags = context.Items.TryGetValue(TraceConstants.InternalKeys.FunctionContextItemsKey, out var tagsObj)
```
jviau: I wonder if the `Activity.Current` is available at this point? Can we just access it directly instead of needing to access `Items`?

satvu: It's unfortunately not available at this point (it is `null`), leading to this workaround.

src/DotNetWorker.Core/Diagnostics/TraceConstants.cs
kshyju: Did you consider FrozenSet?

src/DotNetWorker.Core/FunctionsApplication.cs
satvu: Originally converted to dictionary w/ concerns over availability post disposal - discussed offline and tested, this should be a low risk and tags will be used directly.

src/DotNetWorker.Core/Diagnostics/TraceConstants.cs
satvu: This project still targets net6.0 and netstandard2.0 where FrozenSet is not available (but HashSet is)

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
```diff
@@ -189,6 +188,14 @@ private void AddTraceContextTags(InvocationResponse response, FunctionContext co
                 {
                     if (!known.Contains(key))
                     {
+                        // Duplicate tags can exist for an Activity, but the gRPC message type
+                        // Map<string, string>  does not allow for duplicate keys.
+                        // Log and overwrite in this scenario.
+                        if (response.TraceContextAttributes.ContainsKey(key))
+                        {
+                            _logger.LogDebug("Trace context attribute '{key}' already exists on the invocation response. Overwriting value.", key);
```
jviau: This is giving me flashbacks of an issue we just had. We had the exact same log statement about duplicate tags, but it was coupled with another bug that caused the tag list to grow indefinitely with duplicate values.

End result, the log statement tanked performance to the point the app effectively hung.

So, do we really want this? Or should it be `LogTrace` (the most minimal verbosity).

satvu: I'm good with changing this to the most minimal verbosity - just felt that this should be logged somehow? Or if it isn't logged, it should be called out in the documentation for adding custom tags since this different from the regular behavior of adding tags to an `Activity` (which allows duplicates).

jviau: I feel that documenting "last added duplicate tag wins" is sufficient. I _think_ `Activity` allows duplicates for performance reasons. Instead of scanning and replacing duplicates, just append the new tag and let duplicates be handled later.

This would be the time to handle duplicates, which I think last-tag wins is good enough. No log needed in my opinion.

src/DotNetWorker.Core/Diagnostics/TraceConstants.cs
```diff
@@ -40,8 +43,21 @@ public static class OTelAttributes_1_37_0
         public const string SchemaVersion = "https://opentelemetry.io/schemas/1.37.0";
     }
 
+    public static class KnownAttributes
+    {
+        public static ImmutableHashSet<string> All { get; } = ImmutableHashSet.Create<string>(
```
kshyju: Could we add summary comments explaining what this collection is returning?

src/DotNetWorker.Core/Diagnostics/TraceConstants.cs
```diff
@@ -1,6 +1,9 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using System.Collections.Generic;
```
kshyju: Remove unused.

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
```diff
@@ -163,5 +168,30 @@ public bool TryCancel(string invocationId)
 
             return false;
         }
+
+        private void AddTraceContextTags(InvocationResponse response, FunctionContext context)
```
kshyju: mark as `static` method.

src/DotNetWorker.Core/FunctionsApplication.cs
```diff
@@ -74,6 +74,13 @@ public async Task InvokeFunctionAsync(FunctionContext context)
             try
             {
                 await _functionExecutionDelegate(context);
+
+                var tags = invokeActivity?.Tags;
+
+                if (tags is not null && context.Items is not null)
```
kshyju: Should this be done in a `finally` block? Are there cases where this is being accessed somewhere in the pipeline (ex: a middleware) even in the case of failure?

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
```diff
@@ -163,5 +168,30 @@ public bool TryCancel(string invocationId)
...
+            if (context.Items is null)
+            {
+                return;
+            }
+
+            var tags = context.Items.TryGetValue(TraceConstants.InternalKeys.FunctionContextItemsKey, out var tagsObj)
+                ? tagsObj as IEnumerable<KeyValuePair<string, string>>
+                : null;
+
+            if (tags is not null)
```
kshyju: nit: return if tags is null to reduce nesting.

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
```diff
@@ -163,5 +168,30 @@ public bool TryCancel(string invocationId)
...
+            {
+                var known = TraceConstants.KnownAttributes.All;
+
+                foreach (var (key, value) in tags)
+                {
+                    if (!known.Contains(key))
```
kshyju: should we add a comment saying we are not overwriting known/system tags?

src/DotNetWorker.Core/FunctionsApplication.cs
jviau: Great point. A finally block would be better.

src/DotNetWorker.Core/FunctionsApplication.cs
```diff
@@ -83,6 +83,15 @@ public async Task InvokeFunctionAsync(FunctionContext context)
 
                 throw;
             }
+            finally
+            {
+                var tags = invokeActivity?.Tags;
+
+                if (tags is not null && context.Items is not null)
+                {
+                    context.Items[TraceConstants.InternalKeys.FunctionContextItemsKey] = tags;
```
RohitRanjanMS: Activity.Tags is a lazy iterator, those tags remain referenced even after the activity is disposed, preventing GC. I don’t see this as a major issue since the context goes out of scope soon, but we can still optimize by only performing this work when we detect a tag that isn’t part of our known set. That way, if the feature isn’t being used, the invocation handler has essentially no overhead.

src/DotNetWorker.Core/Diagnostics/TraceConstants.cs
```diff
@@ -40,8 +42,24 @@ public static class OTelAttributes_1_37_0
         public const string SchemaVersion = "https://opentelemetry.io/schemas/1.37.0";
     }
 
+    public static class KnownAttributes
+    {
+        /// <summary>
+        /// Returns protected attribute names that are set by Azure functions that should not be overriden.
+        /// </summary>
+        public static ImmutableHashSet<string> All { get; } = ImmutableHashSet.Create<string>(
```
RohitRanjanMS: This should also include all the attributes from the `TraceConstants.InternalKeys`.

src/DotNetWorker.Core/Diagnostics/TraceConstants.cs
```diff
@@ -27,7 +33,8 @@ public static class OTelAttributes_1_17_0
         // v1.17.0
         public const string InvocationId = "faas.execution";
         public const string SchemaUrl = "az.schema_url";
-        public const string SchemaVersion = "https://opentelemetry.io/schemas/1.17.0";
+
+        internal static readonly string[] All = { InvocationId, SchemaUrl };
```
jviau: Array itself is not readonly. For readonly collections, I do one of the following:

1. Expose as `IReadOnlyCollection<T>`. Implementation depends on usage. If I will be running `.Contains`, then I will use `HashSet<T>` as the implementation.
2. Expose an immutable collection. Whichever type you think is fitting.

src/DotNetWorker.Core/Diagnostics/TraceConstants.cs
```diff
@@ -1,11 +1,17 @@
...
+using System.Collections.Immutable;
+using System.Linq;
+
 namespace Microsoft.Azure.Functions.Worker.Diagnostics;
 
 internal static class TraceConstants
 {
-   public static class ActivityAttributes
+    public const string OTel_1_37_0_SchemaVersion = "https://opentelemetry.io/schemas/1.37.0";
+    public const string OTel_1_17_0_SchemaVersion = "https://opentelemetry.io/schemas/1.17.0";
```
jviau, line 12: Curious why you moved these?

src/DotNetWorker.Core/Diagnostics/TraceConstants.cs
```diff
@@ -37,7 +44,20 @@ public static class OTelAttributes_1_37_0
...
+
+        internal static readonly string[] All = { InvocationId, FunctionName, Instance, SchemaUrl };
+    }
+
+    public static class KnownAttributes
+    {
+        /// <summary>
+        /// Returns protected attribute names that are set by Azure functions that should not be overriden.
+        /// </summary>
+        public static readonly ImmutableHashSet<string> All =
```
jviau, line 56: I wonder if collection initializer would work also? No need to change if you like what you have, but just so you are aware:

``` CSharp
public static readonly ImmutableHashSet<string> All =
    [ ..OTelAttributes_1_17_0.All, ..OTelAttributes_1_37_0.All, ..InternalKeys.All ];
```

src/DotNetWorker.Core/FunctionsApplication.cs
```diff
@@ -3,10 +3,12 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using Azure;
```
RohitRanjanMS: This seems unused.

src/DotNetWorker.Core/FunctionsApplication.cs
```diff
@@ -83,6 +85,30 @@ public async Task InvokeFunctionAsync(FunctionContext context)
...
                 throw;
             }
+            finally
+            {
+                var tags = invokeActivity?.Tags;
+
+                if (tags is not null && context.Items is not null)
+                {
+                    var known = TraceConstants.KnownAttributes.All;
+                    var validTags = new List<KeyValuePair<string, string>>();
```
RohitRanjanMS: We can make validTags lazy. Only allocate `validTags` if needed, check if any non-known tags exist, and if yes, then allocate. You can also try to guess the size of the list to avoid resizing.

src/DotNetWorker.Core/Diagnostics/TraceConstants.cs
satvu: We don't want to count them as part of the Otel attributes checked for when filtering tags, so this was moved out since we test # of attributes against the `All` array to catch when new attributes are added. cc: @RohitRanjanMS

jviau (approved):

PR #3319, opened 2026-02-07
Add support for baggage propagation

### LINKED ISSUE #3338: [Otel] Add support for baggage propagation
Add support for baggage propagation in the .NET worker. Related to https://github.com/Azure/azure-functions-host/issues/11026 which tracks the host support for this feature.

satvu:
resolves #3338

Related to the following issue: https://github.com/Azure/azure-functions-host/issues/11026

Related host PR: https://github.com/Azure/azure-functions-host/pull/11598

### Pull request checklist

### Additional Details

If this runs with an older version of the host (that does not propagate baggage), this will no-op because no baggage values are passed over gRPC and the worker treats the situation as there is no baggage to handle/propagate. 

If new host runs with old version of worker (without this change), then the host populates the gRPC message, but worker never does anything with it so that is also a no-op w/o breaking changes.

### DIFF
--- a/src/DotNetWorker.OpenTelemetry/DefaultBaggagePropagator.cs
+++ b/src/DotNetWorker.OpenTelemetry/DefaultBaggagePropagator.cs
@@ -17,11 +17,17 @@ namespace Microsoft.Azure.Functions.Worker.OpenTelemetry
                 Baggage.SetBaggage(kv.Key, kv.Value);
             }
 
-            return new BaggageScope();      
+            return BaggageScope.Instance;      
         }
 
         private sealed class BaggageScope : IDisposable
         {
+            public static readonly BaggageScope Instance = new BaggageScope();
+
+            private BaggageScope()
+            {
+            }
+
             public void Dispose()
             {
                 Baggage.ClearBaggage();


protos/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
```diff
@@ -406,6 +406,9 @@ message RpcTraceContext {
 
   // This corresponds to Activity.Current?.Tags
   map<string, string> attributes = 3;
+
+  // This corresponds to Activity.Current?.Baggage
+  map<string, string> baggage = 4;
```
satvu, line 411: TODO: pull in final protobuf changes later

Here's the PR: https://github.com/Azure/azure-functions-language-worker-protobuf/pull/99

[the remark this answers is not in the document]
satvu: Could define a new type that has key/value in gRPC (let's call it KeyValuePair) and add a field called `repeated KeyValuePair`, but according to [OpenTelemetry](https://opentelemetry.io/docs/specs/otel/baggage/api/), "Each name in Baggage MUST be associated with exactly one value."

We will use a last one wins strategy

src/DotNetWorker.OpenTelemetry/OpenTelemetryWorkerBuilderExtensions.cs
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Hosting;
+
+namespace Microsoft.Azure.Functions.Worker.OpenTelemetry
+{
+    public static class OpenTelemetryWorkerBuilderExtensions
+    {
+        public static IFunctionsWorkerApplicationBuilder EnableBaggagePropagation(this IFunctionsWorkerApplicationBuilder builder)
```
jviau: Is the expectation that customers call this manually? If so, there is going to be an ordering problem with it. Extension auto-registration will already have ran by the time a customer calls this, and so the baggage middleware will always be _after_ any extension middleware. This means extension middleware misses out on all the baggage.

I think we will need a way to insert this at the start. Do we have middleware insert semantics? If no, we might need a way to supply this to the app builder so it gets registered before any extensions.

src/DotNetWorker.OpenTelemetry/TraceConstants.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.Functions.Worker.OpenTelemetry
+{
+    internal static class TraceConstants
+    {
+        public const string BaggageKeyName = "AzureFunctions_TraceContext_Baggage";
```
jviau: This key is defined twice. Any chance it can be consolidated? Or is it an access issue?

satvu: It's an access issue. If the constant is moved to something that can be shared by `DotNetWorker.Core` and the Otel package, then it can be consolidated but they don't currently share a dependency.

The HTTP packages reference the core package so it wouldn't be new for the Otel package to do the same (and have the key conslidated).

src/DotNetWorker.OpenTelemetry/OpenTelemetryWorkerBuilderExtensions.cs
satvu: Yes, this would be called manually in this design - let me see if there's anything existing that we can leverage to make sure this goes before extensions.

jviau: I wonder if we need this to be a middleware approach at all. What if this sets a capability which we flow back to the host, and that turns host flowing baggage on/off. And then our implementation worker side can always populate baggage.

src/DotNetWorker.OpenTelemetry/TraceConstants.cs
jviau: I would ask @RohitRanjanMS if having OTel package reference core is acceptable or not.

src/DotNetWorker.OpenTelemetry/OpenTelemetryWorkerBuilderExtensions.cs
kshyju: What if we set this in [FunctionsApplication.InvokeFunctionAsync](https://github.com/Azure/azure-functions-dotnet-worker/blob/01793bab0251e2887a4b1c4667846d501862996c/src/DotNetWorker.Core/FunctionsApplication.cs#L69) method , before calling `_functionExecutionDelegate`? That way baggage is available to all middlewares. We are doing similar thing to setup Activity for the invocation there.

I like the capability idea @jviau suggested above. With this, we can send this capability with a new version of worker package (_with an opt-out feature if necessary_) and host can send this or not based on it. (Yea, this requires another host update, but will provide flexibility for workers to control this behavior).

Baggage comes from OTEL package and we can abstract this behavior (to set baggage) as a Feature and DotnetWorker.Otel project/package can implement the code (to set it if present in the invocation request)

src/DotNetWorker.OpenTelemetry/BaggageMiddleware.cs
```diff
@@ -0,0 +1,36 @@
...
+        public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next)
+        {
+            try
+            {
+                if (context.Items.TryGetValue(TraceConstants.BaggageKeyName, out var value) &&
+                    value is IEnumerable<KeyValuePair<string, string>> dict)
+                {
+                    foreach (var kv in dict)
+                    {
+                        Baggage.SetBaggage(kv.Key, kv.Value);
```
kshyju: Could `kv.Key` be `null`? Do we need a null check before using it for SetBaggage method?

src/DotNetWorker.OpenTelemetry/TraceConstants.cs
satvu: Refactored to use existing `TraceContext` abstraction - no longer using Items collection

src/DotNetWorker.OpenTelemetry/BaggageMiddleware.cs
satvu: No longer using Items, leveraging `TraceContext` instead and passing Baggage along through [here](https://github.com/Azure/azure-functions-dotnet-worker/pull/3319/changes#diff-9fec3ab7a75fb9297b1b586f41b999a7b5d5590c5f9d4c714bdafbf4b1379751). In grpc, keys cannot be `null` and this is passed is as map<string, string>.

src/DotNetWorker.OpenTelemetry/OpenTelemetryWorkerBuilderExtensions.cs
satvu: My first iteration used that approach (set baggage inside `InvokeFunctionAsync`) but it would require taking an otel dependency in core.

Moving the baggage propagation into middleware inside the otel extension was to avoid taking an otel dependency in core (not to control enablement/disablement). The host PR that was merged flows the baggage any time the conditions necessary are met (otel enabled, baggage exists) since we didn't have a strong case for disabling baggage (if those values are sent, why should they be blocked).

src/DotNetWorker.Grpc/Handlers/InvocationHandler.cs
```diff
@@ -7,6 +7,7 @@
 using System.Threading.Tasks;
 using Microsoft.Azure.Functions.Worker.Context.Features;
 using Microsoft.Azure.Functions.Worker.Core;
+using Microsoft.Azure.Functions.Worker.Diagnostics;
```
RohitRanjanMS, line 11: Unused namespace.

src/DotNetWorker.OpenTelemetry/OpenTelemetryWorkerBuilderExtensions.cs
kshyju: >but it would require taking an otel dependency in core.

You can create a `IBaggagePropogator` feature in core and have OTEL project have the implementation of that interface/abstraction.

jviau: If we are okay with this being always-on, no way to disable from the customer, then @kshyju suggestion sounds great to me.

If we are worried about making it configurable, we can look into adding a way to disable it in the app startup. But maybe it's best to have on-by-default behavior.

RohitRanjanMS: Another option is to attach a listener to the internal activity and set Baggage.Current in the OnStart event, but we still don’t have a clean way to propagate that baggage to the ActivitySource listener. This approach also has a limitation: Baggage.Current won’t be visible to a span processor because it must be set before the activity starts. 

Let’s go with @kshyju's suggestion instead—it’s much simpler to set the baggage before starting the activity.

jviau: @RohitRanjanMS I don't think we can go with anything activity dependent here, because activities may be sampled out or disabled. Baggage needs to run all the time, regardless of activity state.

src/DotNetWorker.Core/Diagnostics/Telemetry/IBaggagePropagator.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// underlying system.</remarks>
+        /// <param name="baggage">The collection of key-value pairs representing the baggage items to be set. Each key must be a non-empty
+        /// string, and the corresponding value can be null or an empty string.</param>
+        public void SetBaggage(IEnumerable<KeyValuePair<string, string>> baggage);
+
+        /// <summary>
+        /// Clears the specified baggage items from the current context.
+        /// </summary>
+        /// <param name="baggage">The collection of key-value pairs representing the baggage items to be cleared.</param>
+        void ClearBaggage(IEnumerable<KeyValuePair<string, string>> baggage);
```
jviau: Does the baggage need to be explicitly supplied here? The OTel API usage you had just called `Baggage.ClearBaggage()`, it didn't provide anything explicit.

src/DotNetWorker.Core/Diagnostics/Telemetry/IBaggagePropagator.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// <summary>
+        /// Sets the baggage items for the current context, allowing for additional metadata to be associated with the
+        /// operation.
+        /// </summary>
+        /// <remarks>This method can be used to attach contextual information that may be useful for
+        /// tracing or logging purposes. Ensure that the baggage items do not exceed any size limits imposed by the
+        /// underlying system.</remarks>
+        /// <param name="baggage">The collection of key-value pairs representing the baggage items to be set. Each key must be a non-empty
+        /// string, and the corresponding value can be null or an empty string.</param>
+        public void SetBaggage(IEnumerable<KeyValuePair<string, string>> baggage);
```
jviau: If you return an `IDisposable?` here, which clears the baggage on disposal, then you could simplify usage via `using (propagator.SetBaggage(...)`

Also by being nullable, implementors can opt to return `null` if they have no baggage clearing semantics.

satvu: Implemented this

src/DotNetWorker.OpenTelemetry/DefaultBaggagePropagator.cs
```diff
@@ -0,0 +1,31 @@
...
+    internal class DefaultBaggagePropagator : IBaggagePropagator
+    {
+        public IDisposable? SetBaggage(IEnumerable<KeyValuePair<string, string>> baggage)
+        {
+            foreach (var kv in baggage)
+            {
+                Baggage.SetBaggage(kv.Key, kv.Value);
+            }
+
+            return new BaggageScope();
```
jviau: Could use a singleton instance as well since it has no state itself.

jviau (approved):

PR #3327, opened 2026-02-21
Improve/fix input binding performance in JsonPocoConverter

### LINKED ISSUE #3272: FunctionInputConverterException when binding integer path parameters with leading zeros in Isolated Worker
### Description

I am encountering an issue when binding a route parameter to an int argument in an Azure Function using the Isolated Worker model. If the path parameter in the URL contains a leading zero (e.g., 01), the binding fails with a FunctionInputConverterException wrapping a System.Text.Json.JsonException.

It appears that the input converter is treating the path parameter as a JSON number and System.Text.Json is rejecting the leading zero, whereas standard URL routing usually handles this gracefully by parsing the integer value.

### Steps to reproduce

1. Create a new Azure Functions project using the Isolated Worker model (.NET 8).
2. Create a function with an HTTP trigger and a route constraint for an integer:

  ```csharp
 [Function("Function1")]
public IActionResult Run([HttpTrigger(AuthorizationLevel.Function, "get", "post", Route = "test/{id:int}")] HttpRequest req, int id)
{
    return new OkObjectResult("Welcome to Azure Functions!");
}
  ```

3. Run the function locally.
4. Send a request with a leading zero in the ID: `curl -X GET "http://localhost:7071/api/test/01"`

**Expected behavior** 

The function should execute successfully, and the id parameter should be bound to the integer value 1.

**Actual behavior**

The function invocation fails with the following exception:

```
[2025-12-03T16:24:39.551Z] Function 'Function1', Invocation id '...': An exception was thrown by the invocation.
Type:
Exception: Microsoft.Azure.Functions.Worker.FunctionInputConverterException: Error converting 1 input parameters for Function 'Function1': Cannot convert input parameter 'id' to type 'System.Int32' from type 'System.String'. Error:System.Text.Json.JsonException: Invalid leading zero before '0'. Path: $ | LineNumber: 0 | BytePositionInLine: 1.
[2025-12-03T16:24:39.557Z]  ---> System.Text.Json.JsonReaderException: Invalid leading zero before '0'. LineNumber: 0 | BytePositionInLine: 1.
   at System.Text.Json.ThrowHelper.ThrowJsonReaderException(Utf8JsonReader& json, ExceptionResource resource, Byte nextByte, ReadOnlySpan`1 bytes)
   at System.Text.Json.Utf8JsonReader.ConsumeZero(ReadOnlySpan`1& data, Int32& i)[2025-12-03T16:24:39.551Z] Function 'Function1', Invocation id '...': An exception was thrown by the invocation.
Type:
Exception: Microsoft.Azure.Functions.Worker.FunctionInputConverterException: Error converting 1 input parameters for Function 'Function1': Cannot convert input parameter 'id' to type 'System.Int32' from type 'System.String'. Error:System.Text.Json.JsonException: Invalid leading zero before '0'. Path: $ | LineNumber: 0 | BytePositionInLine: 1.
[2025-12-03T16:24:39.557Z]  ---> System.Text.Json.JsonReaderException: Invalid leading zero before '0'. LineNumber: 0 | BytePositionInLine: 1.
   at System.Text.Json.ThrowHelper.ThrowJsonReaderException(Utf8JsonReader& json, ExceptionResource resource, Byte nextByte, ReadOnlySpan`1 bytes)
   at System.Text.Json.Utf8JsonReader.ConsumeZero(ReadOnlySpan`1& data, Int32& i)
```

### Environment
- **Worker Model:** Isolated Worker
- **Framework:** .NET 8.0
- **Package version:** :
  - Microsoft.Azure.Functions.Worker, Version "2.51.0"
  - Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore, Version "2.1.0"
  - Microsoft.Azure.Functions.Worker.Sdk, Version "2.0.7"

victor-lubn:

Resolves #3272

This PR improves primitive input binding performance in the worker input conversion pipeline by introducing a fast-path for primitive numeric types in `JsonPocoConverter`.

Primitive values originating from bindings may already represent scalar values but were previously always processed through JSON deserialization. This change adds direct parsing for primitive numeric types while preserving existing serializer-based behavior as a fallback.

Additionally, this change resolves an issue where numeric values containing leading zeros (for example `"01"`) could fail primitive binding when processed through JSON deserialization #3272. Direct primitive parsing aligns the behavior with standard .NET numeric parsing semantics and correctly binds such values to their target numeric types.

The PR also improves handling of `ReadOnlyMemory<byte>` inputs by avoiding unnecessary byte array allocations when the underlying buffer can be accessed directly, reducing allocation overhead for byte payload scenarios commonly used for JSON message bodies.

### Pull request checklist

### Additional information

The optimization is additive and preserves backward compatibility by falling back to the existing serializer-based conversion when primitive parsing is not applicable.

Unit tests validate primitive conversion paths, and HTTP trigger end-to-end tests verify behavior through the worker input conversion pipeline. Since `JsonPocoConverter` is shared across bindings, the HTTP scenario serves as a representative end-to-end validation for other triggers relying on the same conversion mechanism.

Happy to adjust implementation details if a different converter ordering or placement is preferred.

### DIFF
--- a/protos/azure-functions-language-worker-protobuf/CODEOWNERS
+++ b/protos/azure-functions-language-worker-protobuf/CODEOWNERS
@@ -11,4 +11,4 @@
 # include the following people in the PRs.
 # Language owners should get notified of any new changes to the proto file.
 
-src/proto/FunctionRpc.proto @vrdmr @gavin-aguiar @YunchuWang @surgupta-msft @satvu @ejizba @alrod @anatolib @kaibocai @shreyas-gopalakrishna @amamounelsayed @Francisco-Gamino
+src/proto/FunctionRpc.proto @vrdmr @gavin-aguiar @YunchuWang @surgupta-msft @satvu @alrod @anatolib @kaibocai @shreyas-gopalakrishna @amamounelsayed @Francisco-Gamino
--- a/protos/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
+++ b/protos/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
@@ -437,11 +437,15 @@ message InvocationResponse {
   // Output binding data
   repeated ParameterBinding output_data = 2;
 
+  // Status of the invocation (success/failure/canceled)
+  StatusResult result = 3;
+
   // data returned from Function (for $return and triggers with return support)
   TypedData return_value = 4;
 
-  // Status of the invocation (success/failure/canceled)
-  StatusResult result = 3;
+  // Enables propagation of tags from worker -> host.
+  map<string, string> trace_context_attributes = 5;
+
 }
 
 message WorkerWarmupRequest {
--- a/release_notes.md
+++ b/release_notes.md
@@ -6,12 +6,13 @@
 
 ### Microsoft.Azure.Functions.Worker (metapackage) <version>
 
-- <entry>
+- Update `Microsoft.Azure.Functions.Worker.Core` to <version>
+- Update `Microsoft.Azure.Functions.Worker.Grpc` to <version>
 
 ### Microsoft.Azure.Functions.Worker.Core <version>
 
-- <entry>
+- Add support for propagating trace context tags from worker to host (#3303)
 
 ### Microsoft.Azure.Functions.Worker.Grpc <version>
 
-- <entry>
+- Update protobuf version to v1.12.0-protofile and add support for propagating tags from the worker to the functions host (#3303).
--- a/src/DotNetWorker.Core/Diagnostics/Telemetry/TelemetryProviderV1_17_0.cs
+++ b/src/DotNetWorker.Core/Diagnostics/Telemetry/TelemetryProviderV1_17_0.cs
@@ -9,7 +9,7 @@ namespace Microsoft.Azure.Functions.Worker.Diagnostics;
 internal sealed class TelemetryProviderV1_17_0 : TelemetryProvider
 {
     private static readonly KeyValuePair<string, object> SchemaUrlAttribute =
-        new(TraceConstants.OTelAttributes_1_17_0.SchemaUrl, TraceConstants.OTelAttributes_1_17_0.SchemaVersion);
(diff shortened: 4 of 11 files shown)


victor-lubn:

jviau (changes requested):
Blocking PR for further discussion with the team.

I am concerned about potential breaking behavior changes with this PR. 

1. The custom parsing will short-circuit any user registered JSON converters when the type is a number primitive.
2. New inputs are now accepted which previously failed. For your use case this is desired, but for other users it may not be desired. Again, they may be relying on some custom JSON converters with validation that will be entirely skipped now.

There are also some e2e experience concerns:

1. Only integers are supported. Other primitives like float, double, and decimal are not. If we do accept a change like this, I think supporting those other primitives should be considered.

With that said, you could workaround your issue yourself with a custom JSON converter to handle parsing integers with leading 0s. Have you tried that already?

victor-lubn:
Hello,

I am already using a custom JSON converter on my side.
I also wanted to ask about the optimization around `ReadOnlyMemory<byte>` handling. Specifically this part:

```c#
if (MemoryMarshal.TryGetArray(sourceMemory, out ArraySegment<byte> segment) && segment.Array != null)
{
    return await GetConversionResultFromDeserialization(segment.Array, segment.Offset, segment.Count, context.TargetType);
}

var bytes = sourceMemory.ToArray();
```

My understanding is that in some cases context.Source may already arrive as `ReadOnlyMemory<byte>` backed by an existing `array`, so this would let us avoid creating another copy before deserialization. I was thinking this could be beneficial for scenarios where the payload is already materialized as bytes by the binding pipeline, potentially including some trigger paths such as HTTP and other bindings that pass byte-based input through the converter.

Would you consider this optimization acceptable in `JsonPocoConverter`, or would you prefer to always materialize with `ToArray()` here for consistency and simplicity?

If it would be better to keep that discussion separate from the primitive parsing change, I would be happy to remove it from this PR and open a separate PR for the `ReadOnlyMemory<byte>` optimization.

jviau:
> I am already using a custom JSON converter on my side.

Is this working for your scenario?

@victor-lubn a separate PR for the span optimization would be preferred. I haven't looked at it closely yet, but will take a look when you open the PR.

Thank you!

PR #3330, opened 2026-03-03
[MSBUILD SDK] Add ZipDeployTask and tests

### LINKED ISSUE #3140: [SDK] Port ZipDeploy task to new SDK

jviau:

resolves #3140

### Pull request checklist

### Additional information

Adds the ZipDeploy task, target, and tests to support publish via dotnet/vscode/visual studio with the new SDK.

### DIFF
--- a/src/Azure.Functions.Sdk/Tasks/Publish/ZipDeploy.cs
+++ b/src/Azure.Functions.Sdk/Tasks/Publish/ZipDeploy.cs
@@ -8,10 +8,6 @@ using System.Runtime.InteropServices;
 using Azure.Functions.Sdk.ZipDeploy;
 using Microsoft.Build.Framework;
 
-// IMPORTANT: Do not modify this file directly with major changes
-// This file is a copy from this project (with minor updates) -- https://github.com/Azure/azure-functions-vs-build-sdk/blob/b0e54a832a92119e00a2b1796258fcf88e0d6109/src/Microsoft.NET.Sdk.Functions.MSBuild/Microsoft.NET.Sdk.Functions.MSBuild.csproj
-// Please make any changes upstream first.
-
 namespace Azure.Functions.Sdk.Tasks.Publish;
 
 public sealed class ZipDeploy(IFileSystem fileSystem, DeploymentClient? client = null)


src/Azure.Functions.Sdk/Tasks/Publish/ZipDeploy.cs
```diff
@@ -0,0 +1,174 @@
...
+
+using System.Diagnostics.CodeAnalysis;
+using System.IO.Abstractions;
+using System.Net.Http.Headers;
+using System.Runtime.InteropServices;
+using Azure.Functions.Sdk.ZipDeploy;
+using Microsoft.Build.Framework;
+
+// IMPORTANT: Do not modify this file directly with major changes
+// This file is a copy from this project (with minor updates) -- https://github.com/Azure/azure-functions-vs-build-sdk/blob/b0e54a832a92119e00a2b1796258fcf88e0d6109/src/Microsoft.NET.Sdk.Functions.MSBuild/Microsoft.NET.Sdk.Functions.MSBuild.csproj
```
brettsam: Is most of this still a copy of what we've already got? It looks pretty different at first glance so maybe remove the note about making upstream changes first?

src/Azure.Functions.Sdk/Tasks/Publish/ZipDeploy.cs
```diff
@@ -0,0 +1,174 @@
...
+// Please make any changes upstream first.
+
+namespace Azure.Functions.Sdk.Tasks.Publish;
+
+public sealed class ZipDeploy(IFileSystem fileSystem, DeploymentClient? client = null)
```
brettsam, line 13: Was just doing some manual comparison and remembered that we needed to do this in order for VS to work correctly... is this not needed anymore? So much has changed since we originally did this so I have no idea!

```
#if NET472
    [LoadInSeparateAppDomain]
    public class ZipDeployTask : AppDomainIsolatedTask
#else
    public class ZipDeployTask : Task
#endif
```

jviau: Not sure! I'll be doing more E2E testing of this scenario before GA'ing for sure.

Are you okay with this going into preview as is? Understanding this is remaining on our own nuget feeds for now still (and not on nuget.org).

src/Azure.Functions.Sdk/Tasks/Publish/ZipDeploy.cs
jviau: Behavior wise it is a copy, but completely refactored. I got rid of the 3rd party compression library and use reflection to access built in .NET APIs for setting permissions (they are available in both netfx and netcoreapp, just not netstandard 🤷 ). I also completely re-did the client approach to make it more unit testable.

I'll remove this comment.

src/Azure.Functions.Sdk/Tasks/Publish/ZipDeploy.cs
jviau: Som I _think_ this isn't needed. As this task no longer loads any extra libraries that I would see needing a separate app domain.

PR #3359, opened 2026-04-09
Add compile-time duplicate function name detection (AZFW0017)

### LINKED ISSUE #3158: Add a check for duplicate function names
Add a check for duplicate function names and display a warning/error if there are any. The host does not allow this, and we want to add a fail-fast approach here on the worker side when the worker creates the function metadata payload.

MO2k4:
## Summary
- Adds a new Roslyn analyzer diagnostic **AZFW0014** that reports an error at compile time when two or more functions share the same name (case-insensitive), preventing a confusing runtime failure from the Azure Functions host.
- The check runs inside `FunctionMetadataProviderGenerator` after collecting all function metadata; if duplicates are found the generator emits the diagnostic and skips code generation.
- Includes two new test cases covering exact-match and case-insensitive duplicate detection across all supported C# language versions.

Closes #3158

## Test plan
- [x] `DuplicateFunctionNamesFails` — two functions with identical name `"MyFunc"` produce AZFW0014
- [x] `DuplicateFunctionNamesCaseInsensitiveFails` — `"MyFunc"` vs `"myfunc"` also produce AZFW0014
- [x] Existing generator tests continue to pass (no regressions)

### DIFF
--- a/docs/analyzer-rules/AZFW0017.md
+++ b/docs/analyzer-rules/AZFW0017.md
@@ -1,0 +1,41 @@
+# AZFW0017: Duplicate function name
+
+| | Value |
+|-|-|
+| **Rule ID** |AZFW0017|
+| **Category** |[FunctionMetadataGeneration]|
+| **Severity** |Error|
+
+## Cause
+
+This rule is triggered when two or more functions in the same project share the same name (case-insensitive).
+
+## Rule description
+
+Each function in an Azure Functions app must have a unique name. The Azure Functions host treats function names as case-insensitive identifiers, so `MyFunc` and `myfunc` are considered duplicates. When duplicate names are present, the host fails to start.
+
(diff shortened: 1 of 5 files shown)


satvu (approved):
Looks good - error level makes sense since it matches the severity of this failure in the host (host fails to start up if there are duplicate functions).

MO2k4:
@satvu i rebased my branch to be up2date which should fix the errors from the pipeline, could you trigger the pipeline run again?

satvu:
@MO2k4 - It's been a while, so I forgot part of the process.

Could you add a doc page for this rule in [this folder](https://github.com/Azure/azure-functions-dotnet-worker/tree/main/docs/analyzer-rules). You'll also have to update the rule number to AZFW0017 (as the prev numbers are taken and defined in other analyzer projects).

MO2k4:
@satvu Oh thanks for pointing this out, i've updated the code and added the doc page.

satvu (approved):

PR #3361, opened 2026-04-10
Fix response cookie duplication in ASP.NET Core integration

### LINKED ISSUE #3353: Response cookies are duplicated
### Description

When you set multiple cookies in an HTTP trigger function using `Cookies.Append`, each one after the first duplicates all of the previous ones, including any `Set-Cookie` headers that were added manually. For example, if you set `Cookie1`, `Cookie2`, and then `Cookie3`, you end up with a `Set-Cookie` header containing `Cookie1, Cookie1, Cookie2, Cookie1, Cookie1, Cookie2, Cookie3`.

### Steps to reproduce

Starting from the default Azure Functions project in Visual Studio, replace `Function1` with this:

```cs
[Function("Function1")]
public async Task<HttpResponseData> Run([HttpTrigger(AuthorizationLevel.Anonymous, "get")] HttpRequestData request)
{
	var response = request.CreateResponse();
	response.Headers.Add("Set-Cookie", "ManualCookie=AlsoDuplicated");
	response.Cookies.Append("Cookie1", "one");
	response.Cookies.Append("Cookie2", "two");
	response.Cookies.Append("Cookie3", "three");

	await response.WriteStringAsync($"The response headers are now: {response.Headers}");
	return response;
}
```

When you run the function, you get the following:

```
The response headers are now: Set-Cookie: ManualCookie=AlsoDuplicated, ManualCookie=AlsoDuplicated, Cookie1=one; path=/, ManualCookie=AlsoDuplicated, ManualCookie=AlsoDuplicated, Cookie1=one; path=/, Cookie2=two; path=/, ManualCookie=AlsoDuplicated, ManualCookie=AlsoDuplicated, Cookie1=one; path=/, ManualCookie=AlsoDuplicated, ManualCookie=AlsoDuplicated, Cookie1=one; path=/, Cookie2=two; path=/, Cookie3=three; path=/
```

MO2k4:
## Summary

Fixes #3353 — calling `response.Cookies.Append()` multiple times duplicates all previous `Set-Cookie` header values exponentially.

- **Root cause:** The `IHeaderDictionary` indexer setters in `AspNetCoreHttpHeadersCollection` used `TryAddWithoutValidation()` which *appends* values to an existing key. The contract requires *replace* semantics. When ASP.NET Core's internal `ResponseCookies.Append()` reads all existing `Set-Cookie` values, appends the new cookie, and sets the combined result back via the indexer, the old values are preserved AND the combined values are added — causing exponential duplication.
- **Fix:** Call `Remove(key)` before `TryAddWithoutValidation(key, value)` in both indexer setters (explicit `IHeaderDictionary` and public).
- **Change surface:** 2 lines of production code in `AspNetCoreHttpHeadersCollection.cs`.

## Note for reviewers

The `Add(string key, StringValues value)` method (line 157) in the same class also uses `TryAddWithoutValidation` without removing first. This means `IHeaderDictionary.Add` has append behavior rather than throwing on duplicate keys. This is **not** the cause of the cookie bug (ASP.NET Core uses the indexer, not `Add`, for cookie header manipulation), but it may warrant a follow-up to align with standard `IDictionary.Add` semantics.

## Test plan

- [x] `IndexerSetter_ShouldReplaceValues_NotAppend` — sets a header twice via the indexer, asserts only the second value remains
- [x] `CookiesAppend_ShouldNotDuplicateSetCookieHeaders` — reproduces exact issue: 1 manual `Set-Cookie` header + 3 `Cookies.Append` calls = 4 values (was 15 before fix)
- [x] `MultipleCookiesAppend_ShouldNotDuplicate` — 3 `Cookies.Append` calls = 3 values
- [x] All 26 existing `Worker.Extensions.Http.AspNetCore.Tests` pass
- [x] All 295 `DotNetWorker.Tests` pass

### DIFF
--- a/extensions/Worker.Extensions.Http.AspNetCore/src/HttpDataModel/AspNetCoreHttpHeadersCollection.cs
+++ b/extensions/Worker.Extensions.Http.AspNetCore/src/HttpDataModel/AspNetCoreHttpHeadersCollection.cs
@@ -96,7 +96,10 @@ namespace Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore
             set
             {
                 Remove(key);
-                TryAddWithoutValidation(key, (IEnumerable<string>)value);
+                if (value.Count > 0)
+                {
+                    TryAddWithoutValidation(key, (IEnumerable<string>)value);
+                }
             }
         }
 
@@ -144,7 +147,10 @@ namespace Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore
             set
             {
                 Remove(key);
-                TryAddWithoutValidation(key, (IEnumerable<string>)value);
+                if (value.Count > 0)
+                {
+                    TryAddWithoutValidation(key, (IEnumerable<string>)value);
+                }
             }
         }
 
--- a/extensions/Worker.Extensions.Http.AspNetCore/src/HttpDataModel/AspNetCoreHttpHeadersCollection.cs
+++ b/extensions/Worker.Extensions.Http.AspNetCore/src/HttpDataModel/AspNetCoreHttpHeadersCollection.cs
@@ -95,6 +95,7 @@ namespace Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore
             }
             set
             {
+                Remove(key);
                 TryAddWithoutValidation(key, (IEnumerable<string>)value);
             }
         }
@@ -142,6 +143,7 @@ namespace Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore
             }
             set
             {
+                Remove(key);
                 TryAddWithoutValidation(key, (IEnumerable<string>)value);
             }
         }
--- a/extensions/Worker.Extensions.Http.AspNetCore/release_notes.md
+++ b/extensions/Worker.Extensions.Http.AspNetCore/release_notes.md
@@ -6,7 +6,7 @@
 
 ### Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore <version>
 
-- Updating ASP.NET Core integration to support the metadata transformation feature (#3172)
+- Fix response cookie duplication in ASP.NET Core integration by giving `IHeaderDictionary` indexer setters replace semantics, including removing the header when assigned `StringValues.Empty` (#3353)
 
 ### Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore.Analyzers  <version>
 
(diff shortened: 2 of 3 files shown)


extensions/Worker.Extensions.Http.AspNetCore/src/HttpDataModel/AspNetCoreHttpHeadersCollection.cs
```diff
@@ -95,6 +95,7 @@ StringValues IHeaderDictionary.this[string key]
             }
             set
             {
+                Remove(key);
```
jviau, line 98: Looking at aspnetcore's implementation, that will perform a remove-only if value is empty.

https://github.com/dotnet/aspnetcore/blob/765e192789d3809bd1729fbcf0980dc221a6b386/src/Http/Http/src/HeaderDictionary.cs#L90

Maybe this should do the same? Or rather skip the `TryAddWithoutValidation` if value is empty:

``` CSharp
if (value.Count > 0)
{
    TryAddWithoutValidation(key, (IEnumerable<string>)value);
}
```

MO2k4: @jviau you're right i missed that case. With this case we treat it as just removal if the values are empty. i've also added another test case to cover this.

jviau:
@MO2k4 can you update the release notes https://github.com/Azure/azure-functions-dotnet-worker/blob/main/extensions/Worker.Extensions.Http.AspNetCore/release_notes.md to reflect this change? Remove the existing entry there, that has already shipped.

MO2k4:
@jviau release notes are added and i removed the old entry

extensions/Worker.Extensions.Http.AspNetCore/src/HttpDataModel/AspNetCoreHttpHeadersCollection.cs
```diff
@@ -142,7 +146,11 @@ public StringValues this[string key]
             }
             set
             {
-                TryAddWithoutValidation(key, (IEnumerable<string>)value);
+                Remove(key);
```
fabiocav, line 149: The explicit `IHeaderDictionary.this[string]` setter and the public `this[string]` setter are now 
byte‑identical. Marking this as a non-blocking code duplication, but it would be good to consider having one implementation delegate to the other here to avoid unintentional drift.

fabiocav (approved):

extensions/Worker.Extensions.Http.AspNetCore/src/HttpDataModel/AspNetCoreHttpHeadersCollection.cs
jviau: I think they were identical before also, unless I am reading it wrong. But yeah, not a blocker. We can consider a separate change for consolidating the implementation.

jviau (approved):

PR #3362, opened 2026-04-13
Resource detector update

RohitRanjanMS:
Brings the isolated worker's FunctionsResourceDetector in line with the equivalent changes made to the Azure Functions host in Azure/azure-functions-host#11506.

  FunctionsResourceDetector
  - service.name now follows priority order: OTEL_SERVICE_NAME → OTEL_RESOURCE_ATTRIBUTES[service.name] →
  WEBSITE_SITE_NAME → assembly name ("unknown" as final fallback). When a higher-priority source is present the detector   skips adding service.name, letting the OTel SDK pick it up instead.
  - service.version now follows priority order: OTEL_RESOURCE_ATTRIBUTES[service.version] → assembly version. There is no OTEL_SERVICE_VERSION per OTel spec (https://github.com/open-telemetry/semantic-conventions/issues/2669).
  - New attribute deployment.environment.name populated from WEBSITE_SLOT_NAME (Azure only).
  - New attribute azure.functions.site.update_id populated from FUNCTIONS_SITE_UPDATE_ID (Azure only).
  - GetAzureResourceUri uses span-based parsing for the subscription ID and returns null instead of string.Empty when env vars are missing.
  - OTEL_RESOURCE_ATTRIBUTES parsing uses ReadOnlySpan<char> to avoid allocations; key matching is case-sensitive   ordinal. TODO left to replace with MemoryExtensions.Split once the library targets .NET 10.
  - AssemblyVersion is cached as private static readonly fields — computed once at class load rather than on every Detect() call.

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/DotNetWorker.OpenTelemetry/FunctionsResourceDetector.cs
+++ b/src/DotNetWorker.OpenTelemetry/FunctionsResourceDetector.cs
@@ -15,8 +15,6 @@ namespace Microsoft.Azure.Functions.Worker.OpenTelemetry
         private static readonly string AssemblyVersion =
             typeof(FunctionsResourceDetector).Assembly.GetName().Version?.ToString() ?? "unknown";
 
-        private static readonly int ProcessId = Process.GetCurrentProcess().Id;
-
         public Resource Detect()
         {
             try
@@ -24,7 +22,7 @@ namespace Microsoft.Azure.Functions.Worker.OpenTelemetry
                 var attributes = new List<KeyValuePair<string, object>>(capacity: 10)
                 {
                     new(ResourceSemanticConventions.AISDKPrefix, $"{OpenTelemetryConstants.SDKPrefix}:{AssemblyVersion}"),
-                    new(ResourceSemanticConventions.ProcessId, ProcessId)
+                    new(ResourceSemanticConventions.ProcessId, Process.GetCurrentProcess().Id)
                 };
 
                 string? siteName = Environment.GetEnvironmentVariable(OpenTelemetryConstants.SiteNameEnvVar);
--- a/src/DotNetWorker.OpenTelemetry/FunctionsResourceDetector.cs
+++ b/src/DotNetWorker.OpenTelemetry/FunctionsResourceDetector.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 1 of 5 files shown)


src/DotNetWorker.OpenTelemetry/FunctionsResourceDetector.cs
```diff
@@ -12,64 +12,143 @@ namespace Microsoft.Azure.Functions.Worker.OpenTelemetry
...
         {
-            List<KeyValuePair<string, object>> attributeList = new(8);
             try
             {
-                string serviceName = Environment.GetEnvironmentVariable(OpenTelemetryConstants.SiteNameEnvVar);
-                string version = Assembly.GetEntryAssembly().GetName().Version.ToString();
+                var attributes = new List<KeyValuePair<string, object>>(capacity: 10)
+                {
+                    new(ResourceSemanticConventions.AISDKPrefix, $"{OpenTelemetryConstants.SDKPrefix}:{AssemblyVersion}"),
+                    new(ResourceSemanticConventions.ProcessId, Process.GetCurrentProcess().Id)
```
satvu, line 25: PR description mentions process id getting cached, but I don't think that's the case here (computed every call)?

src/DotNetWorker.OpenTelemetry/FunctionsResourceDetector.cs
```diff
@@ -12,64 +12,143 @@ namespace Microsoft.Azure.Functions.Worker.OpenTelemetry
...
-                attributeList.Add(new KeyValuePair<string, object>(ResourceSemanticConventions.AISDKPrefix, 
-                    $@"{OpenTelemetryConstants.SDKPrefix}:{typeof(FunctionsResourceDetector).Assembly.GetName().Version}"));
-                attributeList.Add(new KeyValuePair<string, object>(ResourceSemanticConventions.ProcessId, Process.GetCurrentProcess().Id));
+                // Priority: OTEL_RESOURCE_ATTRIBUTES[service.version] > AssemblyVersion
+                // OTel decided to not have OTEL_SERVICE_VERSION, so we only check OTEL_RESOURCE_ATTRIBUTES.
+                // https://github.com/open-telemetry/semantic-conventions/issues/2669
+                if (!IsResourceAttributeConfigured(ResourceSemanticConventions.ServiceVersion, resourceAttributes))
+                {
+                    attributes.Add(new(ResourceSemanticConventions.ServiceVersion,
+                        Assembly.GetEntryAssembly()?.GetName().Version?.ToString() ?? "unknown"));
```
satvu, line 44: nit: could store `Assembly.GetEntryAssembly()` since it is used here and line 35.

src/DotNetWorker.OpenTelemetry/FunctionsResourceDetector.cs
RohitRanjanMS: I have updated the description. I started with it but later changed the code.

src/DotNetWorker.OpenTelemetry/FunctionsResourceDetector.cs
RohitRanjanMS: `Assembly.GetEntryAssembly()` returns a cached assembly instance, and `Detect()` is only called during startup.

satvu (approved):

PR #3357, opened 2026-04-09
Add package README for Microsoft.Azure.Functions.Worker.Sdk

MO2k4:
## Summary

- Adds a `README.md` to the `Microsoft.Azure.Functions.Worker.Sdk` package so NuGet/Visual Studio displays documentation instead of "no README available"
- Wires the README into the `.nupkg` via `PackageReadmeFile` and a `Content` item in `Sdk.csproj`
- README covers what the SDK provides: build/publish targets, source generators, analyzers (with diagnostic IDs), and user-configurable MSBuild properties

Closes #3249

## Test plan

- [x] `dotnet build sdk/Sdk/Sdk.csproj` succeeds
- [x] `dotnet pack sdk/Sdk/Sdk.csproj` produces a `.nupkg` containing `README.md` at the package root

PR #3370, opened 2026-04-16
Add AZFW0017 to analyzer table in Sdk README

MO2k4:
## Summary

Adds the missing `AZFW0017` (Duplicate function name) diagnostic to the analyzer table in the `Microsoft.Azure.Functions.Worker.Sdk` package README.

This diagnostic was shipped as part of #3158 but was not included in the README table added by #3357.

## Changes

- [sdk/Sdk/README.md](sdk/Sdk/README.md): Added `AZFW0017 | Error | Duplicate function name` row to the Analyzers table.

### DIFF
--- a/sdk/Sdk/README.md
+++ b/sdk/Sdk/README.md
@@ -32,14 +32,7 @@ Three Roslyn source generators run at compile time:
 
 ### Analyzers
 
-| Diagnostic | Severity | Description |
-|------------|----------|-------------|
-| AZFW0001 | Error | WebJobs binding attributes (in-process) used instead of isolated worker attributes |
-| AZFW0002 | Error | Async function method returns `void` instead of `Task` |
-| AZFW0009 | Error | `SupportsDeferredBinding` applied to an output binding |
-| AZFW0010 | Warning | Parameter type not supported by the binding attribute |
-| AZFW0011 | Error | Blob container path requires an iterable type or `BlobContainerClient` |
-| AZFW0017 | Error | Duplicate function name — function names must be unique across the app |
+This package includes analyzers (AZFW0001–AZFW0013, AZFW0017) that catch common issues at compile time. For the full list of rules and detailed documentation, see the [analyzer rules reference](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/docs/analyzer-rules/index.md).
 
 ## Configuration
 
--- a/sdk/Sdk/README.md
+++ b/sdk/Sdk/README.md
@@ -39,6 +39,7 @@ Three Roslyn source generators run at compile time:
 | AZFW0009 | Error | `SupportsDeferredBinding` applied to an output binding |
 | AZFW0010 | Warning | Parameter type not supported by the binding attribute |
 | AZFW0011 | Error | Blob container path requires an iterable type or `BlobContainerClient` |
+| AZFW0017 | Error | Duplicate function name — function names must be unique across the app |
 
 ## Configuration
 
--- a/sdk/Sdk/README.md
+++ b/sdk/Sdk/README.md
@@ -32,7 +32,7 @@ Three Roslyn source generators run at compile time:
 
 ### Analyzers
 
-This package includes analyzers (AZFW0001–AZFW0013, AZFW0017) that catch common issues at compile time. For the full list of rules and detailed documentation, see the [analyzer rules reference](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/docs/analyzer-rules/index.md).
+This package includes analyzers (AZFW0001–AZFW0013, AZFW0017) that catch common issues during development. For the full list of rules and detailed documentation, see the [analyzer rules reference](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/docs/analyzer-rules/index.md).
 
 ## Configuration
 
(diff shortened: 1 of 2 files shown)


sdk/Sdk/README.md
```diff
@@ -39,6 +39,7 @@ Three Roslyn source generators run at compile time:
 | AZFW0009 | Error | `SupportsDeferredBinding` applied to an output binding |
 | AZFW0010 | Warning | Parameter type not supported by the binding attribute |
 | AZFW0011 | Error | Blob container path requires an iterable type or `BlobContainerClient` |
+| AZFW0017 | Error | Duplicate function name — function names must be unique across the app |
```
Frulfump: Nit: Other descriptions don't contain the extra text after a hyphen.

Are there other rules that would make sense to add? https://github.com/Azure/azure-functions-dotnet-worker/blob/main/docs/analyzer-rules/ ?

Frulfump (approved):

sdk/Sdk/README.md
satvu: Maybe add another line to this section that links to the full set of analyzer rules? Not sure how to prioritize the subset of rules that get highlighted on this page (we don't have telemetry on which of these is most commonly run into, most of the error level ones are no more or less fatal than the others). 

I think [this index file](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/docs/analyzer-rules/index.md) in the analyzer docs directory was meant to document the analyzer rules the way this table is doing. Perhaps we can leverage that file as the "full" index of rule and link it in this section.

MO2k4: @satvu i've updated the general index and linked the readme to the index. What do you think?

sdk/Sdk/README.md
```diff
@@ -32,13 +32,7 @@ Three Roslyn source generators run at compile time:
...
 ### Analyzers
 
-| Diagnostic | Severity | Description |
-|------------|----------|-------------|
-| AZFW0001 | Error | WebJobs binding attributes (in-process) used instead of isolated worker attributes |
-| AZFW0002 | Error | Async function method returns `void` instead of `Task` |
-| AZFW0009 | Error | `SupportsDeferredBinding` applied to an output binding |
-| AZFW0010 | Warning | Parameter type not supported by the binding attribute |
-| AZFW0011 | Error | Blob container path requires an iterable type or `BlobContainerClient` |
+This package includes analyzers (AZFW0001–AZFW0013, AZFW0017) that catch common issues at compile time. For the full list of rules and detailed documentation, see the [analyzer rules reference](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/docs/analyzer-rules/index.md).
```
Frulfump: Nit: Some analyzers I think would trigger on design time build in VS(and Rider and VS Code with C# extension(?) or Dev Kit) which isn't really strictly compile time.

Frulfump (approved):

jviau (approved):

sdk/Sdk/README.md
```diff
@@ -32,13 +32,7 @@ Three Roslyn source generators run at compile time:
...
 ### Analyzers
 
-| Diagnostic | Severity | Description |
-|------------|----------|-------------|
-| AZFW0001 | Error | WebJobs binding attributes (in-process) used instead of isolated worker attributes |
-| AZFW0002 | Error | Async function method returns `void` instead of `Task` |
-| AZFW0009 | Error | `SupportsDeferredBinding` applied to an output binding |
-| AZFW0010 | Warning | Parameter type not supported by the binding attribute |
-| AZFW0011 | Error | Blob container path requires an iterable type or `BlobContainerClient` |
+This package includes analyzers (AZFW0001–AZFW0013, AZFW0017) that catch common issues during development. For the full list of rules and detailed documentation, see the [analyzer rules reference](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/docs/analyzer-rules/index.md).
```
jviau, line 35: The contents of this file can change relative to the point in time of this SDK release. The link may eventually diverge from what is in this SDK.

Ideally, we would capture the tag of this release (or git commit) and create a permalink. This would be a build step to transform the README.md before including in the package.

That is probably beyond the scope of this PR though. But I do get the idea of wanting to move that documentation external to the README for easier maintenance. I'll have to think about the tradeoffs for a bit and get back to you.

sdk/Sdk/README.md
jviau: The nuance here between DTB and compile time is mostly irrelevant and not worth pointing out.

sdk/Sdk/README.md
satvu: Looks good, Jacob raised a good point [here](https://github.com/Azure/azure-functions-dotnet-worker/pull/3370#discussion_r3102086186) about this approach though. We'll follow up in that comment thread instead.

sdk/Sdk/README.md
MO2k4: Good point about the link drifting. I looked into a few ways to pin it at pack time:

1. MSBuild target before GenerateNuspec — rewrite blob/main/ to blob/<tag-or-sha>/ in a temp copy of the README. There's already URL rewriting in RepositoryInfo.targets for SourceLink, so the pattern fits.
2. sed replacement in the CI pipeline — simpler, but only works in CI, not local dotnet pack.

Agree this is out of scope for this PR. Is it ok if I file an issue to track the permalink work and keep blob/main for now?

jviau: Yeah, filing a separate issue for that is fine. Thank you.

PR #3387, opened 2026-04-27
Merge `feature/msbuild-sdk` to `main`

jviau:

resolves #issue_for_this_pr

### Pull request checklist

  *[ ] Otherwise: Documentation issue linked to PR

### Additional information

Merges the `feature/msbuild-sdk` branch into `main`. This is in preparation for cutting a public preview of the package.

#### Extra Changes

While the vast majority of the changes in this feature branch have already been reviewed via previous PRs, there are some direct changes pushed for this merge:

1. ZipDeploy copilot feedback -- removed `CharSet` and `ContentEncoding` from `ZipDeployRequest`. Tested without these and the app still deployed from VS with no issues.
2. Updated sample README
3. Fixed retry-count extra increment in deployment log.
4. Addressed rest of copilot feedback (extra semicolon, `sdk-rules/RULE_TEMPLATE.md` inconsistency, typos)


samples/FunctionApp.MSBuildSdk/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+Your code! No changes to function app code -- only to the csproj.
+
+## Publish/Deploy
+
+ZipDeploy has not been added yet. To deploy a dotnet app using this SDK, follow these steps:
+
+1. `dotnet publish -c release`
+2. Zip up the published contents (zip the contents, not the folder itself)
+3. `az functionapp deployment source config-zip -g <resource-group> -n <function-app-name> --src <zip>`
```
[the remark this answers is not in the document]
jviau: Will fix these docs up later.

liliankasem (approved):

src/Azure.Functions.Sdk/ZipDeploy/DeploymentClient.cs
```diff
@@ -0,0 +1,149 @@
...
+
+        private static async Task<StatusResult> ParseResponseAsync(
+            HttpResponseMessage response, CancellationToken cancellation)
+        {
+            if (response.Content is null)
+            {
+                return Default;
+            }
+
+            using Stream stream = await response.Content.ReadAsStreamAsync();
```
satvu, line 144: does this need to pass in the cancellation token?

satvu (approved):

src/Azure.Functions.Sdk/ZipDeploy/DeploymentClient.cs
jviau: No, the cancellation token accepting overload is not available in this project (net472, netstandard2.0). I think it is only in >= net5.0

src/Azure.Functions.Sdk/Tasks/Publish/CreateZipFile.cs
```diff
@@ -0,0 +1,88 @@
...
+        using ZipArchive archive = ZipFile.Open(zipFilePath, ZipArchiveMode.Update);
+        ZipArchiveEntry? entry = archive.GetEntry(entryName) ?? archive.GetEntry(entryName + ".exe");
+        entry?.SetUnixFilePermissions(UnixExecutablePermissions);
+    }
+
+    internal string CreateZipFileFromDirectory()
+    {
+        string zipFileName = ZipName + ".zip";
+        string destination = Path.Combine(PublishIntermediateTempPath, zipFileName);
+        ZipFile.CreateFromDirectory(SourceFolder, destination);
```
fabiocav, line 74: `ZipFile.CreateFromDirectory` throws if the destination already exists, is the name here stable? Will we run into issues with repeated deployments?

src/Azure.Functions.Sdk/Tasks/Extensions/WriteExtensionProject.cs
```diff
@@ -0,0 +1,198 @@
...
+        }
+
+        _fileSystem.File.Delete(project.Path);
+        using Stream stream = _fileSystem.File.OpenWrite(project.Path);
+        using XmlWriter writer = XmlWriter.Create(stream, _xmlSettings);
+        writer.WriteComment(string.Format(CultureInfo.InvariantCulture, Disclaimer, _time.GetUtcNow(), hash));
+        writer.WriteStartElement("Project");
+
+        // Omit version so the outer projects version is implicitly used.
+        writer.WriteAttributeString("Sdk", ThisAssembly.Name);
```
fabiocav, line 103: Provided similar feedback on this potential issue in a different PR: There is an incrementality corruption risk in this write path: the hash is updated before the `XmlWriter`/stream are disposed. If the build is killed after the hash write but before the csproj is fully flushed, future builds can see a current hash and skip regenerating a truncated/invalid generated project. Please write to a temp file, close it, atomically replace the csproj, and only then update the hash.

src/Azure.Functions.Sdk/Targets/Azure.Functions.Sdk.props
```diff
@@ -0,0 +1,37 @@
...
+  <Import Project="$(MSBuildThisFileDirectory)Azure.Functions.Sdk.Analyzers.props" />
+  <Import Project="$(MSBuildThisFileDirectory)extensions/Azure.Functions.Sdk.Extensions.props" />
+
+  <!-- Enable Azure Functions project capability to enable tools -->
+  <ItemGroup>
+    <ProjectCapability Include="AzureFunctions" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="2.51.0" IsImplicitlyDefined="true" />
```
fabiocav, line 30: Did you validate this with CPM enabled? Is this implicit versioned `PackageReference` cause problems with that? If so, we should be validating `ManagePackageVersionsCentrally=true` here and emitting package references with no no `Version`?

src/Azure.Functions.Sdk/Targets/Extensions/Azure.Functions.Sdk.Extensions.targets
```diff
@@ -0,0 +1,158 @@
...
+  <Target Name="WriteExtensionProject">
+    <ValidateExtensionPackages ExtensionPackages="@(_AzureFunctionPackageReference)">
+      <Output TaskParameter="FilteredPackages" ItemName="_ValidatedAzureFunctionPackageReference" />
+    </ValidateExtensionPackages>
+    <WriteExtensionProject Projects="@(_AzureFunctionsTargetFramework)" Packages="@(_ValidatedAzureFunctionPackageReference)" />
+  </Target>
+
+  <Target Name="RestoreExtensionProject">
+    <MSBuild Projects="@(_AzureFunctionsTargetFramework->'%(ProjectFile)')" Targets="Restore"
+      RemoveProperties="$(_AzureFunctionsExtensionRemoveProps)" Properties="IsRestoring=true;RestoreSources=$(_OutputSources)" />
```
fabiocav, line 85: Any concern with the fact `$(_OutputSources)` is a NuGet-internal/private restore property? Does this target need to introduce a dependency on `_GetRestoreSettings`? Is there a concern about future behavior changes introducing reliability problems here, or do we have guidance to use this property? Might be helpful to ensure this property is not empty and error if it is, with a clear message.

src/Azure.Functions.Sdk/Targets/Extensions/Azure.Functions.Sdk.Extensions.targets
```diff
@@ -0,0 +1,158 @@
...
+  <!--
+    Until there is a consistent post-restore hook which gives a guarantee to run our targets after restore,
+    _VerifyRestoredFunctionsExtensions will verify the restore hook has run. If extension restore hook has not
+    run, this will emit a warning and try to perform the restore hook during build.
+  -->
+  <Target Name="_VerifyRestoredFunctionsExtensions" Condition="'$(DesignTimeBuild)' != 'true'"
+    BeforeTargets="BeforeBuild"
+    Inputs="@(_AzureFunctionsTargetFramework->'%(ProjectFile)');$(ProjectAssetsFile)" Outputs="@(_AzureFunctionsTargetFramework->'%(MarkerFile)')">
+    <!-- Visual Studio does restore manually via design-time-build outputs. We have no way to hook into it and will have to rely on build fallback. -->
+    <FuncSdkLog Resource="Warning_ExtensionsNotRestored" Condition="'$(BuildingInsideVisualStudio)' != 'true'" />
```
fabiocav, line 154: This suppresses AZFW0108 inside Visual Studio but still runs the side-effecting inner NuGet restore during `Build`. Could this lead to some confusion in VS? should we avoid running restore in the build phase or fail with an actionable diagnostic when VS restore has not produced the generated project?

src/Azure.Functions.Sdk/Targets/Extensions/Azure.Functions.Sdk.Extensions.props
```diff
@@ -0,0 +1,27 @@
...
+          created a backup copy.  Incorrect changes to this file will make it
+          impossible to load or build your projects from the command-line or the IDE.
+
+***********************************************************************************************
+-->
+
+<Project>
+
+  <PropertyGroup>
+    <_AzureFunctionsExtensionRemoveProps>TargetFramework;RuntimeIdentifier;SelfContained;PublishSingleFile;PublishReadyToRun;UseCurrentRuntimeIdentifier;WebPublishMethod;PublishProfile;DeployOnBuild;PublishDir;OutDir;OutputPath;ManagePackageVersionsCentrally;</_AzureFunctionsExtensionRemoveProps>
```
fabiocav, line 15: How was this list defined? What is the expectation for maintenance of this list moving forward? Do we need to ensure things like `PublishAot`, `PublishTrimmed`, `RuntimeIdentifiers`, `BaseIntermediateOutputPath`, `IntermediateOutputPath`, `Configuration`, `Platform`, etc., are here as well?

src/Azure.Functions.Sdk/Targets/Inner/Azure.Functions.Sdk.Inner.RuntimePackages.props
```diff
@@ -0,0 +1,610 @@
...
+WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and have
+          created a backup copy.  Incorrect changes to this file will make it
+          impossible to load or build your projects from the command-line or the IDE.
+
+***********************************************************************************************
+-->
+
+<Project>
+
+  <!-- The list of assemblies to exclude from extensions payload. -->
```
fabiocav, line 14: Still going through the changes and might have missed it, but is this referencing the runtime assemblies manifest as part of the SDK build?

src/Azure.Functions.Sdk/Targets/Azure.Functions.Sdk.targets
```diff
@@ -0,0 +1,140 @@
...
+
+  <Import Sdk="Microsoft.NET.Sdk.Worker" Project="Sdk.targets" />
+  <UsingTask TaskName="$(AzureFunctionsTaskNamespace).FuncSdkLog" AssemblyFile="$(AzureFunctionsSdkTasksAssembly)" />
+  <UsingTask TaskName="$(AzureFunctionsTaskNamespace).GenerateWorkerConfig" AssemblyFile="$(AzureFunctionsSdkTasksAssembly)" />
+
+  <PropertyGroup>
+    <_WorkerConfigOutputPath>$(IntermediateOutputPath)$(_WorkerConfigFileName)</_WorkerConfigOutputPath>
+    <_AzureFunctionsVersionStandardized>$(AzureFunctionsVersion.ToLowerInvariant().Split('-')[0])</_AzureFunctionsVersionStandardized>
+    <_FunctionsRuntimeMajorVersion>$(_AzureFunctionsVersionStandardized.TrimStart('vV'))</_FunctionsRuntimeMajorVersion>
+    <_ToolingSuffix Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp'">net$([MSBuild]::GetTargetFrameworkVersion('$(TargetFramework)', 1))-isolated</_ToolingSuffix>
```
fabiocav, line 22: This opens TFM support much more broadly than the existing Worker SDK: any future `.NETCoreApp` TFM will produce a tooling suffix such as `net11-isolated` before Functions tooling and container images necessarily support it. This would also fail to resolve the configuration entries for the TFM. 

The explicit model we had before required updates, but ensured safety as it would only reference supported tooling entries. 

Did you have a chance to try this with .NET 11?

src/Azure.Functions.Sdk/Targets/Azure.Functions.Sdk.targets
```diff
@@ -0,0 +1,140 @@
...
+      <Output TaskParameter="ExitCode" PropertyName="_FuncCliExitCode" />
+    </Exec>
+
+    <FuncSdkLog Condition="'$(_FuncCliExitCode)' != '0'" Resource="Error_CannotRunFuncCli" />
+  </Target>
+
+  <Target Name="_FunctionsComputeRunArguments" DependsOnTargets="_FunctionsCheckForCoreTools" BeforeTargets="ComputeRunArguments" >
+    <!-- Windows Configuration -->
+    <PropertyGroup Condition="'$(OS)' == 'Windows_NT'">
+      <RunCommand>cmd</RunCommand>
```
fabiocav: `RunCommand`, `RunArguments`, and `RunWorkingDirectory` are overwritten unconditionally. Customers/tests that intentionally set `RunCommand` (for example to run something other than `func`, or in containers without Core Tools) will be silently overridden. Please only set these defaults when the properties are empty and consider an explicit opt-out such as `FunctionsLaunchWithFuncCli=false`.

src/Azure.Functions.Sdk/Targets/Azure.Functions.Sdk.targets
```diff
@@ -0,0 +1,140 @@
...
+    </PropertyGroup>
+  </Target>
+
+  <Target Name="_PreGenerateWorkerConfig" Returns="@(_WorkerConfig)">
+    <ItemGroup>
+      <_WorkerConfig Include="$(_WorkerConfigOutputPath)" TargetPath="$(_WorkerConfigFileName)" />
+    </ItemGroup>
+  </Target>
+
+  <Target Name="_CoreGenerateWorkerConfig" Inputs="@(IntermediateAssembly)" Outputs="$(_WorkerConfigOutputPath)">
```
fabiocav, line 103: Will this work if you have a build followed by a self-contained publish without source changes, or will that use a stale worker config (`dotnet` vs `{WorkerRoot}<app>`)?

src/Azure.Functions.Sdk/Targets/Azure.Functions.Sdk.targets
```diff
@@ -0,0 +1,140 @@
...
+  <!--
+    Adds worker.config.json to ResolvedFileToPublish for the publish pipeline.
+    Sources from the intermediate output path; for NoBuild, the file must exist from a prior build.
+  -->
+  <Target Name="_AddWorkerConfigToPublish"
+    BeforeTargets="ComputeFilesToPublish"
+    DependsOnTargets="_PrepareWorkerConfig"
+    Condition="'$(DesignTimeBuild)' != 'true'">
+    <ItemGroup>
+      <ResolvedFileToPublish Include="$(_WorkerConfigOutputPath)" RelativePath="$(_WorkerConfigFileName)" CopyToPublishDirectory="PreserveNewest" />
```
fabiocav: For `dotnet publish --no-build`, `_GenerateWorkerConfig` is skipped, but `_AddWorkerConfigToPublish` still adds `$(_WorkerConfigOutputPath)` to `ResolvedFileToPublish`. If the prior build wrote outputs to a different intermediate/output location (common in CI build-then-publish staging), publish will fail later with a generic copy error. Please add an explicit existence check with an AZFW diagnostic, or ensure publish can generate/locate the config for no-build scenarios.

src/Azure.Functions.Sdk/Tasks/GenerateWorkerConfig.cs
```diff
@@ -0,0 +1,93 @@
...
+    {
+        public string Language => "dotnet-isolated";
+
+        public IEnumerable<string> Extensions => [".dll"];
+
+        public string DefaultExecutablePath => executable;
+
+        public string DefaultWorkerPath => entry;
+
+        public string WorkerIndexing => "true";
```
fabiocav, line 82: It's great that `workerIndexing` will be set to `true`. The existing Worker SDK has a `FunctionsEnableWorkerIndexing` property that enables customers to control this. Would be good to flag this so our migration guidance mentions the property. Perhaps we should also consider some validation with the new model.

src/Azure.Functions.Sdk/ZipDeploy/DeploymentClient.cs
```diff
@@ -0,0 +1,149 @@
...
+
+    private readonly Logger? _logger = logger;
+    private readonly HttpClient _client = Throw.IfNull(client);
+
+    public virtual async Task<DeployStatus> ZipDeployAsync(ZipDeployRequest request, CancellationToken cancellation)
+    {
+        Throw.IfNull(request);
+        using HttpRequestMessage httpRequest = request.CreateRequestMessage();
+        using HttpResponseMessage response = await _client.SendAsync(httpRequest, cancellation);
+        StatusResult status = await StatusResult.ParseAsync(response, cancellation);
```
fabiocav: This parses the initial response body before checking for an async `Location` header. Kudu can validly return `202 Accepted` + `Location` with an empty body (or non-JSON body), which would throw before polling starts even though deployment was accepted. Please check success/`Location` first, treat empty bodies as default status, and convert non-JSON/transport failures into clean MSBuild task errors.

src/Azure.Functions.Sdk/ZipDeploy/DeploymentClient.cs
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    private async Task<StatusResult> GetDeploymentStatusAsync(Uri location, CancellationToken cancellation)
+    {
+        int retry = 0;
+        while (true)
+        {
+            try
+            {
+                using HttpResponseMessage response = await _client.GetAsync(location, cancellation);
```
fabiocav, line 69: We set the authentication headers on the request for the initial POST, but since those headers are not set on the client, and they don't flow here, will this result in a 401? Is this polling working as intended?

src/Azure.Functions.Sdk/Tasks/Publish/ZipDeploy.cs
```diff
@@ -0,0 +1,170 @@
...
+public sealed class ZipDeploy(IFileSystem fileSystem, DeploymentClient? client = null)
+    : Microsoft.Build.Utilities.Task, ICancelableTask, IDisposable
+{
+    internal static readonly ProductInfoHeaderValue SdkUserAgentHeader = new(
+        ThisAssembly.Name, ThisAssembly.Version.ToString(fieldCount: 3));
+
+    internal static readonly ProductInfoHeaderValue OsUserAgentHeader = new(
+        $"({RuntimeInformation.OSDescription}; {RuntimeInformation.OSArchitecture})");
+
+    private static readonly TimeSpan DeployTimeout = TimeSpan.FromMinutes(3);
```
fabiocav, line 22: Does this match what we had before? Do we need to make this configurable?

src/Azure.Functions.Sdk/ZipDeploy/ZipDeployRequest.cs
```diff
@@ -0,0 +1,67 @@
...
+    private static readonly MediaTypeHeaderValue ZipContentHeader = new(MediaTypeNames.Application.Zip);
+
+    public bool UseBlobContainer { get; init; }
+
+    public Uri GetUri(Uri? baseUri = null)
+    {
+        Uri path = UseBlobContainer ? PublishPath : ZipDeployPath;
+        return baseUri is null
+            ? path
+            : new Uri(baseUri, path);
```
fabiocav: When `UseBlobContainerDeploy=true`, composing `new Uri(baseUri, "api/publish?RemoteBuild=false")` can drop the container path/query if `PublishUrl` is actually a SAS/container URL. Separately, `ZipDeploy` logs raw `PublishUrl`, which can leak SAS query tokens. Please add realistic SAS/blob-container deploy tests, preserve the intended URL shape, and redact query strings from logs.

src/Azure.Functions.Sdk/FunctionsAssemblyScanner.cs
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    /// <summary>
+    /// Checks if the given package name should be scanned or not.
+    /// </summary>
+    /// <param name="name">The name of the package.</param>
+    /// <returns><c>true</c> if package should be scanned, <c>false</c> otherwise.</returns>
+    public static bool ShouldScanPackage(string name)
+    {
+        return !string.IsNullOrEmpty(name) && !ExcludedPackagesRegex.IsMatch(name);
```
fabiocav: This is a broad negative filter: anything not matching these prefixes is scanned as a potential Functions extension. Combined with hard errors when Cecil cannot read a candidate assembly, an unrelated package with unusual/mixed/native assets can break Functions extension discovery. Please narrow the blast radius by positively identifying extension candidates where possible, or skip unreadable non-extension assemblies unless the package is already known to be a Functions extension.

src/Azure.Functions.Sdk/Targets/Publish/Azure.Functions.Sdk.Publish.ZipDeploy.targets
```diff
@@ -0,0 +1,43 @@
...
+  </PropertyGroup>
+
+  <Target Name="CreateZipFile" DependsOnTargets="_ResolveFunctionExecutable">
+    <CreateZipFile SourceFolder="$(PublishIntermediateOutputPath)" ZipName="$(MSBuildProjectName)"
+      PublishIntermediateTempPath="$(PublishIntermediateTempPath)" Executable="$(_FunctionsExecutable)">
+      <Output TaskParameter="CreatedZipPath" PropertyName="ZippedPublishContentsPath"/>
+    </CreateZipFile>
+  </Target>
+
+  <Target Name="ZipDeploy" DependsOnTargets="CreateZipFile">
```
fabiocav: `ZipDeploy` and `CreateZipFile` are public target names and `ZipDeploy` in particular is a canonical publish target name used by other publish targets. Once this SDK ships, these become customer-visible extension points/collision points. Please consider renaming implementation targets to `_FunctionsZipDeploy` / `_FunctionsCreateZipFile` and only exposing intentionally documented public targets.

src/Azure.Functions.Sdk/Tasks/Publish/ZipDeploy.cs
```diff
@@ -0,0 +1,170 @@
...
+            {
+                UserAgent = { SdkUserAgentHeader, dotnetSdkHeader, OsUserAgentHeader }
+            },
+        };
+    }
+
+    private async Task<bool> DeployAsync(DeploymentClient client, CancellationToken cancellation)
+    {
+        Log.LogMessageFromResources(
+            MessageImportance.High, nameof(Strings.Deploy_BeginPublish), ZipContentsPath, PublishUrl);
```
fabiocav, line 99: Could this potentially log sensitive information? Do we need to sanitize this URL or ensure we're not logging query strings?

src/Azure.Functions.Sdk/Tasks/Publish/CreateZipFile.cs
jviau: Opened issue to check. https://github.com/Azure/azure-functions-dotnet-worker/issues/3390

src/Azure.Functions.Sdk/Tasks/Extensions/WriteExtensionProject.cs
jviau: Opened an issue to check. https://github.com/Azure/azure-functions-dotnet-worker/issues/3391

src/Azure.Functions.Sdk/Targets/Azure.Functions.Sdk.props
jviau: https://github.com/Azure/azure-functions-dotnet-worker/issues/3392

CPM "works". This will have broader impact when mixed with projects that don't use this SDK, reference this package, and use CPM. Opened the above issue to re-evaluate this approach. I like the idea of the SDK providing a default, but the implicit package flow is bringing in some other less-than-idea approaches. 

> If so, we should be validating ManagePackageVersionsCentrally=true here and emitting package references with no no Version?

This might help - but we should also check that there is a `PackageVersion` for the item.

src/Azure.Functions.Sdk/Targets/Extensions/Azure.Functions.Sdk.Extensions.targets
jviau: https://github.com/Azure/azure-functions-dotnet-worker/issues/3393

src/Azure.Functions.Sdk/Targets/Extensions/Azure.Functions.Sdk.Extensions.targets
jviau: I was originally checking for this in VS, but I found VS had NO way of hooking into restore, so it was impossible to avoid this warning when developing via VS. I added this check to avoid emitting an unsolvable warning. When we get official generic restore hooks this will be addressed.

src/Azure.Functions.Sdk/Targets/Azure.Functions.Sdk.targets
jviau: https://github.com/Azure/azure-functions-dotnet-worker/issues/3394

src/Azure.Functions.Sdk/Tasks/GenerateWorkerConfig.cs
jviau: https://github.com/Azure/azure-functions-dotnet-worker/issues/3395

src/Azure.Functions.Sdk/ZipDeploy/DeploymentClient.cs
jviau: https://github.com/Azure/azure-functions-dotnet-worker/issues/3396

src/Azure.Functions.Sdk/Tasks/Publish/ZipDeploy.cs
jviau: Yes, I believe I copied the values we had before: https://github.com/Azure/azure-functions-dotnet-worker/blob/main/sdk/Sdk/Tasks/ZipDeploy/ZipDeploymentStatus.cs#L84

Issue to evaluate having this be configurable: https://github.com/Azure/azure-functions-dotnet-worker/issues/3397

src/Azure.Functions.Sdk/Tasks/Publish/ZipDeploy.cs
jviau: The current SDK logs the url. 

https://github.com/Azure/azure-functions-dotnet-worker/issues/3398

src/Azure.Functions.Sdk/Targets/Extensions/Azure.Functions.Sdk.Extensions.props
jviau: Copied from the current SDK. It is mostly done manually. We can add more values as needed. It is a lot less important in this SDK as the inner project is never built (only specific targets called on it). I'll look into expanding this list.

https://github.com/Azure/azure-functions-dotnet-worker/issues/3399

src/Azure.Functions.Sdk/Targets/Azure.Functions.Sdk.targets
jviau: Yeah, I intentionally did this change. Do you see much value in the SDK needing a hard update to support new TFMs? To me this doesn't seem valuable, particularly if we ever support BYOC. Customer's could be deploying their own container with the new TFM before we do - why block the build? Also as different SKUs add new TFM support when do we opt to enable it in the SDK? We technically haven't 'fixed' the problem at all, some skus may work, others won't. Also why block local development in preparation for a TFM being supported before it is added?

I don't think this SDK is the right place to be deciding if a sku supports a TFM or not.

src/Azure.Functions.Sdk/Targets/Inner/Azure.Functions.Sdk.Inner.RuntimePackages.props
jviau: Manually done for now.

https://github.com/Azure/azure-functions-dotnet-worker/issues/3400

fabiocav (approved):
Approving as we have issues tracking the concerns raised. We should be able to cut a preview while iterating on those.

PR #3401, opened 2026-05-04
Remove LeaderEpoch from KafkaRecord and update host extension to 4.3.2

TsuyoshiUshio:
## Summary

Remove `LeaderEpoch` from `KafkaRecord` and update host extension dependency to 4.3.2, aligning with host-side changes in [azure-functions-kafka-extension#640](https://github.com/Azure/azure-functions-kafka-extension/pull/640).

`LeaderEpoch` represents consumer fetch/offset metadata (from `ConsumeResult.LeaderEpoch` in Confluent.Kafka), not stored Kafka record data. Per customer feedback in [azure-functions-kafka-extension#639](https://github.com/Azure/azure-functions-kafka-extension/issues/639), it does not belong on the `KafkaRecord` user-facing type.

### Proto schema
- Removed `leader_epoch` (field 8) from `KafkaRecordProto`
- Reserved field number 8 and name `leader_epoch` to prevent future reuse
- Schema now matches the host-side copy in azure-functions-kafka-extension 4.3.2

### KafkaRecord POCO
- Removed `public int? LeaderEpoch { get; set; }` property

### KafkaRecordConverter
- Removed `LeaderEpoch` mapping from Protobuf deserialization

### Tests
- Removed `ConvertAsync_PreservesLeaderEpoch` and `ConvertAsync_NoLeaderEpoch_ReturnsNull` tests
- Removed `LeaderEpoch` from test helper `CreateTestKafkaRecordProto()` and assertion helper `AssertKafkaRecord()`

### Dependencies
- Updated `WebJobsExtension` reference from 4.3.1 to 4.3.2

### Sample
- Added `KafkaRecordSample` isolated worker sample under `extensions/Worker.Extensions.Kafka/samples/`
- Demonstrates KafkaRecord deferred binding with Produce, Consume, and Status endpoints

## Testing
- Unit tests: 10/10 passed
- E2E manual test: `func start` with local Kafka broker, produce/consume/status flow verified
- `KafkaRecord` correctly binds with topic, partition, offset, key, value, timestamp, headers — no `LeaderEpoch`

## Related
- azure-functions-kafka-extension#639
- azure-functions-kafka-extension#640

### DIFF
--- a/extensions/Worker.Extensions.Kafka/src/KafkaRecord.cs
+++ b/extensions/Worker.Extensions.Kafka/src/KafkaRecord.cs
@@ -43,10 +43,5 @@ namespace Microsoft.Azure.Functions.Worker
         /// Gets or sets the record headers.
         /// </summary>
         public KafkaHeader[] Headers { get; set; }
-
-        /// <summary>
-        /// Gets or sets the leader epoch, if available. Null if not provided by the broker.
-        /// </summary>
-        public int? LeaderEpoch { get; set; }
     }
 }
--- a/extensions/Worker.Extensions.Kafka/release_notes.md
+++ b/extensions/Worker.Extensions.Kafka/release_notes.md
@@ -4,6 +4,11 @@
 - My change description (#PR/#issue)
 -->
 
+### Microsoft.Azure.Functions.Worker.Extensions.Kafka 4.3.0
+
+- **Breaking**: Remove `LeaderEpoch` from `KafkaRecord` — it is consumer fetch metadata, not stored record data (azure-functions-kafka-extension#639)
+- Update `Microsoft.Azure.WebJobs.Extensions.Kafka` dependency from 4.3.1 to 4.3.2
+
 ### Microsoft.Azure.Functions.Worker.Extensions.Kafka 4.2.0
 
 - Add `KafkaRecord` type for raw Apache Kafka record binding with full metadata access (topic, partition, offset, key/value as raw bytes, headers, timestamp, leader epoch) via Protobuf deserialization (#3356)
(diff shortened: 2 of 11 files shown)


extensions/Worker.Extensions.Kafka/src/KafkaRecord.cs
```diff
@@ -43,10 +43,5 @@ public class KafkaRecord
         /// Gets or sets the record headers.
         /// </summary>
         public KafkaHeader[] Headers { get; set; }
-
-        /// <summary>
-        /// Gets or sets the leader epoch, if available. Null if not provided by the broker.
-        /// </summary>
-        public int? LeaderEpoch { get; set; }
```
satvu, line 50: I feel like this is technically a breaking change (since it was already introduced/released) but since the WebJobs version seems to have already been released with this change I guess we will just have to emphasize this in the notes.

satvu (approved):

extensions/Worker.Extensions.Kafka/src/KafkaRecord.cs
TsuyoshiUshio: Thank you @satvu . Technically yes, but actual impact is nothing. For the context, We released last version that includes `KafkaRecrod` But issue was when user use the `KafkaRecord` that mean no user can use this feature, and we found that `LeaderEpoch` should be removed. So that it is very good chance to remove it without introducing any impact.

TsuyoshiUshio: Other users who doesn't use `KafkaRecord` will be no impacted.

PR #3413, opened 2026-05-13
Set working directory during placeholder reuse specialization

### LINKED ISSUE #3374: [HOST] Functions native host is running worker with wrong CWD after specialization
### Description

This needs more verification, but it looks like after specialization a customer's app will be running with the wrong working directory. 

In flex, the CWD will be `/tmp/functions/standby/wwwroot/` when it should be `/home/site/wwwroot/`. The root cause appears to be that the native does not respect `FunctionEnvironmentReloadRequest.FunctionAppDirectory`. The host _should_ update the current working directory with that request.

### Steps to reproduce

Can be reproduced with specialization simulation in the host. On app specialization, the working directory remains as the placeholder temp dir.

### Possible Fix

The functions native host may be able to just update CWD before loading the customer's app.

Tonewall:
resolves #3374 

**Fix: Set working directory during placeholder reuse specialization**

During placeholder reuse (`UsePlaceholderDotNetIsolated`), the FunctionsNetHost worker process starts with its working directory set to `/tmp/functions/standby/wwwroot/`. When the `FunctionEnvironmentReloadRequest` arrives during specialization, the working directory was not updated to `/home/site/wwwroot/` before loading the customer app. This causes apps that load config files via relative paths (e.g., `builder.Configuration.AddJsonFile("sharedsettings.json", optional: false)`) to crash with `FileNotFoundException` (exit code 134).

**Impact:**
- Every cold start for affected apps drops 1 HTTP request (30s timeout → 404) before the host recycles and starts a fresh worker
- The crash triggers the CV1 migration revert action runner to pin apps back to ACI despite the app running healthy after recovery
- Observed in production on multiple customer apps (dotnet-isolated 9.0 on Legion)

**Fix:**
- In `IncomingGrpcMessageHandler`, set `Directory.SetCurrentDirectory(envReloadRequest.FunctionAppDirectory)` before calling `_appLoader.RunApplication()` 
- This matches the behavior of the fresh worker process path, where `ProcessStartInfo.WorkingDirectory` is set to `/home/site/wwwroot/` at spawn time
- The fix is scoped to the reload path — it does not disable the placeholder reuse cold start optimization
- Only apps loading config files via relative paths are affected; apps using environment variables or absolute paths are unaffected
- No risk to existing apps: fresh worker processes already run with CWD = `/home/site/wwwroot/`, this change makes placeholder reuse consistent

**Root cause analysis:** The `FunctionAppDirectory` field was already sent by the host in the `FunctionEnvironmentReloadRequest` proto message but was not used to update the process working directory. ACI does not hit this bug because it uses `UsePlaceholderDotNetIsolated: False` (kills the placeholder and starts a fresh process).

### Pull request checklist

### Additional information

- **Affected runtime:** dotnet-isolated 9.0 on Legion (Linux Consumption CV1 Migration)
- **Not affected:** ACI (uses fresh worker process), apps without relative-path config files
- **Reproduction:** Deploy a dotnet-isolated 9.0 function app with `AddJsonFile("sharedsettings.json", optional: false)` to a Legion-backed Linux Consumption stamp with `WEBSITE_USE_PLACEHOLDER_DOTNETISOLATED=1`
- **Validated by investigation:** [giftup-inbound](https://github.com/serverless-paas-balam/dobby/blob/main/agents/cv1-functions/reports/cv1-app-investigation-giftup-inbound-2026-05-06.md) — 72,557 successful executions after recovery, reverted by action runner due to transient crash log

### DIFF
--- a/host/src/FunctionsNetHost/Grpc/IncomingGrpcMessageHandler.cs
+++ b/host/src/FunctionsNetHost/Grpc/IncomingGrpcMessageHandler.cs
@@ -67,6 +67,18 @@ namespace FunctionsNetHost.Grpc
 
                     var envReloadRequest = msg.FunctionEnvironmentReloadRequest;
 
+                    // Set working directory to the function app's content root before loading the app.
+                    // During placeholder reuse, the worker process was started with a standby working
+                    // directory (/tmp/functions/standby/wwwroot). The reload request provides the real
+                    // content path (/home/site/wwwroot) which must be set before the app loads,
+                    // otherwise config files loaded via relative paths will fail.
+                    if (!string.IsNullOrEmpty(envReloadRequest.FunctionAppDirectory)
+                        && Directory.Exists(envReloadRequest.FunctionAppDirectory))
+                    {
+                        Directory.SetCurrentDirectory(envReloadRequest.FunctionAppDirectory);
+                        Logger.LogTrace($"Set working directory to {envReloadRequest.FunctionAppDirectory}");
+                    }
+
                     var workerConfig = await WorkerConfigUtils.GetWorkerConfig(envReloadRequest.FunctionAppDirectory);
 
                     if (workerConfig?.Description is null)


jviau:
@fabiocav @kshyju -- this PR addresses that issue I discussed the other week, where we don't update the CWD for the function app post specialization.

I wasn't sure where we landed with the fix. It is a straightforward fix, but are we worried about a breaking behavior change? If so, how can we scope this to a specific update?

Tonewall:
Abandoning to address separately for cv1 migration via dockerfile update

jviau:
Re-opening PR, rebased on main, removed the CV1-scoping commit.

We have decided to take this change

brettsam (approved):
Verified the change locally in docker w/ full specialization. Looks good!

fabiocav (approved):

brettsam:
Note that if you are hitting this issue -- look into the workaround here: https://github.com/Azure/azure-functions-dotnet-worker/issues/3374#issuecomment-4984606378

PR #3448, opened 2026-07-01
fix(sdk): redact sensitive info from ZipDeploy logs

### LINKED ISSUE #3398: [SDK] Evaluate ZipDeploy logs
https://github.com/Azure/azure-functions-dotnet-worker/pull/3387#discussion_r3179482832

jviau:

resolves #3398

Follow-up to the review question on #3387 ([discussion](https://github.com/Azure/azure-functions-dotnet-worker/pull/3387#discussion_r3179482832)): _"Could this potentially log sensitive information? Do we need to sanitize this URL or ensure we're not logging query strings?"_

The concern is valid. Publish/deployment URLs can carry secrets:
- **Credentials** embedded as user info (`https://user:password@host`).
- **Tokens** in the query string (e.g. a SAS `?sig=...`), notably for blob-container / pre-signed scenarios.

These were being written verbatim to build logs (frequently captured by CI systems and shared). I also confirmed that `new Uri(baseUri, relativePath)` **preserves the base URI's user info**, so the resolved deployment URI logged on success/failure would leak embedded credentials even though the query is replaced.

**Fix:** Added `UriLogExtensions.ToLogSafeString()` (`Uri` and `string` overloads) that retains only scheme, host, optional port, and path — stripping user info, query, and fragment (best-effort for values that don't parse as an absolute URI, and for relative URIs). Applied it to every URL log site:
- `ZipDeploy.cs`: `Deploy_BeginPublish`, `Deploy_CompletedSuccess`, `Deploy_CompletedFailure`, `Deploy_Failed`, `Deploy_InvalidPublishUrl`, `Deploy_InsecurePublishUrl`.
- `DeploymentClient.cs`: `Deploy_AsyncDeployment`, `Deploy_PollFailure` (server-provided `Location` header).

### Pull request checklist

### Additional information

- New `UriLogExtensionsTests` unit tests cover user-info/query/fragment stripping, non-default vs default ports, http scheme, relative URIs, null/empty, and best-effort redaction of non-absolute strings.
- Added a regression test in `ZipDeployTaskTests` proving credentials and query in `PublishUrl` are redacted from the logged deployment URL.
- Updated two existing `ZipDeployTaskTests` expectations (the logged URL no longer includes the `?isAsync=true` query).
- Test secret placeholders use `PLACEHOLDER` to avoid secret scanning.
- Builds clean on `netstandard2.0` and `net472`; full `Azure.Functions.Sdk.Tests` suite passes on `net472` and `net10.0`.

### DIFF
--- a/src/Azure.Functions.Sdk/UriLogExtensions.cs
+++ b/src/Azure.Functions.Sdk/UriLogExtensions.cs
@@ -71,7 +71,8 @@ internal static class UriLogExtensions
 
     /// <summary>
     /// Best-effort redaction for values that do not parse as an absolute URI. Removes any
-    /// query string or fragment (which may contain secrets) and any user info component.
+    /// query string or fragment (which may contain secrets) and any user info component
+    /// within an authority.
     /// </summary>
     private static string RedactRawUrl(string url)
     {
@@ -79,13 +80,22 @@ internal static class UriLogExtensions
         int queryOrFragment = url.IndexOfAny(['?', '#']);
         string result = queryOrFragment >= 0 ? url[..queryOrFragment] : url;
 
-        // Strip user info (anything between the "//" authority marker and the "@").
+        // Strip user info only when it appears in the authority.
         int authorityStart = result.IndexOf("//", StringComparison.Ordinal);
-        int start = authorityStart >= 0 ? authorityStart + 2 : 0;
-        int at = result.IndexOf('@', start);
-        if (at >= 0)
+        if (authorityStart >= 0)
         {
-            result = result[..start] + result[(at + 1)..];
+            int start = authorityStart + 2;
+            int authorityEnd = result.IndexOf('/', start);
+            if (authorityEnd < 0)
+            {
+                authorityEnd = result.Length;
+            }
+
+            int at = result.IndexOf('@', start);
+            if (at >= start && at < authorityEnd)
+            {
+                result = result[..start] + result[(at + 1)..];
+            }
         }
 
         return result;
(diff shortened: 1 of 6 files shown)


src/Azure.Functions.Sdk/UriLogExtensions.cs
```diff
@@ -0,0 +1,93 @@
...
+        string result = queryOrFragment >= 0 ? url[..queryOrFragment] : url;
+
+        // Strip user info (anything between the "//" authority marker and the "@").
+        int authorityStart = result.IndexOf("//", StringComparison.Ordinal);
+        int start = authorityStart >= 0 ? authorityStart + 2 : 0;
+        int at = result.IndexOf('@', start);
+        if (at >= 0)
+        {
+            result = result[..start] + result[(at + 1)..];
+        }
```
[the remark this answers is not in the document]
jviau: Good catch, accepted. `RedactRawUrl` now only strips user-info when a `//` authority marker is present and the `@` falls within the authority region (before the first path `/`), so values like `mailto:user@example.com`, the relative path `api/@me`, and `@` characters in a path are preserved. Added regression tests covering these cases plus the real user-info stripping case (`//user:PLACEHOLDER@host/path` -> `//host/path`); all UriLogExtensions tests pass on net472 and net10.0.

kshyju (approved):

src/Azure.Functions.Sdk/UriLogExtensions.cs
```diff
@@ -0,0 +1,103 @@
...
+        if (authorityStart >= 0)
+        {
+            int start = authorityStart + 2;
+            int authorityEnd = result.IndexOf('/', start);
+            if (authorityEnd < 0)
+            {
+                authorityEnd = result.Length;
+            }
+
+            int at = result.IndexOf('@', start);
```
satvu: Any concern that a password will have '@' in it?

jviau: Good catch. This would only partially redact the password. Fixed by looking for the last `@` before the host segment.

satvu (approved):

PR #3455, opened 2026-07-15
fix(grpc): avoid capturing ambient SynchronizationContext in worker stream loops

### LINKED ISSUE #3454: Isolated .NET worker's outbound gRPC write pump can hang indefinitely when it captures a short-lived single-threaded SynchronizationContext
### Description

The .NET isolated worker uses a **single process-level outbound gRPC write pump** (`GrpcWorkerClient.StartWriterAsync`) to send *all* messages to the host — both `InvocationResponse` (the "function complete" signal) and forwarded logs (`RpcLog`) travel over the **same** outbound channel.

That pump `await`s **without `ConfigureAwait(false)`**. If application code runs part of its work under a short-lived, single-threaded **`SynchronizationContext`** (for example **Nito.AsyncEx `AsyncContext`**) and emits a log on that thread, the channel (which uses `AllowSynchronousContinuations = true`, the SDK default) resumes the idle write pump *inline* onto that context, and the pump's next `await WriteAsync(...)` **captures** it. When that short-lived context is then disposed, the captured continuation is posted to a now-dead context, **silently dropped, and never resumes**.

Because this is the only outbound pump, the worker then stops sending *everything* to the host — including the already-produced `InvocationResponse`. The invocation **hangs indefinitely (the worker never recovers on its own)**, and the host eventually fails it once the **configured `functionTimeout`** elapses. The worker process otherwise stays healthy, so there is no crash and no error log — the outbound direction simply goes silent. Because a **log** is what wakes the pump onto the app context, the failure is **intermittent and correlates with log volume**.

Example of application code that can trigger it (a sync-over-async wrapper that logs inside the context):

```csharp
// Runs the async work on a short-lived, single-threaded SynchronizationContext.
// AsyncContext.Run installs the context, then disposes it when it returns.
public void SendEvent(MyEvent evt)
{
    Nito.AsyncEx.AsyncContext.Run(async () =>
    {
        // A log emitted here runs under the single-threaded context.
        // It is forwarded over the shared outbound gRPC channel and can
        // inline-resume the worker's write pump onto THIS context.
        _logger.LogInformation("Publishing event {Id}", evt.Id);

        await _publisher.PublishAsync(evt); // any awaited work
    });
    // <-- context is disposed here; if the write pump captured it above,
    //     its continuation is now posted to a dead context and lost.
}
```

**Fix:** the worker's background stream loops should `await ... .ConfigureAwait(false)` so the write pump never captures an ambient application `SynchronizationContext`.

**Environment:**
- Package: `Microsoft.Azure.Functions.Worker.Grpc`
- Model: .NET isolated worker (out-of-process)
- Affects versions whose `StartWriterAsync` awaits **without** `ConfigureAwait(false)`.
- Affected path: `GrpcWorkerClient.StartWriterAsync` (and the sibling stream loops that share the pattern).

### Steps to reproduce

1. Create a .NET isolated Functions app (`Microsoft.Azure.Functions.Worker.Grpc` transport).
2. From function code, run some work inside a single-threaded `SynchronizationContext` that is disposed promptly — e.g. the `AsyncContext.Run(...)` wrapper shown above — and **emit a log inside that scope**.
3. Run under realistic load / log volume to hit the timing window (context disposed before the pump's inline-captured write completes).
4. **Observed:** the write pump's continuation is lost; the worker stops sending to the host, the `InvocationResponse` never leaves, and the invocation hangs until the host fails it at the **configured `functionTimeout`**. Subsequent invocations on the same instance also hang.
5. **Expected:** the outbound write pump runs independently of any ambient `SynchronizationContext` and keeps draining regardless of application-installed contexts.

feiqian:

resolves #3454

### Summary

The worker's single process-level outbound gRPC write pump
(`GrpcWorkerClient.StartWriterAsync`) awaits without `ConfigureAwait(false)`.
Under a specific but real combination of conditions it captures a short-lived,
application-installed single-threaded `SynchronizationContext` (for example
**Nito.AsyncEx `AsyncContext`**). When that context is torn down, the pump's
captured continuation is posted to the now-dead context, silently dropped, and
never resumes. Because this is the **only** outbound pump, the worker stops
sending *all* messages to the host — including `InvocationResponse` — so the
invocation **hangs indefinitely** (the worker never recovers on its own) and
the host eventually fails it once the **configured `functionTimeout`** elapses.

### Root cause

The hang requires four conditions to line up — the missing `ConfigureAwait(false)`
is only one of them, which is why it is intermittent:

1. **Synchronous continuations** — the outbound channel is created with
   `AllowSynchronousContinuations = true` (the SDK default), so a writer can
   resume the pump's parked read-continuation *inline* on the writing thread.
2. **An app single-threaded context** — the outbound gRPC channel is shared by
   both `InvocationResponse` and forwarded `RpcLog` messages. When application
   code runs under a single-threaded, *derived* `SynchronizationContext`
   (e.g. `AsyncContext`) and writes a **log** on that thread, `TryWrite`
   inline-resumes the write pump onto that context.
   - Note: the pump does **not** capture at its first suspension — that happens
     on the thread pool at `ReadAllAsync()` where `Current == null`. Capture
     occurs only *after* the inline resume described here moves the pump onto
     the app context.
3. **No `ConfigureAwait(false)`** — now running on the app context, the pump's
   next `await requestStream.WriteAsync(...)` captures that context as its
   continuation target. (This is what the fix removes.)
4. **The context is disposed first** — the short-lived context completes and is
   disposed before the awaited write finishes. When the write completes, the
   continuation is `Post`-ed to the dead context; the post throws internally and
   is swallowed, so the continuation is physically lost and the pump stalls
   permanently.

Higher log volume from the affected thread increases the probability of the
inline wake-up in (2), which is why the failure is intermittent and correlates
with logging activity.

### Fix

Add `ConfigureAwait(false)` to the awaits in the worker's background stream
loops so continuations resume on the thread pool:

- `GrpcWorkerClientFactory.GrpcWorkerClient`: `StartAsync` →
  `SendStartStreamMessageAsync`; `SendStartStreamMessageAsync` → `WriteAsync`;
  `StartWriterAsync` (channel enumeration + per-message `WriteAsync`);
  `StartReaderAsync` (`MoveNext` + `ProcessMessageAsync`).
- `NativeWorkerClient`: `ProcessInbound` (channel enumeration +
  `ProcessMessageAsync`) and `ProcessOutbound` (channel enumeration).

**Regression safety:** these loops only do channel reads, gRPC stream I/O,
native sends, and message dispatch — none reads `SynchronizationContext.Current`
or thread-static state, so removing the accidental capture is
behavior-preserving and is the intended scheduling for background loops.
`ChannelReaderExtensions.ReadAllAsync` already uses `ConfigureAwait(false)`
internally; the missing piece was the consuming `await foreach` at the call
sites (required on all TFMs since net6.0+ binds the BCL `ReadAllAsync`).
`SendMessageAsync` is a `ValueTask` passthrough with no local await and is
unchanged.

### Tests

Adds `GrpcWorkerClientContinuationTests`, which constructs the real internal
`GrpcWorkerClient` and drives the actual private `StartWriterAsync` under a
single-threaded `SynchronizationContext` that is shut down mid-loop. It
asserts both queued messages still drain. The test **fails (times out) if the
`ConfigureAwait(false)` calls are removed from production**, so it guards
against regression of the exact hang.

### Pull request checklist

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -14,4 +14,4 @@
 
 ### Microsoft.Azure.Functions.Worker.Grpc <version>
 
-- <entry>
+- Ensured gRPC worker background stream loops do not capture an ambient SynchronizationContext, preventing a worker hang when the application uses a single-threaded context (e.g. Nito.AsyncEx AsyncContext).
--- a/release_notes.md
+++ b/release_notes.md
@@ -14,4 +14,4 @@
 
 ### Microsoft.Azure.Functions.Worker.Grpc <version>
 
-- Ensured gRPC worker background stream loops do not capture an ambient SynchronizationContext, preventing a worker hang when the application uses a single-threaded context (e.g. Nito.AsyncEx AsyncContext).
+- Ensured gRPC worker background stream loops do not capture an ambient SynchronizationContext, preventing a worker hang when the application uses a single-threaded context (e.g. Nito.AsyncEx AsyncContext). (#3455)
--- a/test/DotNetWorker.Tests/GrpcWorkerClientContinuationTests.cs
+++ b/test/DotNetWorker.Tests/GrpcWorkerClientContinuationTests.cs
@@ -1,0 +1,221 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Threading;
+using System.Threading.Channels;
+using System.Threading.Tasks;
+using Grpc.Core;
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Azure.Functions.Worker.Grpc;
+using Microsoft.Azure.Functions.Worker.Grpc.Messages;
+using Xunit;
+
+namespace Microsoft.Azure.Functions.Worker.Tests;
+
+public class GrpcWorkerClientContinuationTests
+{
+    [Fact]
+    public void StartWriterAsyncContinuationDoesNotRequireOriginalSynchronizationContext()
+    {
+        SynchronizationContext originalContext = SynchronizationContext.Current;
+        using var context = new DroppingSingleThreadSynchronizationContext();
+
+        try
+        {
+            SynchronizationContext.SetSynchronizationContext(context);
+
+            Channel<StreamingMessage> channel = Channel.CreateUnbounded<StreamingMessage>();
+            var outputChannel = new GrpcHostChannel(channel);
+            object client = CreateGrpcWorkerClient(outputChannel);
+            var requestStream = new FakeClientStreamWriter();
+
+            Task writerTask = StartWriterAsync(client, requestStream);
+
+            var firstMessage = new StreamingMessage { RequestId = "1" };
+            var secondMessage = new StreamingMessage { RequestId = "2" };
+
+            Assert.True(channel.Writer.TryWrite(firstMessage));
+            TaskCompletionSource<bool> firstWrite = WaitFor(requestStream.FirstWriteStarted.Task, "The first write did not start.");
+
+            context.Shutdown();
+            firstWrite.SetResult(true);
+
+            Assert.True(channel.Writer.TryWrite(secondMessage));
+            channel.Writer.Complete();
+
+            TaskCompletionSource<bool> secondWrite = WaitFor(requestStream.SecondWriteStarted.Task, "The second write did not start after the context was shut down.");
+            secondWrite.SetResult(true);
+
+            WaitFor(writerTask, "The production write loop did not drain both messages.");
+
+            Assert.Collection(requestStream.WrittenMessages,
+                message => Assert.Same(firstMessage, message),
+                message => Assert.Same(secondMessage, message));
+        }
+        finally
+        {
+            SynchronizationContext.SetSynchronizationContext(originalContext);
+        }
+    }
+
+    private static object CreateGrpcWorkerClient(GrpcHostChannel outputChannel)
+    {
+        Type clientType = typeof(GrpcWorkerClientFactory).GetNestedType("GrpcWorkerClient", BindingFlags.NonPublic);
+        Assert.NotNull(clientType);
+
+        var startupOptions = new GrpcWorkerStartupOptions
+        {
+            HostEndpoint = new Uri("http://localhost:5000"),
+            WorkerId = "test-worker",
+            GrpcMaxMessageLength = 1024
+        };
+
+        object client = Activator.CreateInstance(
+            clientType,
+            BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic,
+            binder: null,
+            args: new object[] { outputChannel, startupOptions, new NoOpMessageProcessor() },
+            culture: null);
+
+        Assert.NotNull(client);
+        return client;
+    }
+
+    private static Task StartWriterAsync(object client, IClientStreamWriter<StreamingMessage> requestStream)
+    {
+        MethodInfo method = client.GetType().GetMethod("StartWriterAsync", BindingFlags.Instance | BindingFlags.NonPublic);
+        Assert.NotNull(method);
+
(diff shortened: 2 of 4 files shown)


feiqian:
@microsoft-github-policy-service agree company="Microsoft"

release_notes.md
```diff
@@ -14,4 +14,4 @@
 
 ### Microsoft.Azure.Functions.Worker.Grpc <version>
 
-- <entry>
\ No newline at end of file
+- Ensured gRPC worker background stream loops do not capture an ambient SynchronizationContext, preventing a worker hang when the application uses a single-threaded context (e.g. Nito.AsyncEx AsyncContext).
```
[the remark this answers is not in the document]
feiqian: Thanks! Fixed both points:

1. **Format** — added the PR reference `(#3455)` to the release-note bullet so it matches the documented format at the top of the file (commit `1cf74d1`).
2. **Checklist conflict** — updated the PR description: since this change *is* added to the release notes, I unchecked "My changes should not be added to the release notes" and kept the "I've added my notes to `release_notes.md`" item checked. The checklist now matches the release-note entry.

test/DotNetWorker.Tests/GrpcWorkerClientContinuationTests.cs
```diff
@@ -0,0 +1,221 @@
...
+    public void StartWriterAsyncContinuationDoesNotRequireOriginalSynchronizationContext()
+    {
+        SynchronizationContext originalContext = SynchronizationContext.Current;
+        using var context = new DroppingSingleThreadSynchronizationContext();
+
+        try
+        {
+            SynchronizationContext.SetSynchronizationContext(context);
+
+            Channel<StreamingMessage> channel = Channel.CreateUnbounded<StreamingMessage>();
```
fabiocav: This test does not exercise the production path described by the PR. Can you please construct the channel with the production options: 

https://github.com/Azure/azure-functions-dotnet-worker/blob/bb5d998d26dd0283f87144a939df64c732bed792/src/DotNetWorker.Grpc/GrpcServiceCollectionExtensions.cs#L21-L32

Start the pump with SynchronizationContext.Current == null, and perform the first write on the context thread.

Ideally use separate cases for an incomplete gRPC write and a synchronously completed write followed by an empty channel read, so each ConfigureAwait(false) is independently guarded.

fabiocav (commented):
Thank you for the contribution, @feiqian 

The changes look good, but I've added a comment to the test to make sure we're correctly exercising the surface that was modified.

test/DotNetWorker.Tests/GrpcWorkerClientContinuationTests.cs
feiqian: Thanks, this is a great catch — you're right that the original test bypassed the real ignition path. Reworked in `58e9110`:

1. **Production channel options** — the channel is now built via the production `RegisterOutputChannel()` registration (so `AllowSynchronousContinuations = true`, `SingleReader = true`, `SingleWriter = false`), instead of default `Channel.CreateUnbounded()`. This also guards against future drift in the registered options.
2. **Real ignition sequence** — `StartWriterAsync` is started with `SynchronizationContext.Current == null`, so the pump first parks at the empty channel read *without* capturing (asserted via the returned task being incomplete). The first write is then performed **on the single-thread context thread**, and `AllowSynchronousContinuations = true` inline-resumes the pump onto that context. Each test asserts the first `WriteAsync` actually observed the custom context, so it fails loudly if the ignition path isn't reproduced.
3. **Separate cases per await** —
   - `...IncompleteWriteContinuation...`: the fake `WriteAsync` returns a pending task, so the captured continuation is on `await WriteAsync(...)` — independently guards that `ConfigureAwait(false)`.
   - `...SecondReadContinuation...`: the fake `WriteAsync` completes synchronously, so the pump loops back and parks at the next empty read; the continuation is on the `await foreach (...ReadAllAsync())` — independently guards that `ConfigureAwait(false)`.

Both cases shut the context down before completing the awaited operation, so against the unfixed code the dropped continuation would time out. All gates are deterministic (task-completion signals, not sleeps). PTAL.

jviau (approved):

fabiocav (approved):

fabiocav:
Thank you for the contribution, @feiqian !

PR #3466, opened 2026-07-20
Add configurable MaxTelemetryBufferDelay option for Application Insights

### LINKED ISSUE #3465: Make Application Insights ServerTelemetryChannel flush interval (MaxTelemetryBufferDelay) configurable in the worker
### Summary

The Application Insights `ServerTelemetryChannel` (registered by `AddApplicationInsightsTelemetryWorkerService()`) buffers telemetry in memory and flushes it to the ingestion endpoint based on `MaxTelemetryBufferDelay` (default **30 seconds** in the App Insights SDK, or sooner once the buffer fills). The `Microsoft.Azure.Functions.Worker.ApplicationInsights` integration does not expose this, so customers cannot tune how quickly telemetry becomes visible without reaching into the channel themselves.

### Request

Expose the channel flush interval as a configurable option on the Functions worker Application Insights integration, and apply it to the `ServerTelemetryChannel`.

### Proposed behavior

- New public `FunctionsApplicationInsightsOptions.MaxTelemetryBufferDelay` (`TimeSpan`), default **8 seconds**, minimum **5 seconds**.
- Configurable via a new `ConfigureFunctionsApplicationInsights(Action<FunctionsApplicationInsightsOptions>)` overload:

```csharp
services
    .AddApplicationInsightsTelemetryWorkerService()
    .ConfigureFunctionsApplicationInsights(o => o.MaxTelemetryBufferDelay = TimeSpan.FromSeconds(8));
```

- Only applied when the configured telemetry channel is a `ServerTelemetryChannel`. Values below the 5s minimum throw `ArgumentOutOfRangeException`.

RohitRanjanMS:
Fixes #3465

Adds a public `FunctionsApplicationInsightsOptions` with a `MaxTelemetryBufferDelay` setting (default 8s, minimum 5s) that is applied to the Application Insights `ServerTelemetryChannel`, letting customers tune how quickly telemetry is flushed to the ingestion endpoint.

### Usage

```csharp
services
    .AddApplicationInsightsTelemetryWorkerService()
    .ConfigureFunctionsApplicationInsights(o => o.MaxTelemetryBufferDelay = TimeSpan.FromSeconds(8));
```

### Changes
- New public `FunctionsApplicationInsightsOptions.MaxTelemetryBufferDelay` (`TimeSpan`, default 8s, min 5s).
- New `ConfigureFunctionsApplicationInsights(Action<FunctionsApplicationInsightsOptions>)` overload; the existing parameterless overload is unchanged and source-compatible.
- Applied to the channel in `TelemetryConfigurationSetup` (only when it is a `ServerTelemetryChannel`); values below 5s throw `ArgumentOutOfRangeException`.
- Added `Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel` package reference.
- Added unit tests (defaults, override, below-minimum, null-guard, registration parity across both overloads).
- Bumped the package version to **2.52.0** and updated `release_notes.md`.

### DIFF
--- a/src/DotNetWorker.ApplicationInsights/DotNetWorker.ApplicationInsights.csproj
+++ b/src/DotNetWorker.ApplicationInsights/DotNetWorker.ApplicationInsights.csproj
@@ -6,7 +6,7 @@
     <AssemblyName>Microsoft.Azure.Functions.Worker.ApplicationInsights</AssemblyName>
     <RootNamespace>Microsoft.Azure.Functions.Worker.ApplicationInsights</RootNamespace>
     <MajorProductVersion>2</MajorProductVersion>
-    <MinorProductVersion>51</MinorProductVersion>
+    <MinorProductVersion>52</MinorProductVersion>
     <PatchProductVersion>0</PatchProductVersion>
     <PackageReadmeFile>README.md</PackageReadmeFile>    
     <BeforePack>$(BeforePack);GetReleaseNotes</BeforePack>
--- a/src/DotNetWorker.ApplicationInsights/DotNetWorker.ApplicationInsights.csproj
+++ b/src/DotNetWorker.ApplicationInsights/DotNetWorker.ApplicationInsights.csproj
@@ -17,6 +17,7 @@
   <ItemGroup>
     <PackageReference Include="Azure.Identity" Version="1.17.0" />
     <PackageReference Include="Microsoft.ApplicationInsights.PerfCounterCollector" Version="2.23.0" />
+    <PackageReference Include="Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel" Version="2.23.0" />
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="10.0.0" /> 
   </ItemGroup>
 
--- a/src/DotNetWorker.ApplicationInsights/FunctionsApplicationInsightsExtensions.cs
+++ b/src/DotNetWorker.ApplicationInsights/FunctionsApplicationInsightsExtensions.cs
@@ -83,6 +83,7 @@ namespace Microsoft.Azure.Functions.Worker
                     "Application Insights SDK has not been added. Please add and configure the Application Insights SDK. See https://learn.microsoft.com/en-us/azure/azure-monitor/app/worker-service for more information.");
 
             services.AddHostedService<ApplicationInsightsValidationService>();
+            services.AddHostedService<ServerTelemetryChannelConfigurationService>();
 
             // Lets the host know that the worker is sending logs to App Insights. The host will now ignore these.
             services.Configure<WorkerOptions>(workerOptions => workerOptions.Capabilities["WorkerApplicationInsightsLoggingEnabled"] = bool.TrueString);
@@ -100,7 +101,7 @@ namespace Microsoft.Azure.Functions.Worker
         }
 
         /// <summary>
-        /// This services is for a singular purpose: trigger validation of <see cref="ApplicationInsightsSdkValidationMarker" /> on startup.
+        /// This service is for a singular purpose: trigger validation of <see cref="ApplicationInsightsSdkValidationMarker" /> on startup.
         /// </summary>
         private class ApplicationInsightsValidationService : IHostedService
         {
--- a/src/DotNetWorker.ApplicationInsights/FunctionsApplicationInsightsExtensions.cs
+++ b/src/DotNetWorker.ApplicationInsights/FunctionsApplicationInsightsExtensions.cs
@@ -19,6 +19,23 @@ namespace Microsoft.Azure.Functions.Worker
 {
     public static class FunctionsApplicationInsightsExtensions
     {
+        public static IServiceCollection ConfigureFunctionsApplicationInsights(this IServiceCollection services, Action<FunctionsApplicationInsightsOptions> configureOptions)
+        {
+            if (services == null)
+            {
+                throw new ArgumentNullException(nameof(services));
+            }
+
+            if (configureOptions == null)
+            {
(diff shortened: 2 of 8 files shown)


src/DotNetWorker.ApplicationInsights/FunctionsApplicationInsightsExtensions.cs
```diff
@@ -72,25 +90,23 @@ public static IServiceCollection ConfigureFunctionsApplicationInsights(this ISer
...
-        /// When we do have options to configure, this can be moved to be a public top-level class.
-        /// </remarks>
-        private class FunctionsApplicationInsightsOptions
+        private class ApplicationInsightsSdkValidationMarker
         {
         }
 
         /// <summary>
-        /// This services is for a singular purpose: trigger validation of <see cref="FunctionsApplicationInsightsOptions" /> on startup.
+        /// This services is for a singular purpose: trigger validation of <see cref="ApplicationInsightsSdkValidationMarker" /> on startup.
```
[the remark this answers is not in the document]
RohitRanjanMS: Fixed in 442627fa8 — updated the XML doc comment to "This service is".

src/DotNetWorker.ApplicationInsights/Initializers/TelemetryConfigurationSetup.cs
```diff
@@ -22,6 +26,19 @@ public void Configure(TelemetryConfiguration telemetryConfiguration)
...
             }
+
+            if (telemetryConfiguration.TelemetryChannel is ServerTelemetryChannel serverTelemetryChannel)
+            {
+                if (_options.MaxTelemetryBufferDelay < MinTelemetryBufferDelay)
+                {
+                    throw new ArgumentOutOfRangeException(
+                        $"{nameof(FunctionsApplicationInsightsOptions)}.{nameof(FunctionsApplicationInsightsOptions.MaxTelemetryBufferDelay)}",
+                        _options.MaxTelemetryBufferDelay,
+                        $"{nameof(FunctionsApplicationInsightsOptions.MaxTelemetryBufferDelay)} must be at least {MinTelemetryBufferDelay.TotalSeconds} seconds.");
```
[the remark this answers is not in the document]
RohitRanjanMS: Addressed in 442627fa8. The buffer-delay validation moved into a new ServerTelemetryChannelConfigurationService (IHostedService) to fix a registration-order bug, and it now uses nameof(FunctionsApplicationInsightsOptions.MaxTelemetryBufferDelay) as the paramName.

src/DotNetWorker.ApplicationInsights/Initializers/ServerTelemetryChannelConfigurationService.cs
```diff
@@ -0,0 +1,53 @@
...
+
+namespace Microsoft.Azure.Functions.Worker.ApplicationInsights.Initializers
+{
+    /// <summary>
+    /// Applies <see cref="FunctionsApplicationInsightsOptions.MaxTelemetryBufferDelay" /> to the Application Insights
+    /// <see cref="ServerTelemetryChannel" /> on startup. This runs after the container is fully built, so the value is
+    /// applied regardless of whether <c>ConfigureFunctionsApplicationInsights</c> was called before or after
+    /// <c>AddApplicationInsightsTelemetryWorkerService</c>.
+    /// </summary>
+    internal class ServerTelemetryChannelConfigurationService : IHostedService
```
jviau: I believe `TelemetryConfiguration` leverages `IOptions` pattern. We should be able to just register an `IConfigureOptions<TelemetryConfiguration>` which will map the config from `FunctionsApplicationInsightsOptions` onto `TelemetryConfiguration`. A hosted service shouldn't be needed.

jviau (changes requested):

jviau:
@RohitRanjanMS I am unsure we need this change at all. Can customers not just configure `TelemetryConfiguration` themselves? Why do we need to provide our own abstraction?

src/DotNetWorker.ApplicationInsights/DotNetWorker.ApplicationInsights.csproj
```diff
@@ -17,6 +17,7 @@
   <ItemGroup>
     <PackageReference Include="Azure.Identity" Version="1.17.0" />
     <PackageReference Include="Microsoft.ApplicationInsights.PerfCounterCollector" Version="2.23.0" />
+    <PackageReference Include="Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel" Version="2.23.0" />
```
jviau, line 20: Should we be adding this dependency? It is enforcing a specific telemetry channel on all app insight usages of this package now.

src/DotNetWorker.ApplicationInsights/Initializers/ServerTelemetryChannelConfigurationService.cs
RohitRanjanMS: Done.

src/DotNetWorker.ApplicationInsights/DotNetWorker.ApplicationInsights.csproj
RohitRanjanMS: ServerTelemetryChannel  is the SDK default channel, so this package is already present in the dependency graph — the reference doesn't change functionality. The only downside is for apps that swap in a different  ITelemetryChannel , where the added reference is unused overhead. I took the direct reference to keep the channel access type-safe and avoid reflection.

RohitRanjanMS:
We see frequent "missing telemetry" complaints, and reducing the buffer delay should help. It was also an ask from the Azure Monitor team. I considered documenting it instead of changing the default, but I don't think that would help adoption.

PR #3468, opened 2026-07-20
refactor(sdk): replace Mono.Cecil with System.Reflection.MetadataLoadContext

jviau:
Replaces Mono.Cecil with the in-box `System.Reflection.MetadataLoadContext` (MLC) for the assembly metadata inspection in `Azure.Functions.Sdk`. The motivation is to drop a third-party dependency in favor of a first-party, in-box library for the read-only metadata scanning the SDK performs (assembly-level custom attributes plus the attribute base-type inheritance walk used to detect WebJobs startup and extension-information attributes).

### Approach

- Swap the package reference: remove `Mono.Cecil` and add `System.Reflection.MetadataLoadContext` `10.0.10`; bump `Microsoft.Bcl.TimeProvider` to `10.0.10`.
- Port `MonoExtensions` to `ReflectionExtensions`, working over `CustomAttributeData`/`Type` instead of Cecil's object model.
- `FunctionsAssemblyScanner` now owns a `MetadataLoadContext` backed by a `PathAssemblyResolver`. The resolver is built from a flat list of assembly file paths plus the host runtime directory. The runtime directory is required so MLC can resolve the core assembly and terminate the attribute base-type walk at `System.Attribute` -> `object` -> null (Cecil could rely on lazy resolution and did not need this).
- `ResolveExtensionPackages` now builds and owns a per-target scanner instead of the previous stateless static `ExtensionReference.TryGetFromModule` call. MLC eagerly resolves attribute types (for example `ExtensionInformationAttribute`, which lives in a separate `Extensions.Abstractions` package), so the scanner must be given the runtime-assembly paths of every package in the target, not just the assembly being scanned.

### Notes for reviewers

- Attribute-type resolution is eager under MLC, so both scanners catch `FileNotFoundException`/`BadImageFormatException` to preserve Cecil's tolerance of unresolvable types.
- Scanners are created per `LockFileTarget` because search paths differ per target framework, and each owns a disposable `MetadataLoadContext` (`using`).
- Behavior is unchanged: matching is still by attribute type name/inheritance; version/PKT differences in resolved dependencies do not affect base-type identity.
- Full `Azure.Functions.Sdk.Tests` suite passes (277/277) on both `net10.0` and `net472`, including the real-restore `ResolveExtensionPackages` tests that exercise cross-package attribute resolution.

N/A

### Pull request checklist

### Additional information

Internal implementation change only; no public API or behavioral change to the SDK.


jviau:
This approach isn't going to work. MLC eagerly loads all types, and we will never have the full assembly graph in this build and so reading fails due to missing assemblies. Mono.Cecil avoided this by lazily loading types and comparing on name only. Going to try an approach with `System.Reflection.Metadata`

PR #3415, opened 2026-05-14
Improve handling of cancelled HTTP requests

satvu:
### Issue describing the changes in this PR

resolves #3348 , resolves #2485

### Pull request checklist

* [x] My changes **do not** require documentation changes
  * [ ] Otherwise: Documentation issue linked to PR
* [ ] My changes **should not** be added to the release notes for the next release
  * [ ] Otherwise: I've added my notes to `release_notes.md`
* [x] My changes **do not** need to be backported to a previous version
  * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [ ] I have added all required tests (Unit tests, E2E tests)

### Additional information

Unable to reproduce #2485 but #3348 was easily reproducible with a load test that sends and cancels requests immediately. The new behavior is that instead of getting a timeout exception, an `OperationCanceledException` is thrown.

 Changes:

   - Replace SetResult with TrySetResult in both SetHttpContextAsync and SetFunctionContextAsync, with proper handling
   when the TCS is already completed or cancelled
   - Pass cancellation tokens (RequestAborted / CancellationToken) through to WaitAsync calls so cancelled HTTP
  requests propagate promptly instead of waiting for timeout
   - Wrap SetFunctionContextAsync call in FunctionsHttpProxyingMiddleware inside the existing try block so
  cancellation exceptions are handled consistently
   - Add unit tests covering normal flow, cancellation, and timeout scenarios

PR #3472, opened 2026-07-22
Fix: Register RequestAborted cancellation on function completion task

satvu:

Addresses the root cause of HTTP request retention when clients disconnect. PR #3415 (v2.1.1) improved cancellation during the context exchange phase, but the function completion wait (`_functionCompletionTask`) remained uncancellable — causing the ASP.NET Core pipeline to hold requests until function completion or timeout (up to 30 min).

This PR registers `HttpContext.RequestAborted` on `_functionCompletionTask` so the pipeline releases the connection immediately on client disconnect.

### Pull request checklist

### Additional information

**What this changes:**
- `ContextReference.InvokeFunctionAsync()` now accepts a `CancellationToken` and registers a callback that calls `TrySetCanceled()` on disconnect
- `IHttpCoordinator.RunFunctionInvocationAsync()` accepts an optional `CancellationToken`
- `FunctionsHttpContextExtensions.InvokeFunctionAsync()` passes `context.RequestAborted` through the call chain
- `ContextReference.CompleteFunction()` disposes the cancellation registration on normal completion

**What this does NOT change:**
- The function itself still runs to completion on the worker side — only the HTTP connection hold is released
- The `Try*` pattern on `TaskCompletionSource` ensures the race between cancellation and normal completion is safe
- No breaking changes for correctly-written consumers (ASP.NET Core already expects `OperationCanceledException` from endpoint delegates for disconnected clients)

#### Testing

Tested locally, without the fix, the pipeline blocks for full function duration. With the fix it releases in ms.

### DIFF
--- a/extensions/Worker.Extensions.Http.AspNetCore/src/AspNetMiddleware/FunctionsHttpContextExtensions.cs
+++ b/extensions/Worker.Extensions.Http.AspNetCore/src/AspNetMiddleware/FunctionsHttpContextExtensions.cs
@@ -14,7 +14,7 @@ namespace Microsoft.AspNetCore.Http
         {
             var coordinator = context.RequestServices.GetRequiredService<IHttpCoordinator>();
             context.Request.Headers.TryGetValue(Constants.CorrelationHeader, out StringValues invocationId);
-            return coordinator.RunFunctionInvocationAsync(invocationId!); // will fail later if null.
+            return coordinator.RunFunctionInvocationAsync(invocationId!, context.RequestAborted);
         }
     }
 }
--- a/extensions/Worker.Extensions.Http.AspNetCore/src/Coordinator/ContextReference.cs
+++ b/extensions/Worker.Extensions.Http.AspNetCore/src/Coordinator/ContextReference.cs
@@ -32,8 +32,6 @@ namespace Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore
 
         internal Task InvokeFunctionAsync(CancellationToken cancellationToken)
         {
-            _functionStartTask.SetResult(true);
-
             if (cancellationToken.CanBeCanceled)
             {
                 _cancellationRegistration = cancellationToken.Register(
@@ -44,6 +42,8 @@ namespace Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore
                     },
                     (_functionCompletionTask, cancellationToken));
             }
+
+            _functionStartTask.SetResult(true);
 
             return _functionCompletionTask.Task;
         }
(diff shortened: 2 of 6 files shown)


extensions/Worker.Extensions.Http.AspNetCore/src/Coordinator/ContextReference.cs
```diff
@@ -27,14 +30,28 @@ public ContextReference(string invocationId)
 
         public TaskCompletionSource<FunctionContext> FunctionContextValueSource { get => _functionContextValueSource; set => _functionContextValueSource = value; }
 
-        internal Task InvokeFunctionAsync()
+        internal Task InvokeFunctionAsync(CancellationToken cancellationToken)
         {
+            if (cancellationToken.CanBeCanceled)
+            {
+                _cancellationRegistration = cancellationToken.Register(
+                    static state =>
```
fabiocav, line 38: There's a callback overload that takes the token, enabling you to simplify this code a bit: 

``` csharp
_cancellationRegistration = cancellationToken.Register(
    static (state, token) => ((TaskCompletionSource<bool>)state!).TrySetCanceled(token),
    _functionCompletionTask);
```

extensions/Worker.Extensions.Http.AspNetCore/src/AspNetMiddleware/FunctionsHttpContextExtensions.cs
```diff
@@ -14,7 +14,7 @@ internal static Task InvokeFunctionAsync(this HttpContext context)
         {
             var coordinator = context.RequestServices.GetRequiredService<IHttpCoordinator>();
             context.Request.Headers.TryGetValue(Constants.CorrelationHeader, out StringValues invocationId);
-            return coordinator.RunFunctionInvocationAsync(invocationId!); // will fail later if null.
+            return coordinator.RunFunctionInvocationAsync(invocationId!, context.RequestAborted);
```
fabiocav, line 17: `RequestAborted` signals transport loss; it does not make the `HttpContext` safe to release while the worker still uses it. Canceling this task lets the endpoint return, after which ASP.NET can uninitialize/pool the context. The running function or result handling may then access a disposed or reused context.


Although early pipeline release is this PR’s goal, this task currently acts as the context lifetime barrier. The safest incremental approach is to propagate cancellation to the function and keep the endpoint alive until the function cooperatively unwinds. Allowing the function to continue independently requires a detached worker-owned context and broader design discussion.

fabiocav (changes requested):

satvu:
Closing due to concerns w/ running function accessing disposed/reused context - will test another approach that can be taken from customer side to manage cancellation of the function execution.

PR #3473, opened 2026-07-23
chore: bump generators implicit package to 1.4.0

jviau:

Quick bump to this package, had to do in a separate PR so that `1.4.0` was published to nuget.org first.

### Pull request checklist

### DIFF
--- a/src/Azure.Functions.Sdk/Tasks/ResolveWorkerPackageReference.cs
+++ b/src/Azure.Functions.Sdk/Tasks/ResolveWorkerPackageReference.cs
@@ -1,0 +1,164 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Diagnostics.CodeAnalysis;
+using System.IO.Abstractions;
+using Microsoft.Build.Framework;
+using NuGet.LibraryModel;
+using NuGet.Frameworks;
+using NuGet.ProjectModel;
+
+namespace Azure.Functions.Sdk.Tasks;
+
+public class ResolveWorkerPackageReference(IFileSystem fileSystem)
+    : Microsoft.Build.Utilities.Task, ICancelableTask, IDisposable
+{
+    private const string WorkerPackageId = "Microsoft.Azure.Functions.Worker";
+
+    private readonly CancellationTokenSource _cts = new();
+    private readonly IFileSystem _fileSystem = Throw.IfNull(fileSystem);
+
+    public ResolveWorkerPackageReference()
+        : this(new FileSystem())
+    {
+    }
+
+    [Required]
+    public string ProjectAssetsFile { get; set; } = string.Empty;
+
+    public string? TargetFramework { get; set; }
+
+    [Output]
+    public ITaskItem[] WorkerPackages { get; private set; } = [];
+
+    public void Cancel()
+    {
+        _cts.Cancel();
+    }
+
+    public void Dispose()
+    {
+        _cts.Dispose();
+    }
+
+    public override bool Execute()
+    {
+        if (!TryGetLockFile(out LockFile? lockFile))
+        {
+            return !_cts.IsCancellationRequested;
+        }
+
+        List<ITaskItem> workerPackages = [];
+        NuGetFramework? currentFramework = TryParseFramework(TargetFramework);
+        foreach (LockFileTarget target in lockFile!.Targets)
+        {
+            if (_cts.IsCancellationRequested)
+            {
+                return false;
+            }
+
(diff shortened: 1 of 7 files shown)


satvu (approved):

kshyju (approved):

src/Azure.Functions.Sdk/Tasks/ResolveWorkerPackageReference.cs
```diff
@@ -0,0 +1,164 @@
...
+        {
+            // Don't fail when assets are missing. This can happen in DTB before restore.
+            Log.LogMessage(
+                Microsoft.Build.Framework.MessageImportance.Low,
+                "Project assets file '{0}' was not found.",
+                ProjectAssetsFile);
+
+            lockFile = null;
+            WorkerPackages = [];
+            Log.LogMessage(LogMessage.Warning_WorkerPackageNotReferenced);
```
kshyju: This logs `AZFW0111` when the assets file is *missing*, but the comment right above says that happens in design-time builds before restore. Combined with `_ConfigureWorkerSourceGen` not carrying over the `Condition="'$(DesignTimeBuild)' != 'true'"` guard that `_ValidateWorkerPackageReference` had, a fresh clone opened in VS (or `dotnet build --no-restore`) will now show "No Azure Functions worker package was found after restore. Add a package reference..." when nothing was restored yet.

Can we keep the low-importance message here and only emit `AZFW0111` when the assets file exists but the package is absent? We still want the target itself running at DTB so the source-gen properties are right for IntelliSense - it's just the warning that should be gated.

`LockFileDoesNotExist_EmitsWarning_AndReturnsNoPackage` locks in the current behavior, so that test would need to change too.

test/Azure.Functions.Sdk.Tests/Integration/SdkEndToEndTests.Build.cs
```diff
@@ -320,19 +320,17 @@ public void Build_MissingWorkerPackage_Warning()
...
 
         // Act
         BuildOutput output = project.Build(restore: true);
 
         // Assert - build succeeds, but the SDK warns that no worker package was found after restore.
         output.Should().BeSuccessful();
-        output.WarningEvents
-            .Where(x => x.Code == LogMessage.Warning_WorkerPackageNotReferenced.Code)
-            .Should().NotBeEmpty()
+        output.WarningEvents.Should().Contain(x => x.Code == LogMessage.Warning_WorkerPackageNotReferenced.Code)
```
kshyju, line 330: `AndWhichConstraint.And` returns the parent collection assertions, not the matched subset, so `AllSatisfy` here runs over *every* warning in the build rather than just the AZFW0111 ones. I checked this against AwesomeAssertions 9.3.0 - `new[]{"AZFW0111","NU1234"}.Should().Contain(x => x == "AZFW0111").And.AllSatisfy(...)` fails.

That makes this stricter than the original `.Where(...)` version: any unrelated NuGet or Roslyn warning will now break the test. Either keep the `Where` filter or scope through `.Which`.

src/Azure.Functions.Sdk/Tasks/ResolveWorkerPackageReference.cs
```diff
@@ -0,0 +1,164 @@
...
+using NuGet.LibraryModel;
+using NuGet.Frameworks;
+using NuGet.ProjectModel;
+
+namespace Azure.Functions.Sdk.Tasks;
+
+public class ResolveWorkerPackageReference(IFileSystem fileSystem)
+    : Microsoft.Build.Utilities.Task, ICancelableTask, IDisposable
+{
+    private const string WorkerPackageId = "Microsoft.Azure.Functions.Worker";
```
kshyju, line 16: The commit message says "disable generators if core package is missing", but this matches the `Microsoft.Azure.Functions.Worker` metapackage while the `_ValidateWorkerPackageReference` target this replaces checked for `Microsoft.Azure.Functions.Worker.Core`.

Is the narrowing intentional? `FunctionAttribute` lives in `.Core`, so a project that references `Worker.Core` directly (without the metapackage) would now get AZFW0111 *and* silently lose source generation.

kshyju (commented):
Late follow-up on the source-gen change in the second commit - the version bump itself looks fine. Three things worth a look, one of which is a user-facing false positive.

kshyju:
Review comments from GHCP Opus 5

src/Azure.Functions.Sdk/Tasks/ResolveWorkerPackageReference.cs
jviau: Yeah this was an intentional change. I figured it was safer to push users to the meta-package instead of just the core one.

kshyju (approved):

PR #3493, opened 2026-08-27
fix: repair Kafka attribute tests

TsuyoshiUshio:
## Summary

- remove stray closing braces from the new Kafka attribute test file
- add the missing xUnit import

## Validation

- focused Kafka attribute tests: 2 passed
- Worker.Extensions.Tests: 235 passed
- dotnet build: succeeded

PR #3491, opened 2026-08-25
feat: add Kafka OAuth HTTPS CA settings

TsuyoshiUshio:
## Summary
Adds .NET isolated Worker API parity for the OAuth/OIDC HTTPS CA settings introduced in Azure/azure-functions-kafka-extension#650, and updates the bundled WebJobs Kafka extension metadata to 4.3.3.

## Issue
Relates to Azure/azure-functions-kafka-extension#649

## Changes
- Add `HttpsCaLocation` and `HttpsCaPem` to Kafka trigger and output attributes.
- Reference `Microsoft.Azure.WebJobs.Extensions.Kafka` 4.3.3.
- Add attribute contract tests and release notes.

## Breaking Changes
None.

## Testing
- [x] Kafka unit tests: 12 passed, 0 failed
- [x] Release build with `ContinuousIntegrationBuild=true`
- [x] Generated extension validation resolved local Kafka WebJobs extension 4.3.3
- [x] Vulnerability scan: clean
- [x] E2E tests: skipped; no Kafka runtime behavior changes
- [x] Coverage: skipped; direct API contract tests cover the added surface

## Dependency
This PR depends on publishing `Microsoft.Azure.WebJobs.Extensions.Kafka` 4.3.3 from Azure/azure-functions-kafka-extension#672.

## Self-Review Checklist
- [x] No intermediate files or coverage output committed
- [x] No debug statements or TODO comments
- [x] No hardcoded credentials or sensitive data
- [x] Commit message follows conventional format

PR #3495, opened 2026-08-27
feat: add Kafka OAuth HTTPS CA settings

TsuyoshiUshio:
## Summary

- Add `HttpsCaLocation` and `HttpsCaPem` to Kafka trigger and output bindings for OAuth/OIDC token endpoint certificate validation.
- Update `Microsoft.Azure.WebJobs.Extensions.Kafka` from 4.3.2 to 4.3.3.
- Add corrected attribute tests covering both Kafka trigger and output bindings.

Supersedes #3491 and #3493.
Relates to Azure/azure-functions-kafka-extension#649.

## Testing

- `dotnet test test\extensions\Worker.Extensions.Tests\Worker.Extensions.Tests.csproj --filter FullyQualifiedName~KafkaAttributeTests --no-restore --verbosity minimal`

### DIFF
--- a/extensions/Worker.Extensions.Kafka/src/KafkaTriggerAttribute.cs
+++ b/extensions/Worker.Extensions.Kafka/src/KafkaTriggerAttribute.cs
@@ -205,6 +205,19 @@ namespace Microsoft.Azure.Functions.Worker
         public string OAuthBearerTokenEndpointUrl { get; set; }
 
         /// <summary>
+        /// File or directory path to CA certificates for verifying the OAuth/OIDC token endpoint certificate.
+        /// The special value "probe" uses the operating system's default certificate paths.
+        /// https.ca.location in librdkafka
+        /// </summary>
+        public string HttpsCaLocation { get; set; }
+
+        /// <summary>
+        /// CA certificate for verifying the OAuth/OIDC token endpoint certificate in PEM format.
+        /// https.ca.pem in librdkafka
+        /// </summary>
+        public string HttpsCaPem { get; set; }
+
+        /// <summary>
         /// OAuth Bearer extensions.
         /// Allow additional information to be provided to the broker.
         /// Comma-separated list of key=value pairs. E.g., "supportFeatureX=true,organizationId=sales-emea"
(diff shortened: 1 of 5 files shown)


extensions/Worker.Extensions.Kafka/src/KafkaTriggerAttribute.cs
```diff
@@ -204,6 +204,19 @@ public KafkaTriggerAttribute(string brokerList, string topic)
...
+        /// The special value "probe" uses the operating system's default certificate paths.
+        /// https.ca.location in librdkafka
+        /// </summary>
+        public string HttpsCaLocation { get; set; }
+
+        /// <summary>
+        /// CA certificate for verifying the OAuth/OIDC token endpoint certificate in PEM format.
+        /// https.ca.pem in librdkafka
+        /// </summary>
+        public string HttpsCaPem { get; set; }
```
jviau, line 218: Is it expected customers put their full PEM in the attribute itself? That means the PEM will be part of function metadata and sync triggers payload. That seems both unwieldy and also potentially exposing a secret.

TsuyoshiUshio: Thanks, this is good feedback. We should not expect customers to embed the full PEM directly in the attribute. Although this particular value is a CA certificate rather than a private key, putting the literal PEM in function metadata and the sync-triggers payload is still unwieldy and should be avoided.

For this PR, I would like to keep these properties because the change restores a specific customer scenario that can no longer use the Kafka extension, and it is consistent with the existing Kafka certificate settings such as `SslCaPEM`, `SslCertificatePEM`, and `SslKeyPEM`. The intended usage is an app-setting expression such as `%KafkaHttpsCaPem%`, with the app setting backed by an App Service Key Vault reference, rather than a PEM literal in the attribute.

I opened Azure/azure-functions-kafka-extension#678 to track the follow-up work: make the Key Vault-backed app-setting pattern the explicit recommendation in the official documentation and samples, design a connection-based configuration that keeps certificate material out of binding metadata and works with the external Scale Controller, and deprecate the direct PEM-valued properties in a future major version after the replacement has feature parity.

jviau (approved):

